PR #788, opened 2019-03-08
Don't read from errchan if Stop was called

ZymoticB:
https://godoc.org/google.golang.org/grpc#Server.Serve
"Serve will return a non-nil error unless Stop or GracefulStop is called."
"lis will be closed when this method returns."

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**


ZymoticB:
Haven't tested this but it will likely fix #787

Issue #728, opened 2019-02-12
Agents should balance streams across all available SPIRE servers


ZymoticB:
When running SPIRE server in a HA configuration, agents should be balanced across the available servers. There are two primary factors to making this successful: having a balancing strategy, and ensuring that balancing strategy is periodically applied especially after a server failure.

Imagine this scenario, you have 10k agents connected to 2 servers. You have deployed a balancing strategy that is perfect, the agents are evenly balanced {5000, 5000}. Now, one of the servers fails, all the agent streams then move to the server that is still available {10000,0}. The failed server recovers, however, the streams will not rebalance unless the agent is restarted so you are stuck in this {10000,0} configuration.

Consider another scenario, you have your agents perfectly balanced {5000, 5000}. Unfortunately, you planned badly, and a single spire server can only actually handle 7500 agents; if one of your servers fails you're going to have a bad time! You add a new server, your agents are now balanced like {5000,5000,0}. You need to restart all the agents in your fleet to rebalance the streams.

A complete solution here likely involves at least the following check-boxes:
- [ ] Periodically resolve DNS to ensure dynamic membership of SPIRE servers.
- [x] A "reasonably good" balancing strategy, from experience, a randomized response from a DNS resolver for an A record isn't enough. http://www.eecs.umich.edu/techreports/cse/96/CSE-TR-316-96.pdf covers some options, the HRW option is likely appropriate as it doesn't require collecting any information or sharing any state.
- [x] Agent must periodically re-establish their streams to ensure they are balanced across an ever-changing group of servers.

evan2645:
@ZymoticB there have been a number of changes since this issue was opened around connection management, client balancing gRPC algo choice, etc... I believe that the first two boxes here can be checked (as they're handled by the gRPC libs), but I am not sure if the third is also managed for us? Any idea if this is still an issue?

ZymoticB:
3. was my mistake, the agent RPCs are all unary so that's not an issue
2. is also already in a reasonable place because the RPCs are unary to a client side round robin load balancer does a good enough job.
1. is still somewhat of an issue because grpc only re-resolves every 30 minutes which is a terrible default.

Should probably open a new issue specific to #1

evan2645:
I opened https://github.com/spiffe/spire/issues/1192 to address the dns resolution interval. Going to close this out in favor of that one, please let me know if I've missed anything

PR #795, opened 2019-03-13
Releasing connection when there are errors on responses

### LINKED ISSUE #769: Reconnect to SPIRE server on error as appropriate
While investigating #766 in a multiple SPIRE server deployment with a load balancer I saw a case where the gRPC connection from the SPIRE agent to server was in an OK state, but the server is unable to service requests. The SPIRE agent kept retrying the request on the same connection in futility. In this sort of case, the SPIRE agent should try a new connection on the chance that it will get a good server.

It's probably worth looking at all cases where the SPIRE agent might have a failure mode where it does not currently reconnect on error, and a re-connection attempt would never be worse than not trying to reconnect.

marcosy:

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Agent fetching updates from server

**Description of change**

This change releases agent connection to SPIRE server when there is an error on the response. This allows the client to re-connect to another server in case something went wrong with that particular instance.

**Which issue this PR fixes**

Fixes #769


pkg/agent/client/client.go
```diff
@@ -120,7 +120,7 @@ func (c *client) FetchUpdates(ctx context.Context, req *node.FetchX509SVIDReques
 	// Close the stream whether there was an error or not
 	stream.CloseSend()
 	if err != nil {
-		// TODO: should we try to create a new stream?
+		c.Release()
```
evan2645: What kind of errors might cause us to reach this point? Could any of them be recoverable i.e. might work if we asked the same server again? Similar question for below, but perhaps with a slightly different answer

marcosy: The errors that reach the point below (line 136) are those coming from the [server handler](https://github.com/spiffe/spire/blob/51e2e96f0fd8c2e612f884f9015cfca3510928af/pkg/server/endpoints/node/handler.go#L197).

Some of them does not seem to work if asked to the same server again, for example:

```
return status.Error(codes.ResourceExhausted, err.Error())
```


On the other hand, I'm not sure about errors reaching line 122 honestly. However, my intention here was to make sure that any failure leads to a new connection. So in case something unexpected happens on the server, the agent is able to recycle the connection and reach another server.

I found this particular issue difficult to reproduce so I did more focus on the second part.

What do you think? Happy to receive suggestions

scotte: It is very difficult to reproduce these sorts of issues. This is the kind of thing that will surface and cause cascading failures in a large production environment when things go bad.

Essentially, it boils down to not beating your head against the wall, trying the same failing request over and over again to the same server and instead try reconnecting. Assuming a production environment with a scaled-up SPIRE server deployment behind a load balancer, there's a better chance that reconnecting will give you a good server, unless all servers are bad (in which case you have bigger issues anyway).

evan2645: Yea, wasn't suggesting that we _shouldnt_ reconnect. Ultimately, I was a little worried about small transient failures triggering stampeding herd of agent reconnects, because 1) tls negotiation isn't free at scale and 2) this issue https://github.com/spiffe/spire/issues/728

I got some rough numbers from @ZymoticB and @APTy, ran some back of the envelope calculation and think that we're probably OK and perhaps trying to recover before redialing is a premature optimization.

ZymoticB: I would invest in load shedding on the server side before I invested in recovering on the client side.

Exponential backoff+jitter on the client side and making rejecting new connections on the server side under load cheap will solve the stampeding herd concern IMO.

PR #798, opened 2019-03-14
add DNS to registration, use in X509 SVID

### LINKED ISSUE #688: Set CN and DNS SANs in registration API

amoore877:

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
No breaking changes. Addition of DNS List to Registration.

**Description of change**
Add DNS List to Registration API proto and CLI. The first DNS name becomes the `Common Name` of X509 SVIDs, and the entire DNS List goes into `DNSNames` of X509 SVIDs.
Refactoring of `ca.go` due to large overlap of code.
Minor lints.

**Which issue this PR fixes**
Fixes #688


amoore877:
I will be starting vacation tomorrow, so I welcome anyone with the time, interest, and eagerness to address any review comments that arise and submit the code themselves instead (just please link to this PR if a new one is made :) ).
If not, I'll get to comments when I return.

evan2645:
Hi @amoore877, thank you for this. Just wanted to ping here and let you know that this change is not forgotten. I will be reviewing it tomorrow. Very sorry for the delay

amoore877:
Back from vacation. Will fix merge conflict today

pkg/server/endpoints/registration/handler.go
```diff
@@ -740,3 +751,45 @@ func getCallerID(ctx context.Context) string {
 	callerID, _ := ctx.Value(callerIDKey{}).(string)
 	return callerID
 }
+
+func validateDNS(dns string) error {
+	// follow https://tools.ietf.org/html/rfc5280#section-4.2.1.6
+	// do not allow empty or the technically valid DNS " "
```
evan2645: 👍

pkg/server/endpoints/registration/handler.go
```diff
@@ -618,6 +622,13 @@ func (h *Handler) prepareRegistrationEntry(entry *common.RegistrationEntry, forU
 	}
 
 	var err error
+	for _, dns := range entry.DnsList {
+		err = validateDNS(dns)
+		if err != nil {
+			return nil, err
```
evan2645: It would be useful to know which DNS name failed validation e.g. `fmt.Errorf("dns name %v failed validation: %v"), dns, err`

pkg/server/endpoints/registration/handler.go
```diff
@@ -740,3 +751,45 @@ func getCallerID(ctx context.Context) string {
 	callerID, _ := ctx.Value(callerIDKey{}).(string)
 	return callerID
 }
+
+func validateDNS(dns string) error {
+	// follow https://tools.ietf.org/html/rfc5280#section-4.2.1.6
+	// do not allow empty or the technically valid DNS " "
+	dns = strings.TrimSpace(dns)
+	if len(dns) == 0 {
+		return errors.New("dns is empty or only whitespace")
```
evan2645: If we update error message similar to comment above, we can probably drop the "dns" prefixing on all these error messages

pkg/server/plugin/datastore/sql/models.go
```diff
@@ -84,6 +86,19 @@ type Selector struct {
 	Value             string `gorm:"unique_index:idx_selector_entry"`
 }
 
+// A DNS for a registration entry
+type DNS struct {
```
evan2645: `DNSName` feels a good bit clearer here. Table name would also make more sense too I think e.g. `dns_names`

cmd/spire-server/cli/entry/create.go
```diff
@@ -45,6 +45,9 @@ type CreateConfig struct {
 
 	// Expiry of entry
 	EntryExpiry int64
+
+	// DNS entries for SVIDs based on this entry
+	DNSList StringsFlag
```
evan2645: DNSNames feels clearer, more aligned with `Selectors`, and I think will work better in datastore layer too

pkg/server/endpoints/node/handler.go
```diff
@@ -712,7 +712,12 @@ func (h *Handler) signCSRs(ctx context.Context,
 			if err != nil {
 				return nil, err
 			}
-			svid, err := h.buildCASVID(ctx, csr, e.Ttl)
+			svid, err := h.buildCASVID(ctx, csr,
+				ca.X509Params{
+					TTL:     time.Duration(e.Ttl) * time.Second,
+					DNSList: e.DnsList,
```
evan2645: non-blocking comment: Maybe take belt-and-suspenders approach here, and not pass dns since we know we're not supposed to be setting it?

pkg/server/endpoints/node/handler_test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/spiffe/spire/pkg/server/ca"
```
evan2645: nit: grouping

pkg/server/endpoints/node/handler_test.go
```diff
@@ -605,13 +607,55 @@ func (s *HandlerSuite) TestFetchX509SVIDWithWorkloadCSR() {
...
+	})
+
+	s.Equal([]*common.RegistrationEntry{entry}, upd.RegistrationEntries)
+	s.assertBundlesInUpdate(upd)
+	chains := s.assertSVIDsInUpdate(upd, workloadID)
+	s.Equal(dnsList, chains[0][0].DNSNames)
+	s.Equal("somehost1", chains[0][0].Subject.CommonName)
+}
+
+func (s *HandlerSuite) TestFetchX509SVIDWithMultipleDNS() {
```
evan2645: might be good to test the CA bits too to ensure that DNS and CN is not being set?

pkg/server/plugin/datastore/sql/sql.go
```diff
@@ -1274,6 +1300,19 @@ func modelToEntry(tx *gorm.DB, model RegisteredEntry) (*common.RegistrationEntry
 		})
 	}
 
+	var fetchedDNSs []*DNS
+	if err := tx.Debug().Model(&model).Related(&fetchedDNSs).Error; err != nil {
```
evan2645: What does this call to `Debug()` do?

amoore877: forgot to get rid of when I was having sql issues :)

pkg/server/endpoints/node/handler_test.go
amoore877: adding to TestFetchX509SVIDWithDownstreamCSR

amoore877:
`FAIL	github.com/spiffe/spire/pkg/server/endpoints	0.133s`
Looks like more racy tests. Re-attempt/re-build?

evan2645:
@amoore877 I was looking at that this morning... Is it a race? It doesn't seem to be? The stacktrace isn't long enough to reach back into our code, but I did capture it. I will try re-running and see what happens

amoore877:
> @amoore877 I was looking at that this morning... Is it a race? It doesn't seem to be? The stacktrace isn't long enough to reach back into our code, but I did capture it. I will try re-running and see what happens

My guess from looking at the package was the frequent use of timers, which seems to also be combined with a mock clock. The odd thing is that the mock clock appears to have ticker functionality which is not used in the testing.
As you said the trace doesn't go far back enough, but at least has that package name in the failure summary.

costinm:
Maybe I'm missing something - but how does the CA validate the requestor is entitled/owns the DNS entry ?

evan2645:
These names are set via a privileged administration api (i.e. by an RA), not requested by agents or workloads. Callers of that API may be customer-specific integrations or automation, administrative CLI invocations, or k8s controller consuming CRDs that declare which workloads should have which names.

PR #807, opened 2019-03-20
Add gauges metric to measure svid ttl

martincapello:
This PR adds some gauges to measure SVIDs TTLs. Also refactors a bit some metrics counters to make use of the CallCounter struct, which simplifies some of the code.

### DIFF
--- a/pkg/server/endpoints/node/handler.go
+++ b/pkg/server/endpoints/node/handler.go
@@ -142,7 +142,7 @@ func (h *Handler) Attest(stream node.Node_AttestServer) (err error) {
 	svid, err := h.c.ServerCA.SignX509SVID(ctx, request.Csr, ca.X509Params{})
 	if err != nil {
 		h.c.Log.Error(err)
-		return errors.New("failed to to sign CSR")
+		return errors.New("failed to sign CSR")
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/server/endpoints/node/handler.go
```diff
@@ -234,8 +235,16 @@ func (h *Handler) FetchX509SVID(server node.Node_FetchX509SVIDServer) (err error
...
-		for spiffeID := range svids {
+		for spiffeID, svid := range svids {
 			counter.AddLabel("spiffe_id", spiffeID)
+
+			ttl := time.Until(time.Unix(svid.ExpiresAt, 0))
+			metrics.SetGaugeWithLabels(
+				[]string{"node_api", "x509_svid", "fetch", "ttl"},
+				float32(ttl.Seconds()),
+				[]telemetry.Label{
+					{Name: "spiffe_id", Value: spiffeID},
```
evan2645: I think that in practice, measuring this from node api handler will be less useful. First, we already know what the ttl should be because it is set in the registration entry as a matter of policy. Second, I think this might get confusing because it is possible to have more than one registration entry for the same SPIFFE ID, which might have differing TTLs.. in this case, the metric points will be indistinguishable, some backends will actually take the average value for display purposes.

I recommend removing this point and emitting only from workload api

martincapello: Makes sense, will remove this. Thank you!

PR #821, opened 2019-03-27
Add TLS options for MySQL

mbyczkowski:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
None, adds functionality to datastore/sql plugin for MySQL.

**Description of change**
Allows users to specify custom Root CAs and client certificate for the TLS connection to the MySQL server. As an example, [AWS RDS MySQL](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_MySQL.html#MySQL.Concepts.SSLSupport) requires the client to use RDS root certificate for verification.

Unfortunately, [MySQL Go driver](https://github.com/go-sql-driver/mysql#tls) only allows to specify Root CAs and client certificates via a custom TLS Config (and not via a connection string options). This change allows for conditional configuration of the connection string, by simply specifying `root_ca_path` or `client_key_path` and `client_cert_path` in the plugin's config.

PostgreSQL doesn't need these config options, because they can be set via a connection string. Adding this to the datastore/sql plugin config might be somewhat misleading, so I've updated the documentation accordingly.

**Which issue this PR fixes**
None.

### DIFF
--- a/pkg/server/plugin/datastore/sql/mysql.go
+++ b/pkg/server/plugin/datastore/sql/mysql.go
@@ -1,7 +1,10 @@
 package sql
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"errors"
+	"io/ioutil"
 
 	mysqldriver "github.com/go-sql-driver/mysql"
 	"github.com/jinzhu/gorm"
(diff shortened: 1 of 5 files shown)


pkg/server/plugin/datastore/sql/mysql.go
```diff
@@ -1,25 +1,96 @@
...
+	opts, err := mysqldriver.ParseDSN(cfg.ConnectionString)
+	if err != nil {
+		// the connection string should have already been validated by now
+		// (in validateMySQLConfig)
+		return "", sqlError.Wrap(err)
+	}
+
+	// load and configure Root CA if it exists
+	if len(cfg.RootCAPath) > 0 {
+		logrus.Infof("configuring MySQL connection to use Root CA found in %s", cfg.RootCAPath)
```
evan2645: Couple thoughts on this (and subsequent) log line

First is since we are not using a logger passed in from core, if we're configured to log to file or if we are configured for level above info, this won't have any knowledge.  Second, @azdagron is currently working on plugin catalog refactor that will expose logging to plugins as a first class citizen, solving problem 1. Finally, since these messages aren't particularly critical, I might leave them out for now and we can add them later when we have fixed the plugin logging situation (but perhaps as debug?).

What do you think? If you feel strongly about these lines, then I recommend we ensure that info level is suitable, and then pass a logger into datastore plugin... but I am not sure it's worth it at this point. Curious to hear @azdagron thoughts too

mbyczkowski: I think that makes sense. I actually added those log lines as an after-thought and mainly to help with debugging. I'm happy to take them out and revisit once we have loggers available to plugins. 👍

PR #846, opened 2019-04-10
Don't sort ListRegistrationEntries response and remove reflect

### LINKED ISSUE #844: ListRegistrationEntries shouldn't sort entries
Currently ListRegistrationEntries in the sql datastore implementation returns entries in a sorted fashion. This isn't documented and I don't think the performance overhead is not worth it, we should remove it.

```
func modelsToEntries(tx *gorm.DB, fetchedRegisteredEntries []RegisteredEntry) (responseEntries []*common.RegistrationEntry, err error) {
	entries, err := modelsToUnsortedEntries(tx, fetchedRegisteredEntries)
	if err != nil {
		return nil, err
	}
	util.SortRegistrationEntries(entries)
	return entries, nil
}

```

ZymoticB:
Fixes #844

### DIFF
--- a/cmd/spire-server/cli/entry/show.go
+++ b/cmd/spire-server/cli/entry/show.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/spiffe/spire/cmd/spire-server/util"
 	"github.com/spiffe/spire/pkg/common/idutil"
+	commonutil "github.com/spiffe/spire/pkg/common/util"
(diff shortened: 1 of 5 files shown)


ZymoticB:
Great idea, updated. I modified the fixture while I was testing to make sure that things were actually getting sorted. Left it as I saw value in the tests requiring their expected values get sorted.

PR #848, opened 2019-04-10
Add experimental flag

MarcosDY:
Add experimental flag to skip validation done to csr spiffe id, to verify it contains `/spire/agent/*`
Solve: #854


pkg/server/endpoints/node/handler_test.go
```diff
@@ -314,6 +315,36 @@ func (s *HandlerSuite) TestAttestSuccess() {
 	s.Empty(s.getNodeSelectors(agentID))
 }
 
+func (s *HandlerSuite) TestAttestAgentless() {
+	//  Create a registration entry to return with the SVID update
+	s.createRegistrationEntry(&common.RegistrationEntry{
+		ParentId: agentlessID,
+		SpiffeId: workloadID,
+	})
```
evan2645: I'm not sure I understand why we need to create a registration entry for this? It should work without one, no?

pkg/server/endpoints/node/handler_test.go
```diff
@@ -314,6 +315,36 @@ func (s *HandlerSuite) TestAttestSuccess() {
...
+
+	attestor := fakeservernodeattestor.Config{
+		Data:          map[string]string{"data": agentlessID},
+		ReturnLiteral: true,
+	}
+
+	agentlessCrt := s.makeCSR(agentlessID)
```
evan2645: nit: `agentlessCSR`

pkg/server/server.go
```diff
@@ -59,6 +59,9 @@ type Config struct {
 
 	UpstreamBundle bool
 
+	// Skip agent id validation in node attestation
```
evan2645: I think we can yank this comment? The same one appears below in the actual configurable that does this

pkg/server/endpoints/node/handler_test.go
MarcosDY: yes, I forgot to remove it 🤦‍♂️

PR #860, opened 2019-04-16
Make PSAT attestor to use Token Review API

### LINKED ISSUE #815: Add token review API support for k8s PSAT node attestor
The Kubernetes node attestors currently require SPIRE server to validate service account tokens, which is a problem in hosted Kubernetes environments where the token validation certificate is not readily available. An option to use the token review API would be useful in these environments.

marcosy:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

PSAT node attestor

**Description of change**

PSAT node attestor currently requires SPIRE server to validate service account tokens. To do so, the server requieres the validation key, which may not be available in some providers. 

This PR: 
- Moves token validation to Kubernetes side, by making use of the [Token Review API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#tokenreview-v1-authentication-k8s-io).

- Changes the spiffeID format for this attestor (as recommendend in the [issue discussion](https://github.com/spiffe/spire/issues/815#issuecomment-476326779)). 
From: 
`spiffe://<trust domain>/spire/agent/k8s_psat/<cluster>/<pod uid>` 
To: 
`spiffe://<trust domain>/spire/agent/k8s_psat/<cluster>/<node name>` 

- Removes the `issuer` attestor configuration. Since this claim is not supported by Token Review API

**Which issue this PR fixes**

Fixes #815


examples/k8s/simple_psat/spire-server.yaml
```diff
@@ -13,6 +13,40 @@ metadata:
...
+  name: spire-server-role
+rules:
+- apiGroups: [""] # "" indicates the core API group
+  resources: ["pods"]
+  verbs: ["get"]
+- apiGroups: ["authentication.k8s.io"]
+  resources: ["tokenreviews"]
+  verbs: ["get", "create"]
+- apiGroups: ["authentication.k8s.io"]
+  resources: ["subjectaccessreviews "]
```
azdagron: i didn't see this api being used directly.. is this required?

pkg/agent/plugin/nodeattestor/k8s/psat/psat.go
```diff
@@ -83,8 +87,17 @@ func (p *AttestorPlugin) FetchAttestationData(stream nodeattestor.NodeAttestor_F
...
-		return psatError.New("token claim pod UID is empty")
+	if claims.K8s.Namespace == "" {
+		return psatError.New("token claim namespace is empty")
+	}
+
+	if claims.K8s.Pod.Name == "" {
+		return psatError.New("token claim pod name is empty")
+	}
+
+	nodeName, err := p.k8sClient.GetNode(claims.K8s.Namespace, claims.K8s.Pod.Name)
```
azdagron: hmm. wonder if we should be getting this from the kubelet instead of requiring agents to have direct access to the API server? either that or we do the work to remove the requirement for agent's to provide the SPIFFE ID. @evan2645 , thoughts?

examples/k8s/simple_psat/spire-server.yaml
marcosy: You are right, it is not required. I'll remove it :+1:

pkg/agent/plugin/nodeattestor/k8s/psat/psat.go
marcosy: I have been thinking about this and asked @evan2645  for some thoughts. It sounds good to me to get this info from kubelet :+1:

examples/k8s/simple_psat/README.md
```diff
@@ -8,15 +8,26 @@ This configuration is an example of a simple SPIRE deployment for Kubernetes tha
...
 
 ## Usage
 
 ### Configuration
 
-+ Set trust_domain and the cluster name for the k8s PSAT NodeAttestor.
-+ Modify the path in the *k8s-sa-cert* volume for SPIRE server as appropriate
-  for your deployment - this is the certificate used to verify service accounts
-  in the cluster. This example assumes minikube.
+The following flags must be passed to API server to properly run this PSAT attestor example:
```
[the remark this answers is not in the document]
evan2645: Is this still the case now that we are using TokenReview? I'm surprised that all this has to be set in order to use service account tokens

doc/plugin_agent_nodeattestor_k8s_psat.md
```diff
@@ -7,10 +7,11 @@ reads and provides the signed projected service account token (PSAT) to the serv
 In addition to service account data, PSAT embeds the pod name and UID on its claims. This allows
 SPIRE to create more fine-grained attestation policies for agents.
 
-The plugin generates SPIFFE IDs with the form:
+The plugin queries Kubernetes API server to obtain the node in which the agent is running on.
+In this way, it generates SPIFFE IDs with the form:
 
 ```
-spiffe://<trust domain>/spire/agent/k8s_psat/<cluster>/<POD UID>
+spiffe://<trust domain>/spire/agent/k8s_psat/<cluster>/<node name>
```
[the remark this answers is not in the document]
marcosy: Node name is unique at a given time. It could be deleted and created again holding the same name and a new UID. However, two nodes with the same name and different UID can't exist at the same time.

From k8s docs:
> Only one object of a given kind can have a given name at a time. However, if you delete the object, you can make a new object with the same name.

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

examples/k8s/simple_psat/README.md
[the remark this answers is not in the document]
marcosy: Unfortunately yes,

- To allow SPIRE server to access  Kubernetes API server resources we need to set up an authorization module:
```
apiserver.authorization-mode=RBAC
```

- To get the projected service account tokens we need the following [flags](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection):
 ```
apiserver.service-account-signing-key-file=/var/lib/minikube/certs/sa.key
apiserver.service-account-api-audiences=api
apiserver.service-account-issuer=api
```

- For token review, we also need to specify the public key used to verify the tokens:
```
apiserver.service-account-key-file=/var/lib/minikube/certs/sa.pub
```

A couple more flags will be required if we want to securely access kubelet ([kubelet auth/authz](
https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-authentication-authorization/)) in SPIRE agent to get the node name.

File: doc/plugin_agent_nodeattestor_k8s_psat.md (hunk 4 above)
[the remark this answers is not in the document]
Comment (evan2645) [reply doc/plugin_agent_nodeattestor_k8s_psat.md]: Hmmm I didn't know that node has a UID. Thoughts on using node UID instead of name? It would technically be a different node at that point, right? Can we get node UID from kubelet? In this case, node name seems more appropriate as a selector than as the agent ID

Sorry for the back and forth on this :-/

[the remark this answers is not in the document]
Comment (marcosy) [reply doc/plugin_agent_nodeattestor_k8s_psat.md]: No worries! Some thoughts:
- If we use node UID, the spiffeID would be unique, and we will get a new spiffeID every time the node is recreated.
- If we use node name, the agent will re-attest every time the node is recreated, and the spiffe ID will be kept.

After some research, it looks like there is not possible to get node UID from the kubelet. I haven't been able to find documentation about kubelet API but looking into [code](https://github.com/kubernetes/kubernetes/blob/dc071b8e81b33e8f4456ab2edcb03a45a75674ee/pkg/kubelet/server/server.go#L279-L337) it seems that it can not be got.

If we decide to have node UID into spiffeID, I guess we would have to get it from Kubernetes API server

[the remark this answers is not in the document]
Comment (evan2645) [reply doc/plugin_agent_nodeattestor_k8s_psat.md]: Ok. In the interest of moving things along here, I'd vote to just keep the node name rather than take on api server dependency again. Thanks very much for looking into this

File: pkg/server/plugin/nodeattestor/k8s/psat/psat.go [hunk 5]
```diff
@@ -286,24 +255,62 @@ func (p *AttestorPlugin) setConfig(config *attestorConfig) {
...
-
-	// Additional selectors (only if query k8s api server is enabled)
-	if cluster.queryAPIServer {
-		node, err := cluster.k8sClient.GetNode(claims.K8s.Namespace, claims.K8s.Pod.Name)
-		if err != nil {
-			return nil, psatError.New("can't get node name from k8s api: %v", err)
-		}
-		selectors = append(selectors, k8s.MakeSelector(pluginName, "agent_node_name", node))
+// getNamesFromTokenStatus parses a fully qualified k8s username like: 'system:serviceaccount:spire:spire-agent'
+// from tokenStatus. The string is splitted and the last two names are returned: namespace and service account name
```
azdagron: nit: "the string is split and the"

pkg/common/plugin/k8s/apiserver/client.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	// Do request
+	resp, err := clientset.AuthenticationV1().TokenReviews().Create(req)
+	if resp == nil {
+		return nil, errors.New("token review API response is nil")
+	}
+
+	// Evaluate token review response (review server will populate TokenReview.Status field)
+	if err != nil {
```
azdagron: shouldn't we check for error before we check the response?

marcosy: agreed :+1:

PR #876, opened 2019-04-25
common/plugin/sshpop: Add custom templating functions

APTy:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Adds `pipeindex` and `split` functions that may be used as part of the agent spiffe name templating config.

**Description of change**
In some cases, an ssh certificate may contain data that needs to be massaged before it can be used in the agent's spiffe name. Rather than clutter the plugins with this logic, we expose some simple template utilities for string manipulation (See `TestTemplateFuncs` for usage).

This enables cases where the ssh `ValidPrincipal` is a routable DNS name (which is likely if its being used for inbound ssh connections). With this PR, the extraneous parts of a DNS name can be stripped as part of the agent templating process.

### DIFF
--- a/pkg/common/plugin/sshpop/sshpop.go
+++ b/pkg/common/plugin/sshpop/sshpop.go
@@ -13,6 +13,13 @@ import (
 	"golang.org/x/crypto/ssh"
 )
 
+var customFuncs template.FuncMap = map[string]interface{}{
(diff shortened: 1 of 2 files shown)


APTy:
After some offline conversations, I don't think this is the right approach. Will follow-up with some other options.

PR #882, opened 2019-04-30
Add health checking infrastructure to SPIRE Server

mcpherrinm:
We add a new common health package which is a SPIRE-flavored wrapper around
github.com/InVisionApp/go-health

This doesn't actually check anything yet.  I started looking at how to have something that called a Status method on plugins if they exposed it, but I got hung up in the weeds figuring out how the catalog worked.  So I thought it was better to start small and get some early feedback if this approach seems reasonable.

### DIFF
--- a/cmd/spire-server/cli/run/run.go
+++ b/cmd/spire-server/cli/run/run.go
@@ -17,6 +17,7 @@ import (
 	"github.com/imdario/mergo"
 	"github.com/spiffe/spire/pkg/common/catalog"
 	"github.com/spiffe/spire/pkg/common/cli"
+	"github.com/spiffe/spire/pkg/common/health"
 	"github.com/spiffe/spire/pkg/common/idutil"
 	"github.com/spiffe/spire/pkg/common/log"
 	"github.com/spiffe/spire/pkg/common/telemetry"
@@ -34,9 +35,10 @@ const (
 
 // config contains all available configurables, arranged by section
 type config struct {
-	Server    *serverConfig               `hcl:"server"`
-	Plugins   *catalog.HCLPluginConfigMap `hcl:"plugins"`
-	Telemetry telemetry.FileConfig        `hcl:"telemetry"`
+	Server       *serverConfig               `hcl:"server"`
+	Plugins      *catalog.HCLPluginConfigMap `hcl:"plugins"`
+	Telemetry    telemetry.FileConfig        `hcl:"telemetry"`
+	HealthChecks health.Config               `hcl:"health_checks"`
 }
 
 type serverConfig struct {
@@ -312,6 +314,7 @@ func newServerConfig(c *config) (*server.Config, error) {
 
 	sc.PluginConfigs = *c.Plugins
 	sc.Telemetry = c.Telemetry
+	sc.HealthChecks = c.HealthChecks
 
 	return sc, nil
 }
--- a/pkg/common/health/config.go
+++ b/pkg/common/health/config.go
@@ -1,0 +1,55 @@
+package health
+
+import (
+	"fmt"
+)
+
+type Config struct {
+	ListenerEnabled bool `hcl:"listener_enabled"`
+
+	// Address and port to listen on, defaulting to localhost:80
+	BindAddress string `hcl:"bind_address"`
+	BindPort    string `hcl:"bind_port"`
+
+	// Paths for /ready and /live
+	ReadyPath string `hcl:"ready_path"`
+	LivePath  string `hcl:"live_path"`
+}
+
+// getReadyPath returns the configured value or a default
+func (c *Config) getReadyPath() string {
+	if c.ReadyPath == "" {
+		return "/ready"
+	}
+
+	return c.ReadyPath
+}
+
+// getLivePath returns the configured value or a default
+func (c *Config) getLivePath() string {
+	if c.LivePath == "" {
+		return "/live"
+	}
+
+	return c.LivePath
+}
+
+// getAddress returns an address suitable for use as http.Server.Addr, or nil if not configured with
+// a port to listen on
+func (c *Config) getAddress() *string {
+	if c.ListenerEnabled {
+		host := "localhost"
+		if c.BindAddress != "" {
+			host = c.BindAddress
+		}
+
+		port := "80"
+		if c.BindPort != "" {
+			port = c.BindPort
+		}
+
+		address := fmt.Sprintf("%s:%s", host, port)
+		return &address
+	}
+	return nil
+}
--- a/pkg/common/health/health.go
+++ b/pkg/common/health/health.go
@@ -1,0 +1,102 @@
+package health
+
+import (
+	"context"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/InVisionApp/go-health"
+	"github.com/InVisionApp/go-health/handlers"
+	"github.com/sirupsen/logrus"
+)
+
+// health.Checker is responsible for running health checks and serving the healthcheck HTTP paths
+type Checker struct {
+	config Config
+
+	server *http.Server
+
+	hc    *health.Health
+	mutex sync.Mutex // Mutex protects non-threadsafe hc
+
+	log logrus.FieldLogger
+}
+
+func live(w http.ResponseWriter, _ *http.Request) {
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(http.StatusOK)
+}
+
+func NewChecker(config Config, log logrus.FieldLogger) *Checker {
+	hc := health.New()
+
+	var server http.Server
+	// Start HTTP server if address is configured
+	address := config.getAddress()
+	if address != nil {
+		handler := http.NewServeMux()
+
+		handler.HandleFunc(config.getReadyPath(), handlers.NewJSONHandlerFunc(hc, nil))
+		handler.HandleFunc(config.getLivePath(), live)
+
+		server = http.Server{
+			Addr:    *address,
+			Handler: handler,
+		}
+	}
+
+	hc.StatusListener = &statusListener{}
+
+	return &Checker{config: config, server: &server, hc: hc, log: log}
+}
+
+func (c *Checker) AddCheck(name string, checker health.ICheckable, interval time.Duration) error {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	return c.hc.AddCheck(&health.Config{
+		Name:     name,
+		Checker:  checker,
+		Interval: interval,
+		Fatal:    true,
+	})
+}
+
+func (c *Checker) ListenAndServe(ctx context.Context) error {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	if err := c.hc.Start(); err != nil {
+		return err
+	}
+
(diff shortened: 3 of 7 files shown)


pkg/common/health/health.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}
+
+	hc.StatusListener = &statusListener{}
+
+	return &Checker{config: config, server: &server, hc: hc, log: log}
+}
+
+func (c *Checker) AddCheck(name string, checker health.ICheckable) error {
+	return c.hc.AddCheck(&health.Config{
```
evan2645: Quick review of go-health source leads me to believe that AddCheck isn't thread safe. Should we have a mutex on Checker struct?

pkg/common/health/health.go
```diff
@@ -0,0 +1,95 @@
...
+func NewChecker(config Config, log logrus.FieldLogger) *Checker {
+	hc := health.New()
+
+	var server http.Server
+	// Start HTTP server if address is configured
+	if config.Address != "" {
+		handler := http.NewServeMux()
+
+		handler.HandleFunc(config.getHealthPath(), handlers.NewJSONHandlerFunc(hc, nil))
+		handler.HandleFunc(config.getLivePath(), hello)
```
evan2645: Based on this, I assume that Health == k8s readiness and Live == k8s liveness?

Perhaps we should converge on the k8s naming for these two checks to avoid confusion... same for default paths e.g. /ready ?

cmd/spire-server/cli/run/run.go
```diff
@@ -34,6 +35,7 @@ type runConfig struct {
 	Server        serverRunConfig            `hcl:"server"`
 	PluginConfigs catalog.HCLPluginConfigMap `hcl:"plugins"`
 	Telemetry     telemetry.FileConfig       `hcl:"telemetry"`
+	HealthChecks  health.Config              `hcl:"health_checks"`
```
evan2645: I am curious to hear your thoughts on how this might apply to spire agent in the future. I'm personally a little hesitant to allow the agent to bind to a non-localhost address.

If we think that there might be differences in what is made available to the server vs the agent, then this config might better suited to be put directly into the `server { ... }` and `agent { ... }` config blocks.

Lastly (as a general comment) we have been trying to keep configuration complexity to a minimum. The more that we can have sane defaults for, the better. The less we can expose that has the potential for misconfiguration or abuse, the better.

pkg/common/health/config.go
```diff
@@ -0,0 +1,39 @@
+package health
+
+import "time"
+
+type Config struct {
+	Address  string        `hcl:"listen_address"`
```
evan2645: Should we have a localhost default here? In exchange for using listen address we could add an `enabled` bool

pkg/common/health/config.go
```diff
@@ -0,0 +1,39 @@
...
+	Interval time.Duration `hcl:"interval"`
```
evan2645: Any thoughts on what circumstances this might be tweaked? I wouldn't mind a hardcoded value here, a user setting this to 1s or less seems like it could be bad

side-thought: any idea what hcl does for time.Duration type? This old issue seems to suggest that duration is expressed as nanoseconds (!)
https://github.com/hashicorp/hcl/issues/202

cmd/spire-server/cli/run/run.go
evan2645: Here's one possible example:
```
server {
    healthcheck_listener_enabled = true
    healthcheck_bind_address = "0.0.0.0:80"
}

...

agent {
    healthcheck_listener_enabled = true
}

File: pkg/common/health/health.go [hunk 6]
```diff
@@ -0,0 +1,95 @@
...
+	return c.hc.AddCheck(&health.Config{
+		Name:     name,
+		Checker:  checker,
+		Interval: c.config.getInterval(),
+		Fatal:    true,
+	})
+}
+
+func (c *Checker) ListenAndServe(ctx context.Context) error {
+	if err := c.hc.Start(); err != nil {
```
evan2645: Enabling health check listener should be optional.

Given util.RunTasks consideration, perhaps we don't bother setting up any of the health.New() stuff in NewChecker in the event that we are disabled, and AddCheck and ListenAndServe return nil?

pkg/common/health/health.go
mcpherrinm: AddCheck can only be called before healthchecks are started, IE, it must be called during server startup.  I think we're single threaded there, but it's certainly safer to throw a mutex around it.

pkg/common/health/config.go
mcpherrinm: Hardcoding a value seems OK to me too, though I think if we do that, the individual checks should hardcode them (Eg, if somebody pays per request to an upstream CA, you may want to health check every hour instead of every 5 seconds).

pkg/common/health/health.go
mcpherrinm: If no address is specified, it doesn't listen -- it'll still log healthy/unhealthy state transitions though, which seemed valuable to me.  I don't have strong opinions on what the level of configiness should be here.

pkg/common/health/config.go
mcpherrinm: My use-case is fine with plaintext being localhost-only, and this will match the prometheus one (just set a port).

cmd/spire-server/cli/run/run.go
mcpherrinm: My use-case is fine with plaintext being localhost-only, and this will match the prometheus one (just set a port).

pkg/common/health/health.go
mcpherrinm: Yes, I'm happy to use /ready and /live (and corresponding configurable names) to be consistent with k8s.

pkg/common/health/config.go
evan2645: Hmm yes that's a good point. I don't think that should be too problematic. A plugin implementation with extra constraints like that could also decouple internally

pkg/common/health/health.go
evan2645: Aah gotcha sorry I missed the conditional below. Yes logging health state changes seems valuable regardless of whether the listener is enabled or not - thanks for pointing that out.

pkg/common/health/health.go
```diff
@@ -0,0 +1,95 @@
...
+		return err
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(1)
```
evan2645: I think this statement should be one line below, inside the conditional, so we don't add a waiter if c.server is nil.

pkg/common/health/health.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+		}()
+	}
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		<-ctx.Done()
+		if c.server != nil {
+			c.server.Shutdown(ctx)
```
evan2645: Context will already be cancelled at this point - perhaps we should call Close instead?

evan2645:
Quick ping on this @mcpherrinm - did my comments look ok to you and are you planning on pulling this up? It's been open for a while, just want to make sure you're not blocked on us. We can try to pick it up at some point but are a little backlogged at the moment.

mcpherrinm:
I'm planning to pick this up; we're running with this as a patch for the moment but want to get something merged here.  I'm just busy with all the things we need to do for getting SPIRE running right now.

evan2645:
> I'm planning to pick this up

Ok great, no worries just wanted to check in.

> I'm just busy with all the things we need to do for getting SPIRE running right now.

Anything in particular we can help out with? Any pains that could be lessened?

mcpherrinm:
^ most recent change is just a rebase onto current master before I pick this up again

mcpherrinm:
I've addressed most of the feedback I think.  the fixup commits don't have DCOs; but I was planning to squash them.  Should I squash them now, or would it be useful for reviewing to keep them separate temporarily?

pkg/common/health/health.go
mcpherrinm: Done, code uses ready and live now.

cmd/spire-server/cli/run/run.go
mcpherrinm: Current config sample:

```health_checks {
       listener_enabled = true
       bind_address = "127.0.0.1"
       bind_port = "80"
       ready_path = "/ready"
       live_path = "/live"
}```

It's server-only, but we could disallow setting bind_address in the agent if you want (though similarly to the discussion in prometheus metrics, not sure that's worthwhile

File: pkg/common/health/config.go (hunk 4 above)
Comment (mcpherrinm) [reply pkg/common/health/config.go]: this is now configurable as discussed above

File: pkg/common/health/config.go (hunk 5 above)
Comment (mcpherrinm) [reply pkg/common/health/config.go]: Each check now specifies its own interval (though no checks are added yet, except a base dummy one)

File: pkg/common/health/health.go (hunk 6 above)
Comment (mcpherrinm) [reply pkg/common/health/health.go]: It's explicitly configurable now

File: pkg/common/health/health.go (hunk 8 above)
Comment (mcpherrinm) [reply pkg/common/health/health.go]: done

**Comment by mcpherrinm** [comment]
I've rebased again & addressed all the feedback I think.

**Comment by ajessup** [comment]
This is awesome to see. Thanks @mcpherrinm! 

Am I correct in understanding this isn't yet user-facing functionality? (ie. there are no docs to update?)

**Comment by mcpherrinm** [comment]
Yup, this doesn't really do much yet.  I just want to start laying some groundwork.

Subsequent changes can start actually adding some checks, and then we can document how to turn on health checks once they actually do something.

**Comment by ajessup** [comment]
SGTM

**Comment by mcpherrinm** [comment]
@evan2645 Any more feedback here?

File: pkg/common/health/logger.go [hunk 9]
```diff
@@ -0,0 +1,33 @@
...
+type statusListener struct {
+	log logrus.FieldLogger
+}
+
+// Assert statusListener implements IStatusListener
+var _ health.IStatusListener = &statusListener{}
+
+// HealthCheckFailed is triggered when a health check fails the first time
+func (sl *statusListener) HealthCheckFailed(entry *health.State) {
+	log.With("check", entry.Name).
```
azdagron: sl.log...

pkg/common/health/logger.go
```diff
@@ -0,0 +1,33 @@
...
+		With("details", entry.Details).
+		With("error", entry.Err).
+		Warn("Health check failed")
+}
+
+// HealthCheckRecovered is triggered when a health check recovers
+func (sl *statusListener) HealthCheckRecovered(entry *health.State, recordedFailures int64, failureDurationSeconds float64) {
+	log.With("check", entry.Name).
```
azdagron: sl.log...

mcpherrinm: Fixed these two.  This was also dragging in an extra dep on promethus/common/log, intead of using logrus directly.

mcpherrinm:
It looks like the build failed in the MySQL integration test, because it wasn't able to connect to a mysql server.  I think that's probably unrelated?

evan2645:
I restarted the build and recorded the output. I think I've seen it once or
twice before... I'll file an issue tomorrow.

--
evan

Issue #822, opened 2019-03-28
Centralization of metrics naming


amoore877:
Sub-issue of #799 

Metric names should be kept in a common file with documentation, for consistency of use, decreasing redundancy, and decreasing conflicting meanings.
Metric names should be standardized in some hierarchical/logical format (ex. not `attestor_name`, which is very generic, but perhaps something like `agent.attestor.workload.attestor_name`, which provides much more context and does not conflict with other similar names)

PR #884, opened 2019-05-01
centralization of metric names

amoore877:
**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
N/A

**Description of change**
Move all metrics names and labels to common location for consistency and searchability.
A conversation should be had here, in #822, or in a new issue about reducing possibly redundant metric names.
Minor lints fixed.

**Which issue this PR fixes**
Addresses, but does not fully close spirit of, #822. A conversation needs to be had about paring down possibly redundant metric names.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,7 +57,6 @@ require (
 	github.com/pascaldekloe/goe v0.0.0-20180627143212-57f6aae5913c // indirect
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/prometheus/client_golang v0.9.3-0.20190127221311-3c4408c8b829
-	github.com/prometheus/client_model v0.0.0-20190115171406-56726106282f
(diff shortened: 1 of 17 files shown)


amoore877:
still welcoming comments of course, but for posterity posting summary of talk with @evan2645 and @azdagron:
this common extracting should be merged as-is, but a follow-up PR can take this a step further and expose common telemetry functions that form metrics labels and names from base constants, with the functions containing detailed comments on usage. Some functions (such as labels which have associated values) would take input for assigning to the fields
this would further encapsulate telemetry logic and constants and make users of the package more aware of what they should be using where when wanting to emit metrics. It also would provide a good foundation for adding in additional metrics
that future PR can also handle potential label/name redundancies (such as `duration` vs `latency`)

amoore877:
fix merge conflict with `pkg/agent/endpoints/sds/handler.go`

PR #901, opened 2019-05-08
workloadattestor/docker: Add retry logic

APTy:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
This PR adds retry logic to the docker workload attestor in case a transient failure causes the docker daemon to be unavailable.

This PR also adds a somewhat sane rate limit to requests made from this plugin such that it can't accidentally DoS the docker daemon.


pkg/agent/plugin/workloadattestor/docker/http.go
```diff
@@ -0,0 +1,31 @@
+package docker
+
+import (
+	"net/http"
+
+	"golang.org/x/time/rate"
+)
+
+// default number of requests per second
+const defaultRateLimit = 100
```
APTy: When adding retries, my instinct is to also add rate limiting, but it may not be necessary for a local agent like the docker daemon. Also, does a rate limit actually introduce a DoS vector whereby a rogue workload could make enough attestation requests such that legitimate attestation traffic can't make it through?

pkg/agent/plugin/workloadattestor/docker/http.go
```diff
@@ -0,0 +1,31 @@
+package docker
+
+import (
+	"net/http"
+
+	"golang.org/x/time/rate"
```
ZymoticB: This is (IMO) un-testable. We should invest in an internal rate limiting implementation, or define an interface for `*rate.Limiter` and use that in spiffe/spire.

pkg/agent/plugin/workloadattestor/docker/http.go
ZymoticB: Seems like the limiting would need to be per-something rather than for the whole client.

pkg/agent/plugin/workloadattestor/docker/docker.go
```diff
@@ -47,12 +47,14 @@ type DockerPlugin struct {
 	cgroupContainerIndex int
 	fs                   cgroups.FileSystem
 	mtx                  *sync.RWMutex
```
ZymoticB: From another diff but a pointer to a `sync` primitive it a bit of a smell. You are returning a pointer from `New()` so the `sync.RWMutex` won't ever be copied.

pkg/agent/plugin/workloadattestor/docker/http.go
APTy: it could make sense to have both

limiting per container id would prevent a workload in a crash loop from DoSing us and the docker daemon by extension

limiting for the whole client helps guard against cases where workloads in a crash loop are getting new container ids as well, though maybe thats not realistic

pkg/agent/plugin/workloadattestor/docker/docker_test.go
```diff
@@ -200,6 +203,42 @@ func TestDockerError(t *testing.T) {
...
+	p := New()
+	p.docker = mockDocker
+	p.fs = mockFS
+	p.retryer.clock = mockClock
+
+	cgroupFile, cleanup := newTestFile(t, "1:foo:/bar")
+	defer cleanup()
+	mockFS.EXPECT().Open("/proc/123/cgroup").Return(os.Open(cgroupFile))
+	for i := 0; i < 4; i++ {
+		mockDocker.EXPECT().
```
ZymoticB: It doesn't look like you are using `i`. I believe there is a `.Times(4)` you could use.

pkg/agent/plugin/workloadattestor/docker/docker_test.go
```diff
@@ -200,6 +203,42 @@ func TestDockerError(t *testing.T) {
 	require.Nil(t, res)
 }
 
+func TestDockerErrorRetries(t *testing.T) {
```
ZymoticB: Can you add a test to cover the canceled context path?

I really wish there was inline coverage reporting. I don't know if we can get that on github.

pkg/agent/plugin/workloadattestor/docker/retry.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"math"
+	"time"
+
+	"github.com/spiffe/spire/test/clock"
+)
+
+const (
+	defaultNumRetries     = 3
+	defaultInitialBackoff = 100 * time.Millisecond
```
evan2645: We have similar logic in the kubernetes workload attestor as the workload can often times beat the k8s state update. Interestingly, this data comes from the docker event api, so we're probably actually working around the same thing.

The kubernetes workload attestor exposes some of this stuff as undocumented configurables. We recently found ourselves in a situation where we had to tweak them to deal with a demo that turned a bunch of stuff on at the same time. Might be worthwhile to similarly expose these, but this change is a strict improvement over the current posture so I'm happy to accept as-is.

evan2645:
Oops, looks like the last commit is missing the DCO. Normally I'd set it to pass but there is some non-trivial stuff there ... mind updating it quickly? Sorry for the trouble.

PR #938, opened 2019-05-31
UpstreamCA awssecret: Adding AssumeRole feature

mweissbacher:
This `fixes #<936>`

### DIFF
--- a/pkg/server/plugin/upstreamca/awssecret/awslib.go
+++ b/pkg/server/plugin/upstreamca/awssecret/awslib.go
@@ -3,12 +3,15 @@ package awssecret
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/secretsmanager"
+	"github.com/aws/aws-sdk-go/service/sts"
 )
 
 type secretsManagerClient interface {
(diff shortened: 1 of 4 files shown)


pkg/server/plugin/upstreamca/awssecret/awssecret.go
```diff
@@ -42,6 +42,7 @@ type AWSSecretConfiguration struct {
 	AccessKeyID     string `hcl:"access_key_id" json:"access_key_id"`
 	SecretAccessKey string `hcl:"secret_access_key" json:"secret_access_key"`
 	SecurityToken   string `hcl:"secret_token" json:"secret_token"`
+	AssumeRoleARN   string `hcl:"assumerole_arn" json:"assumerole_arn"`
```
evan2645: Perhaps this should be `assume_role_arn`? Trying to get more consistent about naming, case etc

pkg/server/plugin/upstreamca/awssecret/awslib.go
```diff
@@ -40,6 +43,19 @@ func newSecretsManagerClient(config *AWSSecretConfiguration, region string) (sec
...
+	// Optional: Assuming role
+	if config.AssumeRoleARN != "" {
+		staticsess, err := session.NewSession(&aws.Config{Credentials: awsConfig.Credentials})
+		if err != nil {
+			return nil, err
+		}
+		awsConfig.Credentials = credentials.NewCredentials(&stscreds.AssumeRoleProvider{
+			Client:   sts.New(staticsess),
+			RoleARN:  config.AssumeRoleARN,
+			Duration: 30 * time.Minute,
```
evan2645: What impact does this duration have? Will a new client be created on each rotation making this duration the max lifetime of a single operation?

mweissbacher: Yes, it's a cutoff for how long these credentials will be valid. The default is 15 minutes and we could actually make it shorter considering (changed it to 5). This will only be used once to fetch.

mweissbacher: https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds/#pkg-variables

PR #944, opened 2019-06-05
common/plugin/sshpop: Change fingerprint to be url-safe

APTy:
Signed-off-by: R. Tyler Julian <tjulian@uber.com>

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
The sshpop fingerprint previously used a version of base64-encoding that isn't safe for URLs.

**Description of change**
The sshpop fingerprint is now url-safe.


APTy:
> Looks great. I think the doc has a reference to "base64" when discussing the fingerprint portion of the SPIFFE ID that we should probably update to "url-safe base64".

good catch, thanks

Issue #833, opened 2019-03-28
Logging fields, not string formatting


amoore877:
Sub-issue of #799 

Logging must not use string formatting to insert data of interest. Instead, use logging fields. Ex. not `wla.c.L.Debugf("PID %v attested to have selectors %v", pid, selectors)` but instead `wla.c.L.WithField("pid", pid).WithField("selectors", selectors).Debugf("PID attested to have selectors")`

This cuts down on unnecessary uniqueness of logging messages, a general must in log observability platforms, while still enabling search/traceability on specific criteria.

Field names should be kept in a common location as described in #834 

This is potentially a medium to large task, and so it may be appropriate to break down into further sub-issues by file / group of files.

amoore877:
already fixed with #945

PR #945, opened 2019-06-05
centralize logging field keys

### LINKED ISSUE #834: Centralization of log field naming
Sub-issue of #799 

Dependent on acceptance of #833 and in same vein as #822

Log field names must be kept in a common file with documentation, for consistency of use, decreasing redundancy, and decreasing conflicting meanings.

amoore877:
**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Some logs have changed in structure

**Description of change**
Logging makes more liberal use of `field` functionality, with field keys stored in telemetry package alongside metric labels/tags (and often using the same).
Reduce usage of string formatting for logs; string formatting usually implies a field to be used.

**Which issue this PR fixes**
fixes #834
partially addresses #833


pkg/common/telemetry/names.go
```diff
@@ -184,7 +312,7 @@ const (
 	JoinToken = "join_token"
 
 	// JWTKey functionality related to a JWT key; should be used with other tags
-	// to add clarity
+	// to add clarity. Should NEVER actually provide the key itself, use Key ID instead.
```
evan2645: Hmm... Should we take this chance to change calls to this to use Kid instead? This updated comment has got me worried about future misuse, and I don't see any value in keeping `jwt_key` when we have `kid` already and the value of the latter is really what we want anyway

pkg/common/catalog/init.go
```diff
@@ -5,6 +5,8 @@ import (
 	"sort"
 	"sync"
 
+	"github.com/spiffe/spire/pkg/common/telemetry"
```
evan2645: This import is out of order in many places. Normally I'd turn a blind eye but there's about a dozen occurrences in this PR so I felt compelled to mention it

amoore877: oops :)

pkg/common/telemetry/names.go
amoore877: so the only 2 places this is used are in ca_manager.go, declaring "jwt_key" "prepare" and "jwt_key" "activate"; so it is already used safely and only to declare "some activity involving a jwt key"

evan2645: Yes understand, it is the "future misuse" part I was concerned about... the "don't leave knives laying around" rule. In the end though I won't block on it since it's the current behavior.

Thanks for fixing the imports!

PR #948, opened 2019-06-06
Update plugin_server_noderesolver_aws_iid.md

mbarrin:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
The docs for the [AWS IID Node Resolver](https://github.com/spiffe/spire/blob/0.8.0/doc/plugin_server_noderesolver_aws_iid.md) state that the new selector exposed is `iamrole:arn:aws:iam::123456789012:instance-profile/Blog` when in actual fact it is `iamrole:arn:aws:iam::123456789012:role/Blog`

The code can be seen in the following link:
https://github.com/spiffe/spire/blob/master/pkg/server/plugin/noderesolver/aws/iid.go#L256-L267

However the name of the method is arguably still misnamed, as it is returning the role, but it is called `resolveInstanceProfile`

**Description of change**
Updated the `plugin_server_noderesolver_aws_iid.md` file to list `role` instead of `instance-profile`


evan2645:
Oh jeez, thank you for this @mbarrin, sorry you had to track that down!

Looking at the code you linked, it seems that the problematic prefix is coming straight from AWS as part of the ARN? Any idea in what cases the ARN might have the `instance-profile` prefix? Makes me wonder if perhaps the docs should include an template instead (e.g. `aws_iid:iamrole:ROLE_ARN_HERE`), perhaps instead of or in addition to the example?

mbarrin:
So from reading the code I assumed you could have multiple roles associated, but I'd only ever seen a 1:1 mapping.

I looked at the [docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html) based on your question and found.

> An instance profile can contain only one IAM role. This limit cannot be increased. 

Although maybe a role can be part of multiple instance-profiles?

> Amazon EC2 uses an instance profile as a container for an IAM role. When you create an IAM role using the IAM console, the console creates an instance profile automatically and gives it the same name as the role to which it corresponds. If you use the Amazon EC2 console to launch an instance with an IAM role or to attach an IAM role to an instance, you choose the role based on a list of instance profile names. 

So the role name and the instance-profile name don't have to be the same.

So maybe it makes sense to allow both instance-profile and role as different selectors?

But as it is the current code implies it returns the instance-profile when it is returning a role.

I do think that maybe: https://github.com/spiffe/spire/blob/master/pkg/server/plugin/noderesolver/aws/iid.go#L170-L181

```golang
if instance.IamInstanceProfile != nil && instance.IamInstanceProfile.Arn != nil {
	instanceProfileName, err := instanceProfileNameFromArn(*instance.IamInstanceProfile.Arn)
	if err != nil {
		return nil, err
	}
	output, err := client.GetInstanceProfileWithContext(ctx, &iam.GetInstanceProfileInput{
		InstanceProfileName: aws.String(instanceProfileName),
	})
	if err != nil {
		return nil, iidError.Wrap(err)
	}
	addSelectors(resolveInstanceProfile(output.InstanceProfile))
}
```

with

```golang
if instance.IamInstanceProfile.Arn != nil {
        addSelectors(instance.IamInstanceProfile.Arn)
}
```

And then the docs don't need to be changed at all.

Interested to hear your thoughts.

mbarrin:
Also other than the method name, it definitely seems like the code and docs are meant to return the role.

evan2645:
Thanks for the insight @mbarrin. I've dug into this a little bit further, and judging by the commit history, I think this is probably a simple documentation error as you initially suggested. At first I thought that the person who wrote the resolver somehow received an instance profile arn, however it looks like the docs and the resolver were written by different people so the most likely thing is that we simply used the wrong arn type when writing the docs.

The methods are definitely misnamed, which likely contributed to the doc flub.

> So maybe it makes sense to allow both instance-profile and role as different selectors?

I think that makes sense! But that doesn't have to be part of this PR - I will accept this change as-is. If you'd like to clean this bit up a little bit (rename the methods etc) and provide the additional selector, those changes would be welcome!

Thanks very much for this @mbarrin and sorry again for you having to dig around on it.

Issue #907, opened 2019-05-10
Document UpstreamCA spire plugin


evan2645:
We have a plugin named `UpstreamCA "spire"` which we recently enabled but forgot to document. Add a configuration guide in `/doc` like the rest of the plugins have

PR #955, opened 2019-06-10
Add document of Server plugin UpstreamCA "spire"

ryysud:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Nothing.

**Description of change**

Referring to [the source code](https://github.com/spiffe/spire/blob/520f43b95c0a6f9640c87b6ae78f464c7b26da2e/pkg/server/plugin/upstreamca/spire/spire.go) and PRs (https://github.com/spiffe/spire/pull/643 , https://github.com/spiffe/spire/pull/890),
I add a document of `Server plugin: UpstreamCA "spire"` .

**Which issue this PR fixes**

This PR fixes https://github.com/spiffe/spire/issues/907 .


doc/plugin_server_upstreamca_spire.md
```diff
@@ -0,0 +1,24 @@
+# Server plugin: UpstreamCA "spire"
+
+The `spire` plugin loads root CA credentials from upstream SPIRE server in the same trust domain, using
+them to generate intermediate signing certificates for the server's signing authority.
```
evan2645: This kind of makes it sound like the plugin receives the root key from upstream SPIRE servers, which is not the case. Perhaps something like

> The `spire` plugin uses credentials fetched from the Workload API to call an upstream SPIRE server in the same trust domain, requesting an intermediate signing certificate to use as the server's X.509 signing authority.

would make that clearer?

doc/spire_server.md
```diff
@@ -34,6 +34,7 @@ This document is a configuration reference for SPIRE Server. It includes informa
 | Notifier   | [k8sbundle](/doc/plugin_server_notifier_k8sbundle.md) | A notifier that pushes the latest trust bundle contents into a Kubernetes ConfigMap. |
 | UpstreamCA | [disk](/doc/plugin_server_upstreamca_disk.md) | Uses a CA loaded from disk to sign SPIRE server intermediate certificates. |
 | UpstreamCA | [awssecret](/doc/plugin_server_upstreamca_awssecret.md) | Uses a CA loaded from AWS SecretsManager to sign SPIRE server intermediate certificates. |
+| UpstreamCA | [spire](/doc/plugin_server_upstreamca_spire.md) | Uses a CA loaded from upstream SPIRE server in the same trust domain to sign SPIRE server intermediate certificates. |
```
evan2645: Similar comment here: `Uses an upstream SPIRE server in the same trust domain to obtain intermediate signing certificates for SPIRE server`

ryysud:
I didn't understand the specification of UpstreamCA "spire" completely ...
I also think that it is correct to be able to convey the following intentions to spire users.
> The intention is to make it clear that 1) this plugin uses a local workload api to retrieve client credentials, and 2) that the upstream SPIRE server does not divulge any private keys to downstream servers.

I fix it with https://github.com/spiffe/spire/pull/955/commits/0c94beb02119ad1f0b963f439d1586d1922cc094 .

PR #975, opened 2019-06-20
shellcheck-ify build.sh

ZymoticB:
Semi-blindly update build.sh to adhere to shellcheck.

Double quotes galore, never assign and declare on the same line.

I tried to make as much sense of 
```
echo "export PATH=$(ls -d ${BUILD_DIR}/*/bin 2>/dev/null | tr '\n' ':'):${_gr}/bin:${_gp}/bin:${PATH}"
```
as I could. On OSX this depends on a shell option being set, on master the `ls` actually returns nothing but I'm "saved" by the fact that ${_gr} == ${BUILD_DIR}. Additionally, the tr replacement was making it so that on platforms where the `ls` did work, there was a `::` in the path which adds CWD to the path. I'm assuming this isn't intended as it's considered a Bad Idea (TM). I believe I've captured the spirit of what the subcommand was doing.

Other potentially controversial things are the read commands. shellcheck has opinions about making arrarys that seem to boil down to "don't trust word splitting".


ZymoticB:
I'm finished my yak shave of the proto stuff, this is ready to be reviewed.

build.sh
```diff
@@ -3,10 +3,14 @@
 set -o errexit
 [[ -n $DEBUG ]] && set -o xtrace
 
-declare -r ARTIFACT_DIRS="$(find cmd/* functional/* -maxdepth 0 -type d 2>/dev/null)"
-declare -r RELEASE_DIRS="$(find cmd/* -maxdepth 0 -type d 2>/dev/null)"
-declare -r SOURCE_PKGS="$(go list ./cmd/... ./pkg/... 2>/dev/null)"
-declare -r RELEASE_FILES="LICENSE README.md conf"
+ARTIFACT_DIRS="$(find cmd/* functional/* -maxdepth 0 -type d 2>/dev/null)"
```
drrt: Per https://github.com/spiffe/spiffe/blob/master/CONTRIBUTING.md we try to follow google's shell style guide, https://google.github.io/styleguide/shell.xml?showone=Read-only_Variables#Read-only_Variables

build.sh
```diff
@@ -59,41 +63,41 @@ build_env() {
 	_gr="${GOROOT:-$BUILD_DIR}"
 	echo "export GOPATH=${_gp}"
 	echo "export GOROOT=${_gr}"
-	echo "export PATH=$(ls -d ${BUILD_DIR}/*/bin 2>/dev/null | tr '\n' ':'):${_gr}/bin:${_gp}/bin:${PATH}"
+	echo "export PATH=$(test -d "${BUILD_DIR}/bin" && echo "${BUILD_DIR}/bin:")${_gr}/bin:${_gp}/bin:${PATH}"
```
drrt: There used to be multiple and changing bin directories, hence the ugliness

build.sh
```diff
@@ -127,13 +134,15 @@ build_protobuf() {
 ## Create a private copy of generated code and compare it to
 ## what's checked in. Fail if there's a difference
 build_protobuf_verify() {
-	local _n _result _tmp="$(mktemp -d)"
-	eval $(build_env)
+	local _n _result _tmp
+    _tmp="$(mktemp -d)"
```
drrt: indentation?

build.sh
```diff
@@ -41,10 +45,10 @@ _exit_error() { echo "ERROR: $*" 1>&2; exit 1; }
 _log_info() { echo "INFO: $*"; }
 
 _fetch_url() {
-	mkdir -p ${BUILD_CACHE}
+	mkdir -p "${BUILD_CACHE}"
```
drrt: And ya per our own instructions, hella quotes were missing everywhere 👍

build.sh
drrt: Oh and i should add that passing shellcheck is also a fine idea, just need to update the CONTRIBUTING doc

ZymoticB: The google shell style guide doesn't seem to require they are declared and assigned at the same time, in fact the example set the variable readonly after assignment *and* error checking.

Are you asking to move back to the single line version?

build.sh
ZymoticB: Are you happy with this? Is there a better version of this now?

build.sh
drrt: Oh linked to the wrong part of the guide, this is the right link https://google.github.io/styleguide/shell.xml?showone=Constants_and_Environment_Variable_Names#Constants_and_Environment_Variable_Names

And you're right, the guide really isn't consistent. Comment retracted! shellcheck FTW!

build.sh
ZymoticB: Which instructions? The google shell style guide also seems to suggest pretty aggressive quoting
```
Always quote strings containing variables, command substitutions, spaces or shell meta characters, unless careful unquoted expansion is required.
Prefer quoting strings that are "words" (as opposed to command options or path names).
Never quote literal integers.
Be aware of the quoting rules for pattern matches in [[.
Use "$@" unless you have a specific reason to use $*.
```

Just want to make sure there aren't shellcheck rules you would prefer ignored. It would also be great to have shellcheck run in travis. I could add that now, or as a follow up if you prefer.

drrt: Was just noting that this script didn't follow the google style guide on quoting - thanks for adding them all

ZymoticB: 👍

PR #979, opened 2019-06-25
Return GRPC NotFound from datastore plugin where appropriate

### LINKED ISSUE #978: API returns record not found as generic error
I have hit a case where I called `EvictAgent` on a node that was not longer listed in `attested_node_entries` and the error I got was `rpc error: code = Unknown desc = datastore-sql: record not found`. I have only tested this with `datastore/sql` plugin, so not sure if that's also happening with other data stores (a quick code grep suggests the errors are just being passed through, though).

It would be great if I could distinguish between _something went wrong_ and _record not found_ without parsing the error message. IMHO, the API should return appropriate gRPC status code.

ZymoticB:
Fixes #978


test/fakes/fakedatastore/fakedatastore.go
```diff
@@ -19,20 +18,22 @@ import (
...
 
 var (
-	ErrBundleAlreadyExists       = errors.New("bundle already exists")
-	ErrNoSuchBundle              = errors.New("no such bundle")
-	ErrAttestedNodeAlreadyExists = errors.New("attested node entry already exists")
-	ErrNoSuchAttestedNode        = errors.New("no such attested node entry")
-	ErrNoSuchRegistrationEntry   = errors.New("no such registration entry")
-	ErrNoSuchToken               = errors.New("no such token")
-	ErrTokenAlreadyExists        = errors.New("token already exists")
+	ErrBundleAlreadyExists       = status.Error(codes.NotFound, "bundle already exists")
```
azdagron: these "exists" errors seem more like codes.PreconditionFailed than codes.NotFound

pkg/server/plugin/datastore/sql/sql_test.go
```diff
@@ -119,6 +121,18 @@ func (s *PluginSuite) TestBundleCRUD() {
 	s.Require().NotNil(fresp)
 	s.Require().Nil(fresp.Bundle)
 
+	// update non-existant
+	_, err = s.ds.UpdateBundle(ctx, &datastore.UpdateBundleRequest{Bundle: bundle})
+	s.Require().Error(err)
```
azdagron: nit: `spiretest.RequireGRPCStatus` might shrink some of this assertion code (or `s.RequireGRPCStatus()` considering PluginSuite is a `spiretest.Suite`).
```
_, err = s.ds.UpdateBundle(ctx, &datastore.UpdateBundleRequest{Bundle: bundle})
s.RequireGRPCStatus(err, codes.NotFound, "bundle not found")
```

pkg/server/plugin/datastore/sql/sql.go
```diff
@@ -499,6 +501,22 @@ func (ds *SQLPlugin) withTx(ctx context.Context, op func(tx *gorm.DB) error, rea
 	return sqlError.Wrap(tx.Commit().Error)
 }
 
+// gormToGRPCStatus takes an error, and converts it to a GRPC error.
+// If the error is a gorm error type with a known mapping to a GRPC
+// status, that code will be set, otherwise the code will be set to
+// Unknown.
+func (ds *SQLPlugin) gormToGRPCStatus(err error) error {
```
azdagron: nit: consider making this a standalone helper function since it requires no state from `ds`

test/fakes/fakedatastore/fakedatastore.go
ZymoticB: Maybe? Maybe I should do a closer pass as some of these since the real version returns zero values rather than errors in some NotFound cases.

I haven't read the code that's making use of the fake but it seems like there is use in it returning NotFound in the same cases as the real version.

pkg/server/plugin/datastore/sql/sql_test.go
ZymoticB: oh hey that already exists. I'm not used to the Suite style of things :)

test/fakes/fakedatastore/fakedatastore.go
azdagron: The fake should definitely be consistent with the SQL implementation for sure. If it's not, that needs to get fixed.

ZymoticB: Hmm, I was very liberal with this status code, not sure how it escaped me that "already exists" != "not found" :). Fixing.

ZymoticB: I've updated this, in some ways this is now inconsistent, and I'm not converting to FailedPrecondition in the real implementation. Would you like me to do that now?

azdagron: I'd be ok with either not having the fake return FailedPrecondition (yet) or updating the real implementation to return it in the right circumstances, as long as we remain consistent.

azdagron:
Whoops, @ZymoticB. Looks like 04ea286 didn't have DCO. I clicked "Update Branch" before noticing... feel free to fix the commit, rebase (or merge), and force push. Sorry about making it more complicated to fix 😬

ZymoticB:
I believe that fixes it

ZymoticB:
image

azdagron:
😆 ... not sure if you saw the recent Travis CI status page or not...

PR #982, opened 2019-06-26
Fix task stat parsing on Linux to handle task names containing spaces.

skaslev:
The `peertracker` module on Linux doesn't parse task stat correctly given a process with task name containing spaces. This PR addresses this issue by relying on the task name being parenthesized instead of simply splitting stat data on spaces.


ZymoticB:
While you're here, could you please add a unit test for this.

skaslev:
> While you're here, could you please add a unit test for this.

Good point, thanks for the feedback. I added a couple of tests for `parseTaskStat`. Let me know what you think.

PR #987, opened 2019-06-28
Run travis tests on multiple platforms

ZymoticB:

There are some opinions in here about what should be cached to be shared across stages. Additionally, it separates some stuff out of the "release" stage.

I think I could make this change much smaller but I thought separating the stages made some sense and this gives consistency across platform where possible. For example, the test job for macos is the same as linux If the test logic was still in the "release" stage, that stage would differ across platform, rather than not existing for certain platforms.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 sudo: required
-dist: xenial
 
 services:
   - docker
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,24 +8,54 @@ language: go
...
+    - stage: test
+      script:
+        - ./build.sh race-test
+      os: linux
+      dist: xenial
     - stage: build release
       script:
-        - ./build.sh setup
-        - ./build.sh utils
         - ./build.sh protobuf_verify
```
evan2645: I wonder if we should move this into the lint stage?

ZymoticB: great idea.

PR #1007, opened 2019-07-14
datastore: Add indexes to registration_entries and selectors tables

APTy:
This change improves performance of some database operations by adding
a few indexes:
 - `registration_entries` by `parent_id`
 - `registration_entries` by `spiffe_id`
 - `selectors` by `(type, value)`

Currently, the Registration API's [CreateEntry()](https://github.com/spiffe/spire/blob/5951ad20f267fe709c9b91b49d22cb37aea06346/pkg/server/endpoints/registration/handler.go#L46) call checks whether
the entry already exists by querying the registration_entries table
by spiffe_id. Without an index, this results in a full table scan
for each newly created entry. By adding an index, this query becomes
much more efficient.

Listing entries also requires a lookup of selectors by (type, value)
tuple. Without a proper index, this also results in a full table scan.
This commit adds an index to the selectors table for this reason.

This commit also improves the behavior of various callsites where
entries are queried by parent_id (such as regentryutil).

### More information
Sqlite's `EXPLAIN` statement shows the benefit of indexing the `selectors` table:
##### Before Index
```sql
sqlite> EXPLAIN QUERY PLAN SELECT * FROM selectors WHERE type = 'test' AND value = 'foo:bar';
QUERY PLAN
`--SCAN TABLE selectors
```

##### After Index
```sql
sqlite> CREATE INDEX idx_selectors_type_value ON selectors (type, value);
sqlite> EXPLAIN QUERY PLAN SELECT * FROM selectors WHERE type = 'test' AND value = 'foo:bar';
QUERY PLAN
`--SEARCH TABLE selectors USING INDEX idx_selectors_type_value (type=? AND value=?)
```

As well as for the `registration_entries` table:
##### Before Index
```sql
sqlite> EXPLAIN QUERY PLAN SELECT * FROM registered_entries WHERE spiffe_id = 'spiffe://example.org/host/foo';
QUERY PLAN
`--SCAN TABLE registered_entries
```

##### After Index
```sql
sqlite> CREATE INDEX idx_registered_entries_spiffe_id ON registered_entries (spiffe_id);
sqlite> EXPLAIN QUERY PLAN SELECT * FROM registered_entries WHERE spiffe_id = 'spiffe://example.org/host/foo';
QUERY PLAN
`--SEARCH TABLE registered_entries USING INDEX idx_registered_entries_spiffe_id (spiffe_id=?)
```


pkg/server/plugin/datastore/sql/sql_test.go
```diff
@@ -1551,6 +1551,8 @@ func (s *PluginSuite) TestMigration() {
 			s.Require().Len(resp.Entries, 1)
 			s.Require().Len(resp.Entries[0].DnsNames, 1)
 			s.Require().Equal("abcd.efg", resp.Entries[0].DnsNames[0])
+		case 8:
+			// todo
```
APTy: any ideas on how to validate this? the [gorm.Dialect](https://godoc.org/github.com/jinzhu/gorm#Dialect) object has a `HasIndex()` call, but we don't have access to the underling `gorm.DB` object, since its across the plugin boundary

PR #1041, opened 2019-07-31
Run `golangci-lint --enable-all --fix` to tidy up

mcpherrinm:
This fixes a number of small issues, around spelling, codestyle, and
formatting.

No functionality should be affected.  There's one case where an environment variable name is changed to be spelled properly, but I don't see any use of it in this repo or docs, so I think it's just for manual testing.


mcpherrinm:
There's a standalone https://golangci.com/ that's a bit nicer than Travis because it can comment feedback directly on pull requests too. Free for OSS projects.

PR #1045, opened 2019-08-02
[datastore/sql] modify fixtures

ZymoticB:
This removes a dependency on fixtures read relative to the "project root"
and either creates the data in memory, or depends on a local testdata directory


ZymoticB:
Looking into the travis failure.

ZymoticB:
`test/util.GetRegistrationEntries` doesn't check errors from `json.Unmarshal`. The `federates_with` field in `test/fixture/registration/entries.json` was being silently dropped.

I fixed it so that it would unmarshal but then that requires that the bundle exists in the DB. Rather than change the test logic, I've dropped the field from the fixture.

PR #1049, opened 2019-08-05
sanitize labels in metrics

### LINKED ISSUE #1039: sanitize SPIFFE ID in metrics
Right now when used as a label value the SPIFFE ID is printed as is, but that includes `.`'s. This is generally not good for metrics querying which tends to use `.` as a delimiter.

Sanitize this in the `AddSPIFFEID` portion of telemetry package

amoore877:

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
All telemetry calls, specifically in labels.
`IncrValidJWTSVIDErrCounter` label removal.
`fakemetrics` calls.

**Description of change**
All telemetry labels are sanitized, replacing periods with underscores.
`IncrValidJWTSVIDErrCounter` metric no longer has a full error string as a label value.
`fakemetrics` sanitizes labels.
This does not address that fact that the mock metrics will not sanitize labels, but that only affects tests. Some future effort can force mock metrics to be hidden behind a sanitizing wrapper.

**Which issue this PR fixes**
fixes #1039


amoore877:
Much like #1025, this may break people expecting non-sanitized, however **not** fixing this makes dot-based delimiting (the most common from my knowledge) of metrics, especially in `statsd`, very ugly and unusable.

pkg/common/telemetry/agent/manager.go
```diff
@@ -20,10 +23,7 @@ func StartManagerFetchUpdatesCall(m telemetry.Metrics) *telemetry.CallCounter {
 // Takes registration entry SPIFFE ID
 func IncrRegistrationEntryCreatedCounter(m telemetry.Metrics, id string) {
 	labels := []telemetry.Label{
-		{
-			Name:  telemetry.SPIFFEID,
-			Value: id,
-		},
+		common.GetSanitizedLabel(telemetry.SPIFFEID, id),
```
evan2645: Sanitizing strings here is definitely better than requiring the caller of `IncrRegistrationEntryCreatedCounter` to do it, but I think it's best done as a wrapper around the *WithLabels calls to the metrics implementation. Feels like it would be pretty easy to regress on this as is. Any reason _not_ to apply this sanitizing transform to all labels?

pkg/common/telemetry/common/sanitize.go
```diff
@@ -0,0 +1,22 @@
+package common
```
evan2645: This would probably be better put in `pkg/common/telemetry/labelutils` or something similar. The `common` stutter isn't great and the name `telemetry/common` in the presence of other packages like `telemetry/agent` sends signals the package contains metrics that are common to both server and agent, which is not the case here.

If we wrap the *WithLabels functions then this point is moot - these could be defined as unexported functions on the MetricsImpl struct

pkg/common/telemetry/agent/manager.go
amoore877: `WithLabels` is only being called in one space, by SDS endpoint.

I agree it would be much better to have some kind of wrapping that just ensures all labels always get sanitized, but `WithLabels` seems to not be the place, especially since labels can also be added via `CallCounter` `AddLabel`.
I know that spire allows multiple telemetry plugins active at once, so that must mean there is some delegate object out there that fulfills the Metrics API and calls each of the plugins? Is that `WithLabels` and I just didn't search correctly?

amoore877: hmm seems such an object is `MetricsImpl`, with the desired API being imported from a hashicorp lib object. So don't think this could be done there.
Still not sure how `withLabels` object fits into this, unless you mean that we should propagate its usage beyond the SDS endpoint call?

amoore877: spoke offline for clarity, will fix in next push

pkg/common/telemetry/common/sanitize.go
amoore877: spoke offline for clarity, will fix in next push

amoore877:
need to refactor per @evan2645 's comments

amoore877:
had to force push in order to signoff commit. apologies

pkg/common/telemetry/metrics.go
```diff
@@ -77,6 +81,7 @@ func NewMetrics(c *MetricsConfig) (*MetricsImpl, error) {
 	return impl, nil
 }
 
+// ListenAndServe starts the metrics process
```
evan2645: Thank you for these comments

test/fakes/fakemetrics/fakemetrics.go
```diff
@@ -56,7 +56,12 @@ func (m *FakeMetrics) SetGauge(key []string, val float32) {
 func (m *FakeMetrics) SetGaugeWithLabels(key []string, val float32, labels []telemetry.Label) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	m.metrics = append(m.metrics, MetricItem{Type: SetGaugeWithLabelsType, Key: key, Val: val, Labels: labels})
+	m.metrics = append(m.metrics, MetricItem{
+		Type:   SetGaugeWithLabelsType,
+		Key:    key,
+		Val:    val,
+		Labels: telemetry.GetSanitizedLabels(labels),
```
evan2645: One last (small) comment: this (and the other call sites) would be a little easier to read and more idiomatic if this function was named `telemetry.SanitizeLabels` instead

Everything else here looks great

PR #1058, opened 2019-08-08
Adding logging wherever metrics are emitted

### LINKED ISSUE #831: Logging alongside metrics
Sub-issue of #799 

At minimum, wherever a metric is emitted, a log must be as well. Ideally logs should have all relevant information for that instance in time via fields, as long as the information is not a security/privacy risk (ex. private keys).
This protects against minor possibility of either metrics or logging being down, and also allows logs to do their job of providing more context to events than metrics generally do.

This is potentially a medium to large task given it applies to all metrics, so it may be appropriate to break down into further sub-issues for each metric or group of metrics.

rturner3:
Signed-off-by: Ryan Turner <turner@uber.com>

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Should not directly affect any functionality.

**Description of change**
Anywhere metrics are emitted, new log statements have been added where not already present.

**Which issue this PR fixes**
fixes #831

### DIFF
--- a/pkg/agent/endpoints/sds/handler.go
+++ b/pkg/agent/endpoints/sds/handler.go
@@ -224,9 +224,11 @@ func (h *Handler) startCall(ctx context.Context) (int32, []*common.Selector, fun
 		return 0, nil, nil, status.Errorf(codes.Internal, "Is this a supported system? Please report this bug: %v", err)
 	}
 
-	metrics := telemetry.WithLabels(h.c.Metrics, []telemetry.Label{{Name: telemetry.SDSPID, Value: fmt.Sprint(watcher.PID())}})
+	pidStr := fmt.Sprint(watcher.PID())
+	metrics := telemetry.WithLabels(h.c.Metrics, []telemetry.Label{{Name: telemetry.SDSPID, Value: fmt.Sprint(pidStr)}})
 	telemetry_agent.IncrSDSAPIConnectionCounter(metrics)
 	telemetry_agent.IncrSDSAPIConnectionTotalCounter(metrics)
+	h.c.Log.WithField(telemetry.SDSPID, pidStr).Debug("Handling SDS API request")
 
 	selectors := h.c.Attestor.Attest(ctx, watcher.PID())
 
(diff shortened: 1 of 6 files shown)


amoore877:
when re-opened:
-alter title and "fixes" desc, unless these are the only places in the repo where metrics were being sent without an associated log

pkg/agent/endpoints/sds/handler.go
```diff
@@ -224,9 +224,11 @@ func (h *Handler) startCall(ctx context.Context) (int32, []*common.Selector, fun
 		return 0, nil, nil, status.Errorf(codes.Internal, "Is this a supported system? Please report this bug: %v", err)
 	}
 
-	metrics := telemetry.WithLabels(h.c.Metrics, []telemetry.Label{{Name: telemetry.SDSPID, Value: fmt.Sprint(watcher.PID())}})
+	pidStr := fmt.Sprint(watcher.PID())
```
amoore877: using `fmt.Sprint` twice (here and in telemetry label)

may also be more consistent to use `strconv` library

amoore877: also just assign a function-scope logger with the field and carry that through the rest-
```
log := h.c.Log.WithField(telemetry.SDSPID, pidStr)
...
log.Debug("Handling...")
...
log.Debug("Finishing...")
```

following this sort of pattern, a logger within a function can pick up useful field information as the logic continues, rather than re-stating all fields in every call

pkg/agent/endpoints/workload/handler.go
```diff
@@ -130,6 +136,11 @@ func (h *Handler) FetchJWTBundles(req *workload.JWTBundlesRequest, stream worklo
 					telemetry.Seconds: time.Since(start).Seconds,
 					telemetry.PID:     pid,
 				}).Warn("Took >1 second to send JWT bundle to PID")
+			} else {
+				h.Log.WithFields(logrus.Fields{
```
amoore877: with my comment above about "picking up fields as we go", be careful of pitfalls like in this situation; here we would **not** want to keep adding TTL and PID fields for eternity :)

PR #1061, opened 2019-08-13
[spire-agent] use roundrobin grpc balancer

ZymoticB:
This change will ensure that agent->server communication
using the roundrobin grpc balancer. If the server address
resolves to more than one record this will ensure that
each agent will balance its traffic across all available
servers. If the server address resolves to a single record
the agent will continue sending all traffic to that server.

We need to force the grpc target to have a resolver scheme
since passthrough (the default resolver) doesn't seem to use
anything other than pick_first.

### DIFF
--- a/cmd/spire-agent/cli/run/run.go
+++ b/cmd/spire-agent/cli/run/run.go
@@ -214,7 +214,8 @@ func newAgentConfig(c *config) (*agent.Config, error) {
 		return nil, err
 	}
 
-	ac.ServerAddress = net.JoinHostPort(c.Agent.ServerAddress, strconv.Itoa(c.Agent.ServerPort))
+	serverHostPort := net.JoinHostPort(c.Agent.ServerAddress, strconv.Itoa(c.Agent.ServerPort))
+	ac.ServerAddress = fmt.Sprintf("dns:///%s", serverHostPort)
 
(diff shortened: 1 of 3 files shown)


cmd/spire-agent/cli/run/run.go
```diff
@@ -214,7 +214,8 @@ func newAgentConfig(c *config) (*agent.Config, error) {
 		return nil, err
 	}
 
-	ac.ServerAddress = net.JoinHostPort(c.Agent.ServerAddress, strconv.Itoa(c.Agent.ServerPort))
+	serverHostPort := net.JoinHostPort(c.Agent.ServerAddress, strconv.Itoa(c.Agent.ServerPort))
+	ac.ServerAddress = fmt.Sprintf("dns:///%s", serverHostPort)
```
evan2645: For posterity, from offline conversation

> Will gRPC do the right thing if an IP address is provided with the scheme set to DNS as it is here?

Yes, it will.

https://github.com/grpc/grpc-go/blob/ee21c923a2d1b8c5aa62dea4ce93c2fecac5e687/resolver/dns/dns_resolver.go#L113

> Will gRPC periodically re-resolve this address?

Yes.

https://github.com/grpc/grpc-go/blob/ee21c923a2d1b8c5aa62dea4ce93c2fecac5e687/resolver/dns/dns_resolver.go#L105

PR #1078, opened 2019-08-20
add logic for negative workload_size_limit to disable unix sha function

### LINKED ISSUE #1076: Config option to disable unix workload attestor hashing
Hashing of a file can be computationally expensive for a single large file and in aggregate for many files at scale, and is a potential vector for DoS.
Add config option to unix workload attestor to disable the hashing.

amoore877:

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Agent Unix Workload Attestor plugin SHA calculation.

**Description of change**
Add ability to disable agent Unix workload attestor SHA calculation if workload_size_limit is set to negative, since 0 implies not set implies no size limit.

The alternative of no change + simply setting a very low value for size limit was tried out. It still results in large amounts of processing in aggregate, and is still not resistant to a DoS as noted in the plugin's doc.

**Which issue this PR fixes**
fixes #1076

### DIFF
--- a/doc/plugin_agent_workloadattestor_unix.md
+++ b/doc/plugin_agent_workloadattestor_unix.md
@@ -5,7 +5,7 @@ The `unix` plugin generates unix-based selectors for workloads calling the agent
 | Configuration | Description | Default |
 | ------------- | ----------- | ------- |
 | `discover_workload_path` | If true, the workload path will be discovered by the plugin and used to provide additional selectors | false |
-| `workload_size_limit` | The limit of workload binary sizes when calculating certain selectors (e.g. sha256). If zero, no limit is enforced. | 0 |
+| `workload_size_limit` | The limit of workload binary sizes when calculating certain selectors (e.g. sha256). If zero, no limit is enforced. If negative, never calculate the hash. | 0 |
 
 If configured with `discover_workload_path = true`, the plugin will discover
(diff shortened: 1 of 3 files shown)


ZymoticB:
I think it's worth mentioning in the documentation that in an environment with a mixture of unix and containerized workloads (e.g. has both the docker attestor and unix attestor configured) the unix attestor will currently always fail for docker workloads since the /proc/[pid]/exe symlink is broken (or potentially points to the same file outside the container which has no guarantee of being the same!). This will spam your logs and also prevent doing unix user based attestation of workloads in containers.

evan2645:
@amoore877 I assume this means you want the attestor to expose path selectors but not shasum?

This is probably a good opportunity to review the defaults of this attestor and how it is configured, in general. For instance, perhaps we can provide a better experience by detecting whether we have the correct permissions or not by attempting a read

ZymoticB:
I can see unix:user and unix:group (and their id versions) being useful for workloads that are in chroots, the path selector seems a bit iffy, and we don't actually forsee using the SHA selector at all so we'd like to have the ability to opt out of the CPU cost.

The fact that it is also non-functional for chroot-ed workloads and spams the logs leads us to desire the ability to disable it even more since we attest containerized and non-containerized workloads from the same agent.

PR #1085, opened 2019-08-26
cull unnecessary, redundant, or high-cardinality metric labels

amoore877:

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Emitted metrics throughout SPIRE.

**Description of change**
Cull metrics labels that are either:
- Unnecessary (provide no useful aggregate information for the purpose of metric reports)
- Difficult to interpret (selectors in particular are hard to interpret as part of a metric, especially with sanitization)
- Redundant (info is already provided either as part of metric, can be intuitively figured out, or is a logging field)
- High-cardinality (generally metrics systems have trouble with metrics that have a large set of potential values)

### DIFF
--- a/pkg/agent/endpoints/workload/handler.go
+++ b/pkg/agent/endpoints/workload/handler.go
@@ -260,7 +260,9 @@ func (h *Handler) sendX509SVIDResponse(update *cache.WorkloadUpdate, stream work
 
 	log = log.WithField(telemetry.Count, len(resp.Svids))
 
-	// Add all the SPIFFE IDs to the labels array.
+	// log and emit telemetry on each SVID
+	// a response has already been sent so nothing is
+	// blocked on this logic
 	for i, svid := range resp.Svids {
 		ttl := time.Until(update.Identities[i].SVID[0].NotAfter)
 		telemetry_workload.SetFetchX509SVIDTTLGauge(metrics, svid.SpiffeId, float32(ttl.Seconds()))
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,6 +37,8 @@ A Makefile is provided for common actions.
 * `clean` - cleans `vendor` directory
 * `distclean` - removes caches in addition to `make clean`
 * `utils` - installs gRPC related development utilities
+* `protobuf` - regenerates the gRPC pb.go and README_pb.md files
+* `protobuf_verify` - checks that the checked-in generated code is up-to-date
 * `help` - shows makefile targets and description
 
 ## Development in Docker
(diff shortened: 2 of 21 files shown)


pkg/agent/endpoints/workload/handler.go
```diff
@@ -66,21 +66,16 @@ func (h *Handler) FetchJWTSVID(ctx context.Context, req *workload.JWTSVIDRequest
 	var spiffeIDs []string
 	identities := h.Manager.MatchingIdentities(selectors)
 	if len(identities) == 0 {
-		telemetry_common.AddRegistered(counter, false)
```
azdagron: what is the justification for the removal of the registered label? it doesn't seem to have high cardinality nor is it redundant. Seems useful to shed insight into understand situations where workloads aren't registered....

amoore877: - it clogs up non-tagging metric systems like statsd; we now have to search for (simplified example) `spire_agent.workload_api.fetch_jwt_svid.error.p99`,  `spire_agent.workload_api.fetch_jwt_svid.registered.false.error.p99`, `spire_agent.workload_api.fetch_jwt_svid.registered.true.error.p99`, `spire_agent.workload_api.fetch_jwt_svid.registered.true.p99`
This doesn't even get into the complexity added by `count` later
- if an error occurred, probably registration failed
- if errors are occurring, we should look to the logs for more details, not metrics (which should be more high-level for searchability and easier aggregation)

speaking of logging, I see there is only minimal logging here. I will amend that now

ZymoticB: Supporting non-tagging systems is indeed unfortunate as every "tag" you add is yet another `*` you need to add in many cases. Additionally, you need to be extremely consistent with the number of tags and the order of tags you add to all metrics or it becomes a nightmare to ensure you correctly query them all.

Further to this, p99s can't actually be aggregated, so every "tag" you add to a percentile metric the less accurate the overall "what is the p99 of this endpoint regardless of the result" gets.

I agree it seems interesting to know the difference between a registered and non-registered in a metric and I'd be open to keeping it, we would need to be careful about ordering, and ensuring it always exists.

Another note on timers and the tags associated with them, potentially this means that we should be treating counters separately from timers to avoid the aggregation concerns associated with percentiles.

Generally, support "all the things" is going to make a lot of this really restrictive and difficult to do in a performant way. We're passing around non-hashable types (string slices) so we can't even cache the strings we generate in a straight-forward way :(

`</rant>`

amoore877: as an example of what @ZymoticB is talking about, here is one of our Grafana graphs. You can see from the queries that we
- need 2 queries to get the same information (because of `error` key in this case)
- write a cumbersome query because of the `selectors` label (we probably never want to aggregate the selector count in metrics)
Screen Shot 2019-09-09 at 10 04 24 AM

metrics should be fairly general, especially when it comes to non-tagging telemetry systems. Logs can be more specific, giving more context to what is happening in the system.

below is an even more egregious example:
Screen Shot 2019-09-09 at 10 08 37 AM

ZymoticB: In the first example the issue is more that in one "position" in the metric we have both the list of hosts and the string "error". If instead that position was one of {"success", "error"} it would be better.

Having a count be a label isn't useful since it's not aggregatable. Emitting a second metric that is a count/size of the returned data, along side a metrics that is the number of calls would be useful.

amoore877:
> Sorry for the long turn around time on this. There is a small merge conflict to resolve but otherwise looks good.

all good; nice chatting with you today :)
after this PR I'll be putting up one more related to #1106, specifically for standardizing error vs success metrics (make error/status key non-optional). After that, non-tagging systems should be in a much better place.

Issue #1083, opened 2019-08-23
Safer DB migration strategy in spire server cluster env


amoore877:
Currently if SPIRE Server with DB version X starts up and looks to DB with version < X, SPIRE will auto-migrate the DB to X. Backwards compatibility does not exist.

Problems:
- In a distributed environment this is fatal to the other SPIRE Servers. They will be in a < X DB version and fail to start due to not being able to migrate backwards
- The migration is uncontrolled; in such an environment we not commit to every single Server must upgrade, and hope that the new DB schema is good

Some (non-mutually exclusive) Solutions:
- A flag to pass to Server config to **not** perform auto-migration; without other changes this is fatal to the newer server but lets the older ones continue operations
- Proper semantic versioning of the DB versions; when starting up and the DB is at a later version than the Server expects, if the new version does not indicate a breaking change then the old Server should be able to continue operations (not perform backwards migration)
- A method of rolling back the DB version
- A method of a later Server being backwards compatible for older DB versions

ZymoticB:
To start we should add a flag to disabled auto-migration. It will be a very quick change and will significantly increase maintainability.

A simple requirement which leads to this conclusion, a running build should *always* be able to restart, that is not currently true due to the automigration behaviour.

mbyczkowski:
I think the flag to disable auto-migration would be a good low-hanging fruit (and, incidentally, is already being tracked in #786).

In the future, it would be good to have migrations managed explicitly, through command that takes a list of migration files/statements as inputs (something similar to [Flyway](https://flywaydb.org/) in Java or migrations in Rails). This would still make development easy, but also would be helpful for people that use migration solutions that live outside of their apps, e.g. MySQL users would use [pt-online-schema-change](https://www.percona.com/doc/percona-toolkit/LATEST/pt-online-schema-change.html), [gh-ost](https://github.com/github/gh-ost) or similar tools.

evan2645:
This issue (and much more) is under active scrutiny as part of https://github.com/spiffe/spire/issues/896

We've had a few conversations around how to address this problem in particular. Some observations we've made thus far:
* A great deal of the pain is related to lack of discipline around introducing schema changes and subsequently using them.
  * Schema changes should be both forward and backwards compatible
  * This will move pain from operators to developers (as changes must be planned well in advance), however I don't think it is avoidable
  * Downgrades should be trivial with this approach
* It's not clear (to me yet, at least) that we actually need a facility to disable migration management in SPIRE
[...]

PR #1089, opened 2019-08-28
add disable_migration flag and more versioning to SQL plugin

### LINKED ISSUE #786: Option to disable auto-migration on start-up
Although current SPIRE auto-migration mechanism in DataStore SQL plugin is very convenient (especially for development and prototyping), it's not suitable for all use cases.

In a production environment, it's likely that database migrations are handled outside of apps. E.g. MySQL users would use [pt-online-schema-change](https://www.percona.com/doc/percona-toolkit/LATEST/pt-online-schema-change.html), [gh-ost](https://github.com/github/gh-ost) or similar tools.

It would be great if there was a `auto_migration:  true/false` flag in the config file, so auto-migration can be turned off when needed.

amoore877:

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
SQL plugin migration / configuration.
Small fix to migration test dump data.

**Description of change**
Addresses points in https://github.com/spiffe/spire/issues/1083#issuecomment-531012118 + offline discussions.
Add `disable_migration` flag and better version control to SQL plugin.
The flag disables migration of the datastore model, if versioning is 0.9.X or later (else error out). It allows more control over DB changes in a production and/or clustered environment, and prevents one newer versioned SPIRE server from crashing older ones; thus the running versions can always be safe and only the newer is potentially affected.
Adds `code_version` to the `migrations` table.
In order of precedence-
- If disable flag is set but DB code version is empty (pre-0.9.X), error out.
- If schema version for DB has not changed, update `migrations` table with current code version.
- If the DB major version is not 0, this is an unsupported version so error out (in the future, may change to be +/-1 major version).
- If disable flag is set, perform no migration.
- If current schema version is greater than the code's latest, log a warning but continue (we know from new `code_version` that we should be safe, barring improper semantic versioning)
- Else, run migrations per existing logic.

bug fix- Migration test dump data stopped updating `migrations` table data after version 4. Fixed the versioning for each dump set.

**Which issue this PR fixes**
Partially addresses, but **does not fix** #1083 (backwards migrations possibility?)
FIxes #786


amoore877:
An initial pass based on https://github.com/spiffe/spire/issues/1083#issuecomment-531012118 to verify that this is the expected direction, before delving further into any refactors / cleanups / docs

pkg/server/plugin/datastore/sql/migration.go
```diff
@@ -35,25 +53,72 @@ func migrateDB(db *gorm.DB, dbType string, log hclog.Logger) (err error) {
...
+	if dbCodeVersion.Major != 0 {
+		// DB schema versions do not match, and the Code versions are incompatible
+		err = sqlError.New("Code version %s is NOT compatible with DB code version %s", codeVersion.String(), dbCodeVersion.String())
 		log.Error(err.Error())
 		return err
 	}
 
-	if version == codeVersion {
+	if disableMigration {
+		return nil
```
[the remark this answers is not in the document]
amoore877: someday yes, but right now we are bumping to `0.9.0` and if the DB schema is from before that time *and* migration is disabled, we fail in a check above. We also will not be introducing breaking changes until 1.X at earliest, and that is covered by the code major version check.
Therefore I don't think there's anything to check today.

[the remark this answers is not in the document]
amoore877: I guess maybe the part I'm missing here is, given from our talks 1.X may have major changes making 0.X incompatible (hence my original `dbCodeVersion` check above)... is 1.X good or not for compatibility?

pkg/server/plugin/datastore/sql/migration.go
```diff
@@ -35,25 +53,72 @@ func migrateDB(db *gorm.DB, dbType string, log hclog.Logger) (err error) {
...
+	}
+
+	// at this point:
+	// - auto-migration is enabled
+	// - code version difference is not too large to support
+
+	if schemaVersion > latestSchemaVersion {
+		log.Warn(fmt.Sprintf("backwards DB schema migration not supported! (current=%d, latest=%d)", schemaVersion, latestSchemaVersion))
```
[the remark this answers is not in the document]
amoore877: so downgrade from warning to info? or no log at all?

CONTRIBUTING.md
```diff
@@ -83,6 +83,10 @@ In addition to the conventions covered in the SPIFFE project's
 [CONTRIBUTING](https://github.com/spiffe/spiffe/blob/master/CONTRIBUTING.md), the following
 conventions apply to the SPIRE repository:
 
+## SQL Plugin Changes
+
+Datastore changes must be present in at least one full major release cycle prior to introducing code changes that depend on them.
```
amoore877: currently does not match actual logic (checking minor versions)
waiting for @evan2645 for which direction the logic should be pre-1.0

amoore877:
dat travis error tho

pkg/server/plugin/datastore/sql/migration.go
```diff
@@ -27,6 +40,23 @@ func migrateDB(db *gorm.DB, dbType string, log hclog.Logger) (err error) {
 		return initDB(db, dbType, log)
 	}
 
+	// TODO related epic https://github.com/spiffe/spire/issues/1083
+	// We do not allow breaking changes in 0.X beyond version 0.9.0, but
+	// 1.0+ may have such changes. Thus iff Major version is 0, we are
+	// compatible. This logic will change in 1.0+.
+	// With 1.0 code, we will need a proper version check between code and db versions.
```
evan2645: The stance here is that we promise no breaking changes within any given +/- 1 code version window. For example, given version 0.9, we promise that you will be backwards compatible to 0.8 and forwards compatible to 0.10... 0.11 may introduce changes that break 0.9, but not 0.10, and so forth.

Of course, the comparison we make to determine this compatibility is different between pre-1.0 and post-1.0 versions, and we are stopping short of solving for the latter. So, I recommend we update this comment, something like this
```
// TODO related epic https://github.com/spiffe/spire/issues/1083
// The version comparison logic in this package is specific to pre-1.0 versioning semantics.
// It will need to be updated prior to releasing 1.0. Ensure that we're still building a pre-1.0
// version before continuing, and fail cowardly if we're not.
```

pkg/server/plugin/datastore/sql/migration.go
```diff
@@ -27,6 +40,23 @@ func migrateDB(db *gorm.DB, dbType string, log hclog.Logger) (err error) {
...
+		log.Error(err.Error())
+		return err
+	}
+
+	// TODO related epic https://github.com/spiffe/spire/issues/1083
+	// continue doing this automigration for now
+	// eventually SPIRE Code will be bumped to version 0.10, where we can
+	// expect that the `migrations` table exists and has the CodeVersion column,
+	// and then this can be removed. The below `Assign` check may be a new guard
+	// against pre-0.9 versions due to mismatching model
```
evan2645: I think we need to keep this here even post-0.10 because we will need it when starting a fresh cluster?

pkg/server/plugin/datastore/sql/migration.go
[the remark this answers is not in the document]
evan2645: My original comment here is now resolved, and I think your last message is addressed by my new comment at the top of this file... please let me know if it's still unclear

pkg/server/plugin/datastore/sql/migration.go
```diff
@@ -35,25 +65,81 @@ func migrateDB(db *gorm.DB, dbType string, log hclog.Logger) (err error) {
...
+		}
+		log.Warn(fmt.Sprintf("DB schema %d is ahead of code version %s with schema version %d, upgrading SPIRE Server is recommended",
+			schemaVersion, codeVersion.String(), latestSchemaVersion))
+		return nil
+	}
+
+	// at this point:
+	// - auto-migration is enabled
+	// - schema version of DB is behind
+	// - code version difference is not too large to support
```
evan2645: I think this comment line can be removed - code version could be 0 (if it is a fresh cluster). And while it will probably not happen often in practice, we could _technically_ still migrate you up from e.g. code version 0.9 to 0.42 in a safe fashion if we have reached this point.

pkg/server/plugin/datastore/sql/migration.go
```diff
@@ -27,6 +40,23 @@ func migrateDB(db *gorm.DB, dbType string, log hclog.Logger) (err error) {
...
 	}
 
+	// TODO related epic https://github.com/spiffe/spire/issues/1083
+	// We do not allow breaking changes in 0.X beyond version 0.9.0, but
+	// 1.0+ may have such changes. Thus iff Major version is 0, we are
+	// compatible. This logic will change in 1.0+.
+	// With 1.0 code, we will need a proper version check between code and db versions.
+	if codeVersion.Major != 0 {
+		err = sqlError.New("current migration code not compatible with %s release", codeVersion.String())
+		log.Error(err.Error())
```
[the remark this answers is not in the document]
amoore877: for your second comment, this is related to #1138; ideally as the error propagates up the stack it will be wrapped with details at each level, but each level of that stack doesn't have structured field details of the other levels, and so only they can provide such context and so they should log as well. Additionally a called module should not depend on knowing whether the caller will log its errors.

pkg/server/plugin/datastore/sql/migration.go
amoore877: true true

pkg/server/plugin/datastore/sql/migration.go
```diff
@@ -35,25 +65,81 @@ func migrateDB(db *gorm.DB, dbType string, log hclog.Logger) (err error) {
...
+	schemaVersion := migration.Version
+
+	var dbCodeVersion semver.Version
+	// Auto-migrate can only be safely disabled if 0.9.0 migrations have already been run
+	// We know if 0.9.0 migrations have been run because dbCodeVersion will be populated if so
+	// TODO: related epic https://github.com/spiffe/spire/issues/1083
+	// This check can be removed in 0.10.0
+	if migration.CodeVersion == "" {
+		if disableMigration {
+			err := sqlError.New("must upgrade from version %s or higher to disable auto-migration", minimumCodeVersionToDisableMigrate)
```
amoore877: `minimumCodeVersionToDisableMigrate` will not change between release versions (except when eventually removed), so not using logging fields

pkg/server/plugin/datastore/sql/migration.go
```diff
@@ -35,25 +65,81 @@ func migrateDB(db *gorm.DB, dbType string, log hclog.Logger) (err error) {
...
-		err = sqlError.New("backwards migration not supported! (current=%d, code=%d)", version, codeVersion)
-		log.Error(err.Error())
-		return err
+	schemaVersion := migration.Version
+
+	var dbCodeVersion semver.Version
+	// Auto-migrate can only be safely disabled if 0.9.0 migrations have already been run
+	// We know if 0.9.0 migrations have been run because dbCodeVersion will be populated if so
+	// TODO: related epic https://github.com/spiffe/spire/issues/1083
+	// This check can be removed in 0.10.0
```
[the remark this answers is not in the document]
amoore877: won't that be caught by the version compatibility check? this part is basically here just because 0.8 and earlier DBs *must* upgrade

proto/spire/go.mod
```diff
@@ -1,3 +1,9 @@
 module github.com/spiffe/spire/proto/spire
 
 go 1.13
+
+require (
```
evan2645: Was this inclusion accidental?

azdagron: Hmm, the protobuf and grpc deps seem good actually since the repo contains generated code that uses those packages. The spire dependency is in there because of the catalog plugin code.

We added this because we wanted to get at a node.proto from a 3rd party project without having to pull in SPIRE but this will pull in SPIRE anyway....

We really need to separate this out cleanly.

azdagron: I think this can go in as-is. We can revisit what we want to do with the module separately.

amoore877: oops

I'm fine with having this removed, totally accidental from running the `build.sh` setup-related things (though ideally running `make` or `build.sh` commands should result in no change at HEAD)

PR #1150, opened 2019-09-28
plugin/x509pop: Add ability to customize agent spiffe id

APTy:
This enables users to customize agent SPIFFE IDs when
using the `x509pop` plugin. The current behavior is 
maintained, which uses the sha1 fingerprint to construct
the agent ID.


pkg/server/plugin/nodeattestor/x509pop/x509pop_test.go
```diff
@@ -157,7 +177,7 @@ func (s *Suite) TestAttestFailure() {
 		"x509pop: not configured")
 
 	// now configure
-	s.configure()
+	s.configure("")
```
ZymoticB: The empty string isn't my favourite thing in the world but it's not a blocker for me in test code.

PR #1165, opened 2019-10-02
Remove server only code from common/plugin/aws

ZymoticB:
The agent path construction helpers are only used by the server
plugin now. This change moves that code to keep the common module
as small as possible.

### DIFF
--- a/pkg/common/plugin/aws/iid.go
+++ b/pkg/common/plugin/aws/iid.go
@@ -1,16 +1,12 @@
 package aws
 
 import (
-	"bytes"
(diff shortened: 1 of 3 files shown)


amoore877:
this does now introduce a discrepancy that GCP common has agent template but AWS common does not
not sure which is better- paring down AWS common or having providers have similar code organization. That being said Azure and k8s don't follow the pattern.. but should they?

ZymoticB:
I would rather have as little code as possible under a package called "common". I can do a follow up for GCP.

PR #1176, opened 2019-10-09
Upgrade github.com/docker/docker

### LINKED ISSUE #1135: Spire breaks with newer docker versions
Currently, spire is using docker version from March `v0.7.3-0.20190312165151-258edd715d46` , however in April docker introduced a [breaking change](https://github.com/moby/moby/commit/e6c0d19c3a5ebd0d719bb21c29b9d2395c7c3719) which affects spire:
```
cannot use opts (type []func(*client.Client) error) as type []client.Opt in argument to client.NewClientWithOpts
```

This causes a bit of a headache as a bunch of other libraries are using newer docker versions and because `go mod` uses the oldest allowed version, it becomes tricky to support multiple libraries with different docker versions.

ZymoticB:
Fixes #1135

This was re-raised to me as a change that was breaking due to a somewhat edge case
use case in `pkg/agent/plugin/workloadattestor/docker/docker.go` where we were collecting
a slice of options to then pass to the constructor. docker/docker changed to have an exported
type for the options and the fix here is extremely simple. Additionally, its unlikely that
direct consumers of spiffe/spire are also doing this exact thing with docker/docker. If they
are, that is unfortunate but the fix is trivial and this should not hold spiffe/spire to an
ancient version of docker/docker.


ZymoticB:
I'll investigate the build. I only ran `build.sh binaries` locally.

PR #1216, opened 2019-10-24
Fix kubernetes container restart race

JonathanO:
The state show by the Kubelet api can lag the actual state of the system. For a newly created pod this was worked around by polling until all containers in all pods on the system had a container ID, or until a timeout expired.

When a container within a pod is restarted, however, there's a window where the API reports the old container ID. If the application in the new container starts rapidly enough then the agent will be unable to map the new ID to a pod, and since it believes that all pods are "ready" it will immediately return no selectors.

This fix removes the !notAllContainersReady shortcut entirely. If a container ID cannot be looked up then we will poll until MaxPollAttempts is hit, regardless of the reported state of containers on the system.

We also increase the time the attestor will poll the k8s api for to discover pods, as the defaults have proved to be too low to consistently return selectors correctly.


JonathanO:
It's possible the lookup will fail at the containerID stage, but I was really assuming that a k8s attestor config on something that isn't a k8s worker node is unlikely. (I'd go as far as to say it should really be an error.)

We run with a 60s timeout because the consequences of the failed lookup are so catastrophic, but I figured 30s would be an acceptable compromise as a default. We use SDS, where a failed lookup isn't an error, so the client remains connected but unable to retrieve it's full set of SVIDs. I guess this could be resolved (messily) for SDS by making a request for a non existstant name drop the connection. There is potentially a worse problem with the workload API though, as it's possible to create a scenario where a client will retrieve at least 1 SVID, so not an error, but won't receive the full set of SVIDs it should have been entitled to.

evan2645:
> We use SDS, where a failed lookup isn't an error, so the client remains connected but unable to retrieve it's full set of SVIDs. I guess this could be resolved (messily) for SDS by making a request for a non existstant name drop the connection.

@JonathanO this feels like a bug? Do you know by chance Envoy behavior if we return an error - will it retry? What about that feels messy to you?

> There is potentially a worse problem with the workload API though, as it's possible to create a scenario where a client will retrieve at least 1 SVID, so not an error, but won't receive the full set of SVIDs it should have been entitled to.

I agree, this is a problem... it is related to, but different than, the one above IMO.

evan2645:
PS>> To be clear, I'm not suggesting we block this PR. Just digging deeper to see how we can further improve the SDS behavior - trying once and never again is obviously bad

PR #1238, opened 2019-11-03
Bumping go to 1.13.4

mweissbacher:
Signed-off-by: Michael Weissbacher <mweissbacher@squareup.com>


mweissbacher:
@evan2645 I wanted to keep things in sync with go stable, if anything breaks it's probably good to know early :-)

PR #1244, opened 2019-11-05
Add documentation describing SPIRE versioning and upgrade procedures

### LINKED ISSUE #1027: Document the Procedure to Upgrade SPIRE
* **Version**: 
* **Platform**: 
* **Subsystem**: 
Document the procedure to upgrade SPIRE.

Include the following:
Order of components
Backward compatibility guarantees
DB backup
HA concerns

evan2645:
This PR adds documentation relating to managing SPIRE upgrades/downgrades. Included is information about version compatibility and supported version skew, which is an important detail in this context.

I feel we need to sit down and think through a new structure for all SPIFFE/SPIRE documentation that makes sense given everything we know today... but for now, the easiest thing to do is to check this into our `doc/` directory and link to it from the README.

Fixes #1027

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,10 @@ SPIRE provides an implementation of the [Envoy](https://envoyproxy.io)
 (SDS). SDS can be used to transparently install and rotate TLS certificates and
 trust bundles in Envoy. Please see the [SPIRE Agent configuration guide](/doc/spire_agent.md#agent-configuration-file) for more information.
 
+# Upgrading SPIRE
+
+SPIRE server supports zero-downtime upgrades when there's more than one SPIRE server in the cluster. Please seee the [Managing Upgrades/Downgrades](doc/upgrading.md) guide for more information on SPIRE versioning and supported upgrade paths.
+
 # Getting Help
 
 If you have any questions about how SPIRE works, or how to get it up and running, the best place to ask questions is the [SPIFFE Slack Organization](https://slack.spiffe.io/). Most of the maintainers monitor the #spire channel there, and can help direct you to other channels if need be. Please feel free to drop by any time!
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ SPIRE (the [SPIFFE](https://github.com/spiffe/spiffe) Runtime Environment) is a 
 - [Getting started](#getting-started)
 - [Examples](#examples)
 - [Using SPIRE with Envoy](#using-spire-with-envoy)
+- [Upgrading SPIRE](#upgrading-spire)
 - [Getting help](#getting-help)
 - [Community](#community)
 
@@ -45,7 +46,7 @@ trust bundles in Envoy. Please see the [SPIRE Agent configuration guide](/doc/sp
 
 # Upgrading SPIRE
 
-SPIRE server supports zero-downtime upgrades when there's more than one SPIRE server in the cluster. Please seee the [Managing Upgrades/Downgrades](doc/upgrading.md) guide for more information on SPIRE versioning and supported upgrade paths.
+SPIRE Server supports zero-downtime upgrades when there's more than one SPIRE Server in the cluster. Please see the [Managing Upgrades/Downgrades](doc/upgrading.md) guide for more information on SPIRE version compatibility and supported upgrade paths.
 
 # Getting Help
 
--- a/doc/upgrading.md
+++ b/doc/upgrading.md
@@ -2,39 +2,41 @@
 This guide describes how to upgrade your SPIRE deployment, as well as the compatibility guarantees that SPIRE users can expect.
 
 ## SPIRE Versioning
-SPIRE versions are expressed as **x.y.z**, where **x** is the major version, **y** is the minor version, and **z** is the patch version, following Semantic Versioning terminology. The SPIRE project is currently pre-1.0, so minor SPIRE releases act like major releases and patch releases act like minor releases. Despite the fact that the project is pre-1.0, version compatibility guarantees are already in place.
+SPIRE versions are expressed as **x.y.z**, where **x** is the major version, **y** is the minor version, and **z** is the patch version, following Semantic Versioning terminology. Despite the fact that the project is currently pre-1.0, version compatibility guarantees are already in place. This document will be updated in advance of a 1.0.0 release.
 
 ### SPIRE Server Compatibility
-Version skew within a SPIRE server cluster is supported within +/- 1 minor version. In other words, the newest and oldest SPIRE server instances in any given cluster must be within one minor version of each other.
+Version skew within a SPIRE Server cluster is supported within +/- 1 minor version. In other words, the newest and oldest SPIRE Server instances in any given cluster must be within one minor version of each other.
 
 For example:
-* Newest SPIRE server instance is at 0.9.3
-* Other SPIRE server instances are supported at 0.9.x and 0.8.x
+* Newest SPIRE Server instance is at 0.9.3
+* Other SPIRE Server instances are supported at 0.9.x and 0.8.x
 
 ### SPIRE Agent Compatibility
-SPIRE agents must not be newer than the oldest SPIRE server that they communicate with, and may be up to one minor version older.
+SPIRE Agents must not be newer than the oldest SPIRE Server that they communicate with, and may be up to one minor version older.
 
 For example:
-* SPIRE servers are at both 0.9.3 and 0.9.2
-* SPIRE agents are supported at 0.8.0 through 0.9.2
+* SPIRE Servers are at both 0.9.3 and 0.9.2
+* SPIRE Agents are supported at 0.8.0 through 0.9.2
 
 ## Supported Upgrade Paths
 
-The supported version skew between SPIRE servers and agents has implications on the order in which they must be upgraded. SPIRE servers must be upgraded before SPIRE agents, and is limited to a jump of at most one minor version (regardless of patch version). SPIRE server and agent instances may be upgraded in a rolling fashion.
+The supported version skew between SPIRE Servers and agents has implications on the order in which they must be upgraded. SPIRE Servers must be upgraded before SPIRE Agents, and is limited to a jump of at most one minor version (regardless of patch version). Upgrades that jump two or more minor versions (e.g. 0.8.1 to 0.10.0) are not supported.
+
+SPIRE Server and agent instances may be upgraded in a rolling fashion.
 
 For example, if upgrading from 0.8.1 to 0.9.3:
-* Upgrade SPIRE server instances from 0.8.1 to 0.9.3 one at a time
-* Ensure that the SPIRE server cluster is operating as expected
-* Upgrade SPIRE agent instances from 0.8.1 to 0.9.3 one at a time or in batches
+* Upgrade SPIRE Server instances from 0.8.1 to 0.9.3 one instance at a time
+* Ensure that the SPIRE Server cluster is operating as expected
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -43,6 +43,10 @@ SPIRE provides an implementation of the [Envoy](https://envoyproxy.io)
 (SDS). SDS can be used to transparently install and rotate TLS certificates and
 trust bundles in Envoy. Please see the [SPIRE Agent configuration guide](/doc/spire_agent.md#agent-configuration-file) for more information.
 
+# Upgrading SPIRE
+
+SPIRE server supports zero-downtime upgrades when there's more than one SPIRE server in the cluster. Please seee the [Managing Upgrades/Downgrades](doc/upgrading.md) guide for more information on SPIRE versioning and supported upgrade paths.
```
azdagron: typo: `seee` v.s. `see`

doc/upgrading.md
```diff
@@ -0,0 +1,40 @@
...
+For example:
+* SPIRE servers are at both 0.9.3 and 0.9.2
+* SPIRE agents are supported at 0.8.0 through 0.9.2
+
+## Supported Upgrade Paths
+
+The supported version skew between SPIRE servers and agents has implications on the order in which they must be upgraded. SPIRE servers must be upgraded before SPIRE agents, and is limited to a jump of at most one minor version (regardless of patch version). SPIRE server and agent instances may be upgraded in a rolling fashion.
+
+For example, if upgrading from 0.8.1 to 0.9.3:
+* Upgrade SPIRE server instances from 0.8.1 to 0.9.3 one at a time
```
azdagron: the "one at a time" phrasing is a bit ambiguous. one _instance_ at a time? one _minor_ version at a time?

the phrasing occurs in a few spots...

azdagron:
This is great, @evan2645. Looking forward to having this in :)

doc/upgrading.md
evan2645: Would this wording address your concern?

```
Upgrade SPIRE servers from 0.8.1 to 0.9.3 one instance at a time
```

I should add that I went back and forth on whether we should include the rollout methodology on this line or not... the goal is to upgrade from one to the other, any ordering/batching etc will accomplish that, but there's one that we'd recommend over others... wasn't quite sure how/where to capture that in a way that it wouldn't be overlooked

azdagron: Yeah, that's a fair concern. Maybe we're being too prescriptive... maybe we just say something along the lines of what you just said.

doc/upgrading.md
```diff
@@ -0,0 +1,40 @@
+# Managing Upgrades/Downgrades
+This guide describes how to upgrade your SPIRE deployment, as well as the compatibility guarantees that SPIRE users can expect.
+
+## SPIRE Versioning
+SPIRE versions are expressed as **x.y.z**, where **x** is the major version, **y** is the minor version, and **z** is the patch version, following Semantic Versioning terminology. The SPIRE project is currently pre-1.0, so minor SPIRE releases act like major releases and patch releases act like minor releases. Despite the fact that the project is pre-1.0, version compatibility guarantees are already in place.
```
azdagron: Is the version number nomenclature (i.e. major, minor, version) in the rest of the document from a pre-1.0 perspective or using the standard semver meanings? For example, when we say +/- minor version, i'm assuming we mean the semver version? once we've reached 1.0 will we need to come update this document and change minor => major, etc.?

doc/upgrading.md
```diff
@@ -0,0 +1,40 @@
...
+
+### SPIRE Server Compatibility
+Version skew within a SPIRE server cluster is supported within +/- 1 minor version. In other words, the newest and oldest SPIRE server instances in any given cluster must be within one minor version of each other.
+
+For example:
+* Newest SPIRE server instance is at 0.9.3
+* Other SPIRE server instances are supported at 0.9.x and 0.8.x
```
sanderson042: To Andrew H's comment, maybe add this sentence after the bullet items:

"Currently, the minor version refers to the third number in the SPIRE version, such as 1 in 0.8.1. When SPIRE reaches 1.0.0, the minor version will refer to the second number, such as 2 in 1.2.3."

doc/upgrading.md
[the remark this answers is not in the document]
evan2645: This reads much better, but raises a question. I chose here to say "act like" in order to communicate the functional differences between one release type and another without _actually_ renaming minor -> major in the context of this document. Thus, the rest of the sections use the term "minor" to mean 0.x.0.

Here, I think you are suggesting that we use the term "major" instead of "minor" throughout this document even though the true major version number is 0.

There is definitely a lot of confusion about what is minor and what is major in the context of SPIRE and how it relates to semver. Here, I tried to take the absolute definitions and apply them, effectively meaning that there are _no_ major releases... but perhaps that's not the right way to go.

I am happy to take your suggestion as-is, but I also want to be explicit about this decision and apply it everywhere going forward. One last consideration is that pre-1.0 behavior _is_ defined by semver, and I believe our current versioning semantics are in line with what semver defines there...

evan2645: @azdagron would love to get your input on this one https://github.com/spiffe/spire/pull/1244/files#r342855663

doc/upgrading.md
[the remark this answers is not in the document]
evan2645: > Sorry, I'm a bit confused. If 3 is the minor version in 0.9.3, wouldn't +/- one minor version be 0.9.4, 0.9.3, and 0.9.2 and no 0.8.anything?

Related to our above comments, I took the absolute stance, so minor version would be in position 0.x.0. In this context, it would feel weird to say that given version 0.9.3, "9" is the major version (because it is clearly not when prefixed with "0")

> Having these examples is good, but I do prefer the version-skew-policy k8s style where the example versions are listed explicitly like 1.13, 1.12, and 1.11 rather than with "x." Maybe there's a reason you chose that presentation.

I put `x` because I don't know how many releases will fall under them (for instance, today 0.8.4 is the latest but we could cut 0.8.5 in the future, and that version would apply here)

doc/upgrading.md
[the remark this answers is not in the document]
azdagron: I'm very much in favor of using the absolute definitions. I think doing otherwise will be confusing to anybody with experience in versioning schemes. Maybe we say something akin to "SPIRE is pre-1.0, so our compat guarantees are with +/- 1 minor version. After 1.0, our compatability guarantees will align to major versions" and then keep the rest as-is. When we reach 1.0, we can come update the document accordingly.

azdagron: commented :) I'll resolve this convo since the other thread captures the essence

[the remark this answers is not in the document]
evan2645: > I'm very much in favor of using the absolute definitions. I think doing otherwise will be confusing to anybody with experience in versioning schemes.

Yes... the more I read through the document with this in mind, the more it feels like defining the "major" position as 0.x.0 will lead to even more confusion.

Taking @sanderson042's comment though (and @azdagron's previous comment), I think it's also clear that the current wording is confusing. Would it make more sense to leave the "minors are like majors" analogy out altogether?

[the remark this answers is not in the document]
azdagron: Yes, i think it would. I think it would still be worth mentioning that the guarantees layed out in the document are for pre-1.0 only and that the document will be updated when 1.0 lands or something.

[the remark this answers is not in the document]
sanderson042: Ah, I see where I was confused. This has been a productive discussion. I'm on vacation Thursday & Friday, so this'll be my last contribution. This is a complete straw man proposal, but what if we used "big" and "little" instead of major and minor and pared this paragraph down to one sentence. The method to my madness is that big and little don't have preexisting formal definitions like in semver. Anyway, I think you guys have found the confusing bits, so can figure out the wording from here.

"In its current pre-1.0 state, versions in the SPIRE project are expressed as 0.big.little, where big indicates a release version with many changes and little indicates a release with a few changes."

See more below...

doc/upgrading.md
sanderson042: I think we agree we can leave my proposed addition out.

doc/upgrading.md
```diff
@@ -0,0 +1,40 @@
...
+
+### SPIRE Agent Compatibility
+SPIRE agents must not be newer than the oldest SPIRE server that they communicate with, and may be up to one minor version older.
+
+For example:
+* SPIRE servers are at both 0.9.3 and 0.9.2
+* SPIRE agents are supported at 0.8.0 through 0.9.2
```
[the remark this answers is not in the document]
sanderson042: Ok as is.

doc/upgrading.md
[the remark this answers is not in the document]
sanderson042: Continuing on with my big/little crazy idea:

"Version skew within a SPIRE server cluster is supported within one big version. In other words, the newest and oldest SPIRE server instances in any given cluster must be within one big version of each other.

 For example:
 * Newest SPIRE server instance is at 0.9.3
 * Other SPIRE server instances are supported at 0.9.x and 0.8.x 
"

doc/upgrading.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Supported Upgrade Paths
+
+The supported version skew between SPIRE servers and agents has implications on the order in which they must be upgraded. SPIRE servers must be upgraded before SPIRE agents, and is limited to a jump of at most one minor version (regardless of patch version). SPIRE server and agent instances may be upgraded in a rolling fashion.
+
+For example, if upgrading from 0.8.1 to 0.9.3:
```
APTy: Based on `the newest and oldest SPIRE server instances in any given cluster must be within one minor version of each other`, it doesn't sound possible to upgrade two minor versions at a time, e.g. you can't go from `0.8.x` to `0.10.x` at the same time.. is my understanding correct? If so, maybe we call that out explicitly?

APTy:
This is really helpful guidance, thanks for doing it!

README.md
```diff
@@ -43,6 +43,10 @@ SPIRE provides an implementation of the [Envoy](https://envoyproxy.io)
 (SDS). SDS can be used to transparently install and rotate TLS certificates and
 trust bundles in Envoy. Please see the [SPIRE Agent configuration guide](/doc/spire_agent.md#agent-configuration-file) for more information.
 
+# Upgrading SPIRE
```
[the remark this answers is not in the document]
evan2645: 👍 sounds good! I updated all the instances both here and in the upgrade doc

README.md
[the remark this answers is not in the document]
evan2645: Awesome, sounds much better. Fixed

doc/upgrading.md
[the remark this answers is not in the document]
evan2645: I chatted around about this one, and the conclusion I reached is that 1) it is the "x is like y" and "y is like z" talk that is the primary source of confusion, and 2) that detail is not very important because the impact it has on upgrade path is explicitly spelled out through the rest of this document.

I spoke to @sanderson042 offline about this, and we decided to simply strike that statement.

I did, however, add one more sentence here stating that we would update this document prior to a 1.0.0 release in an effort to quell any questions that may pop up around how this will work in the future

doc/upgrading.md
evan2645: Yes that is correct.

I have added a sentence in this section to explicitly state that jumps of two or more minor versions are unsupported

doc/upgrading.md
evan2645: Fixed

Issue #1245, opened 2019-11-05
undocumented options in the k8s workload attestor


bobo333:
* **Version**: 0.8.3
* **Platform**: kubernetes
* **Subsystem**: agent

There are undocumented options in the k8s workload attestor. Specifically `max_poll_attempts` and `poll_retry_interval`. These are useful for dealing with pod deletion race conditions. As mentioned in https://github.com/spiffe/spire/issues/1230 and https://github.com/spiffe/spire/pull/1216

evan2645:
Hey @bobo333 I left a small comment on your PR, please let me know what you think... thanks again!

bobo333:
closed the PR as the new defaults cover my use case, thanks!

PR #1246, opened 2019-11-05
add k8s workload attestor polling configuration option to docs

bobo333:
**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
No functionality changes.

**Description of change**
Adds documentation for `max_poll_attempts` and `poll_retry_interval` options for the `k8s workload attestor`. These are useful for dealing with pod deletion race conditions (see info in issue).

Only reason I can think of for not including this as docs is if it's not planned for these options to be available in the future, but I don't see why that would be the case either :)

**Which issue this PR fixes**
https://github.com/spiffe/spire/issues/1245

### DIFF
--- a/doc/plugin_agent_workloadattestor_k8s.md
+++ b/doc/plugin_agent_workloadattestor_k8s.md
@@ -41,6 +41,8 @@ that can impact permission revocation.
 | `private_key_path` | The path on disk to client key used for kubelet authentication |
 | `node_name_env` | The environment variable used to obtain the node name. Defaults to `MY_NODE_NAME`. |
 | `node_name` | The name of the node. Overrides the value obtained by the environment variable specified by `node_name_env`. |
+| `poll_retry_interval` | The time to wait between retries when attesting a workload. Defaults to `300ms`. |
+| `max_poll_attempts` | The number of attempts to attest a workload before giving up. Defaults to `5`. |
 
 | Selector | Value |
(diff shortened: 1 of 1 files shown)


evan2645:
Thank you for sending this update @bobo333 

We had intentionally left these configurables undocumented because changing them can be kind of brittle (e.g. lead to problems if the values aren't right) and we didn't want to complicate the configuration.

That being said, we're aware that the original defaults were too aggressive and were leading to false negatives in some cases. We [accepted a PR](https://github.com/spiffe/spire/pull/1216) recently that updates these defaults, and that change will be released this week as part of version 0.9.0. I am curious to hear 1) what values you found to work best for you in your environment, and 2) if you feel that the new defaults satisfy your requirements and/or negate the need to document them? In an ideal world, this kind of stuff wouldn't need to be configured at all - if we can get it "right", perhaps we can even deprecate the configurables altogether

bobo333:
Hey @evan2645 thanks for the feedback. I went with polling every 1 second, for a total of 60 attempts. However, the new defaults would cover my case as well. Mostly I just needed it to wait more than a second or 2 before completely giving up, as there was some delay in the pod being returned in the `/pods` endpoint from `kubelet` every now and then.

Going to close this out for now, thank you again!

evan2645:
Ok great! Thanks again @bobo333 it is always appreciated ❤️

Issue #1040, opened 2019-07-31
change telemetry library used


amoore877:
Currently telemetry package used is https://github.com/armon/go-metrics, which provides telemetry via prometheus, statsd, dogstatsd, and inmem. I propose getting off of this library either entirely or partially in favor of another.
- It is not very active.
- It has no official releases.
- Despite origins, it does not appear to be a Hashicorp-supported project, at least not for community. SLA on fairly important issues is long- https://github.com/armon/go-metrics/issues/85 , https://github.com/armon/go-metrics/issues/78
- Implementation is problematic, such as
-- Use of `goto` statements
-- Prometheus impl is broken
--- https://github.com/armon/go-metrics/issues/78
--- When scraping prometheus endpoint, found that errors occur such as
```
8 error(s) occurred:
* collected metric "spire_server_node_api_x509_svid_fetch_elapsed_time" { label:<name:"host" value:"someObscuredHost" > label:<name:"spiffe_id" value:"spiffe://someObscuredSpiffe" > summary:<sample_count:2 sample_sum:11515.1171875 quantile:<quantile:0.5 value:nan > quantile:<quantile:0.9 value:nan > quantile:<quantile:0.99 value:nan > > } was collected before with the same name and label values
```

Overall it may be more effective to transition to a different library.

amoore877:
I propose the alternative of https://github.com/uber-go/tally, which provides `prometheus`, `m3`, and `statsd` implementations.
Additionally:
- It is active
- It is an official Uber-supported project
- It has actual release cycles
- It it generally more configurable than the current library

amoore877:
As a side issue, it is inconvenient to have `go-metrics` be a forced common middleware for all metrics plugins. It is understandable since the existing built-in metrics plugins are all also from `go-metrics`, but this also forces all plugins into `go-metrics` logic. For instance, `MeasureSince` is actually a `go-metrics` wrapper around the plugins' `AddSample`, which is deceptive since we'd expect when bringing in a plugin that `MeasureSince` calls should be going to the plugin's `MeasureSince`.

evan2645:
I'm not quite sure what you mean by `metrics plugins`.

The interface that the telemetry package exposes is modeled after the go-metrics interface because that was the simplest thing to do and it seemed likely that such an interface could map easily onto other libraries in the even that we decided to switch it out. It was specifically meant to act as a layer of indirection over whatever metrics library we want to use

amoore877:
[...]

PR #1262, opened 2019-11-14
fix m3 duration and value bucketing

amoore877:

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
M3 telemetry bucketing

**Description of change**
Existing bucketing defines no buckets, resulting in a single `minus infinity to infinity` bucket for durations and values. This is not useful.
Define actual latency and value bucketing (use defaults).

**Which issue this PR fixes**
N/A


pkg/common/telemetry/m3.go
```diff
@@ -11,6 +11,68 @@ import (
 	"github.com/uber-go/tally/m3"
 )
 
+var (
+	// buckets with decreasing granularity as latency increases
+	// given these values, it is much easier to read in this format
+	// than to use the iterative `MakeBuckets` functionality from tally
+	durationBuckets = tally.DurationBuckets{
```
amoore877: this is an initial suggestion for latency buckets, open to suggestions on different bucket cut offs

pkg/common/telemetry/m3.go
```diff
@@ -11,6 +11,68 @@ import (
...
+		2500 * time.Millisecond,
+		3000 * time.Millisecond,
+		4000 * time.Millisecond,
+		5000 * time.Millisecond,
+		10000 * time.Millisecond,
+	}
+
+	// buckets for orders of magnitude of values, up to 1M
+	// given nature of SPIRE, we do not expect negative values
+	valueBuckets = append(tally.ValueBuckets{0}, tally.MustMakeExponentialValueBuckets(1, 10, 6)...)
```
amoore877: initial suggestion for buckets. open to suggestion
this is applicable only to `AddSample` which is today used only  for:
- `DiscoveredSelectors` on workload attestation
- `ExpiringSVIDs` on agent cache manager

pkg/common/telemetry/m3.go
zmt: It looks like a decent guess on the face of it. Varying bucket size can get really complicated.

pkg/common/telemetry/m3.go
```diff
@@ -11,6 +11,67 @@ import (
...
+		time.Second,
+		1500 * time.Millisecond,
+		2 * time.Second,
+		3 * time.Second,
+		4 * time.Second,
+		5 * time.Second,
+		10 * time.Second,
+	}
+
+	// buckets for orders of magnitude of values, up to 1M
```
jkanywhere: what is 1M in this context? Minute? Meter? Megabyte?

pkg/common/telemetry/m3.go
jkanywhere: `valueBuckets` is a vague sounding name. What is it intended to be used for? `exponentialValueBuckets` is slightly better.

pkg/common/telemetry/m3.go
```diff
@@ -11,6 +11,67 @@ import (
...
+	// given nature of SPIRE, we do not expect negative values
```
jkanywhere: Does it makes sense to add a `-Infinity to 0` bucket so these unexpected cases can be detected?

jkanywhere:
I also envision a duration bucket for SVID TTL, something with granularity as low as minutes and as high as days.
That way we can log and detect when agents or workloads are not getting SVID refresh.

Maybe something like
```
0,
5 * time.Minute,
30 * time.Minute,
1 * time.Hour,
2 * time.Hour,
4 * time.Hour,
6 * time.Hour,
8 *time.Hour,
12 * time.Hour,
1 * time.Day,
2 * time.Day,
5 * time.Day,
```
Might even make sense to have some negative buckets to detect expired certs. You may want it to be symmetric with the positive buckets, or have fewer larger buckets with lower granularity.
```
-Infinity,
-5 * time.Day,
-1 * time.Day,
-6 * time.Hour,
-1 * time.Hour,
-5 * time.Minute, 
```

Does it make sense to add those in the same diff?

jkanywhere:
> I also envision a duration bucket for SVID TTL
> ```
...
> 12 * time.Hour,
> 1 * time.Day,
> 2 * time.Day,
> 5 * time.Day,
> ```

Assuming we log whenever we receive and SVID, and periodically thereafter.
If expected SVID validity period is 1 day, it probably makes sense to add some additional more granular buckets around that time range. Also maybe around the one-half TTL range.
If those values are configurable, maybe the buckets would change too.
```
...
4 * time.Hour,
5 * time.Hour,
6 * time.Hour,
7 *time.Hour,
8 *time.Hour,
12 * time.Hour,
23 * time.Hour,
24 * time.Hour,
25 * time.Hour,
2 * time.Day,
5 * time.Day,
```
We would expect things to stay largely within the 1/2 TTL to 1 TTL range.

pkg/common/telemetry/m3.go
ZymoticB: -Infinity to your lowest bucket implicitly exist. As does your highest bucket to infinity

ZymoticB:
There is utility in making these useful in the short term, but the real problem is that the metrics facade prevents the callsite that is recording the duration sample from providing an expectation of the buckets. 

For example, an SVID TTL is a duration, as is the response time of making a request to the datastore plugin. The granularity of buckets in these two situations will vastly differ. Trying to pick one set of buckets to rule them all will likely be problematic in the future.

The preferred usage for tally is to provide your own buckets to  Scope.Histogram and cache that object locally. We'll need to investigate going forward how to best accomplish this without sacrificing the ability to use other metrics backends.

amoore877:
> For example, an SVID TTL is a duration, as is the response time of making a request to the datastore plugin. The granularity of buckets in these two situations will vastly differ. Trying to pick one set of buckets to rule them all will likely be problematic in the future.

minor nit that if we get in a change to have TTL as a metric, `AddSample` would make more sense than `MeasureSince`. Granted the bucketing for that is also not great for TTLs today but it would be a more intuitive call. It may also make sense to have % categories of TTL remaining or similar instead of precise TTLs, or to be emitting such metrics on the user integration end of things.

> The preferred usage for tally is to provide your own buckets to Scope.Histogram and cache that object locally. We'll need to investigate going forward how to best accomplish this without sacrificing the ability to use other metrics backends.

this somewhat wraps back to the future of spire metrics libraries and design #1040

pkg/common/telemetry/m3.go
amoore877: million

amoore877:
just using default buckets

evan2645:
@amoore877 ICYMI, the tests failed with the below error

```
--- FAIL: TestAddSampleForValueHistogram (0.00s)
    m3_test.go:116: 
        	Error Trace:	m3_test.go:116
        	Error:      	Should NOT be empty, but was map[]
        	Test:       	TestAddSampleForValueHistogram
    m3_test.go:117: 
        	Error Trace:	m3_test.go:117
        	Error:      	Should be empty, but was map[0s:0 10ms:0 25ms:0 50ms:0 75ms:0 100ms:0 200ms:0 300ms:0 400ms:0 500ms:0 600ms:0 800ms:0 1s:0 2s:0 5s:0 2562047h47m16.854775807s:1]
        	Test:       	TestAddSampleForValueHistogram
FAIL
FAIL	github.com/spiffe/spire/pkg/common/telemetry	0.157s
```

amoore877:
> @amoore877 ICYMI, the tests failed with the below error
> 
> ```
> --- FAIL: TestAddSampleForValueHistogram (0.00s)
>     m3_test.go:116: 
>         	Error Trace:	m3_test.go:116
>         	Error:      	Should NOT be empty, but was map[]
>         	Test:       	TestAddSampleForValueHistogram
>     m3_test.go:117: 
>         	Error Trace:	m3_test.go:117
>         	Error:      	Should be empty, but was map[0s:0 10ms:0 25ms:0 50ms:0 75ms:0 100ms:0 200ms:0 300ms:0 400ms:0 500ms:0 600ms:0 800ms:0 1s:0 2s:0 5s:0 2562047h47m16.854775807s:1]
>         	Test:       	TestAddSampleForValueHistogram
> FAIL
> FAIL	github.com/spiffe/spire/pkg/common/telemetry	0.157s
> ```

it's always the time you forget to run the tests :) sorry about that

amoore877:
talked offline with @azdagron 
since tally's default buckets don't include proper value buckets, we will have our own for now of an exponential bucket set:
0-1, 1-10, 10-100, 100-1000,.. up to 10^5

Issue #1257, opened 2019-11-13
Agents should re-attest if its current SVID is invalid


APTy:
I couldn't find an existing issue for this, but I know I've talked about it with some people before. Currently, agent attestation logic is fairly simple:

1. If a cached SVID/key pair exists and is not expired, use it.
2. Else, do node attestation

As a result, the agent fails permanently in the case of a `spire-server agent evict`. The current mitigation is to manually delete the agent's `.data` dir and then restart it. The agent could be more robust and automatically re-attest in these cases.

ZymoticB:
I intend to pick this up next week while everyone else is at kubecon :)

evan2645:
Thoughts on how to detect this condition? The unauthenticated error code comes to mind but I think that we might use that in some other spots as well.

ZymoticB:
The plan was permission denied on fetchx509svids

evan2645:
Copy/pasting some chatter from #sig-integration-aws slack channel as fodder for thought:

> Jonathan Oddy:  Does anyone know why the iid node attestor rejects attestation if the IID has already been used to attest a node before? (https://github.com/spiffe/spire/blob/master/pkg/server/plugin/nodeattestor/aws/iid.go#L173-L179)
> 
> Evan Gilman:  IID (and all other methods in that family like GCP IIT etc) are all trust on first use because anything running on the host can fetch it... Without special non-SPIRE configuration, it poses a problem for deployments with multiple workloads on a single host as any of them can fetch the document and masquerade as the agent (thereby obtaining SVIDs for colocated workloads) (edited) 
> 
> With TOFU, the window is reduced to racing the agent, and at least you'll get to know if you lost
> 
> ...
> 
> Jonathan Oddy: In our particular case we're not really that worried about other workloads masquerading as the agent, so perhaps the quick fix is a config option to disable that guard. (On a related note, we've also been tempted to add a noop workload attestor that'll attest any workload on a given host with a fixed selector)
> 
> These nodes all run a single workload and nothing else, so we aren't worried if anything on the node can attest as that node
> 
[...]

PR #1281, opened 2019-11-22
remove superfluous node attest metric

amoore877:

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
`node_attest` metric

**Description of change**
There are currently metrics of `node_attest` and `node_attestor_new_svid`.
`Attest` function in agent starts with `node_attest`, which captures attempting loading of keys / bundles from the key manager and on failure attest with the server to get a new svid
`node_attestor_new_svid` captures the attempt to attest with the server to get a new svid

`node_attest` should be removed. It is superfluous given the `node_attestor_new_svid` metric, measuring only file operations in addition to measuring what `node_attestor_new_svid` already takes care of.

**Which issue this PR fixes**
N/A


APTy:
@amoore877 Thanks for bringing this up. It seems like the `node_attestor_new_svid` metric provides meaningful insight into the node attestation process. Meanwhile, the `node_attest` metric is a superset of node attestation, and it also includes things that are probably not worth measuring, like how long it takes to load an svid from disk. Given this, what do you think about removing the `node_attest` metric entirely?

amoore877:
> @amoore877 Thanks for bringing this up. It seems like the `node_attestor_new_svid` metric provides meaningful insight into the node attestation process. Meanwhile, the `node_attest` metric is a superset of node attestation, and it also includes things that are probably not worth measuring, like how long it takes to load an svid from disk. Given this, what do you think about removing the `node_attest` metric entirely?

I think it's valuable to keep, if only because it can indicate specific failure conditions on agent startup

APTy:
> I think it's valuable to keep, if only because it can indicate specific failure conditions on agent startup. We may also introduce logic later that adds more, or takes away, some of this init processing (true overhead task, key manager changes...), and it would be valuable to know the latency differences.

What is it really measuring though? It seems like only file read latencies, which aren't super meaningful to operators. If we later introduce logic that adds latency that is meaningful to observe, we should consider adding init metrics at that time. But as we do these metrics tweaks, if we find things that aren't useful, we should clean them up. I think this instance is a case of a metric that was added early on, and now only serves as cruft that can be cleaned up.

As a concrete example, #1257 suggests adding the ability for agents to re-attest, which implies that `Attest()` could be invoked at times other than init. Do we rename the metric at that point? Or move it down to just measure the file reads? Either way, the future is uncertain, and removing this metric provides us with more flexibility for whatever future comes.

APTy:
Thanks @amoore877!

PR #1285, opened 2019-11-25
Makefile: assume control of Google .proto downloads from build.sh

alrs:
This moves the installation of the Google .proto files out of `build.sh` and makes it part of the `Makefile`. Included in the improved Makefile is awareness of the existence of an already-downloaded .zip of `protoc-3.3.0-linux-x86_64.zip`.


alrs:
Apologies for the CI spam, it was a three-way matchup between Travis, bash, and make.

alrs:
A full `Makefile` revision is imminent, away this goes.

Issue #1239, opened 2019-11-04
MySQL read / write connection strings


amoore877:
To take advantage of read replicas, MySQL plugin should allow configuring a Read-only connection string.

- Write connection string is mandatory, and if no Read-only is set then the Write string is used for Read as well
- If Read is set, use for Read ops

APTy:
Strawman config option for [plugin_server_datastore_sql](https://github.com/spiffe/spire/blob/master/doc/plugin_server_datastore_sql.md) proposed in bold:

| Configuration     | Description                                                                |
| ------------------| -------------------------------------------------------------------------- |
| database_type     | database type                                                              |
| connection_string | connection string                                                          |
| **ro_connection_string** | **read-only connection string, used for read-only queries if set**                                                         |

evan2645:
> Strawman config option for plugin_server_datastore_sql proposed in bold:

👍 that looks about right!

There are some features in SPIRE that depend on datastore consistency for security - join tokens (race to use) come to mind, as does node attestor TOFU.

IME read replicas are usually async and can fall behind. We will probably need to revisit some security-sensitive spots to make sure that the guarantees remain the same if an async read replica is in use.

I think there is an assumption that the SPIRE dataset is 1) relatively small and 2) not very demanding. It leads me to wonder how much gain we can really make by supporting read replicas. Are there aspects beyond performance that drive folks in this direction? Curious to hear thoughts on this and the above

kirutthika:
Thanks for the feedback above. The direction that led us to this change was the idea of having HA server clusters that connect to the database with high throughput, which could be benefitted by having read-only replica for some operations. As an example, AWS clusters support read-only replicas for high throughput workloads and we could benefit by having this change. Having said that, I agree with your point on concerns related to read-after-write consistency. In those cases, we can resort to reading from main DB, otherwise we should be having an option to spread out the load and read from one of the read replicas. If this sounds good, I could work on identifying operations that are not critical to have read-after-write consistency guarantees and have it rely on RO connection string with a fallback to default RW connection string.

kirutthika:
On taking a look at the SPIRE codebase, operations such as ListBundles, ListAttestedNodes, ListRegistrationEntries can perhaps live with eventual consistency guarantees with the tradeoff for scalability while using read-only replicas. While there is currently no existing patterns for read-after-write for these operations yet, we could add documentation that states that these operations will be eventually consistent with the usage of read-only replicas. Thoughts around this proposal?

[...]

PR #1321, opened 2019-12-24
Adding read_only database connection string that can be used for read…

kirutthika:
Adding read_only database connection string that can be used for read-only replicas.

Signed-off-by: Kirutthika Raja <kirutthika.raja@uber.com>

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Adding read_only database connection string that can be used for read only replicas

**Description of change**
To take advantage of read replicas, enabling MySQL plugin to allow configuration of a read-only connection string.
Write connection string is mandatory, and if no Read-only is set then the Write string is used for Read as well. If Read is set, then the read connection string is use for Read operations.

**Which issue this PR fixes**
https://github.com/spiffe/spire/issues/1239

### DIFF
--- a/conf/server/server.conf
+++ b/conf/server/server.conf
@@ -19,6 +19,7 @@ plugins {
         plugin_data {
             database_type = "sqlite3"
             connection_string = "./.data/datastore.sqlite3"
+            ro_connection_string = "./.data/datastore.sqlite3"
         }
     }
 
--- a/pkg/server/plugin/datastore/sql/mysql.go
+++ b/pkg/server/plugin/datastore/sql/mysql.go
@@ -23,8 +23,8 @@ const (
 	tlsConfigName = "spireCustomTLS"
 )
 
-func (my mysql) connect(cfg *configuration) (*gorm.DB, error) {
-	connString, err := configureConnection(cfg)
+func (my mysql) connect(cfg *configuration, connectionString string) (*gorm.DB, error) {
+	connString, err := configureConnection(cfg, connectionString)
 	if err != nil {
 		return nil, err
 	}
@@ -38,15 +38,15 @@ func (my mysql) connect(cfg *configuration) (*gorm.DB, error) {
 
 // configureConnection modifies the connection string to support features that
 // normally require code changes, like custom Root CAs or client certificates
-func configureConnection(cfg *configuration) (string, error) {
+func configureConnection(cfg *configuration, connectionString string) (string, error) {
 	if !hasTLSConfig(cfg) {
 		// connection string doesn't have to be modified
-		return cfg.ConnectionString, nil
+		return connectionString, nil
 	}
 
 	tlsConf := tls.Config{}
 
-	opts, err := mysqldriver.ParseDSN(cfg.ConnectionString)
+	opts, err := mysqldriver.ParseDSN(connectionString)
 	if err != nil {
 		// the connection string should have already been validated by now
 		// (in validateMySQLConfig)
@@ -93,7 +93,24 @@ func hasTLSConfig(cfg *configuration) bool {
 }
 
 func validateMySQLConfig(cfg *configuration) error {
-	opts, err := mysqldriver.ParseDSN(cfg.ConnectionString)
+	var err error
+
+	if cfg.ConnectionString != "" {
+		err = validateConfig(cfg.ConnectionString)
+		if err != nil {
+			return err
+		}
+	}
+
+	if cfg.RoConnectionString != "" {
+		err = validateConfig(cfg.RoConnectionString)
+	}
+
+	return err
+}
+
+func validateConfig(connectionString string) error {
+	opts, err := mysqldriver.ParseDSN(connectionString)
 	if err != nil {
 		return sqlError.Wrap(err)
 	}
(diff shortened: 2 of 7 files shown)


evan2645:
Thank you for this contribution @kirutthika!

Before we review this PR, could you please take a moment to address my comment in https://github.com/spiffe/spire/issues/1239#issuecomment-558845537? I'm not opposed to adding the feature, however I'd like to understand a little bit more about what problems you're aiming to solve, as well as ensure we are not opening up any consistency-related security bugs in doing so.

conf/server/server.conf
```diff
@@ -19,6 +19,7 @@ plugins {
         plugin_data {
             database_type = "sqlite3"
             connection_string = "./.data/datastore.sqlite3"
```
zmt: Should we be pedantic and rename this `rw_connection_string`? That would be noise in this commit, so I suggest doing it separately if we decide to do so.

pkg/server/plugin/datastore/sql/sql.go
```diff
@@ -549,22 +584,24 @@ func (*SQLPlugin) GetPluginInfo(context.Context, *spi.GetPluginInfoRequest) (*sp
...
 }
 
 func (ds *SQLPlugin) withReadTx(ctx context.Context, op func(tx *gorm.DB) error) error {
-	return ds.withTx(ctx, op, true, nil)
-}
-
-func (ds *SQLPlugin) withTx(ctx context.Context, op func(tx *gorm.DB) error, readOnly bool, opts *sql.TxOptions) error {
+	if ds.roDb != nil {
+		return ds.withTx(ctx, ds.roDb, op, true, nil)
+	}
```
zmt: Obviously, we need to enable use of the read-only connection to test any performance gains. However, pending resolution of open questions about consistency-sensitive operations in #1239, we should probably leave this using the read/write connection in this commit. It might not wind up being this simple if we need special logic for some reads not to use r/o.

pkg/server/plugin/datastore/sql/sql_test.go
```diff
@@ -7,6 +7,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/spiffe/spire/pkg/common/util"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
zmt: This appears to be a misconfiguration in IDE for `go imports`.

pkg/server/plugin/datastore/sql/mysql.go
```diff
@@ -93,7 +93,24 @@ func hasTLSConfig(cfg *configuration) bool {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	if cfg.RoConnectionString != "" {
+		err = validateConfig(cfg.RoConnectionString)
+	}
+
+	return err
```
zmt: nit: idiomatic `go` would be:
```lang=go
	if cfg.ConnectionString != "" {
		if err = validateConfig(cfg.ConnectionString); err != nil {
			return err
		}
	}
	if cfg.RoConnectionString != "" {
		if err = validateConfig(cfg.RoConnectionString); err != nil {
			return err
		}
	}
	return nil
```

pkg/server/plugin/datastore/sql/sql.go
```diff
@@ -506,41 +508,74 @@ func (ds *SQLPlugin) Configure(ctx context.Context, req *spi.ConfigureRequest) (
 	ds.mu.Lock()
 	defer ds.mu.Unlock()
 
-	if ds.db == nil ||
-		config.ConnectionString != ds.db.connectionString ||
-		config.DatabaseType != ds.db.databaseType {
+	err := ds.openConnection(config, config.ConnectionString, false)
+	if err != nil  {
+		return nil, err
+	}
```
zmt: nit: use the syntactic sugar to reduce scope of `err` here:
```lang=go
	if err := ds.openConnection(config, config.ConnectionString, false); err != nil  {
		return nil, err
	}
```

pkg/server/plugin/datastore/sql/sql.go
```diff
@@ -506,41 +508,74 @@ func (ds *SQLPlugin) Configure(ctx context.Context, req *spi.ConfigureRequest) (
...
+
+	if config.RoConnectionString == "" {
+		return &spi.ConfigureResponse{}, nil
+	}
+
+	err = ds.openConnection(config, config.RoConnectionString, true)
+	if err != nil  {
+		return nil, err
+	}
```
zmt: nit: here as well
```lang=go
	if err := ds.openConnection(config, config.RoConnectionString, false); err != nil  {
		return nil, err
	}
```

pkg/server/plugin/datastore/sql/sql.go
```diff
@@ -506,41 +508,74 @@ func (ds *SQLPlugin) Configure(ctx context.Context, req *spi.ConfigureRequest) (
...
-	ds.db.LogMode(config.LogSQL)
+	if readOnly {
+		ds.roDb = sqlDb
+	} else {
+		ds.db = sqlDb
+	}
 
-	return &spi.ConfigureResponse{}, nil
+	sqlDb.LogMode(config.LogSQL)
+	return nil
```
zmt: nit: avoid `else` wherever possible in `go`
```lang=go
	sqlDb.LogMode(config.LogSQL)
	if readOnly {
		ds.roDb = sqlDb
		return nil
	}
	ds.db = sqlDb
	return nil
```

kirutthika:
Closing this in favor of https://github.com/spiffe/spire/pull/1363

PR #1347, opened 2020-01-21
add agent manager Store metrics

amoore877:
**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Agent Sync Manager metrics

**Description of change**
Add `CallCounter` metrics for Agent Manager storing of SVID, Key, and Bundle.
No tests for Agent Manager metrics currently exist; unit tests are just using `Blackhole`.

These metrics can be helpful for alerting on / triaging issues with the related functionality.

**Which issue this PR fixes**
N/A


pkg/agent/manager/manager.go
```diff
@@ -235,31 +236,41 @@ func (m *manager) runBundleObserver(ctx context.Context) error {
...
 func (m *manager) storeSVID(svidChain []*x509.Certificate) {
-	err := StoreSVID(m.svidCachePath, svidChain)
+	var err error
+	counter := agent_telemetry.StartManagerStoreSVIDCall(m.c.Metrics)
+	defer counter.Done(&err)
+	err = StoreSVID(m.svidCachePath, svidChain)
 	if err != nil {
 		m.c.Log.WithError(err).Warn("could not store SVID")
 	}
 }
```
amoore877: in these cases, the latency metric is not quite useful for anything IMO. What would we think about creating a new Telemetry package functionality for `Counters` (not `CallCounters`) that can indicate success status, much in the way as `CallCounter`~Done does today? (https://github.com/spiffe/spire/blob/master/pkg/common/telemetry/call.go#L66)

pkg/agent/manager/manager.go
```diff
@@ -235,31 +236,41 @@ func (m *manager) runBundleObserver(ctx context.Context) error {
...
+
 	km := m.c.Catalog.GetKeyManager()
 	keyBytes, err := x509.MarshalECPrivateKey(key)
 	if err != nil {
 		return err
 	}
 
-	if _, err := km.StorePrivateKey(ctx, &keymanager.StorePrivateKeyRequest{PrivateKey: keyBytes}); err != nil {
+	if _, err = km.StorePrivateKey(ctx, &keymanager.StorePrivateKeyRequest{PrivateKey: keyBytes}); err != nil {
 		m.c.Log.WithError(err).Error("could not store new agent key pair")
```
amoore877: how long is this backwards compatibility going to be needed?

pkg/common/telemetry/agent/manager.go
```diff
@@ -14,6 +14,24 @@ func StartManagerFetchUpdatesCall(m telemetry.Metrics) *telemetry.CallCounter {
 	return telemetry.StartCall(m, telemetry.Manager, telemetry.Sync, telemetry.FetchUpdates)
 }
 
+// StartManagerStoreBundleCall returns metric for when agent's
+// synchronization manager stores its bundle due to update.
+func StartManagerStoreBundleCall(m telemetry.Metrics) *telemetry.CallCounter {
+	return telemetry.StartCall(m, telemetry.Manager, telemetry.Store, telemetry.Bundle)
+}
```
evan2645: These three points feel a little excessive, and I wonder how much value they will have as a duration. If the goal is monitoring/triage around agent SVID rotation (or lack thereof), feels like all we really need is a `agent_svid_rotated` counter that only fires on success?

amoore877: I think that mostly fits with my previous comment, at least as far as durations are concerned.
However, for the rest:

- A counter that only fires on success is not very meaningful without also having a failure count. With only a success counter, absence of data is ambiguous (were attempts actually made?)
- Each of these are events that can and do independently fail, so having observability into each rather than just the hottest path is helpful

pkg/agent/manager/manager.go
evan2645: Yikes looks like we forgot to yank this one. It's safe to back it out now if you want

pkg/common/telemetry/agent/manager.go
evan2645: > alternatively these particular functions may be small enough to just have in-function logic for status, rather than abstracting to telemetry package

IIRC the biggest reason for centralizing emission was to keep naming etc organized... so even if the logic is small/compartmentalized I think it should probably still live in the telemetry package(s) somewhere.

> A counter that only fires on success is not very meaningful without also having a failure count. With only a success counter, absence of data is ambiguous (were attempts actually made?)

Fair enough. `agent_svid_rotated.error`? Or tag with error bool? I'm pretty sure we've set a precedent here elsewhere

> Each of these are events that can and do independently fail, so having observability into each rather than just the hottest path is helpful

IMO we don't need to emit every small action and error condition as a dedicated timeseries. I have a hard time seeing how it would ever be useful on a dashboard, and managing the alerts for each individual timeseries also feels like high overhead low return. At what point do you simply look at the logs?

amoore877: > Fair enough. agent_svid_rotated.error? Or tag with error bool? I'm pretty sure we've set a precedent here elsewhere

we examine errors for gRPC status codes, nil error having OK status and all else making the attempt at actual status code but defaulting to Unknown

> At what point do you simply look at the logs?
today, I believe consistent failures of these operations would only be noticed once it actually becomes fatal, such as the agent restarting but expected on-disk data being absent or expired (re-attestation being possible only in some environments). Ideally we would detect such problems at an earlier point so intervention is possible.
if the problem is things being too minute, perhaps pulling out / merging some metrics would be good? For instance, a metric just for SVID Observer (combines store private key and store SVID) and Bundle Observer (or abandoning bundle observer metric?).

Underlying issues that can cause these problems are things like the host's disk, or the agent's granted partition of the disk, running out of space, changes in permissions, running out of inodes, or just transient failures (in which case aggregation / retries should prevent false positive alerts). Some of these issues are not necessarily detectable through dedicated host monitoring alone without being very specific, so it is easier to be able to find a problem early via the software itself and then triage the host, rather than think of and monitor for every host-specific problem.

pkg/agent/manager/manager.go
amoore877: I'll make that a separate PR

amoore877: #1367

amoore877:
other end-to-end / usage metrics + logs should be sufficient

PR #1353, opened 2020-01-22
Allow configuring agent sync interval

mcpherrinm:
Pull Request check list

 - [x] Documentation updated?

Affected functionality
SPIRE Agent syncing

Description of change
This exposes a new sync_interval configuration option in SPIRE Agent. It is a string parsed as a time.Duration.

### DIFF
--- a/cmd/spire-agent/cli/run/run_test.go
+++ b/cmd/spire-agent/cli/run/run_test.go
@@ -657,7 +657,7 @@ func TestNewAgentConfig(t *testing.T) {
 		{
 			msg: "sync_interval parses a duration",
 			input: func(c *config) {
-				c.Agent.SyncInterval = "2s45ms"
+				c.Agent.Experimental.SyncInterval = "2s45ms"
 			},
 			test: func(t *testing.T, c *agent.Config) {
 				require.EqualValues(t, 2045000000, c.SyncInterval)
@@ -667,7 +667,7 @@ func TestNewAgentConfig(t *testing.T) {
 			msg:         "invalid sync_interval returns an error",
 			expectError: true,
 			input: func(c *config) {
-				c.Agent.SyncInterval = "moo"
+				c.Agent.Experimental.SyncInterval = "moo"
 			},
 			test: func(t *testing.T, c *agent.Config) {
 				require.Nil(t, c)
--- a/cmd/spire-agent/cli/run/run_test.go
+++ b/cmd/spire-agent/cli/run/run_test.go
@@ -654,6 +654,25 @@ func TestNewAgentConfig(t *testing.T) {
 				require.Nil(t, c)
 			},
 		},
+		{
+			msg: "sync_interval parses a duration",
+			input: func(c *config) {
+				c.Agent.SyncInterval = "2s45ms"
+			},
+			test: func(t *testing.T, c *agent.Config) {
+				require.EqualValues(t, 2045000000, c.SyncInterval)
+			},
+		},
+		{
+			msg:         "invalid sync_interval returns an error",
+			expectError: true,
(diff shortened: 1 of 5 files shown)


mcpherrinm:
While the default's likely fine, It would be helpful to be able to scale up and down load (eg, in an emergency where we hit an unexpected overload, dropping from 5s to 10s or 30s is a potentially good hot fix for us)

evan2645:
I am curious to hear more about your motivation to turn this knob. I don't see any problems with the change itself, however I'm generally wary of adding configurables that can induce a negative experience when turned inappropriately (one example could be a deployment using k8s-workload-registrar that sporadically fails to converge in time after goosing the sync interval).

I can understand the desire to tune this variable based on the cost associated with the entry lookups we do as a result, but on the other hand, it makes me wonder if reducing lookup overhead is a better solution (albeit an admittedly more difficult one).

>It would be helpful to be able to scale up and down load (eg, in an emergency where we hit an unexpected overload, dropping from 5s to 10s or 30s is a potentially good hot fix for us)

We've recently put in context deadlines and retry backoff in the agent logic to deal with this kind of condition. Do you think there are cases where that's insufficient?

mcpherrinm:
Backoffs and retries would help in the case where the agent notices the server is overloaded, but that may not help in all circumstances.

For example, if we're approaching the limits of our CPU usage on a database server, we may take several hours or even have to wait for the next day to provision a new server with more CPU capacity -- we're not fully in the cloud.  Being able to set the sync interval to 10s would potentially be a critical outlet to reduce load.

Support for read replicas is another case that would help us a lot, since there's way more read than write traffic in SPIRE and adding horizontal scale is a bit easier than vertical scale.  I see that's ongoing work in another PR.

SPIRE is by far the busiest database at Square, and not having a knob to turn here makes me incredibly operationally concerned.  I'd love to improve it (for example by reducing load in polling, as discussed on slack earlier today), but we need short-term incident response tools in any circumstance.

evan2645:
>  I'd love to improve it (for example by reducing load in polling, as discussed on slack earlier today), but we need short-term incident response tools in any circumstance.

Fair enough. If/when solved in the future, would you be open to deprecating this configurable? Seems that if the call was low cost it wouldn't matter as much as it does currently. Relatedly, how would you feel about leaving this configurable undocumented?

mcpherrinm:
I'd be fine leaving it undocumented, and maybe tucking it away in an advanced or experimental config section?

evan2645:
>I'd be fine leaving it undocumented, and maybe tucking it away in an advanced or experimental config section?

Using the existing experimental section sounds good to me

APTy:
Big +1 to adding this as experimental so we can deprecate it when it's no longer a necessity!

mcpherrinm:
There's an experimental section in server config but not in the agent.  I'll add one that mirrors it, and leave it undocumented.

mcpherrinm:
Another argument in favor of making it experimental:  If the sync model ever changes, the config value might have no sensible meaning anymore (say, the server supported used a grpc push and there wasn't any polling on the agent side)

cmd/spire-agent/cli/run/run_test.go
```diff
@@ -654,6 +654,25 @@ func TestNewAgentConfig(t *testing.T) {
...
+			},
+			test: func(t *testing.T, c *agent.Config) {
+				require.EqualValues(t, 2045000000, c.SyncInterval)
+			},
+		},
+		{
+			msg:         "invalid sync_interval returns an error",
+			expectError: true,
+			input: func(c *config) {
+				c.Agent.Experimental.SyncInterval = "moo"
```
evan2645: moo

Issue #1060, opened 2019-08-12
log to multiple locations + stdout/stderr


amoore877:
It would be nice to be able to configure logging to multiple file locations + have stdout/stderr still as an option. This would follow similar pattern as the telemetry configuration, which allows multiple metrics plugins at once.

amoore877:
still does print to `std` regardless; excellent :)

amoore877:
creating different issue specifically for the multiple locations, ignoring stdout/err to keep focus

PR #1391, opened 2020-02-13
add multiple-file logging to logging package

amoore877:

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Common Logging package configuration

**Description of change**
Instead of allowing only a single file destination, use hooks instead to allow multiple file destinations. Also add a `stdout` option.
This adjusts only the logging package in a non-breaking way. No parts of SPIRE use the new functionality.

**Which issue this PR fixes**
Enables, but does not fix, #1060 and #1074


pkg/common/log/options.go
```diff
@@ -9,38 +9,46 @@ import (
...
+	JSONFormat = "JSON"
+	// TextFormat signifies Text logging format
+	TextFormat = "TEXT"
 )
 
 // An Option can change the Logger to apply desired configuration in NewLogger
 type Option func(*Logger) error
 
+// WithStdOut configures logger to log with a `stdout` hook
+func WithStdOut() Option {
```
APTy: I would also be fine with providing the special string `"stdout"` or `"stderr"` and interpreting that in `WithOutputFile()`, zap does something similar https://godoc.org/go.uber.org/zap#Open

amoore877: it's slightly more hard to read I think (though could be missing something with os/io libs), as we'd have to have special handling of the constants both in `WithOutputFile` (early return, don't try to open the file) and in the Logger construction or closer (don't try to close it)

I'm ultimately fine with either, just saying why I went with this more explicit option

pkg/common/log/log.go
```diff
@@ -2,32 +2,90 @@ package log
...
+		logger.AddHook(&writerHook{
+			writer:    os.Stdout,
+			logLevels: hookLevels,
+		})
+	}
+
+	if len(logger.Hooks) == 0 {
+		// no hooks (outputs) have been defined, default to standard out
+		logger.SetOutput(os.Stdout)
+	} else {
```
zmt: nit: unnecessary else - should just return early inside the if

evan2645:
Thanks for this @amoore877! I'm curious to hear a little more about your requirements here

Log rotation can be pretty tricky business. In early SPIRE development, there was a decision to lean heavily towards STDOUT/STDERR as systemd et al solved a lot of the log management problems this way. Is there something you're trying to do that isn't already well solved by systemd or other supervisory softwares?

zmt:
> Thanks for this @amoore877! I'm curious to hear a little more about your requirements here
> 
> Log rotation can be pretty tricky business.

Agreed.
> In early SPIRE development, there was a decision to lean heavily towards STDOUT/STDERR as systemd et al solved a lot of the log management problems this way. Is there something you're trying to do that isn't already well solved by systemd or other supervisory softwares?

I am a fan of this factor of 12factor: https://12factor.net/logs. I am uncertain what obstacles caused us to configure spire-agent logging to a file in a well-known location, but enabling logging to both stdout and a file at a well-known location would help any transition to systemd-journald (or similar).

evan2645:
>I am a fan of this factor of 12factor: https://12factor.net/logs. I am uncertain what obstacles caused us to configure spire-agent logging to a file in a well-known location, but enabling logging to both stdout and a file at a well-known location would help any transition to systemd-journald (or similar).

Fully agreed. We do have support for logging to a file currently, with exactly that intention.

This change introduces the ability to log to _multiple_ log files, which is why I ask this question. I don't think there is any problem with the support per se, but it does add "another thing" that will need to be maintained, and I wonder if this change is something which is actually needed/useful or something that is better solved in a different way.

If there are Good Reasons to log to multiple files, I further wonder if this PR goes "far enough" as there doesn't seem to be any flexibility in saying what goes into which files etc... but you can see how we just slip further down the dealing-with-log-files path

amoore877:
closing as out of scope

PR #1395, opened 2020-02-18
server/handler: Log agent spiffe id on attestation failure

APTy:
When looking at server logs, encountering the `Agent is not attested or no longer valid` error is not sufficient to debug a faulty agent without looking at other data sources (metrics, agent logs). By adding the agent's SPIFFE ID on a best-effort basis, it becomes easier to look at server logs and understand the issue.


mcpherrinm:
Thanks! This has been bugging me a bit :)

PR #1406, opened 2020-02-24
Add compatibility layer to SPIRE Server catalog for UpstreamAuthority

### LINKED ISSUE #1374: Add compatibility layer to SPIRE Server catalog for UpstreamAuthority

MarcosDY:
**Description of change**

Add PortedUpstreamCA list to keep ported UpstreamCAs list
Munge HCL configuration, to allow ported UpstreamCA keep load as UpstreamCA, but it is really an UpstreamAuthority.
catelog sever Load method now, support UpstreamAuthority.
In case UpstreamAuthority and UpstreamCA are configured at same time, Load() returns an error

**Which issue this PR fixes**
fixes #1374

### DIFF
--- a/pkg/server/catalog/catalog.go
+++ b/pkg/server/catalog/catalog.go
@@ -2,6 +2,7 @@ package catalog
 
 import (
 	"context"
+	"errors"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spiffe/spire/pkg/common/catalog"
@@ -36,36 +37,10 @@ import (
 	up_spire "github.com/spiffe/spire/pkg/server/plugin/upstreamca/spire"
 )
 
-type Catalog interface {
-	GetDataStore() datastore.DataStore
-	GetNodeAttestorNamed(name string) (nodeattestor.NodeAttestor, bool)
-	GetNodeResolverNamed(name string) (noderesolver.NodeResolver, bool)
-	GetKeyManager() keymanager.KeyManager
-	GetNotifiers() []Notifier
-	GetUpstreamAuthority() (upstreamauthority.UpstreamAuthority, bool)
-}
+var (
+	portedUpstreamCA = map[string]bool{}
 
-type GlobalConfig = catalog.GlobalConfig
-type HCLPluginConfig = catalog.HCLPluginConfig
-type HCLPluginConfigMap = catalog.HCLPluginConfigMap
-
-func KnownPlugins() []catalog.PluginClient {
-	return []catalog.PluginClient{
-		datastore.PluginClient,
-		nodeattestor.PluginClient,
-		noderesolver.PluginClient,
-		upstreamca.PluginClient,
-		keymanager.PluginClient,
-		notifier.PluginClient,
-	}
-}
-
-func KnownServices() []catalog.ServiceClient {
-	return []catalog.ServiceClient{}
-}
-
-func BuiltIns() []catalog.Plugin {
-	return []catalog.Plugin{
+	builtIn = []catalog.Plugin{
 		// DataStores
 		ds_sql.BuiltIn(),
 		// NodeAttestors
@@ -93,6 +68,39 @@ func BuiltIns() []catalog.Plugin {
 		no_k8sbundle.BuiltIn(),
 		no_gcs_bundle.BuiltIn(),
 	}
+)
+
+type Catalog interface {
+	GetDataStore() datastore.DataStore
+	GetNodeAttestorNamed(name string) (nodeattestor.NodeAttestor, bool)
+	GetNodeResolverNamed(name string) (noderesolver.NodeResolver, bool)
+	GetKeyManager() keymanager.KeyManager
+	GetNotifiers() []Notifier
+	GetUpstreamAuthority() (upstreamauthority.UpstreamAuthority, bool)
+}
+
+type GlobalConfig = catalog.GlobalConfig
+type HCLPluginConfig = catalog.HCLPluginConfig
+type HCLPluginConfigMap = catalog.HCLPluginConfigMap
+
+func KnownPlugins() []catalog.PluginClient {
+	return []catalog.PluginClient{
+		datastore.PluginClient,
+		nodeattestor.PluginClient,
+		noderesolver.PluginClient,
+		upstreamauthority.PluginClient,
+		upstreamca.PluginClient,
+		keymanager.PluginClient,
+		notifier.PluginClient,
+	}
+}
+
+func KnownServices() []catalog.ServiceClient {
+	return []catalog.ServiceClient{}
+}
+
+func BuiltIns() []catalog.Plugin {
+	return builtIn
 }
 
 type Notifier struct {
@@ -108,8 +116,7 @@ type Plugins struct {
 	KeyManager    keymanager.KeyManager
 	Notifiers     []Notifier
 
-	// It is unexported to prevent to be processed by Fill, it is handled by ourselves
-	upstreamAuthority upstreamauthority.UpstreamAuthority
+	UpstreamAuthority *upstreamauthority.UpstreamAuthority
 }
 
 var _ Catalog = (*Plugins)(nil)
@@ -137,7 +144,10 @@ func (p *Plugins) GetNotifiers() []Notifier {
 }
 
 func (p *Plugins) GetUpstreamAuthority() (upstreamauthority.UpstreamAuthority, bool) {
-	return p.upstreamAuthority, p.upstreamAuthority != nil
+	if p.UpstreamAuthority != nil {
+		return *p.UpstreamAuthority, true
+	}
+	return nil, false
 }
 
 type Config struct {
@@ -155,8 +165,27 @@ type Repository struct {
 	catalog.Closer
 }
 
+// updatePluginConfig update ported UpstreamCA plugins into UpstreamAuthority
+func updatePluginConfig(pluginConfig catalog.HCLPluginConfigMap, log logrus.FieldLogger) {
+	for k, v := range pluginConfig[upstreamca.Type] {
+		if portedUpstreamCA[k] {
+			pluginConfig[upstreamauthority.Type] = map[string]catalog.HCLPluginConfig{
+				k: v,
+			}
+			log.Warnf("The UpstreamCA plugin %q is now an UpstreamAuthority plugin. "+
+				"Please update your configuration to load UpstreamAuthority %q instead.", k, k)
+			delete(pluginConfig[upstreamca.Type], k)
+			if len(pluginConfig[upstreamca.Type]) == 0 {
+				delete(pluginConfig, upstreamca.Type)
+			}
+		}
+	}
+}
+
 func Load(ctx context.Context, config Config) (*Repository, error) {
-	pluginConfig, err := catalog.PluginConfigFromHCL(config.PluginConfig)
+	updatePluginConfig(config.PluginConfig, config.Log)
+
+	pluginConfigs, err := catalog.PluginConfigFromHCL(config.PluginConfig)
 	if err != nil {
 		return nil, err
 	}
@@ -165,7 +194,7 @@ func Load(ctx context.Context, config Config) (*Repository, error) {
 	closer, err := catalog.Fill(ctx, catalog.Config{
 		Log:           config.Log,
 		GlobalConfig:  config.GlobalConfig,
-		PluginConfig:  pluginConfig,
+		PluginConfig:  pluginConfigs,
 		KnownPlugins:  KnownPlugins(),
 		KnownServices: KnownServices(),
 		BuiltIns:      BuiltIns(),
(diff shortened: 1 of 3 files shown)


pkg/server/catalog/catalog.go
```diff
@@ -36,6 +37,39 @@ import (
 	up_spire "github.com/spiffe/spire/pkg/server/plugin/upstreamca/spire"
 )
 
+var (
+	portedUpstreamCA = map[string]bool{}
+
+	builtIn = []catalog.Plugin{
```
martincapello: I'm not sure about the impact of using this variable. Before this change, the array was returned inside the BuiltIns function, which in turn called each BuiltIn function of each package, so initialization of the array was done every time the BuiltIns function was called. Now, the BuiltIns functions will be called only on initialization time...which might be a good thing, but I just want to confirm that we are all on the same page.

pkg/server/catalog/catalog.go
```diff
@@ -155,8 +165,37 @@ type Repository struct {
 	catalog.Closer
 }
 
+// parsePortedUpstreamCA update configuration, parsing ported UpstreamCA plugins into UpstreamAuthority
+func parsePortedUpstreamCA(pluginConfig catalog.HCLPluginConfigMap, log logrus.FieldLogger) error {
```
martincapello: What do you think if we rename this function to "portUpstreamCA"? I don't see that the "parse" prefix fits well in this case.

pkg/server/catalog/catalog_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+func TestLoad(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	// Create a configuration with minimal required plugins
+	hclConfig := createDefaultConfig()
+
+	// Create a custom list of BuiltIn plugins, it contains the minimal necessary plugins, and fakes for Datastore, UpstreamCA and UpstreamAuthority
+	builtIn = []catalog.Plugin{
```
azdagron: Creating/using fakes is easy enough I'd rather not use any real built-in plugin implementations here.  Also, the only required plugins are datastore and key manager. We don't need the rest.

pkg/server/catalog/catalog_test.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/spiffe/spire/test/fakes/fakeagentstore"
+	"github.com/spiffe/spire/test/fakes/fakedatastore"
+	"github.com/spiffe/spire/test/fakes/fakeidentityprovider"
+	"github.com/spiffe/spire/test/fakes/fakemetrics"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoad(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
azdagron: Seeing this here makes me wonder what operation could possibly time out in the test...?

pkg/server/catalog/catalog_test.go
```diff
@@ -0,0 +1,252 @@
...
+func createDefaultConfig() HCLPluginConfigMap {
+	return HCLPluginConfigMap{
+		datastore.Type:    map[string]HCLPluginConfig{"sql": {}},
+		nodeattestor.Type: map[string]HCLPluginConfig{"join_token": {}},
+		noderesolver.Type: map[string]HCLPluginConfig{"noop": {}},
+		keymanager.Type:   map[string]HCLPluginConfig{"memory": {}},
+	}
+}
+
+type fakeUpstreamAuthorityPlugin struct {
```
azdagron: Since the catalog only uses the `Configure` method on these plugin implementations, the server interface can be embedded to drastically simplify the implementation (no need to define the unused functions).

```suggestion
type fakeUpstreamAuthority struct { upstreamauthority.UpstreamAuthorityServer }

func (fakeUpstreamAuthority) Config(context.Context, *spi.ConfigureRequest) (*spi.ConfigureResponse, error) {
```

pkg/server/catalog/catalog_test.go
```diff
@@ -0,0 +1,252 @@
...
+				c[upstreamca.Type] = map[string]HCLPluginConfig{"testUpstream": {}}
+				return c
+			},
+			verifyRepo: func(t *testing.T, repository *Repository) {
+				// An upstreamAuthority expected and it is calling fakeUpstreamCA method
+				upstream, ok := repository.Catalog.GetUpstreamAuthority()
+				require.True(t, ok)
+				require.NotNil(t, upstream)
+
+				_, err := upstream.MintX509CA(ctx, &upstreamauthority.MintX509CARequest{})
```
azdagron: just assert that `upstream` isn't a wrapper type... no need to invoke a method here

pkg/server/catalog/catalog_test.go
```diff
@@ -0,0 +1,252 @@
...
+				// No UpstreamAuthority expected
+				upstream, ok := repository.Catalog.GetUpstreamAuthority()
+				require.False(t, ok)
+				require.Nil(t, upstream)
+			},
+		},
+		{
+			name: "unported UpstreamCA",
+			createHclConfig: func() HCLPluginConfigMap {
+				c := hclConfig
```
azdagron: maps are reference types... future modifications to `c` will mutate `hclConfig`. You should initialize this with `createDefaultConfig()` instead.

(most of the `createHCLConfig` methods have this problem)

pkg/server/catalog/catalog_test.go
```diff
@@ -0,0 +1,252 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	// Create a configuration with minimal required plugins
+	hclConfig := createDefaultConfig()
+
+	// Create a custom list of BuiltIn plugins, it contains the minimal necessary plugins, and fakes for Datastore, UpstreamCA and UpstreamAuthority
+	builtIn = []catalog.Plugin{
+		// Fake Datastore to prevent sql connections
+		catalog.MakePlugin("sql",
```
azdagron: I'd prefer names like `fake_ds` or something on these fakes. I don't want to give the impression we are loading real plugins.

pkg/server/catalog/catalog.go
```diff
@@ -155,8 +165,37 @@ type Repository struct {
...
+	// We only expect one UpstreamCA configuration
+	for k, v := range pluginConfig[upstreamca.Type] {
```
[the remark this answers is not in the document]
MarcosDY: done

pkg/server/catalog/catalog.go
```diff
@@ -155,8 +165,37 @@ type Repository struct {
...
+		// in case configured UpstreamCA is ported update configuration to process it as an UpstreamAuthority
+		if portedUpstreamCA[k] {
```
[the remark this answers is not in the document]
MarcosDY: done

pkg/server/catalog/catalog.go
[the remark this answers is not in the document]
MarcosDY: done

pkg/server/catalog/catalog.go
```diff
@@ -155,8 +165,37 @@ type Repository struct {
...
+			if _, ok := pluginConfig[upstreamauthority.Type]; ok {
+				return errors.New("an UpstreamAuthority already exists in plugin's configuration")
```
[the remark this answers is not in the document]
MarcosDY: done

pkg/server/catalog/catalog.go
```diff
@@ -155,8 +165,37 @@ type Repository struct {
...
+			}
+			// Create upstream authority type entry
+			pluginConfig[upstreamauthority.Type] = map[string]catalog.HCLPluginConfig{
+				k: v,
+			}
+			log.Warnf("The UpstreamCA plugin %q is now an UpstreamAuthority plugin. "+
+				"Please update your configuration to load UpstreamAuthority %q instead.", k, k)
+			delete(pluginConfig[upstreamca.Type], k)
+			if len(pluginConfig[upstreamca.Type]) == 0 {
+				delete(pluginConfig, upstreamca.Type)
```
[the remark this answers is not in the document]
MarcosDY: no code is breaked with that I only wanted it to be clean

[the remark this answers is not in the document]
MarcosDY: I'll remove it

pkg/server/catalog/catalog.go
```diff
@@ -155,8 +165,37 @@ type Repository struct {
...
+		// in case configured UpstreamCA is ported update configuration to process it as an UpstreamAuthority
+		if portedUpstreamCA[k] {
+			if _, ok := pluginConfig[upstreamauthority.Type]; ok {
+				return errors.New("an UpstreamAuthority already exists in plugin's configuration")
+			}
+			// Create upstream authority type entry
+			pluginConfig[upstreamauthority.Type] = map[string]catalog.HCLPluginConfig{
+				k: v,
+			}
+			log.Warnf("The UpstreamCA plugin %q is now an UpstreamAuthority plugin. "+
```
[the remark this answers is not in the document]
MarcosDY: done

pkg/server/catalog/catalog.go
```diff
@@ -180,7 +219,12 @@ func Load(ctx context.Context, config Config) (*Repository, error) {
 	}
 
 	if p.UpstreamCA != nil {
-		p.upstreamAuthority = upstreamauthority.Wrap(*p.UpstreamCA)
+		if p.UpstreamAuthority != nil {
```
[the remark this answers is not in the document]
MarcosDY: it will never enter to `case p.UpstreamCA == nil:` because previous line if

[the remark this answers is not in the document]
azdagron: Sorry, my "suggestion" was on the wrong line. My recommendation applies to replace both `if's` with a single switch.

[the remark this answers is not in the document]
MarcosDY: done

pkg/server/catalog/catalog.go
```diff
@@ -180,7 +219,12 @@ func Load(ctx context.Context, config Config) (*Repository, error) {
...
+			return nil, errors.New("only one UpstreamCA or UpstreamAuthority is allowed. Please remove one of them")
```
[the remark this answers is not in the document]
MarcosDY: done

pkg/server/catalog/catalog.go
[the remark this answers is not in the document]
MarcosDY: done

pkg/server/catalog/catalog_test.go
MarcosDY: done

pkg/server/catalog/catalog_test.go
MarcosDY: none, I'll remove it, it was added for may first approach on those tests, but I moved to fakes and it is no longer needed

pkg/server/catalog/catalog_test.go
MarcosDY: done

pkg/server/catalog/catalog_test.go
MarcosDY: done

pkg/server/catalog/catalog.go
martincapello: I just wanted to make sure that the change in semantics was ok, if it is, then no need to change anything I believe.

pkg/server/catalog/catalog.go
```diff
@@ -155,8 +166,36 @@ type Repository struct {
 	catalog.Closer
 }
 
+// reclassifyPortedUpstreamCAs update configuration, parsing ported UpstreamCA plugins into UpstreamAuthority
```
martincapello: comment got outdated.

PR #1413, opened 2020-02-26
Add SPIRE Server Node API RPC to inject JWKs into the bundle from Downstream servers

### LINKED ISSUE #1372: Add SPIRE Server Node API RPC to inject JWKs into the bundle from Downstream servers
As downstream SPIRE Servers generate new JWKs to be used for signing JWT-SVIDs, they need to be communicated upstream so that the public keys can be present in the bundle of all other SPIRE Servers. Since all nested SPIRE Servers inherit their bundle from the root SPIRE Servers, by pushing new JWKs upstream we can ensure that they propagate downwards throughout the entire topology.

We will need a new RPC on the SPIRE Server Node API in order to do this. Only identities which can be authenticated as "downstream" identities should be allowed to call it.

One catch to this is that the new JWK needs to reach the root SPIRE Servers. This means that if there are more than two levels in the nested topology, then the key needs to be passed upwards (rather than stored in the local bundle). So, if a SPIRE Server receives a JWK via this RPC, it needs to first check whether or not it itself has an UpstreamAuthority configured. If it does, then we need to pass the JWK to it, and pass its response back to our caller. If it _doesn't_ have an UpstreamAuthority configured, then this is a root SPIRE Server, and we should instead append the key to our local bundle.

It could be the case that an UpstreamAuthority plugin is loaded, however the UpstreamAuthority does not support JWK. In this case, we should receive a gRPC NotImplemented error code from the plugin when we attempt to pass it the JWK. When that happens, we want to append the JWK to our local bundle instead of worrying about continuing to pass it upstream. A warning should also be logged in this event, something like "UpstreamAuthority \"foo\" does not support JWT-SVIDs. Workloads managed by this server or downstream servers may have trouble communicating with workloads outside this zone when using JWT-SVIDs." The warning should be logged just once, perhaps guarded by something like sync.Once.

The response emitted by this RPC should include an up-to-date copy of the JWKS bundle, including the newly-appended JWK.

* Ensure that only "downstream" workloads are authorized to call this new RPC
* When called, pass the received JWK to the UpstreamAuthority plugin (if present)
* If UpstreamAuthority is not present, or NotImplemented error is returned, append the received JWK to the local bundle
* The downstream server should receive an up-to-date bundle in the response

Blocked on #1371

martincapello:
Fixes #1372 
Still working a bit on the server's ca manager tests, will push a commit as soon as possible.


martincapello:
Closed because it was not being updated despite having pushed newer commits to my branch.

PR #1417, opened 2020-02-27
Implements the new Node API's service PushJWTKeyUpstream

martincapello:
Impements the Node API's service as defined on #1372


pkg/server/ca/manager.go
```diff
@@ -315,7 +320,32 @@ func (m *Manager) prepareJWTKey(ctx context.Context, slot *jwtKeySlot) (err erro
 		return err
 	}
 
-	if err := m.appendBundle(ctx, nil, publicKey); err != nil {
+	publicKeys := []*common.PublicKey{publicKey}
+	upstreamAuth, useUpstream := m.c.Catalog.GetUpstreamAuthority()
+	if useUpstream {
+		publishCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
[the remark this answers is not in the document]
martincapello: Oh yeah, I will do it. Since there is another place where we use this same value, do you think it worth to have only one constant for both usages? If so, where do you think that constant should be placed?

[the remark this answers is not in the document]
martincapello: Never mind, I have an idea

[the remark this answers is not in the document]
martincapello: Just added the constant in server.go, and passed it as needed. This way only one constant is defined.

pkg/server/ca/manager.go
```diff
@@ -74,6 +77,9 @@ type Manager struct {
 	nextJWTKey    *jwtKeySlot
 
 	journal *Journal
+
+	// Used to log a warning only once.
+	warnOnce sync.Once
```
MarcosDY: can you put a better name here? so we can be sure we are talking about warning when upstream authority does not support jwt?

pkg/server/endpoints/node/handler.go
```diff
@@ -480,7 +488,70 @@ func (h *Handler) FetchJWTSVID(ctx context.Context, req *node.FetchJWTSVIDReques
...
+	}
+
+	err = h.limiter.Limit(ctx, PushJWTKey, 1)
+	if err != nil {
+		log.WithError(err).Error("Rejecting request due to JWK push rate limiting")
+		return nil, status.Error(codes.ResourceExhausted, err.Error())
+	}
+
+	upstreamAuth, useUpstream := h.c.Catalog.GetUpstreamAuthority()
+	if useUpstream { // We are an intermediate SPIRE server
```
MarcosDY: it is pretty the same than we have on manage.prepareJWTKey() and we are exposiing append to do almost the same, may it worth creating a new method that call append form inside manage, and add all validations the same way we are doing in both places

pkg/server/endpoints/node/handler.go
```diff
@@ -480,7 +488,70 @@ func (h *Handler) FetchJWTSVID(ctx context.Context, req *node.FetchJWTSVIDReques
...
+		}
+	}
+
+	if err := h.c.Manager.AppendBundle(ctx, nil, []*common.PublicKey{req.JwtKey}); err != nil {
+		return nil, err
+	}
+
+	// Ensure we delete any cached bundle to get the latest one.
+	h.dsCache.DeleteBundleEntry(h.c.TrustDomain.String())
+	bundle, err := h.getBundle(ctx, h.c.TrustDomain.String())
```
MarcosDY: will it try to remove x509 bundle too?

pkg/server/util/grpccodesutil/checks.go
```diff
@@ -0,0 +1,14 @@
+package grpccodesutil
+
+import (
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func IsUnimplementedError(err error) bool {
```
azdagron: This helper can be implemented in one line (and is probably therefore unnecessary):

```
status.Code(err) == codes.Unimplemented
```

pkg/server/endpoints/node/handler.go
azdagron: I agree. I think there is an opportunity here to say that the manager is in charge of publishing JWT key's upstream, whether that key originated from a downstream server, or the server itself during rotation. The manager can do the same steps either way:
1) try and publish and warn if unsupported
2) append returned JWT keys into the bundle
3) return the appended bundle (which manager internals can ignore, but this RPC handler can use as the response)

pkg/server/endpoints/node/handler.go
```diff
@@ -480,7 +488,70 @@ func (h *Handler) FetchJWTSVID(ctx context.Context, req *node.FetchJWTSVIDReques
...
+		default:
+			// Send back to the caller the received JWKs from upstream,
+			// this will change soon according to https://github.com/spiffe/spire/issues/1372#issuecomment-589469449
+			return &node.PushJWTKeyUpstreamResponse{
+				JwtSigningKeys: res.UpstreamJwtKeys,
+			}, nil
+		}
+	}
+
+	if err := h.c.Manager.AppendBundle(ctx, nil, []*common.PublicKey{req.JwtKey}); err != nil {
```
azdagron: I wonder if this call method can return the appended bundle. Then we can just return it directly and not bother with the extra call to the datastore.

We might still want to evict the cache since we know it changed.

_Wonders if we should have provided a way for the manager to invalidate the node cache.... anyway that can wait for another PR_

pkg/server/ca/manager.go
```diff
@@ -220,7 +226,7 @@ func (m *Manager) prepareX509CA(ctx context.Context, slot *x509CASlot) (err erro
 		return err
 	}
 
-	if err := m.appendBundle(ctx, trustBundle, nil); err != nil {
+	if err := m.AppendBundle(ctx, trustBundle, nil); err != nil {
```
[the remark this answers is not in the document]
martincapello: I have exported the CA Manager's appendBundle function because I needed to append a JWK to the bundle from the Node API handler (github.com/spiffe/spire/pkg/server/endpoints/node/handler.go). My first thought was to use the Datastore plugin directly, but then I realized that the CA Manager does some notification stuff that I wouldn't get triggered if I did that. So, it seemed to me that exporting the function was the way to go. I'm open to suggestion on how to improve (or reimplement) this, and happy to make all the necessary changes.

pkg/server/ca/manager.go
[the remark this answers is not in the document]
martincapello: I put some thought on the name of this...and actually this was the best I got...
Something like warnOnceUpstreamAuthorityDoesNotSupportJWT would work? :eyes: 
Everyone feel free to suggest an alternative name, I will be happy as hell to use it.

pkg/server/endpoints/node/handler.go
martincapello: The datastore cache caches datastore.FetchBundleResponse objetcs, so if you are referring to the Root CA certificates list that are included in common.Bundle, then yes, it will.

pkg/server/ca/manager.go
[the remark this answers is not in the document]
MarcosDY: if we implement a new method to do all validations as suggested before, we can avoid calling append from outside, and hava validations in a single place

pkg/server/ca/manager.go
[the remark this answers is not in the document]
MarcosDY: mmm I was thinking something like jwtUpstremAuthorityWarnOnce or something like that, we'll have another var like this when we implement the same validations on x509

pkg/server/endpoints/node/handler.go
MarcosDY: can it affect us on somthing? basically we are updating RootCAs when calling an endpoint

pkg/server/endpoints/node/handler.go
martincapello: The logic in both cases is not exactly the same...and this was done according to the requirements defined on https://github.com/spiffe/spire/issues/1372. But I believe it makes sense to have the same behavior for each case, so I will make the suggested changes anyway. Also this is likely to change due to some issues when dealing with clusters of spire servers that are on a different branch than the one the JWKs traverses.

pkg/server/ca/manager.go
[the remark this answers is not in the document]
MarcosDY: or jwtUnimplementedWarnOnce, then we can have another x509UnimplementedWarnOnce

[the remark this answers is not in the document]
martincapello: Alright...I will make the change, but personally I would prefer to have a warnOnce1 and warnOnce2 with a nice comment on their declarations.

pkg/server/endpoints/node/handler.go
martincapello: I don't believe so, it just invalidates an entry in a cache, so next time a caller uses dsCache the entry, it will be created from fresh data.

pkg/server/endpoints/node/handler.go
martincapello: > I wonder if this call method can return the appended bundle. Then we can just return it directly and not bother with the extra call to the datastore.

Yes it can! Thanks :smiley: 

> We might still want to evict the cache since we know it changed.

That's right.

> Wonders if we should have provided a way for the manager to invalidate the node cache.... anyway that can wait for another PR

Maybe...

> anyway that can wait for another PR

:tada:

pkg/server/ca/manager.go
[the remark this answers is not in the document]
martincapello: Yep, this is included on my next commit that I will push soon.

evan2645:
The overall logic etc looks good to me!

pkg/server/ca/manager.go
```diff
@@ -334,6 +342,53 @@ func (m *Manager) prepareJWTKey(ctx context.Context, slot *jwtKeySlot) (err erro
...
+		res, err := upstreamAuth.PublishJWTKey(
+			publishCtx,
+			&upstreamauthority.PublishJWTKeyRequest{
+				JwtKey: publicKey,
+			})
+
+		switch {
+		case status.Code(err) == codes.Unimplemented:
+			m.jwtUnimplementedWarnOnce.Do(func() {
+				m.c.Log.Warn("UpstreamAuthority does not support JWT-SVIDs. Workloads managed " +
```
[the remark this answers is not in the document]
martincapello: This change was bigger than expected, I would recommend to take a look at the commit that includes this.

test/fakes/fakeupstreamauthority/fakeupstreamauthority.go
```diff
@@ -45,7 +47,7 @@ type UpstreamAuthority struct {
 	config     Config
 }
 
-func New(t *testing.T, config Config) *UpstreamAuthority {
+func New(t *testing.T, config Config) *catalog.UpstreamAuthority {
```
azdagron: I think you can revert these changes. The established pattern is to have the fakeservercatalog package provide a helper for constructing a `catalog.UpstreamAuthority` from a `name` and an `upstreamauthority.UpstreamAuthority`. See the following function:
https://github.com/spiffe/spire/blob/4ccc31d85e4d1775c1f0f3d6ee9d07bdc0c9da8a/test/fakes/fakeservercatalog/catalog.go#L55

This makes it easy to use any `upstreamauthority.UpstreamAuthority` implementation without requiring that it also implement `catalog.PluginInfo`.

pkg/server/ca/manager_test.go
```diff
@@ -208,25 +210,27 @@ func (s *ManagerSuite) TestUpstreamIntermediateSignedWithUpstreamBundle() {
 	})
 	s.initUpstreamSignedManager(upstreamAuthority, true)
 
+	root := upstreamAuthority.UpstreamAuthority.(*fakeupstreamauthority.UpstreamAuthority).Root()
```
azdagron: These changes can all be reverted if you follow the suggestion in fakeupstreamauthority.New.

pkg/server/ca/manager_test.go
```diff
@@ -178,24 +179,25 @@ func (s *ManagerSuite) TestUpstreamSignedWithUpstreamBundle() {
...
```
azdagron: These changes can all be reverted if you follow the suggestion in fakeupstreamauthority.New.

PR #1435, opened 2020-03-06
Optimize List Registrations operation in FetchX509SVID by introducing Type field to Registration Entry

### LINKED ISSUE #980: Reduce extra calls to datastore plugin for workload entries

rturner3:
**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Registration entry creation

**Description of change**
There is currently no reliable way in SPIRE to distinguish between node and workload registrations. This PR introduces a new `Type` property on a RegistrationEntry, which has possible values of {`UNKNOWN`, `NODE`, `WORKLOAD`} in order to provide a distinction between different types of registration entries.

Being able to distinguish between node and workload registrations will enable at least two noticeable improvements:

1. Optimization of [regentryutil/fetch](https://github.com/spiffe/spire/blob/master/pkg/server/util/regentryutil/fetch.go) algorithm to avoid recursive registration entry queries when the current registration is a workload registration (see Issue #980). This algorithm causes a large number of unnecessary queries today when the registration entry set is reasonably large, which significantly increases the latency of FetchX509SVID. These extraneous queries have been observed to be a bottleneck in expanding the workload registration set for a node.

1. Ability to query for registrations by a specific type of registration. Most often when listing registrations, a client is only interested in one type of registration, node or workload. Performing this filtering on the query side is non-trivial due to the possibility of large registration entry sets and because the filtering logic today is highly based on consumer-specific SPIFFE ID path conventions.

This PR addresses 1) and lays the foundation for 2) to be done later.

Existing registration entries will assume the default value of `UNKNOWN`. SPIRE Server accepts this Type value of `UNKNOWN` in a Create Registration Entry request, but later may be changed in future releases to not accept it.

These changes also impose two restrictions around the usage of the `Type` field:
1. `WORKLOAD` type registrations cannot be the parent of other registrations. I could not think of a clear use case where such a registration hierarchy would be needed. It is based on this assumption that we can optimize the issue described in #980.
1. The `Type` of a registration cannot be updated. In the real world, a Node never becomes a Workload, or vice versa. There is not a clear reason I could think of for why such a scenario would need to be supported. By not supporting it, we avoid handling a lot of complexity. Some questions that would need to be answered to support updating of Registration Entry `Type` are:

- How does SPIRE Server interpret the `UNKNOWN` value in an Update Registration Entry request? Was it deliberately set to `UNKNOWN` in the request, or was it left unset and Protobuf just used it as the default value? Mostly a problem for the initial releases before SPIRE Server starts enforcing the `Type` value to be set to a non-`UNKNOWN` value.
- What does it mean when a `NODE` registration for an attested Agent is updated to become a `WORKLOAD` registration? Do we evict the Agent? Do we need to make the client aware that their Update request triggered an Agent eviction?

More detailed descriptions of changes in this PR are present in each commit's message.

**Which issue this PR fixes**
fixes #980


evan2645:
Thanks for this pull @rturner3!

I've been chatting a bit with @APTy and @azdagron on this problem in general. It's been a painful one for quite a while and I don't see it going away any time soon. This change definitely helps the situation, but I am not sure that it goes far enough.

As far as performance wins against authz lookups go, I would really like to get https://github.com/spiffe/spire/pull/1439 in and see how we are looking after that. Updating the data model is more difficult and will take longer to see a payoff. I also wonder if the addition of `NODE` and `WORKLOAD` to the entry will cause additional confusion due to the fact that it's not strictly required. There are also use cases in which a workload would need to be labeled `NODE` (for instance, if a workload is authorized to mint subsequent identities, like a helper that lives inside a CI/CD pipeline).

IMO the ideal fix for this problem is a larger change in our data model. These kinds of crawls have proven expensive and the traditional approaches don't seem well suited for them. @azdagron has done some exploration around graph storage which looks promising, but that work is still very early.

We should probably put this change on hold until #1439 is in to see if that is "enough" for now. I don't think that in the long run we'll be able to completely avoid a large change in the data model in... whether or not we want to put this change in before then or not, I guess we need to figure that out. Further, if you have any ideas as to how we can solve this entry relationship crawl problem "once and for all", that would be very useful!

Once and for all!

rturner3:
Thanks for the feedback, @evan2645. I think it's fair to pause before introducing this change given its reach and impact on registrants in the SPIRE ecosystem.

I think #1439 will definitely help minimize this query crawl problem to a certain degree, but I'm concerned it won't be enough for larger registration entry sets. When the number of registrations managed by an Agent is greater than 100, we're still looking at 100+ unnecessary database queries for registration entry set cache misses. For HA deployments of the Server with `N` instances, that could result in `N * 100` unnecessary queries every few seconds (assuming a 5 second Agent sync interval and short-lived cache entries). That being said, I do think it's reasonable to let that change get absorbed first, then evaluate where we are with respect to performance based on some actual observed latencies.

Regarding the scenario you mentioned of a deployment orchestrator minting SVIDs, that is an interesting one that I hadn't considered. In this case, would the registration for the orchestrator have its `Downstream` field set to `true`? If that were so, the registration could still be a `WORKLOAD` type, and we could change the recursion termination condition to:

```go
if !directEntry.Downstream && directEntry.GetType() == common.RegistrationEntryType_WORKLOAD {
    continue
}
```
Delving a little deeper here though, I think the root of these performance problems is this dynamic registration entry graph discovery/traversal performed during each Agent sync. If we could persist all the child + selector subset relationships between registration entries into a single table on create/update/delete of registrations, we might be able to simplify the FetchX509SVID registration entry lookup to only a couple of queries. Doing this would make create/update/delete operations more expensive and complex and would require a little extra storage, but it would heavily optimize the read case, which is generally expected to be the more frequently executed path. My idea here is basically to move the logic of [regentryutil/fetch](https://github.com/spiffe/spire/blob/master/pkg/server/util/regentryutil/fetch.go) into the create/update/delete registration entry code path, then persist the graph of relationships into the database.

We could model the registration entry relationships as a many-to-many relation through a secondary relation table. Here's some DDL showing an example of how the relationship data could be persisted (tested on MySQL 5.6+):

```sql
CREATE TABLE registered_entries (
  id INT NOT NULL AUTO_INCREMENT,
  spiffe_id VARCHAR(2048),
  parent_id VARCHAR(2048),
  PRIMARY KEY (id)
  /* Omitting other DDL of registered_entries here for brevity */
);

CREATE TABLE registered_entry_relation (
  registered_entry1_id INT NOT NULL,
  registered_entry2_id INT NOT NULL,
  PRIMARY KEY (registered_entry1_id, registered_entry2_id),
  FOREIGN KEY (registered_entry1_id) REFERENCES registered_entries(id),
  FOREIGN KEY (registered_entry2_id) REFERENCES registered_entries(id)
);

DROP TRIGGER IF EXISTS registered_entry_relation_order;

/*
    Create a trigger that enforces that registered_entry1_id < registered_entry2_id.
    This prevents possible duplicate relations from being persisted.
*/
DELIMITER //
CREATE TRIGGER registered_entry_relation_order
BEFORE INSERT ON registered_entry_relation 
FOR EACH ROW
BEGIN 
  DECLARE x int; 
  IF (NEW.registered_entry1_id>NEW.registered_entry2_id)
  THEN 
    SET x:=NEW.registered_entry1_id; 
    SET NEW.registered_entry1_id:=NEW.registered_entry2_id; 
    SET NEW.registered_entry2_id:=x; 
  END IF; 
END;//
DELIMITER ;

INSERT INTO registered_entries (parent_id, spiffe_id) VALUES
('spiffe://example.org/spire-server', 'spiffe://example.org/node1'),
('spiffe://example.org/node1', 'spiffe://example.org/workload1'),
('spiffe://example.org/node1', 'spiffe://example.org/workload2');

INSERT INTO registered_entry_relation (registered_entry1_id, registered_entry2_id)
VALUES (2, 1), (3, 1);
```

Then [regentryutil/fetch](https://github.com/spiffe/spire/blob/master/pkg/server/util/regentryutil/fetch.go) could effectively be replaced with the following SQL for the registration entry lookup in FetchX509SVID:

```sql
SELECT * FROM registered_entry_relation
WHERE registered_entry1_id=<AGENT_ID>
OR    registered_entry2_id=<AGENT_ID>;
```

The business logic could process the results of the query and filter them to a distinct set of IDs, then basically do:

```sql
SELECT * FROM registered_entries r
WHERE r.id IN (...)
...
```

Not sure if there's a more sophisticated way with SQL to simplify this further into one query, but still, two queries is much more performant than the current approach.

Long term, I think using a graph database may be a good option if the performance characteristics are reasonable. It makes sense to evaluate it since the registration entry data is inherently modeled as a graph.

rturner3:
Just to clarify a bit further, `registered_entry_relation` would contain all reachable registered entries from a SPIFFE ID, not just direct parent-child relationships. Take the following scenario for example:

- There is an Agent `A` with selectors `S1` mapping to a node registration `N1`.
- There is a selector set `S2`, which is a subset of `S1`.
- There is a node registration `N2` with selectors `S2`.
- There is a workload registration `W1` with parent ID `N1`.
- There is a workload registration `W2` with parent ID `N2`.
- Based on these registrations, `A` should receive the registration entry set `{N2, W1, W2}` when it calls FetchX509SVIDs.

In this scenario, the contents of `registered_entry_relation` would be:
```
{(N1, N2), (N1, W1), (N1, W2)}
```

PR #1445, opened 2020-03-16
Remove spiffeID label from spire agent caching metrics

prasadborole1:
**Pull Request check list**

- [X] Documentation updated?

**Description of change**

Currently spire agent cache manager metrics use spiffeID as label which causes very high cardinality. As the number of spire agents and registrations per agent goes up, it'll flood the telemetry and querying it becomes challenging. 
With this PR, we are removing these metrics:

```
{4 [cache_manager registration_entry create] 1 [{spiffe_id spiffe_domain_test_reg1}] 0001-01-01 00:00:00 +0000 UTC} 
{4 [cache_manager registration_entry create] 1 [{spiffe_id spiffe_domain_test_reg2}] 0001-01-01 00:00:00 +0000 UTC} 
{4 [cache_manager registration_entry delete] 1 [{spiffe_id spiffe_domain_test_reg3}] 0001-01-01 00:00:00 +0000 UTC} 
{4 [cache_manager registration_entry update] 1 [{spiffe_id spiffe_domain_test_reg4}] 0001-01-01 00:00:00 +0000 UTC}
```


pkg/agent/manager/cache/cache.go
```diff
@@ -228,6 +228,10 @@ func (c *Cache) Update(update *Update, checkSVID func(*common.RegistrationEntry,
 	fedRem, fedRemDone := allocStringSet()
 	defer fedRemDone()
 
+	var deleteCounter float32
```
APTy: Rather than have much of this telemetry behavior leak into the caching logic, i'd prefer to see an API like:

```
metricsHelper := newCacheMetricsHelper(c.metrics)
defer metricsHelper.Finish()

...
if _, ok := update.RegistrationEntries[id]; !ok {
    ...
    metricsHelper.IncCacheDelete()
}

...
if existingEntry != nil {
    metricsHelper.IncCacheUpdate()
} else {
    metricsHelper.IncCacheCreate()
}
```

APTy:
Hey @prasadborole1 thanks for this! One of the things about metrics in SPIRE is that the initial layer of metrics were added at various areas in the codebase on the basis of "this seems like a good place for metrics." Now that we have the opportunity to look more closely at certain metrics, like this one in particular, it's a good chance to consider: "what is this metric useful for?"

So, rather than rename the metric (which causes a previous metric to break, and then introduces another metric that now must be supported), it may be good for us to consider whether we can get rid of the metric entirely if it's not useful.

I think this change is fine overall if we have a good use case for the metric, but just want to confirm that. Thanks!

prasadborole1:
> Hey @prasadborole1 thanks for this! One of the things about metrics in SPIRE is that the initial layer of metrics were added at various areas in the codebase on the basis of "this seems like a good place for metrics." Now that we have the opportunity to look more closely at certain metrics, like this one in particular, it's a good chance to consider: "what is this metric useful for?"
> 
> So, rather than rename the metric (which causes a previous metric to break, and then introduces another metric that now must be supported), it may be good for us to consider whether we can get rid of the metric entirely if it's not useful.
> 
> I think this change is fine overall if we have a good use case for the metric, but just want to confirm that. Thanks!

I could not come up with convincing use case for keeping counters of cached entries in spire agent. Removing them and updated description.

pkg/agent/manager/cache/cache.go
prasadborole1: removed this part.

zmt:
@evan2645 Can we get this merged now that it's approved?

PR #1448, opened 2020-03-18
If a bundle doesn't exist yet when creating a federation reg. entry, create it

dfeldman:
**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Creation of federation bundle entries 

**Description of change**
Currently when creating a new federation registration entry, if no bundle is present for the federated SPIFFE ID, the operation fails. This is actually the normal case because typically you don't have the bundle bootstrapped. With this change we will auto-create a bundle. 

Note: These empty bundles will show up in bundle list operations, and won't auto-delete when the federation relationship is deleted. One could argue these should happen. I'm not sure. 

**Which issue this PR fixes**
#1256

### DIFF
--- a/conf/server/server.conf
+++ b/conf/server/server.conf
@@ -17,8 +17,9 @@ server {
 plugins {
     DataStore "sql" {
         plugin_data {
-            database_type = "sqlite3"
-            connection_string = "./.data/datastore.sqlite3"
+            database_type = "postgres"
+#            connection_string = "./.data/datastore.sqlite3"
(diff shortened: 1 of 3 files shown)


evan2645:
@dfeldman and I spoke offline about this change. I was mistaken in my original assumptions when I opened the related issue, and now it seems that we'd be better off not accepting this change. Sorry again for the trouble, please see https://github.com/spiffe/spire/issues/1256#issuecomment-610038303 for more information.

PR #1479, opened 2020-04-06
Cherry pick k8s client upgrade

mcpherrinm:
This cherry-picks part of #1454 

I ran:

    git cherry-pick 745667780f754b114f96632624bcb5b79d97c141
    git reset --soft HEAD^
    git restore --staged support
    git restore --staged Makefile
    go mod tidy
    git add go.sum

This omitted the makefile & k8s-workload-registrar changes, keeping only the go.mod/sum and mock changes


evan2645:
Thank you for this @mcpherrinm

I assume you ran into something that required this bump? It will be useful to know exactly what so we can document in the changelog, backport the commit to 0.9.x, etc.

evan2645:
Looks like I missed a Slack message from @mcpherrinm that said:
```
This makes testing that easier for me, as I don't have to have a separate dependencies for spire vs our pulled in copy of the k8s workload registrar
```

Any idea if the change enables/disables support for certain versions of Kubernetes? Looks fine otherwise!

mcpherrinm:
I want to run upstream spire but use the k8s-workload-registrar forked.  So having a single set of dependencies simplifies that for us

mcpherrinm:
The v0.15.11 client-go will work perfectly with kubernetes v15.

It may include APIs which won't work on older clusters, and older clusters may contain features no longer available in client-go.

This is inevitable unless you pin client-go to the exact cluster version you're using.  Since there were no changes other than upgrading the version, that implies there's no APIs removed that affect SPIRE, or new APIs being used.

It sounds like #1454 wanted some new or changed APIs, so if there's any requirements changes, it would come in via that PR (ie, the informer mode may require a newer kubernetes version than what we have today)

mcpherrinm:
The current client-go version supports kubernetes 1.13.  It is deprecated and in "severe bugs only" maintenance mode.

mcpherrinm:
Turns out I don't need this -- I'm going to close it for now.  We can revisit in #1454

PR #1505, opened 2020-04-15
Fix error in Docker workload attestor documentation

scotte:
Signed-off-by: Scott Emmons <scott@scytale.io>

### DIFF
--- a/doc/plugin_agent_workloadattestor_docker.md
+++ b/doc/plugin_agent_workloadattestor_docker.md
@@ -43,5 +43,16 @@ spire-server entry create \
     -parentID spiffe://example.org/host \
     -spiffeID spiffe://example.org/host/foo \
     -selector docker:label:com.example.name:foo
-    -selector docker:label:com.example.env:prod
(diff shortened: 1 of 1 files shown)


ZymoticB:
That's not an error. This documentation was written by someone working at a company which labelled their container at launch with an "environment". That's a separate concept from the environment variables on that container.

scotte:
In that case, using 'env' as a label example is a poor choice as it conflates the use of `env` between 'env' and 'label' examples. Given that, the example should be rewritten to be more clear:

+ Avoid use of `env` in a "label" example, as it causes confusion.
+ Add an example using environment variables

ZymoticB:
Sure, it was written before the environment variable attestation was available. I agree its confusing at this point. Both of your suggestions would be great.

scotte:
I'm glad you commented that I missed the whole point of the original example. I'll update this PR with the above in mind. Thanks!

scotte:
@ZymoticB does this make more sense? Thanks!

ZymoticB:
LGTM

PR #1525, opened 2020-04-29
Require >= TLS 1.2 in the endpoint API server

### LINKED ISSUE #1524: TLS 1.0 enabled in SPIRE apis
* **Version**: v0.10.0
* **Platform**: Linux
* **Subsystem**: server apis

Verify with:

openssl s_client -tls1 -connect localhost:8081 < /dev/null

(where localhost:8081 is a spire server)

A connection is established with TLS 1.0:

```SSL-Session:
    Protocol  : TLSv1
    Cipher    : ECDHE-ECDSA-AES256-SHA
```

This is a relatively minor risk given that spire agents will support higher versions, and likely other API clients.
For compliance and risk mitigation reasons, we do not allow servers to listen on TLS 1.0 on our network.
I need to disable TLS 1.0 (and, ideally 1.1) in SPIRE

mcpherrinm:
There's no good reason to have older protocols enabled, especially in
SPIRE APIs where we control the clients.

The risk mitigated here is small, as clients will generally support 1.2 and not
be vulnerable to downgrade attacks.

This will avoid us requiring an exception to security policy though, as we
scan continuously to detect servers accepting TLS 1.0 connections as part of
our PCI-DSS compliance regimen.

**Pull Request check list**

- [  ] Documentation updated?

**Which issue this PR fixes**
fixes #1524

### DIFF
--- a/pkg/server/endpoints/endpoints.go
+++ b/pkg/server/endpoints/endpoints.go
@@ -258,6 +258,8 @@ func (e *Endpoints) getTLSConfig(ctx context.Context) func(*tls.ClientHelloInfo)
 
 			Certificates: certs,
 			ClientCAs:    roots,
+
+			MinVersion: tls.VersionTLS12,
 		}
 		return c, nil
 	}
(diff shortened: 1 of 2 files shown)


mcpherrinm:
I suppose I mean by "control the clients":  Clients of the registration API are specifically written for the purpose of communicating with the SPIRE server; we don't need compatibility with any sort of legacy ecosystem

PR #1581, opened 2020-05-22
Add CRD mode for k8s-workload-registrar 

faisal-memon:
Adds a `"crd"` mode to the k8s-workload-registrar to auto generate and manage SVIDs. The CRD works with a standard Reconcile () function to make sure SPIRE entries match the state of the system. It functions similarly to the webhook based implementation and can be used mostly as a drop in replacement (the crd requires some additional configuration detailed in the README.md). More details on what's provided:

- A namespace scoped SpiffeID CRD is defined. A controller watches for create, update, delete, etc. events for the CRD and creates entries on the SPIRE Server accordingly.
- An option pod controller (`pod_controller`) watches for POD events and creates/deletes SpiffeID CRDs accordingly. The pod controller sets the pod as the controller owner of the SPIFFE ID CRD so it is automatically garbage collected if the POD is deleted. The pod controller add the pod name as the first DNS name, which makes it also populate the CN field of the SVID.
- An optional endpoint controller (`add_svc_dns_name`) watches for endpoint events and adds the Service Name as a SAN DNS name to the SVID for all pods that are endpoints of the service. A pod can be an endpoint of multiple services and as a result can have multiple Service Names added as SAN DNS names. If a service is removed, the Service Name is removed from the SVID of all endpoint Pods. The format of the DNS name is `<service_name>.<namespace>.svc`
- A new option to disable namespaces from auto SVID generation (`disabled_namespaces`). By default `kube-system` is disabled for auto-injection.
- A node reconciler so that generated spiffe ids are parented to the node, rather than cluster wide parent.


support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	return ctrl.Result{}, nil
+}
+
+// createSpiffeId attempts to create the SPIFFE ID object. It uses the pod name appeneded with hash of the SPIFFE ID Spec
+// to create a unique name.
+func (r *PodReconciler) createSpiffeId(ctx context.Context, podName string, spiffeId *spiffeidv1beta1.SpiffeID) error {
+	var collisionCount int32
+	var existing spiffeidv1beta1.SpiffeID
+	for {
+		spiffeId.ObjectMeta.Name = podName + "-" + computeHash(&spiffeId.Spec, &collisionCount)
```
JonathanO, line 200: Would using the pod UID itself instead of a hash work?

faisal-memon: Yea, and it would things a lot simpler, but its not human readable looking through a list of spiffeids. Using the hash the output is easier to read and determine what the spiffeid resource is for:
```
$ kubectl get spiffeids
NAME                                         AGE
details-v1-74f858558f-j5vrj-6744667c5c       5h35m
productpage-v1-8554d58bff-jgclf-679c9f954d   5h35m
ratings-v1-7855f5bcb9-hllkv-59f494bbd        5h35m
reviews-v1-59fd8b965b-t4zst-69f9b57fbf       5h35m
reviews-v2-d6cfdb7d6-566xd-86fbc9765         5h35m
reviews-v3-75699b5cfb-gbzz7-646fb55cd4       5h35m
```

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		} else {
+			// No relevant annotation
+			return ctrl.Result{}, nil
+		}
+	}
+
+	spiffeId := &spiffeidv1beta1.SpiffeID{
+		ObjectMeta: v1.ObjectMeta{
+			Namespace:   pod.Namespace,
+			Annotations: make(map[string]string),
```
azdagron, line 140: This empty map initialization doesn't seem necessary... (no annotations are being set below)

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		},
+		Spec: spiffeidv1beta1.SpiffeIDSpec{
+			SpiffeId: spiffeIdUri,
+			DnsNames: make([]string, 0),
```
azdagron, line 144: This empty slice initialization doesn't seem necessary... (append works on nil slices)

Alternatively, since the pod name seems already available...:
```suggestion
			DnsNames: []string{pod.Name},
```

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,221 @@
...
+			},
+		},
+	}
+
+	// Set pod as owner of new SPIFFE ID
+	err := controllerutil.SetControllerReference(&pod, spiffeId, r.Scheme)
+	if err != nil {
+		r.c.Log.WithFields(logrus.Fields{
+			"SpiffeID.Name": spiffeId.Name,
+		}).WithError(err).Error("Failed to set pod as owner of new SpiffeID")
```
azdagron, line 157:
```suggestion
		}).WithError(err).Error("Failed to set pod as owner of new SpiffeID CRD")
```

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,221 @@
...
+				Namespace: pod.Namespace,
+			},
+		},
+	}
+
+	// Set pod as owner of new SPIFFE ID
+	err := controllerutil.SetControllerReference(&pod, spiffeId, r.Scheme)
+	if err != nil {
+		r.c.Log.WithFields(logrus.Fields{
+			"SpiffeID.Name": spiffeId.Name,
```
azdagron, line 156: nit: logging field names should be lower-case-seperated-by-dashes
```suggestion
			"spiffe-id-crd-name": spiffeId.Name,
```

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,221 @@
...
+			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
+			if err := r.Get(ctx, req.NamespacedName, &pod); err != nil {
+				r.c.Log.WithError(err).Error("Failed to get latest version of Pod")
+				return err
+			}
+			pod.ObjectMeta.Labels["spiffe.io/spiffeid"] = spiffeId.ObjectMeta.Name
+
+			err = r.Update(ctx, &pod)
+
+			return err
```
azdagron, line 182: This assignment is setting the `err` variable outside of the closure....
```suggestion
			return r.Update(ctx, &pod)
```

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		})
+		if retryErr != nil {
+			r.c.Log.WithError(retryErr).Error("Update failed")
+			return ctrl.Result{}, retryErr
+		}
+		r.c.Log.Info("Added label to pod")
```
azdagron, line 188: Does this need any fields in order to be useful?

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
azdagron: Each pod gets a single SpiffeID CRD, correct? Isn't the pod name in the object meta (i.e. `podName`, e.g., `details-v1-74f858558f-j5vrj`), already unique? I don't understand what the collision bit is for.

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	TrustDomain  string
+	Cluster      string
+	Mgr          ctrl.Manager
+}
+
+// SpiffeIDReconciler holds the runtime configuration and state of this controller
+type SpiffeIDReconciler struct {
+	client.Client
+	c                  SpiffeIDReconcilerConfig
+	myId               *string
```
azdagron, line 51: nit: unless you need to distinguish between "unset" and "empty", pointers to string aren't very ergonomic. In this case, I don't see any benefit to this being a pointer.

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// Reconcile ensures the SPIRE Server entry matches the corresponding CRD
+func (r *SpiffeIDReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+	spiffeID := spiffeidv1beta1.SpiffeID{}
+
+	if err := r.Get(ctx, req.NamespacedName, &spiffeID); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			r.c.Log.WithError(err).Error("unable to fetch SpiffeID")
```
azdagron, line 93:
```suggestion
			r.c.Log.WithError(err).Error("Unable to fetch SpiffeID CRD")
```

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+			return ctrl.Result{}, err
+		}
+
+		// Delete event
+		if err := r.ensureDeleted(ctx, r.spiffeIDCollection[req.NamespacedName.String()]); err != nil {
+			r.c.Log.WithFields(logrus.Fields{
+				"entryid": r.spiffeIDCollection[req.NamespacedName.String()],
+			}).WithError(err).Error("unable to delete spire entry", )
```
azdagron, line 101:
```suggestion
			}).WithError(err).Error("Unable to delete registration entry")
```

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	Cluster      string
+	Mgr          ctrl.Manager
+}
+
+// SpiffeIDReconciler holds the runtime configuration and state of this controller
+type SpiffeIDReconciler struct {
+	client.Client
+	c                  SpiffeIDReconcilerConfig
+	myId               *string
+	spiffeIDCollection map[string]string
```
azdagron, line 52: This state seems important for proper cleanup. What happens if the registrar is restarted? How does this behave in the face of multiple registrars?

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+			"entry": req.NamespacedName.String(),
+		}).Info("Finalized entry")
+
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	entryId, err := r.updateOrCreateSpiffeID(ctx, &spiffeID)
+	if err != nil {
+		r.c.Log.WithFields(logrus.Fields{
+			"request": req,
```
azdagron, line 115: how does this show up in the logs?

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		}).WithError(err).Error(err, "unable to get or create spire entry")
```
azdagron, line 116:
```suggestion
		}).WithError(err).Error(err, "Unable to get or create registration entry")
```

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+
+	oldEntryId := spiffeID.Status.EntryId
+	if oldEntryId == nil || *oldEntryId != entryId {
+		// We need to update the Status field
+		if oldEntryId != nil {
+			// entry resource must have been modified, delete the hanging one
+			if err := r.ensureDeleted(ctx, *spiffeID.Status.EntryId); err != nil {
+				r.c.Log.WithFields(logrus.Fields{
+					"entryid": *spiffeID.Status.EntryId,
+				}).WithError(err).Error("unable to delete old spire entry")
```
azdagron, line 128:
```suggestion
				}).WithError(err).Error("Unable to delete old registration entry")
```

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+				return ctrl.Result{}, err
+			}
+		}
+		spiffeID.Status.EntryId = &entryId
+		if err := r.Status().Update(ctx, &spiffeID); err != nil {
+			r.c.Log.WithError(err).Error("unable to update SpiffeID status")
```
azdagron, line 134:
```suggestion
			r.c.Log.WithError(err).Error("Unable to update SpiffeID CRD status")
```

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+
+	return ctrl.Result{}, nil
+}
+
+// ensureDeleted deletes the specified entry on the SPIRE Server
+func (r *SpiffeIDReconciler) ensureDeleted(ctx context.Context, entryId string) error {
+	if _, err := r.c.R.DeleteEntry(ctx, &registration.RegistrationEntryID{Id: entryId}); err != nil {
+		if status.Code(err) != codes.NotFound {
+			if status.Code(err) == codes.Internal {
+				// Spire server currently returns a 500 if delete fails due to the entry not existing.
```
azdagron, line 147: we should fix this :facepalm:

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+						}).Info("Entry already deleted")
+						return nil
+					}
+				}
+			}
+			return err
+		}
+	}
+	r.c.Log.WithFields(logrus.Fields{
+		"entry": entryId,
```
azdagron, line 162: We should be consistent on this field name "entryid" v.s. "entry" v.s. "entry-id"

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	spiffeId := instance.Spec.SpiffeId
+	selectors := toCommonSelector(instance.Spec.Selector)
+	commonRegistrationEntry :=  &common.RegistrationEntry{
+		Selectors: selectors,
+		ParentId:  *r.myId,
+		SpiffeId:  spiffeId,
+		DnsNames:  instance.Spec.DnsNames,
+	}
+
+	createEntryIfNotExistsResponse, err := r.c.R.CreateEntryIfNotExists(ctx, commonRegistrationEntry)
```
azdagron, line 178: nit:
```suggestion
	resp, err := r.c.R.CreateEntryIfNotExists(ctx, commonRegistrationEntry)
```

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		"entry": entryId,
+	}).Info("Deleted entry")
+	return nil
+}
+
+// updateOrCreateSpiffeID attempts to create a new entry. if the entry already exists, it updates it.
+func (r *SpiffeIDReconciler) updateOrCreateSpiffeID(ctx context.Context, instance *spiffeidv1beta1.SpiffeID) (string, error) {
+	spiffeId := instance.Spec.SpiffeId
+	selectors := toCommonSelector(instance.Spec.Selector)
+	commonRegistrationEntry :=  &common.RegistrationEntry{
```
azdagron, line 171: nit: there doesn't seem to be other "entry"'s in this function, I don't see the need for the prefix.
```suggestion
	entry :=  &common.RegistrationEntry{
```

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		existing := createEntryIfNotExistsResponse.Entry
+		if !equalStringSlice(existing.DnsNames, instance.Spec.DnsNames) {
+			r.c.Log.Info("Updating Spire Entry")
+
+			commonRegistrationEntry.EntryId = entryId
+			_, err := r.c.R.UpdateEntry(ctx, &registration.UpdateEntryRequest{
+				Entry: commonRegistrationEntry,
+			})
+			if err != nil {
+				r.c.Log.WithError(err).Error("unable to update SpiffeID with new DNS names")
```
azdagron, line 195:
```suggestion
				r.c.Log.WithError(err).Error("Unable to update SpiffeID CRD with new DNS names")
```

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	commonRegistrationEntry :=  &common.RegistrationEntry{
+		Selectors: selectors,
+		ParentId:  *r.myId,
+		SpiffeId:  spiffeId,
+		DnsNames:  instance.Spec.DnsNames,
+	}
+
+	createEntryIfNotExistsResponse, err := r.c.R.CreateEntryIfNotExists(ctx, commonRegistrationEntry)
+	if err != nil {
+		r.c.Log.WithError(err).Error("Failed to create spire entry")
```
azdagron, line 180:
```suggestion
		r.c.Log.WithError(err).Error("Failed to create registration entry")
```

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
JonathanO: While pod names are unique they can be rapidly reused (the obvious case being stateful sets), so the reconciliation process can end up with two SpiffeId resources existing for the same pod name (but different resource UIDs.)

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
JonathanO: Should probably be "resource" rather than "CRD." The CRD is the definition and schema for the resource, not the resource itself.

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	spiffeID := spiffeidv1beta1.SpiffeID{}
+
+	if err := r.Get(ctx, req.NamespacedName, &spiffeID); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			r.c.Log.WithError(err).Error("unable to fetch SpiffeID")
+			return ctrl.Result{}, err
+		}
+
+		// Delete event
+		if err := r.ensureDeleted(ctx, r.spiffeIDCollection[req.NamespacedName.String()]); err != nil {
```
JonathanO, line 98: Using a map to store the resource->entry mapping for deletes means that entries will be orphaned for any deletes that occur while the controller isn't running. Reintroducing the finaliser that was in the original controller guarantees that cleanup happens, and I think you can avoid  the pod deletion hanging if you set blockOwnerDeletion=false on the ownerReference for the pod.

JonathanO:
I have a couple of more general comments:
The CRD leaves open the possibility to create two resources that map to the same underlying spire entry, but deletion of one of those resources will have unintended effects on the other.

Spire considers uniqueness across (parent ID, spiffe ID, selectors), however the kubernetes resource is only unique across name (which is unrelated to spire entry.)
If I create SpiffeId resource A as (parentID=x, spiffeID=y, selectors=[z]) and resource B also as (parentID=x, spiffeID=y, selectors=[z]) then I end up with two resources pointing at the same spire server entry. If I delete one of these resources the controller will blindly delete the underlying entry shared by both. Since there's no mechanism in this PR to reconcile spire entries, this will go unnoticed leaving a SpiffeID resource without an entry.

This is a real problem that we ran into using an early version of the controller. For example, when someone decided to change the pattern for how we named SpiffeId resources in our helm charts. To resolve this we'd implemented an extra layer of resources to manage the many-one relationship, but I agree that it was ugly, which is why you forked that work to create this PR. I think this is actually solvable simply by introducing a simple in-use check in a finalizer on the SpiffeId, assuming certain threading assumptions are true (and a that a brief period where an entry might go missing in certain corner cases is acceptable.)

What seems less solvable is that DNS names, TTLs, and similar don't make an entry unique. If I create SpiffeId A as (parentID=x, spiffeID=y, selectors=[z], dns=foo) and B as (parentID=x, spiffeID=y, selectors=[z], dns=bar) then what should happen? Answering that question has made me start to seriously doubt the CRD based approach.

These problems are hidden if you're using the pod controller, as it uses the pod UID as part of the selector, which guarantees uniqueness per pod, but it makes direct use of the custom resource (which is, after all, the point of introducing the resource) challenging.

Finally, I know it's convenient for garbage collection, but it seems odd for this resource to be namespace scoped. The fact that it's currently unconstrained in what it allows you to create seems to make it a better fit as a Cluster scoped resource. As they can be used to create any spire entry it doesn't seem that you would want to allow people to create these in their own namespaces, and there's nothing inherently namespace bound about them.

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
faisal-memon: Yea, thats the reason for the extra hash

go.mod
```diff
@@ -66,22 +59,26 @@ require (
...
-	k8s.io/klog v1.0.0 // indirect
-	sigs.k8s.io/yaml v1.1.0 // indirect
+	sigs.k8s.io/controller-runtime v0.6.0
 )
+
+replace k8s.io/client-go => k8s.io/client-go v0.18.2
+
+replace k8s.io/api => k8s.io/api v0.18.2
+
+replace k8s.io/apimachinery => k8s.io/apimachinery v0.18.2
```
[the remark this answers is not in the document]
faisal-memon: Something weird with the go modules. Those three have to be in sync. Without the replace's one of them ends up using an older version

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
faisal-memon: Good idea, ill try that out.

support/k8s/k8s-workload-registrar/controllers/endpoint_controller.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	svcName := req.NamespacedName.Name + "." + req.NamespacedName.Namespace + ".svc"
+	for _, subset := range endpoints.Subsets {
+		for _, address := range subset.Addresses {
+			if address.TargetRef == nil {
+				continue
+			}
+			pod := corev1.Pod{}
+			if err := e.Get(ctx, types.NamespacedName{Name: address.TargetRef.Name, Namespace: address.TargetRef.Namespace}, &pod); err != nil {
```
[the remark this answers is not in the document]
faisal-memon: done

support/k8s/k8s-workload-registrar/controllers/endpoint_controller.go
```diff
@@ -0,0 +1,193 @@
...
+				continue
+			}
+			dnsNameAdded := false
+			retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				if err := e.Get(ctx, types.NamespacedName{Name: spiffeidname, Namespace: address.TargetRef.Namespace}, existing); err != nil {
+					e.c.Log.WithError(err).Error("Failed to get latest version of SPIFFE ID")
+					return err
+				}
+
+				if existing != nil && !containsString(existing.Spec.DnsNames[1:], svcName) {
```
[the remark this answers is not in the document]
faisal-memon: Yes, the pod name is always added as the first dns name: https://github.com/faisal-memon/spire/blob/03a6f2249adf88154dd1b77aae25ba8ef42562be/support/k8s/k8s-workload-registrar/controllers/pod_controller.go#L161

support/k8s/k8s-workload-registrar/controllers/endpoint_controller.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// deleteExternalResources remove the service name from the list of DNS Names when the service is removed
+func (e *EndpointReconciler) deleteExternalResources(ctx context.Context, namespacedName types.NamespacedName) error {
+	svcName := namespacedName.Name + "." + namespacedName.Namespace + ".svc"
```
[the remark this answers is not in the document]
faisal-memon: this is done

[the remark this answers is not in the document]
faisal-memon: https://github.com/spiffe/spire/pull/1581/commits/f4468eaf30a9266c07f69f0a1d6b7bbbbfd4e497

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	DisabledNamespaces []string
+}
+
+// PodReconciler holds the runtime configuration and state of this controller
+type PodReconciler struct {
+	client.Client
+	c              PodReconcilerConfig
+	Scheme         *runtime.Scheme
+	Mode           PodReconcilerMode
+	Value          string
```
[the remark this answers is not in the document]
faisal-memon: Mistake, this is now fixed: https://github.com/spiffe/spire/pull/1581/commits/c2e8101c52a9f22a6807820ad6d505873c65817d

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
faisal-memon: removed. https://github.com/spiffe/spire/pull/1581/commits/274047fa47f36086fc6b2cc1b6db1fc3db5dadcc

faisal-memon:
> Thanks, @faisal-memon. This will be a great addition. I took a first pass here. I'd love for some other folks with deeper K8s knowledge than myself to chime in.
> 
> Architecturally this seemed mostly ok. I had some concerns with the following:
> 
> 1. Instead of registering all workloads against a cluster node id could we register against the specific node the pod is scheduled on? This wasn't possible with the webhook, but seems possible here.
> 2. There are a few in-memory maps that hold onto some state that seems very important for cleanup. This is worrisome in the face of the registrar itself getting bounced and/or in the presence of multiple registrars.
> 
> On the more nitpicky side of things, I had quite a few comments on logging... I don't even know if I caught them all but in general 1) logging should start with upper case and 2) There were some inconsistencies referring to the SpiffeID CRD in the log lines... I don't care what we choose but we should be consistent ("SPIFFE ID", "SpiffeID", etc.). Also we should make sure each log line has the details (i.e. fields) it needs to be actionable.

Definitely 1 too many in-memory maps, let me look into removing those.

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
faisal-memon: fixed: https://github.com/spiffe/spire/pull/1581/commits/5dec988899e64c3afcf2472b753a5ff65c65e140

support/k8s/k8s-workload-registrar/config.go
```diff
@@ -72,6 +89,17 @@ func ParseConfig(hclConfig string) (*Config, error) {
 	if c.PodLabel != "" && c.PodAnnotation != "" {
 		return nil, errs.New("workload registration mode specification is incorrect, can't specify both pod_label and pod_annotation")
 	}
+	if c.DisabledNamespaces == nil {
+		c.DisabledNamespaces = defaultDisabledNamespaces()
+	}
+
+	if c.Mode != modeCRD && c.Mode != modeWebhook {
```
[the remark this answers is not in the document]
faisal-memon: Do you like the general way its done in that PR, so just copy it over? Or would you like something done differently?

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
faisal-memon: this seems to mostly work. pod deletion wont hang with `blockOwnerDeletion` set to false. namespace deletion will still hang if there any deleted spiffeids in the namespace. I think its better than using in-memory maps regardless.

faisal-memon: committed changes to remove the spiffeIdCollection map and replaced with (mostly) non-blocking finalizer

faisal-memon:
> I have a couple of more general comments:
> The CRD leaves open the possibility to create two resources that map to the same underlying spire entry, but deletion of one of those resources will have unintended effects on the other.
> 
> Spire considers uniqueness across (parent ID, spiffe ID, selectors), however the kubernetes resource is only unique across name (which is unrelated to spire entry.)
> If I create SpiffeId resource A as (parentID=x, spiffeID=y, selectors=[z]) and resource B also as (parentID=x, spiffeID=y, selectors=[z]) then I end up with two resources pointing at the same spire server entry. If I delete one of these resources the controller will blindly delete the underlying entry shared by both. Since there's no mechanism in this PR to reconcile spire entries, this will go unnoticed leaving a SpiffeID resource without an entry.
> 
> This is a real problem that we ran into using an early version of the controller. For example, when someone decided to change the pattern for how we named SpiffeId resources in our helm charts. To resolve this we'd implemented an extra layer of resources to manage the many-one relationship, but I agree that it was ugly, which is why you forked that work to create this PR. I think this is actually solvable simply by introducing a simple in-use check in a finalizer on the SpiffeId, assuming certain threading assumptions are true (and a that a brief period where an entry might go missing in certain corner cases is acceptable.)
> 
> What seems less solvable is that DNS names, TTLs, and similar don't make an entry unique. If I create SpiffeId A as (parentID=x, spiffeID=y, selectors=[z], dns=foo) and B as (parentID=x, spiffeID=y, selectors=[z], dns=bar) then what should happen? Answering that question has made me start to seriously doubt the CRD based approach can work in the way that I originally implemented it (as copied into this PR.)
> 
> These problems are hidden if you're using the pod controller, as it uses the pod UID as part of the selector, which guarantees uniqueness per pod, but it makes direct use of the custom resource (which is, after all, the point of introducing the resource) challenging.
> 
> Finally, I know it's convenient for garbage collection, but it seems odd for this resource to be namespace scoped. The fact that it's currently unconstrained in what it allows you to create seems to make it a better fit as a Cluster scoped resource. As they can be used to create any spire entry it doesn't seem that you would want to allow people to create these in their own namespaces, and there's nothing inherently namespace bound about them.


Thanks for the feedback. There are definitely some benefits to allowing multiple resources to map to the same underlying SPIRE entry. Ultimately I think supporting that use case adds too much complexity to the implementation for the value it brings. Its better IMO to not allow duplication, so each resource is 1:1 mapped to a SPIRE entry.

Regarding namespace scoping, it helps a lot with garbage collection. Id like to add something that would prepend the spiffeid with a `/ns/<namespace>/` or something like that. Do you think that would solve the issue?

support/k8s/k8s-workload-registrar/controllers/endpoint_controller.go
```diff
@@ -0,0 +1,193 @@
...
+			if !errors.IsNotFound(err) {
+				e.c.Log.WithFields(logrus.Fields{
+					"name": spiffeidname,
+				}).WithError(err).Error("failed to get SpiffeID")
+				return err
+			}
+
+			return nil
+		}
+		if existing != nil {
```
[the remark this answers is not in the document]
faisal-memon: There's an edge case where the resource is deleted while going through this loop, so in that scenario it will be nil

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
faisal-memon: Removed this in memory map

support/k8s/k8s-workload-registrar/controllers/endpoint_controller.go
```diff
@@ -0,0 +1,193 @@
...
+
+			if dnsNameAdded {
+				e.c.Log.WithFields(logrus.Fields{
+					"service": svcName,
+					"pod": pod.ObjectMeta.Name,
+				}).Info("added DNS names for")
+
+				if e.svcNametoSpiffeID[svcName] == nil {
+					e.svcNametoSpiffeID[svcName] = make([]string, 0)
+				}
```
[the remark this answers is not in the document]
faisal-memon: This has been removed

support/k8s/k8s-workload-registrar/controllers/spiffeid_controller.go
faisal-memon: Changed to use `EntryID`.

go.mod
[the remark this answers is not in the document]
azdagron: I think it's because you haven't changed `k8s.io/client-go` to `v0.18.2`? It's still set to `v10.0.0+incompatible`. In any case, I spent some time trying to update these dependencies to discover what you were hitting (before I noticed `client-go` not getting bumped) and uncovered an update that is required in some of our code that interacts with `golang.org/x/crypto` which is also being bumped when moving to v0.18.2. If you are ok with it, I'll open up a separate PR that bumps this dependency while fixing up our code that interacts with golang.org/x/crypto.

support/k8s/k8s-workload-registrar/controllers/node_controller.go
```diff
@@ -0,0 +1,147 @@
...
+	response, err := n.c.R.CreateEntryIfNotExists(ctx, entry)
+	if err != nil {
+		n.c.Log.WithError(err).Error("Failed to create registration entry")
+		return ctrl.Result{}, err
+	}
+	if !response.Preexisting {
+		n.c.Log.WithFields(logrus.Fields{
+			"entryID":  response.Entry.EntryId,
+			"spiffeID": n.nodeID(node.ObjectMeta.Name),
+		}).Info("Created entry")
```
azdagron, line 105: nit: it might be easier to visually distinguish this from workload entries if we qualify the kind of entry:
```suggestion
		}).Info("Created node entry")
```

support/k8s/k8s-workload-registrar/controllers/node_controller.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// deleteNodeEntry remove the service name from the list of DNS Names when the service is removed
+func (n *NodeReconciler) deleteNodeEntry(ctx context.Context, nodeName string) error {
+	entries, err := n.c.R.ListByParentID(ctx, &registration.ParentID{
```
azdagron, line 113: I think it would be much more performant to use `ListBySpiffeID` with the expected node ID then you don't have to get back all node entries just to scan for the one you want.

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,224 @@
...
+	if err := r.Get(ctx, req.NamespacedName, &pod); err != nil {
+		if !errors.IsNotFound(err) {
+			r.c.Log.WithError(err).Error("Unable to get Pod")
+			return ctrl.Result{}, err
+		}
+
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Pod neeeds to be assigned a node before it can get a SPIFFE ID
```
azdagron, line 93:
```suggestion
	// Pod needs to be assigned a node before it can get a SPIFFE ID
```

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,224 @@
...
+
+	_, err := r.createPodEntry(ctx, &pod)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// getOrCreatePodEntry attempts to create a new SpiffeID resource. if the entry already exists, it returns it.
```
azdagron, line 106: stale comment (method name)

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,224 @@
...
+func (r *PodReconciler) createPodEntry(ctx context.Context, pod *corev1.Pod) (*spiffeidv1beta1.SpiffeID, error) {
```
azdagron, line 107: The returned SpiffeID is not used. How would you feel about not returning it?

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
```diff
@@ -0,0 +1,224 @@
...
+	var existing spiffeidv1beta1.SpiffeID
+	for {
+		spiffeId.ObjectMeta.Name = podName + "-" + computeHash(&spiffeId.Spec, &collisionCount)
+		err := r.Create(ctx, spiffeId)
+		if errors.IsAlreadyExists(err) {
+			spiffeIdNamespacedName := types.NamespacedName{
+				Name:      spiffeId.ObjectMeta.Name,
+				Namespace: spiffeId.ObjectMeta.Namespace,
+			}
+			r.Get(ctx, spiffeIdNamespacedName, &existing)
```
azdagron, line 202: no error checking?

support/k8s/k8s-workload-registrar/controllers/utils.go
```diff
@@ -0,0 +1,138 @@
...
+func NewManager() (ctrl.Manager, error) {
+	scheme := runtime.NewScheme()
+	_ = clientgoscheme.AddToScheme(scheme)
+	_ = spiffeidv1beta1.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		MetricsBindAddress: "0",
+		Scheme:             scheme,
+		Port:               9443,
```
azdagron, line 46: Is this likely to change? does it need to be configurable?

support/k8s/k8s-workload-registrar/controllers/endpoint_controller.go
```diff
@@ -0,0 +1,193 @@
...
+		for _, address := range subset.Addresses {
+			if address.TargetRef == nil {
+				continue
+			}
+			pod := corev1.Pod{}
+			if err := e.Get(ctx, types.NamespacedName{Name: address.TargetRef.Name, Namespace: address.TargetRef.Namespace}, &pod); err != nil {
+				e.c.Log.WithError(err).Error("Error retreiving pod")
+				return ctrl.Result{}, err
+			}
+			spiffeidname := pod.ObjectMeta.Labels["spiffe.io/spiffeid"]
```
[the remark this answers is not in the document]
faisal-memon: This label no longer exists.

support/k8s/k8s-workload-registrar/controllers/endpoint_controller.go
```diff
@@ -0,0 +1,193 @@
...
+	Log           logrus.FieldLogger
+	Mgr           ctrl.Manager
+	DisabledNamespaces []string
+}
+
+// EndpointReconciler holds the runtime configuration and state of this controller
+type EndpointReconciler struct {
+	client.Client
+	c                  EndpointReconcilerConfig
+	svcNametoSpiffeID  map[string][]string
```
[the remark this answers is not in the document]
faisal-memon: map removed

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
faisal-memon: This whole code path has been removed.

support/k8s/k8s-workload-registrar/controllers/endpoint_controller.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// EndpointReconciler holds the runtime configuration and state of this controller
+type EndpointReconciler struct {
+	client.Client
+	c EndpointReconcilerConfig
+}
+
+// NewEndpointReconciler creates a new EndpointReconciler object
+func NewEndpointReconciler(config EndpointReconcilerConfig) (*EndpointReconciler, error) {
```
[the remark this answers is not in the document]
faisal-memon: Makes sense. This change is in. https://github.com/spiffe/spire/pull/1581/commits/6c7315fa6b37e1f14a12f4ca4c43db8fa674feef

support/k8s/k8s-workload-registrar/config_webhook.go
```diff
@@ -0,0 +1,92 @@
...
+		return err
+	}
+	defer log.Close()
+
+	log.WithField("socket_path", c.ServerSocketPath).Info("Dialing server")
+	serverConn, err := grpc.DialContext(ctx, "unix://"+c.ServerSocketPath, grpc.WithInsecure())
+	if err != nil {
+		return errs.New("failed to dial server: %v", err)
+	}
+	defer serverConn.Close()
```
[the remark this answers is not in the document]
faisal-memon: done.

support/k8s/k8s-workload-registrar/config.go
```diff
@@ -1,77 +1,87 @@
...
+		return errs.New("workload registration mode specification is incorrect, can't specify both pod_label and pod_annotation")
+	}
+	if c.Mode != modeCRD && c.Mode != modeWebhook {
+		return errs.New("invalid mode \"%s\", valid values are %s and %s", c.Mode, modeCRD, modeWebhook)
+	}
+
+	return nil
+}
+
+func LoadConfig(path string) (Config, error) {
```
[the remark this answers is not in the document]
faisal-memon: Done, though i left ParseConfig in the interface. The code is slight more condensed that way without having to call ParseConfig() separately for each mode.

support/k8s/k8s-workload-registrar/controllers/node_controller.go
faisal-memon: Done. https://github.com/spiffe/spire/pull/1581/commits/bdad0f7e4e00599bb60c147e226e08996b8ea10a

support/k8s/k8s-workload-registrar/config.go
```diff
@@ -1,77 +1,87 @@
...
+	hclBytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, errs.New("unable to load configuration: %v", err)
+	}
+
+	c := &CommonConfig{}
+	c.ParseConfig(string(hclBytes))
```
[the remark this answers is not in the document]
faisal-memon: done

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
faisal-memon: fixed: https://github.com/spiffe/spire/pull/1581/commits/b60690d2b925bf1cfb8cb45efd6a4ea6bd9e2ff3

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
faisal-memon: Makes sense, updated: https://github.com/spiffe/spire/pull/1581/commits/b60690d2b925bf1cfb8cb45efd6a4ea6bd9e2ff3

support/k8s/k8s-workload-registrar/controllers/pod_controller.go
faisal-memon: Fixed: https://github.com/spiffe/spire/pull/1581/commits/b60690d2b925bf1cfb8cb45efd6a4ea6bd9e2ff3

support/k8s/k8s-workload-registrar/controllers/utils.go
faisal-memon: It might, kubebuilder has it configurable. I made the change here, remove the Port because we don't have a webhook in here yet: https://github.com/spiffe/spire/pull/1581/commits/137f7dd0c496939fa904911c66b63961cf567ce2

api/spiffecrd/v1beta1/groupversion_info.go
```diff
@@ -0,0 +1,35 @@
+/*
```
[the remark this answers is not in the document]
faisal-memon: Moved everything to mode-crd/ directory. Would it make sense to move the webhook files to a mode-webhook/ directory?

https://github.com/spiffe/spire/pull/1581/commits/c5e8f291604bb78301c55485504e2ea3179c755d

support/k8s/k8s-workload-registrar/mode-crd/controllers/endpoint_controller.go
```diff
@@ -0,0 +1,209 @@
...
+			// Its possible this reconcile loop ran before the SPIFFE ID resource was generated
+			if len(spiffeIDList.Items) == 0 {
+				return ctrl.Result{
+					Requeue: e.requeue(address.TargetRef.Name, address.TargetRef.Namespace),
+				}, nil
+			}
+
+			// Iterate through the list of SPIFFE ID resources and update to add the DNS name
+			for _, spiffeID := range spiffeIDList.Items {
+				if !containsString(spiffeID.Spec.DnsNames[1:], svcName) {
```
azdagron, line 118: why is the first element of DnsNames in the spec not considered when determining if the service name exists in the list?

faisal-memon: It is statically populated with the name of the pod: https://github.com/faisal-memon/spire/blob/master/support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go#L121

I can remove the `1:` optimization to make the the code a little safer.

azdagron: Yeah, i think that would be much less surprising.

support/k8s/k8s-workload-registrar/mode-crd/controllers/endpoint_controller.go
```diff
@@ -0,0 +1,209 @@
...
+			}).WithError(err).Error("Unable to delete DNS names in SpiffeID CRD")
+			return err
+		}
+	}
+
+	return nil
+}
+
+// requeue determines if the reconcile needs to be requeued. If the controller has been configured with a
+// pod label/annotation and the pod has the label/annotation then yes. If the controller has not been
```
JonathanO, line 177: What about the case where it's in ServiceAccount mode?

faisal-memon: That will requeue each time because a spiffe id is generated for each pod since each pod has a service account. its not optional like the pod label/annotation.

support/k8s/k8s-workload-registrar/mode-crd/controllers/endpoint_controller.go
faisal-memon: done: https://github.com/spiffe/spire/pull/1581/commits/bd840d62c70f6dc87b4ba8f706c7effa583e75df

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	return nil
+}
+
+func (c *CRDMode) Run(ctx context.Context) error {
+	if err := c.SetupLogger(); err != nil {
+		return errs.New("error setting up logging: %v", err)
+	}
+	if err := c.Dial(ctx); err != nil {
```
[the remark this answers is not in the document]
faisal-memon: done: https://github.com/spiffe/spire/pull/1581/commits/bd840d62c70f6dc87b4ba8f706c7effa583e75df

support/k8s/k8s-workload-registrar/README.md
```diff
@@ -44,6 +64,49 @@ server_socket_path = "/run/spire/sockets/registration.sock"
...
+## CRD Mode Configuration
+
+The following configuration is required before `"crd"` mode can be used:
+
+1. The SPIFFE ID CRD needs to be applied: `kubectl apply -f config/spiffeid.spiffe.io_spiffeids.yaml`
+1. The appropriate ClusterRole need to be applied. `kubectl apply -f config/crd_role.yaml`
+   * This creates a new ClusterRole named `spiffe-crd-role`
+1. The new ClusterRole needs a ClusterRoleBinding to the SPIRE Server ServiceAccount. Change the name of the ServiceAccount and then: `kubectl apply -f config/crd_role_binding.yaml` 
+   * This creates a new ClusterRoleBinding named `spiffe-crd-rolebinding`
+
```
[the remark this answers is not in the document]
faisal-memon: I added the webhook config and updated the README.md to reflect it. https://github.com/spiffe/spire/pull/1581/commits/bd840d62c70f6dc87b4ba8f706c7effa583e75df

support/k8s/k8s-workload-registrar/mode-crd/controllers/node_controller.go
```diff
@@ -0,0 +1,148 @@
...
+		For(&corev1.Node{}).
+		Complete(r)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Reconcile steps through the endpoints for each service and adds the name of the service as
```
JonathanO, line 68: Comment copied from the endpoint controller?

support/k8s/k8s-workload-registrar/mode-crd/controllers/node_controller.go
```diff
@@ -0,0 +1,148 @@
...
+		n.c.Log.WithFields(logrus.Fields{
+			"entryID":  response.Entry.EntryId,
+			"spiffeID": n.nodeID(node.ObjectMeta.Name),
+		}).Info("Created node entry")
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// deleteNodeEntry remove the service name from the list of DNS Names when the service is removed
```
JonathanO, line 112: Comment copied from endpoint controller?

support/k8s/k8s-workload-registrar/mode-crd/controllers/node_controller.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Reconcile steps through the endpoints for each service and adds the name of the service as
+// a DNS name to the SPIFFE ID CRD
+func (n *NodeReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
```
JonathanO, line 70: This controller is going to leak entries for nodes deleted while the controller is unavailable. The entire reason to use the custom resource behind the pod controller is to avoid this occurring for pods, the mechanism should exist for Nodes too. That probably means allowing arbitrary parent IDs on SpiffeId, and this controller creating SpiffeId resources in its own namespace for the nodes. (it's permissable for a namespaced resource to be owned by a cluster scoped resource, so this should work.)

faisal-memon: Good catch, thank you.

faisal-memon: This is now fixed

support/k8s/k8s-workload-registrar/mode-crd/controllers/node_controller.go
faisal-memon: Removed this function with the bug fix to the node reconciler

support/k8s/k8s-workload-registrar/mode-crd/controllers/node_controller.go
faisal-memon: yep. thanks for catching this. comment is now updated.

PR #1606, opened 2020-06-01
API Refactor: BatchCreateFederatedBundle RPC

marcosy:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

API Refactor

**Description of change**

Implements `BatchCreateFederatedBundle()` RPC

**Which issue this PR fixes**

Fixes: #1553


pkg/server/api/bundle/v1/service.go
```diff
@@ -161,17 +155,85 @@ func (s *Service) GetFederatedBundle(ctx context.Context, req *bundle.GetFederat
...
 
 func (s *Service) BatchCreateFederatedBundle(ctx context.Context, req *bundle.BatchCreateFederatedBundleRequest) (*bundle.BatchCreateFederatedBundleResponse, error) {
-	return nil, status.Errorf(codes.Unimplemented, "method BatchCreateFederatedBundle not implemented")
+	log := rpccontext.Logger(ctx)
+
+	results := []*bundle.BatchCreateFederatedBundleResponse_Result{}
+	for _, b := range req.Bundle {
+		td, err := spiffeid.TrustDomainFromString(b.TrustDomain)
+		if err != nil {
+			log.Errorf("Trust domain argument is not a valid SPIFFE ID: %q", b.TrustDomain)
```
amartinezfayo, line 176: The trust domain string could be either the host part (e.g. example.org) or the SPIFFE ID URI (e.g. spiffe://example.org). We actually refer to "the name of the trust domain the bundle belongs to": https://github.com/spiffe/spire/blob/master/proto/spire-next/types/bundle.proto#L7
So I would just say in the error that the trust domain argument is not valid.

pkg/server/api/bundle/v1/service.go
```diff
@@ -161,17 +155,85 @@ func (s *Service) GetFederatedBundle(ctx context.Context, req *bundle.GetFederat
...
+			results = append(results, &bundle.BatchCreateFederatedBundleResponse_Result{
+				Status: CreateStatus(codes.InvalidArgument, "trust domain argument is not a valid SPIFFE ID: %q", b.TrustDomain),
```
amartinezfayo, line 178: The same comment about the trust domain string applies here.

pkg/server/api/bundle/v1/service.go
```diff
@@ -161,17 +155,85 @@ func (s *Service) GetFederatedBundle(ctx context.Context, req *bundle.GetFederat
...
+		if err != nil {
+			log.WithError(err).Error("Unable to convert bundle")
+			results = append(results, &bundle.BatchCreateFederatedBundleResponse_Result{
+				Status: CreateStatus(codes.Internal, "unable to convert bundle: %v", err),
+			})
+			continue
+		}
+		applyBundleMask(protoBundle, req.OutputMask)
+
+		log.Infof("Bundle created successfully for trust domain: %q", td.String())
```
MarcosDY, line 227: not sure we want an info log here, it can spam if we add several bundles

pkg/server/api/bundle/v1/service.go
```diff
@@ -186,49 +248,36 @@ func (s *Service) BatchDeleteFederatedBundle(ctx context.Context, req *bundle.Ba
...
-		}
-		out.JwtAuthorities = authorities
+	if !mask.JwtAuthorities {
+		b.JwtAuthorities = nil
 	}
+}
 
-	return out, nil
+// CreateStatus creates a proto Status
+func CreateStatus(code codes.Code, format string, a ...interface{}) *types.Status {
```
MarcosDY, line 278: I dont think here is a good place for this method you have status methods on api: https://github.com/spiffe/spire/blob/master/pkg/server/api/status.go#L12

pkg/server/api/bundle/v1/service_test.go
```diff
@@ -358,9 +339,153 @@ func createBundle(t *testing.T, test *serviceTest, td string) *common.Bundle {
...
+
+	for _, tt := range []struct {
+		name            string
+		bundlesToCreate []*types.Bundle
+		outputMask      *types.BundleMask
+		expectedResults []*bundlepb.BatchCreateFederatedBundleResponse_Result
+		expectedLogMsgs []spiretest.LogEntry
+		dsError         error
+	}{
+		{
```
MarcosDY, line 354: maybe you can add a test case where you create more than one bundle and someone fails and another not, and verify that we always keep order, and expected logs+results

MarcosDY: another test case to add is a mask with all false, what happens in that case? and when it is nil?

MarcosDY: another test case when bundles is empty?

pkg/server/api/bundle/v1/service.go
```diff
@@ -161,17 +155,85 @@ func (s *Service) GetFederatedBundle(ctx context.Context, req *bundle.GetFederat
...
+	applyBundleMask(b, req.OutputMask)
+
 	return b, nil
 }
 
 func (s *Service) BatchCreateFederatedBundle(ctx context.Context, req *bundle.BatchCreateFederatedBundleRequest) (*bundle.BatchCreateFederatedBundleResponse, error) {
-	return nil, status.Errorf(codes.Unimplemented, "method BatchCreateFederatedBundle not implemented")
+	log := rpccontext.Logger(ctx)
+
+	results := []*bundle.BatchCreateFederatedBundleResponse_Result{}
```
MarcosDY, line 172: what happens when req is emtpy? must we do nothing? or return an error?

pkg/server/api/bundle/v1/service_test.go
```diff
@@ -358,9 +339,153 @@ func createBundle(t *testing.T, test *serviceTest, td string) *common.Bundle {
...
+
+			resp, err := test.client.BatchCreateFederatedBundle(context.Background(), &bundlepb.BatchCreateFederatedBundleRequest{
+				Bundle:     tt.bundlesToCreate,
+				OutputMask: tt.outputMask,
+			})
+			require.NoError(t, err)
+			require.NotNil(t, resp)
+			spiretest.AssertLogs(t, test.logHook.AllEntries(), tt.expectedLogMsgs)
+
+			require.Equal(t, len(tt.expectedResults), len(resp.Results))
```
MarcosDY, line 467: instead of this you can use spiretest.RequireProtoEqual

MarcosDY: and you have have an expected where you play with mask, to be sure response has exactly what we want

pkg/server/api/bundle/v1/service.go
```diff
@@ -161,17 +155,85 @@ func (s *Service) GetFederatedBundle(ctx context.Context, req *bundle.GetFederat
...
+
+		dsBundle, err := api.ProtoToBundle(b)
+		if err != nil {
+			log.WithError(err).Error("Unable to convert bundle")
+			results = append(results, &bundle.BatchCreateFederatedBundleResponse_Result{
+				Status: CreateStatus(codes.Internal, "unable to convert bundle: %v", err),
+			})
+			continue
+		}
+		resp, err := s.ds.CreateBundle(ctx, &datastore.CreateBundleRequest{
```
MarcosDY, line 199: I think before we call create bundle we must validate that rootca and public keys are valid

pkg/server/api/bundle.go
```diff
@@ -0,0 +1,78 @@
...
+
+	td, err := spiffeid.TrustDomainFromString(b.TrustDomain)
+	if err != nil {
+		return nil, err
+	}
+
+	rootCas := []*common.Certificate{}
+	for _, certificate := range b.X509Authorities {
+		rootCas = append(rootCas, &common.Certificate{
+			DerBytes: certificate.Asn1,
```
azdagron, line 57: We should be validating that the incoming ASN.1 certificate data is well formed. See the helpers added in the AppendBundle PR.

pkg/server/api/bundle/v1/service.go
azdagron: Hmm, good question. I think I'd lean towards returning an empty response instead of failing.

Also, I'd probably recommend either initializing this slice to nil, or preallocating the slice to the number of incoming bundles.

pkg/server/api/bundle/v1/service_test.go
marcosy: > maybe you can add a test case where you create more than one bundle and someone fails and another not, and verify that we always keep order, and expected logs+results

This scenario is covered by "Create fails if bundle already exists"

> another test case to add is a mask with all false, what happens in that case? and when it is nil?

Will do

pkg/server/api/bundle.go
marcosy: Right, I'm thinking about moving the helpers added in AppendBundle PR to this file. And then, modify AppendBundle to use `ProtoToBundle` instead of the individual helpers. 

Sounds good?

azdagron: Yes, although the `api` package is where we've been sticking the `*ToProto` and `ProtoTo*` methods.

marcosy:
Thank you all for the feedback! I think I fixed all the comments. Please let me know if I missed any of them 🙂

pkg/server/api/bundle.go
```diff
@@ -40,3 +42,69 @@ func BundleToProto(b *common.Bundle) (*types.Bundle, error) {
...
 	}, nil
 }
+
+func ProtoToBundle(b *types.Bundle) (*common.Bundle, error) {
+	if b == nil {
+		return nil, errors.New("no bundle provided")
+	}
+
+	td, err := spiffeid.TrustDomainFromString(b.TrustDomain)
+	if err != nil && b.TrustDomain != "" {
```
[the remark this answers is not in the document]
marcosy: In the AppendBundle RPC the `InputMask` is applied before calling `ProtoToBundle(...)`. So, it could be a case where `TrustDomain` flag is not set in the input mask and it is passed as empty to `ProtoToBundle`.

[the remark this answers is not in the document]
MarcosDY: maybe we can apply mask after parsing? that makes sense to me, and prevent us to add this kind of code?

PR #1617, opened 2020-06-04
[API Refactor] Initial implementation of NewDownstreamX509CA

dfeldman:

Fixes #1614


pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +328,96 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
+	if err != nil {
+		return nil, err
+	}
+
+	entry, ok := entriesMap[req.EntryId]
+	if !ok {
+		err := status.Errorf(codes.NotFound, "error not found or not authorized")
+		log.WithError(err).Error("Rejecting NewDownstreamX509CA request due to invalid entry ID")
+		return nil, err
+	}
```
amartinezfayo, line 408: We may have a helper function for this. This is also done in `NewJWTSVID` RPC.

pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +328,96 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
 	}, nil
 }
 
-func (s *Service) NewDownstreamX509CA(context.Context, *svid.NewDownstreamX509CARequest) (*svid.NewDownstreamX509CAResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+func (s *Service) newX509CASVID(ctx context.Context, entry *types.Entry, csrBytes []byte) (resp *svid.NewDownstreamX509CAResponse, err error) {
```
azdagron, line 331: nit: this method should be grouped below the exported methods with the other unexported methods.

pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +328,96 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
+	csr, err := x509.ParseCertificateRequest(csrBytes)
+	if err != nil {
+		log.WithError(err).Error("Invalid request: malformed CSR")
+		return nil, status.Errorf(codes.InvalidArgument, "malformed CSR: %v", err)
+	}
+
+	if err := csr.CheckSignature(); err != nil {
+		log.WithError(err).Error("Invalid request: invalid CSR signature")
+		return nil, status.Error(codes.InvalidArgument, "invalid CSR signature")
+	}
```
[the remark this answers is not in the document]
dfeldman: you are right, this should be a shared function

pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +328,96 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
+		return nil, err
+	}
+
+	if !entry.Downstream {
+		err := status.Errorf(codes.NotFound, "error not found or not authorized")
+		log.WithError(err).Error("Rejecting NewDownstreamX509CA request due to downstream flag not being set")
+		return nil, err
+	}
+
+	resp, err := s.newX509CASVID(ctx, entry, req.Csr)
```
azdagron, line 416: How would you feel about inlining the contents of `newX509CASVID` here? That function doesn't seem to provide a lot of benefit. I recognize you were probably following the patterns in the other RPC method implementations.... I think those were carryover when we had a split server/service implementation. If you want to keep it the way it is, that is fine, although if you do, you'll want to get rid of the log line on 418 since it is redundant considering all of the returns under `newX509CASVID` already log.

pkg/server/api/svid/v1/service_test.go
```diff
@@ -921,10 +926,206 @@ func TestServiceBatchNewX509SVID(t *testing.T) {
...
+			fetcherErr: "",
+			entry:      downstreamEntry,
+		},
+	} {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			test.logHook.Reset()
+			test.ef.err = tt.fetcherErr
+			if !tt.failDataStore {
+				_, err := test.ds.AppendBundle(context.TODO(), &datastore.AppendBundleRequest{
```
azdagron, line 1065: nit: `context.TODO()` is typically used when you expect to have a context threaded through at some point but don't have one available now. This isn't the case for tests though. `context.Background()` is more appropriate.

pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +328,96 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
+	}
+
+	entry, ok := entriesMap[req.EntryId]
+	if !ok {
+		err := status.Errorf(codes.NotFound, "error not found or not authorized")
+		log.WithError(err).Error("Rejecting NewDownstreamX509CA request due to invalid entry ID")
+		return nil, err
+	}
+
+	if !entry.Downstream {
```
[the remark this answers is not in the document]
dfeldman: good point

pkg/server/api/svid/v1/service.go
dfeldman: I don't think a shared function would save any code here since it's such a simple check (still, can do it if you think)

pkg/server/api/svid/v1/service_test.go
```diff
@@ -921,10 +926,206 @@ func TestServiceBatchNewX509SVID(t *testing.T) {
...
+			resp, err := test.client.NewDownstreamX509CA(ctx, &svidpb.NewDownstreamX509CARequest{
+				EntryId: tt.entry.Id,
+				Csr:     csr,
+			})
+
+			if tt.err == "" {
+				require.NoError(t, err)
+				require.NotNil(t, resp)
+				require.NotNil(t, resp.CaCertChain)
+				require.NotEmpty(t, resp.CaCertChain)
```
azdagron, line 1106: If what we really want to know is if the have contents, we should just assert `NotEmpty` since it would also cover the `nil` case.
```suggestion
				require.NotEmpty(t, resp.CaCertChain)
```

pkg/server/api/svid/v1/service_test.go
```diff
@@ -921,10 +926,206 @@ func TestServiceBatchNewX509SVID(t *testing.T) {
...
+				require.NotNil(t, resp.X509Authorities)
+				require.NotEmpty(t, resp.X509Authorities)
```
azdagron, line 1108:
```suggestion
				require.NotEmpty(t, resp.X509Authorities)
```

pkg/server/api/svid/v1/service_test.go
```diff
@@ -921,10 +926,206 @@ func TestServiceBatchNewX509SVID(t *testing.T) {
...
+				csr = createCSR(t, tt.csrTemplate)
+			}
+
+			ctx := context.Background()
+			resp, err := test.client.NewDownstreamX509CA(ctx, &svidpb.NewDownstreamX509CARequest{
+				EntryId: tt.entry.Id,
+				Csr:     csr,
+			})
+
+			if tt.err == "" {
```
azdagron, line 1102: nit: maybe invert the conditions and return early so we don't have so much nested code?
```suggestion
			if tt.err != "" {
			    spiretest.RequireGRPCStatusContains(....)
			    return
			}
			
			.... bunch of assertions around response contents...
```

pkg/server/api/svid/v1/service_test.go
```diff
@@ -921,10 +926,206 @@ func TestServiceBatchNewX509SVID(t *testing.T) {
...
+			code:       codes.NotFound,
+		},
+		{
+			name:           "SuccessfulCARequest",
+			rateLimiterErr: nil,
+			err:            "",
+			failSigning:    false,
+			failDataStore:  false,
+			csrTemplate: &x509.CertificateRequest{
+				// Right now, the request will succeed even if URIs is empty. This is
```
azdagron, line 1052: From the API documentation:
```
    // Required. The ASN.1 DER encoded Certificate Signing Request (CSR). The
    // CSR is only used to convey the public key; other fields in the CSR are
    // ignored. The X509-SVID attributes are determined by the entry.
 ```

The CSR is only used to convey the public key. The SPIFFE ID of the returned CA certificate should be the trust domain ID (e.g. `spiffe://example.org`). This is a small departure in behavior from the current node FetchX509CASVID RPC.

File: pkg/server/api/svid/v1/service.go (hunk 1 above)
Comment (azdagron) [reply pkg/server/api/svid/v1/service.go:408]: Just to make sure you guys saw my comment on this section, but this code block isn't the right thing for this RPC (since it isn't doing an authorized entry crawl for an agent).

Comment (amartinezfayo) [reply pkg/server/api/svid/v1/service.go:408]: Just saw it. Thanks!

File: pkg/server/api/svid/v1/service.go [hunk 11]
```diff
@@ -324,6 +328,96 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
 }
 
-func (s *Service) NewDownstreamX509CA(context.Context, *svid.NewDownstreamX509CARequest) (*svid.NewDownstreamX509CAResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+func (s *Service) newX509CASVID(ctx context.Context, entry *types.Entry, csrBytes []byte) (resp *svid.NewDownstreamX509CAResponse, err error) {
+	log := rpccontext.Logger(ctx)
+
+	if !entry.Downstream {
+		err := status.Errorf(codes.NotFound, "error not found or not authorized")
+		log.WithError(err).Error("Rejecting NewDownstreamX509CA request due to downstream flag not being set")
```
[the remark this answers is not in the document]
dfeldman: After thinking about it -- I changed this error to "error not found or not authorized" because I don't necessarily want to reveal to the outside world exactly why it isn't authorized. What do you think?

pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +328,96 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
+		log.WithError(err).Error("Rejecting request due to downstream CA signing rate limit")
+		return nil, err
+	}
+
+	entriesMap, err := s.fetchEntries(ctx, log)
+	if err != nil {
+		return nil, err
+	}
+
+	entry, ok := entriesMap[req.EntryId]
```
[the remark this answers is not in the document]
dfeldman: Sure! Sorry, that parameter is the only reason I thought we had to crawl the entries... I'll change to use the rpc context

pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +328,96 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
+
+func (s *Service) NewDownstreamX509CA(ctx context.Context, req *svid.NewDownstreamX509CARequest) (*svid.NewDownstreamX509CAResponse, error) {
+	log := rpccontext.Logger(ctx)
+
+	if err := rpccontext.RateLimit(ctx, 1); err != nil {
+		log.WithError(err).Error("Rejecting request due to downstream CA signing rate limit")
+		return nil, err
+	}
+
+	entriesMap, err := s.fetchEntries(ctx, log)
```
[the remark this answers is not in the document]
dfeldman: Just to check -- if the caller has multiple downstream entries, we check that at least one of them has a SPIFFE ID that matches the SPIFFE ID in the CSR, right? Otherwise you could sign some other unrelated certificate.

pkg/server/api/svid/v1/service.go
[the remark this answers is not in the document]
azdagron: Awesome, thanks!

pkg/server/api/svid/v1/service.go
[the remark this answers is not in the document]
azdagron: See https://github.com/spiffe/spire/pull/1617#discussion_r435391923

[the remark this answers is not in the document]
azdagron: And for now, just take the first downstream entry. That mimics today's behavior.

pkg/server/api/svid/v1/service.go
dfeldman: sounds good

pkg/server/api/svid/v1/service.go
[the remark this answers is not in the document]
dfeldman: Ok makes sense.

pkg/server/api/svid/v1/service_test.go
```diff
@@ -921,10 +926,186 @@ func TestServiceBatchNewX509SVID(t *testing.T) {
...
+	}
+
+	test := setupServiceTest(t)
+	defer test.Cleanup()
+
+	x509CA := test.ca.X509CA()
+
+	for _, tt := range []downstreamCaTest{
+		{
+			name:           "Malformed CSR",
```
azdagron, line 964: nit: these sub-test names don't have consistent naming (some have spaces, some don't)

pkg/server/api/svid/v1/service_test.go
```diff
@@ -921,10 +926,186 @@ func TestServiceBatchNewX509SVID(t *testing.T) {
...
+		{
+			name:           "SuccessfulCARequest",
+			rateLimiterErr: nil,
+			err:            "",
+			failSigning:    false,
+			failDataStore:  false,
+			csrTemplate: &x509.CertificateRequest{
+				// Right now, the request will succeed even if URIs is empty. This is
+				// probably not the right behavior.
+				URIs: []*url.URL{workloadID.URL()},
```
azdagron, line 1013: I think this comment can be removed.... the URI is only used to convey the public key. The SPIFFE ID is determined by the server.

pkg/server/api/svid/v1/service.go
```diff
@@ -57,6 +61,25 @@ type Service struct {
 	ca ca.ServerCA
 	ef AuthorizedEntryFetcher
 	td spiffeid.TrustDomain
+	ds datastore.DataStore
+}
+
+func (s *Service) parseAndCheckCSR(ctx context.Context, csrBytes []byte) (*x509.CertificateRequest, error) {
```
azdagron, line 67: How would you feel about making this a standalone function? It doesn't use any state on the `*Service`. It should also be grouped below the exported methods.

pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +347,64 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
+		return nil, status.Errorf(codes.Internal, "malformed entry")
+	}
+
+	csr, err := s.parseAndCheckCSR(ctx, req.Csr)
+	if err != nil {
+		return nil, err
+	}
+
+	x509CASvid, err := s.ca.SignX509CASVID(ctx, ca.X509CASVIDParams{
+		SpiffeID:  spiffeID.String(),
```
azdagron, line 379: The SPIFFEID does not come from the request or entry and should be determined by the trust domain:
```suggestion
		SpiffeID:  s.td.IDString(),
```

proto/spire-next/api/server/svid/v1/svid.proto
```diff
@@ -111,13 +111,13 @@ message NewJWTSVIDResponse {
 }
 
 message NewDownstreamX509CARequest {
-    // Required. The entry ID for the identity being requested.
-    string entry_id = 1;
-
     // Required. The ASN.1 DER encoded Certificate Signing Request (CSR). The
     // CSR is only used to convey the public key; other fields in the CSR are
     // ignored. The X509-SVID attributes are determined by the entry.
```
azdagron, line 116:
```suggestion
    // ignored. The X509-SVID attributes are determined by the downstream entry.
```

proto/spire-next/api/server/svid/v1/svid.proto
```diff
@@ -111,13 +111,13 @@ message NewJWTSVIDResponse {
...
-    bytes csr = 2;
+    bytes csr = 1;
+
+    // No entry_id there are a small number of downstream IDs and we can just 
+    // look for one that matches the CSR
```
azdagron, line 120: This comment feels more like an internal detail and is also incorrect (we don't match up anything with the CSR). I'd suggest removing it.

pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +347,64 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
+		log.WithError(err).Error("Rejecting request due to downstream CA signing rate limit")
+		return nil, err
+	}
+
+	downstreamEntries, isDownstream := rpccontext.CallerDownstreamEntries(ctx)
+	if !isDownstream || downstreamEntries == nil || len(downstreamEntries) == 0 {
+		// The real CallerDownstreamEntries will never return empty, but it helps for writing unit tests
+		err := status.Error(codes.NotFound, "error not found or not authorized")
+		log.WithError(err).Error("Rejecting NewDownstreamX509CA request because caller has no downstream entries")
+		return nil, err
```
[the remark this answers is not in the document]
dfeldman: Either way is fine with me. I don't see any harm in unnecessary checks but I don't want to clutter things.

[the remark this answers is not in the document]
amartinezfayo: >I think I'd lean towards just asserting that isDownstream is true, since that would be a misconfiguration. Having no entries would mean a catastrophic bug in the middleware in violation of the contract.

I'm OK just asserting that `isDownstream` is true. Adding more checks seems to be unnecessary, adding no real benefit (at the cost of more code to maintain).

pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +328,81 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
+	}
+
+	entry := downstreamEntries[0]
+
+	spiffeID, err := api.IDFromProto(entry.SpiffeId)
+	if err != nil || spiffeID.String() != s.td.IDString() {
+		log.WithError(err).Error("Downstream entry has malformed SPIFFE ID")
+		return nil, status.Error(codes.Internal, "downstream entry has malformed SPIFFE ID")
+	}
```
azdagron, line 352: The SPIFFE ID on the downstream entry is not relevant to the SPIFFE ID of the downstream X.509 CA. We don't need this check.
```suggestion
```

pkg/server/api/svid/v1/service.go
```diff
@@ -324,6 +328,81 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svid.NewJWTSVIDRequest) (
...
+	if err := rpccontext.RateLimit(ctx, 1); err != nil {
+		log.WithError(err).Error("Rejecting request due to downstream CA signing rate limit")
+		return nil, err
+	}
+
+	downstreamEntries, isDownstream := rpccontext.CallerDownstreamEntries(ctx)
+	if !isDownstream {
+		err := status.Error(codes.NotFound, "error not found or not authorized")
+		log.WithError(err).Error("Rejecting NewDownstreamX509CA request because caller has no downstream entries")
+		return nil, err
```
azdagron, line 343: This is an internal error. Also, the name of the RPC method is already attached to the logger by the middleware.
```suggestion
		log.Error("Caller is not a downstream workload")
		return nil, status.Error(codes.Internal, "caller is not a downstream workload")
```

Issue #1514, opened 2020-04-20
Document Federation configuration options


esweiss:
Update SPIRE Server documentation to include federation options.

evan2645:
I think we probably need an issue that precedes this one, along the lines
of "Determine the most appropriate way to configure and interact with
SPIRE's federation features".

E.g. should we even allow file-driven configuration, or allow both file and
API, what is the behavior in that case, etc

--
evan

azdagron:
The solution we land on will also need to consider how servers coordinate updates to the federated trust domain bundle when running in HA.

martincapello:
If nobody is against, am going to start documenting what we already have for this, and then when we implement an appropriate solution, we can update it. This way anyone that wants to try **today** how federation works, can use this documentation to have some guidance in setting the current experimental configuration, without having to wait for a better configuration experience.

martincapello:
Alright, just saw https://github.com/spiffe/spire/issues/1619, I will document the options according to what that issue specifies.

evan2645:
Done, available at https://github.com/spiffe/spire/blob/master/doc/spire_server.md#federation-configuration

PR #1656, opened 2020-06-15
Refactor federation configuration

### LINKED ISSUE #1619: [RFC] Federation Graduation
Enabling SPIFFE Federation in SPIRE today is done by the use of an experimental configuration flag. With the goal of promoting it to a stable feature, we need to discuss some issues with the current implementation that might be relevant for various members of the community.

Spire Server currently relies on a file based configuration that is only applied during startup. There are consequences associated with this approach:

- Enabling federation on an already running server requires a restart.
- When running in HA mode, it opens the possibility of a configuration mismatch between different servers in the same cluster. This creates issues with federation, like not all servers exposing the bundle endpoint, or when starting from scratch, two servers in the same cluster could be started with a different trust domain. Although the latter might be beyond the scope of graduating federation.

## Configuration mismatch possible fix

Introducing some form of configuration validation could deal with the mismatch issue. For that purpose, we will need a way of knowing the global cluster configuration so we can compare with what we are trying to configure. Doing this will address the problem of adding a new server, or changing an already existing one, into an improper configuration that doesn't align with the rest of the cluster.

In order to enforce changes to the cluster configuration, we will need a way of expressing that we want to apply configuration changes to the whole cluster, so we don't get rejected by the configuration validation. We will also need to implement the propagation of this new configuration to every server in the cluster. A flag that we can pass when starting the server could suffice to express we want a configuration change.

## Restart possible fix

Alternatively, we could move the federation configuration altogether. Instead of handling it in the configuration file, we could define an API for this purpose. Doing this will have the benefit of not needing to restart the server, and removes the possibility of servers with different federation configurations. It also reduces the cluster management, when it comes to federation, to just one entry point. The main difference will be that at least one server should be running in order to enable federation, as opposed to configuring it in advance.

## Request for Comments

Does this sound reasonable? Are there other approaches to consider? Should we only address one of the issues or both of them?

martincapello:
This PR fixes #1619 and #1514.


cmd/spire-server/cli/run/run.go
```diff
@@ -109,6 +105,14 @@ type caSubjectConfig struct {
 	UnusedKeys   []string `hcl:",unusedKeys"`
 }
 
+type bundleEndpointConfig struct {
+	Enabled    bool                      `hcl:"enabled"`
```
azdagron, line 109: How would you feel about not having this? In my opinion, it should be implied by the `bundle_endpoint` section existing.

cmd/spire-server/cli/run/run.go
```diff
@@ -61,22 +61,24 @@ type Config struct {
...
-	UpstreamBundle      *bool              `hcl:"upstream_bundle"`
+	BindAddress         string                        `hcl:"bind_address"`
+	BindPort            int                           `hcl:"bind_port"`
+	BundleEndpoint      bundleEndpointConfig          `hcl:"bundle_endpoint"`
+	CAKeyType           string                        `hcl:"ca_key_type"`
+	CASubject           *caSubjectConfig              `hcl:"ca_subject"`
+	CATTL               string                        `hcl:"ca_ttl"`
+	DataDir             string                        `hcl:"data_dir"`
+	Experimental        experimentalConfig            `hcl:"experimental"`
+	FederateWith        map[string]federateWithConfig `hcl:"federate_with"`
```
azdagron, line 72: I forgot about https://github.com/hashicorp/hcl/issues/322#issuecomment-561919807. One consequence of moving map into the main section is that we can no longer rely on the `unusedKeys` field for detecting misconfiguration since it does not work correctly with the sections containing maps.

I wonder if we should move all of the federation stuff under a `federation` section:
```
federation {
     bundle_endpoint {
     }
     federate_with "A" {
     }
     federate_with "B" {
    }
}

Alternatively, we could just forego using a map here and just do a list where each item also includes a value with the trust domain. We could check for duplicates during validation.

File: cmd/spire-server/cli/run/run.go (hunk 1 above)
Comment (martincapello) [reply cmd/spire-server/cli/run/run.go:109]: Sounds good to me!

File: cmd/spire-server/cli/run/run.go (hunk 2 above)
Comment (martincapello) [reply cmd/spire-server/cli/run/run.go:72]: > I wonder if we should move all of the federation stuff under a federation section:

I was wondering the same yesterday. I think it would be nice to group the federation related configuration inside its own section.
> Alternatively, we could just forego using a map here and just do a list where each item also includes a value with the trust domain. We could check for duplicates during validation.

Wasn't aware of the hcl issue you mentioned. So, this suggestion makes sense to me.

File: cmd/spire-server/cli/run/run.go [hunk 3]
```diff
@@ -109,6 +110,18 @@ type caSubjectConfig struct {
 	UnusedKeys   []string `hcl:",unusedKeys"`
 }
 
+type federationConfig struct {
+	BundleEndpoint *bundleEndpointConfig         `hcl:"bundle_endpoint"`
+	FederateWith   map[string]federateWithConfig `hcl:"federate_with"`
```
amartinezfayo, line 115: We refer in other places (like https://github.com/spiffe/spire/blob/master/proto/spire/common/common.proto#L68 and here https://github.com/spiffe/go-spiffe/blob/master/proto/spiffe/workload/workload.proto#L44) as `federates_with` instead of `federate_with`. I think that we should keep consistency with the `federates_with` name.

martincapello: Sounds good to me.

cmd/spire-server/cli/run/run.go
```diff
@@ -105,8 +110,12 @@ type caSubjectConfig struct {
...
```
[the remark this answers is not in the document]
martincapello: Sure!

cmd/spire-server/cli/run/run.go
```diff
@@ -348,39 +365,44 @@ func NewServerConfig(c *Config, logOptions []log.Option) (*server.Config, error)
...
-			UseWebPKI:        config.BundleEndpoint.UseWebPKI,
+
+		federateWith := map[string]bundleClient.TrustDomainConfig{}
+		for trustDomain, config := range c.Server.Federation.FederateWith {
+			port := defaultBundleEndpointPort
+			if config.BundleEndpoint.Port != 0 {
+				port = config.BundleEndpoint.Port
+			}
+			if config.BundleEndpoint.UseWebPKI && config.BundleEndpoint.SpiffeID != "" {
+				sc.Log.Warn("The `bundle_endpoint.spiffe_id` configurable is ignored when authenticating with Web PKI")
```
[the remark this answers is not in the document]
martincapello: This **is** using the new config.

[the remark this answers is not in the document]
azdagron: Right, my point is that I think the "graduated" configuration should fail if the spiffe ID is set when configured for web pki. The old config warned. I think the new config should fail. I don't see a reason to be permissive.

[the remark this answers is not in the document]
martincapello: :man_facepalming: missed the warning...yes, seems that failing is the right call here.

doc/spire_server.md
```diff
@@ -96,6 +97,76 @@ The following configuration options are available to configure a plugin:
 
 Please see the [built-in plugins](#built-in-plugins) section below for information on plugins that are available out-of-the-box.
 
+## Federation configuration
```
azdagron, line 100: We should mention that both the `federation.bundle_endpoint` and `federation.federates_with` sections are both optional.

PR #1681, opened 2020-06-24
Add ShiftLeft Scan

prabhu:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**
This PR adds ShiftLeft Scan, a free opensource DevSecOps tool. Scan uses gosec and staticcheck for scanning go projects. A scan of this repo has highlighted a number of issues as shown here - https://github.com/prabhu/spire/runs/804387460?check_suite_focus=true#step:4:27

I would recommend triaging the findings and addressing the key findings that are required. It will also be nice if this PR could be accepted so that security scans are performed as part of PR process.

**Which issue this PR fixes**

### DIFF
--- a/.github/workflows/shiftleft-analysis.yml
+++ b/.github/workflows/shiftleft-analysis.yml
@@ -1,3 +1,6 @@
+# This workflow integrates ShiftLeft Scan with GitHub's code scanning feature
+# ShiftLeft Scan is a free open-source security tool for modern DevOps teams
+# Visit https://docs.shiftleft.io/shiftleft/scan/integrations/github-actions for help
 name: ShiftLeft Scan
 
 on: push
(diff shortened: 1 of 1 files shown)


mcpherrinm:
It looks like your scan failed to run? I don't see any issues presented when I click your link.

We already run gosec and staticcheck on our code in CI, as part of golangci-lint.

prabhu:
The scan ran however the upload step failed on that particular run since this repo is not using GitHub code scanning. I subsequently used upload artifact step. The results can download from this link https://github.com/prabhu/spire/actions/runs/146489201

After raising this PR I noticed that golangci, which use various static analysis such as gosec, staticcheck, is using a custom ignore annotation which is not compatible with the individual tools. I will use this finding as an opportunity to improve scan.

It will be nice if opensource projects can start publishing the reports from security tests as well since it is not clear whether a build and deployment would break with security defects and if all of the ignore tags (eg: tls ignore flags) used in the code have been reviewed.

PR #1724, opened 2020-07-13
Add basic telemetry to the KeyManager interfaces (agent and server)

### LINKED ISSUE #1564: KeyManager plugin needs observability metrics
* **Version**: v0.10.0 and earlier
* **Platform**: all
* **Subsystem**: keymanager

The keymanager interface/plugin currently has too few / no metrics. Given that such a plugin may encounter an error, it’d be nice to have some observability into count of operations + latency.

```
// KeyManager is the client interface for the service type KeyManager interface.
type KeyManager interface {
	GenerateKey(context.Context, *GenerateKeyRequest) (*GenerateKeyResponse, error)
	GetPublicKey(context.Context, *GetPublicKeyRequest) (*GetPublicKeyResponse, error)
	GetPublicKeys(context.Context, *GetPublicKeysRequest) (*GetPublicKeysResponse, error)
	SignData(context.Context, *SignDataRequest) (*SignDataResponse, error)
}

```

While other metrics may point to something being broken, it’s will be faster to look at the metrics for a module (like keymanager) than searching logs.

The additional metrics would also help to see count of disk operations for disk keymanager get a view into custom keymanager plugins with more complexity than the upstream plugins.

To be clear, we have not encountered an outage related-to or exacerbated by not having this information. It’s more low priority / forward-looking spot for improvement work.

This issue filed by @ryancnelson , but inspired by @amoore877 , during a phone conversation.

dfeldman:

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Adds metrics to the server and agent KeyManager interfaces 

**Which issue this PR fixes**
fixes #1564

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -70,6 +70,7 @@ func (a *Agent) Run(ctx context.Context) error {
 		HostServices: []common_catalog.HostServiceServer{
 			common_services.MetricsServiceHostServiceServer(metricsService),
 		},
+		Metrics: metrics,
 	})
 	if err != nil {
 		return err
--- a/pkg/agent/catalog/catalog.go
+++ b/pkg/agent/catalog/catalog.go
@@ -21,6 +21,8 @@ import (
 	wa_k8s "github.com/spiffe/spire/pkg/agent/plugin/workloadattestor/k8s"
 	wa_unix "github.com/spiffe/spire/pkg/agent/plugin/workloadattestor/unix"
 	"github.com/spiffe/spire/pkg/common/catalog"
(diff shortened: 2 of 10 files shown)


dfeldman:
I didn't do this in the PR, but I feel like we should probably eliminate or add noise to the ElapsedTime field from several of these calls, since it is a potential vector for side channel attacks.

dfeldman:
> Thanks, @dfeldman !
> 
> As far as the security related question of introducing some jitter on the elapsed time, would you mind opening a separate issue for that?

Sounds good! I opened issue #1729 .

PR #1737, opened 2020-07-22
Federation configuration related fixes

martincapello:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
1) Federation's federate_with configuration section.
2) idutil.TrustDomainURI function.

**Description of change**
1) All the federation related examples and documentation were updated to use trust domain names, instead of trust domain SPIFFE IDs, in the federate_with configuration section. This is made to be consistent with the use of the trust_domain configurable, which is set to a trust domain name instead of a trust domain ID.
2) There was an issue when using trust domain SPIFFE IDs in the federate_with config section, when this ID was received by the idutil.TrustDomainURI function produced a *URL with a scheme = "spiffe" and a host = the trust domain ID (which already includes the "spiffe" scheme). Then at some point this URL produced a string like "spiffe://spiffe://trustdomain" which then when parsed throws an error because it tries to parse "://trustdomain" as the host's port. Then I updated the TrustDomainURI function to accept trust domain IDs, in which case the "spiffe://" part is removed before creating the URL that is returned. This way, we now are able to use trust domain spiffe IDs in the federate_with section.


pkg/common/idutil/spiffeid.go
```diff
@@ -301,13 +301,15 @@ func normalizeSpiffeIDURL(u *url.URL) *url.URL {
 	return &c
 }
 
-// TrustDomainID creates an trust domain SPIFFE ID given a trust domain.
+// TrustDomainID creates an trust domain SPIFFE ID given a trust domain. If the passed
```
amartinezfayo, line 304:
```suggestion
// TrustDomainID creates a trust domain SPIFFE ID given a trust domain name. If the passed
```

pkg/common/idutil/spiffeid.go
```diff
@@ -301,13 +301,15 @@ func normalizeSpiffeIDURL(u *url.URL) *url.URL {
...
+// trust domain already is a trust domain ID, it is returned unchanged.
 func TrustDomainID(trustDomain string) string {
 	return TrustDomainURI(trustDomain).String()
 }
 
-// TrustDomainURI creates an trust domain SPIFFE URI given a trust domain.
+// TrustDomainURI creates an trust domain SPIFFE URI given a trust domain or trust domain ID.
```
amartinezfayo, line 310:
```suggestion
// TrustDomainURI creates a trust domain SPIFFE URI given a trust domain name or trust domain ID.
```

test/integration/suites/ghostunnel-federation/conf/downstream/server/server.conf
```diff
@@ -12,7 +12,7 @@ server {
         bundle_endpoint {
             port = 8443
         }
-        federates_with "spiffe://upstream-domain.test" {
+        federates_with "upstream-domain.test" {
```
APTy, line 15: Should we keep some integration tests that include the `spiffe://` prefix to ensure it's still working and there haven't been any backwards-incompatible changes?

martincapello: Makes sense to me. I will revert this change so we have one server configured using a trust domain name and the other using a trust domain SPIFFE ID. Also, will add some comments to make this point clear.

PR #1753, opened 2020-07-27
API Refactor: Implement integration tests for agent apis

dfeldman:
Signed-off-by: Daniel Feldman <dan.feldman@hpe.com>

### DIFF
--- a/test/integration/setup/itclient/client.go
+++ b/test/integration/setup/itclient/client.go
@@ -2,25 +2,29 @@ package itclient
 
 import (
 	"context"
+	"crypto/ecdsa"
+	"crypto/tls"
+	"crypto/x509"
 	"flag"
 	"log"
 
 	"github.com/spiffe/go-spiffe/v2/spiffeid"
 	"github.com/spiffe/go-spiffe/v2/spiffetls/tlsconfig"
 	"github.com/spiffe/go-spiffe/v2/workloadapi"
-	"github.com/spiffe/spire/proto/spire/api/server/agent/v1"
-	"github.com/spiffe/spire/proto/spire/api/server/bundle/v1"
-	"github.com/spiffe/spire/proto/spire/api/server/entry/v1"
-	"github.com/spiffe/spire/proto/spire/api/server/svid/v1"
+	agent "github.com/spiffe/spire/proto/spire/api/server/agent/v1"
+	bundle "github.com/spiffe/spire/proto/spire/api/server/bundle/v1"
+	entry "github.com/spiffe/spire/proto/spire/api/server/entry/v1"
+	svid "github.com/spiffe/spire/proto/spire/api/server/svid/v1"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 )
 
 var (
-	tdFlag           = flag.String("trustDomain", "domain.test", "server trust domain")
-	socketPathFlag   = flag.String("socketPath", "unix:///tmp/agent.sock", "agent socket path")
-	serverAddrFlag   = flag.String("serverAddr", "spire-server:8081", "server addr")
-	expectErrorsFlag = flag.Bool("expectErrors", false, "client is used to validate permission errors")
+	tdFlag               = flag.String("trustDomain", "domain.test", "server trust domain")
+	socketPathFlag       = flag.String("socketPath", "unix:///tmp/agent.sock", "agent socket path")
+	serverAddrFlag       = flag.String("serverAddr", "spire-server:8081", "server addr")
+	serverSocketPathFlag = flag.String("serverSocketPath", "unix:///tmp/spire-registration.sock", "server socket path")
+	expectErrorsFlag     = flag.Bool("expectErrors", false, "client is used to validate permission errors")
 )
 
 type Client struct {
@@ -53,19 +57,57 @@ func New(ctx context.Context) *Client {
 	return &Client{
 		Td:           td,
 		ExpectErrors: *expectErrorsFlag,
+		connection:   conn,
+		source:       source,
+	}
+}
 
-		connection: conn,
-		source:     source,
+func NewInsecure(ctx context.Context) *Client {
+	flag.Parse()
+	tlsConfig := tls.Config{
+		InsecureSkipVerify: true, // nolint: gosec
+	}
+	conn, err := grpc.DialContext(ctx, *serverAddrFlag, grpc.WithTransportCredentials(credentials.NewTLS(&tlsConfig)))
+	if err != nil {
+		log.Fatalf("Error creating dial: %v", err)
+	}
+
+	return &Client{
+		ExpectErrors: *expectErrorsFlag,
+		connection:   conn,
+	}
+}
+
+func NewWithCert(ctx context.Context, cert *x509.Certificate, key *ecdsa.PrivateKey) *Client {
+	flag.Parse()
+
+	tlsConfig := tls.Config{
+		GetClientCertificate: func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+			return &tls.Certificate{
+				Certificate: [][]byte{cert.Raw},
+				PrivateKey:  key,
+			}, nil
+		},
+		InsecureSkipVerify: true, // nolint: gosec
+	}
+	conn, err := grpc.DialContext(ctx, *serverAddrFlag, grpc.WithTransportCredentials(credentials.NewTLS(&tlsConfig)))
+	if err != nil {
+		log.Fatalf("Error creating dial: %v", err)
+	}
+
+	return &Client{
+		ExpectErrors: *expectErrorsFlag,
+		connection:   conn,
 	}
 }
 
 func (c *Client) Release() {
-	c.connection.Close()
-	c.source.Close()
-}
-
-func (c *Client) AgentClient() agent.AgentClient {
-	return agent.NewAgentClient(c.connection)
+	if c.connection != nil {
+		c.connection.Close()
+	}
+	if c.source != nil {
+		c.source.Close()
+	}
 }
 
 func (c *Client) BundleClient() bundle.BundleClient {
@@ -75,6 +117,37 @@ func (c *Client) BundleClient() bundle.BundleClient {
 func (c *Client) EntryClient() entry.EntryClient {
 	return entry.NewEntryClient(c.connection)
 }
+
 func (c *Client) SVIDClient() svid.SVIDClient {
 	return svid.NewSVIDClient(c.connection)
+}
+
+func (c *Client) AgentClient() agent.AgentClient {
+	return agent.NewAgentClient(c.connection)
+}
+
+// Open a client ON THE SPIRE-SERVER container
+// Used for creating join tokens
+type LocalServerClient struct {
+	connection *grpc.ClientConn
+}
+
+func (c *LocalServerClient) AgentClient() agent.AgentClient {
+	return agent.NewAgentClient(c.connection)
+}
+
+func (c *LocalServerClient) Release() {
+	c.connection.Close()
+}
+
+func NewLocalServerClient(ctx context.Context) *LocalServerClient {
+	flag.Parse()
+	conn, err := grpc.DialContext(ctx, *serverSocketPathFlag, grpc.WithInsecure())
+	if err != nil {
+		log.Fatalf("Error creating dial: %v", err)
+	}
+
+	return &LocalServerClient{
+		connection: conn,
+	}
 }
--- a/test/integration/setup/node-attestation/client.go
+++ b/test/integration/setup/node-attestation/client.go
@@ -1,0 +1,159 @@
+package main
+
+import (
+	"context"
+	"crypto/rand"
+	x509 "crypto/x509"
+	"encoding/pem"
+	"flag"
+	"fmt"
+	"log"
+	mathrand "math/rand"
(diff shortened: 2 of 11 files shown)


test/integration/suites/node-attestation/conf/agent/agent.conf
```diff
@@ -0,0 +1,27 @@
+agent {
```
amartinezfayo, line 1: This config file is not needed

test/integration/setup/node-attestation/client.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	s1 := mathrand.NewSource(time.Now().UnixNano())
+	tokenID := mathrand.New(s1).Intn(1000000)
+
+	tokenName := fmt.Sprintf("test_token_%v", tokenID)
+
+	// Create a join token using the local socket connection (simulating the CLI running on the spire-server)
+	agentClient := c.AgentClient()
+	_, err = agentClient.CreateJoinToken(ctx, &agent.CreateJoinTokenRequest{Ttl: 1000, Token: tokenName})
```
MarcosDY, line 48: must not you use the token from create join token? (take a look to admin rpcs ITs)

test/integration/suites/node-attestation/README.md
```diff
@@ -0,0 +1,6 @@
+# Node attestation Suite
+
+## Description
+
+Basic tests of the node attestation APIs using a simple fake agent
```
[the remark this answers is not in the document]
dfeldman: Hmm I will look into it
I don't want to re-implement too much of the agent since that would take a long time and not help with our testing

test/integration/suites/node-attestation/conf/agent/agent.conf
dfeldman: will delete :)

test/integration/setup/node-attestation/client.go
dfeldman: This does actually work since the join token is fixed

test/integration/setup/node-attestation/client.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+	c := itclient.NewLocalServerClient(ctx)
+	defer c.Release()
+
+	// Validate call to New Downstream X509 CA
+	err := testAgentAPI(ctx, c)
```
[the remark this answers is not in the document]
dfeldman: Good point

test/integration/suites/node-attestation/conf/server/server.conf
```diff
@@ -0,0 +1,30 @@
...
+}
+
+plugins {
+	DataStore "sql" {
+		plugin_data {
+			database_type = "sqlite3"
+			connection_string = "/opt/spire/data/server/datastore.sqlite3"
+		}
+	}
+	NodeAttestor "x509pop" {
```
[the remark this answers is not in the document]
dfeldman: Hmm it seems to work like this since join token doesn't need a plugin
I'll try adding other node attestation if I can

test/integration/setup/itclient/client.go
```diff
@@ -2,25 +2,29 @@ package itclient
...
 	"github.com/spiffe/go-spiffe/v2/spiffetls/tlsconfig"
 	"github.com/spiffe/go-spiffe/v2/workloadapi"
-	"github.com/spiffe/spire/proto/spire/api/server/agent/v1"
-	"github.com/spiffe/spire/proto/spire/api/server/bundle/v1"
-	"github.com/spiffe/spire/proto/spire/api/server/entry/v1"
-	"github.com/spiffe/spire/proto/spire/api/server/svid/v1"
+	agent "github.com/spiffe/spire/proto/spire/api/server/agent/v1"
+	bundle "github.com/spiffe/spire/proto/spire/api/server/bundle/v1"
+	entry "github.com/spiffe/spire/proto/spire/api/server/entry/v1"
+	svid "github.com/spiffe/spire/proto/spire/api/server/svid/v1"
```
dfeldman, line 17: Aliases needed to make go vet pass completely... I'm not totally sure why

test/integration/suites/node-attestation/conf/agent/agent.conf
amartinezfayo: It seems that agent.conf is still there?

test/integration/suites/node-attestation/03-test-node-attestation
```diff
@@ -0,0 +1,32 @@
...
+docker-compose exec -u 1001 -T spire-server /opt/spire/conf/server/node-attestation -testStep ban -tokenName ${jointoken} 
+if [[ $? -ne 0 ]];
+then
+	fail-now "Failed to do initial join token attestation"
+fi
+echo "Agent banned"
+
+docker-compose exec -u 1001 -T spire-server /opt/spire/conf/server/node-attestation -testStep renew -certificate "${svid2}"
+if [[ $? -eq 0 ]];
+then
```
amartinezfayo, line 30: I believe that this will make the test to fail, since it will exit with a non-0 code here.

```suggestion
if docker-compose exec -u 1001 -T spire-server /opt/spire/conf/server/node-attestation -testStep renew -certificate "${svid2}" ; then
```

test/integration/setup/node-attestation/client.go
```diff
@@ -0,0 +1,160 @@
...
+		log.Fatalf("failed to parse cert: %v\n", err)
+	}
+
+	agentRemoteConn := itclient.NewWithCert(ctx, *cert, key)
+	defer agentRemoteConn.Release()
+	agentRemoteClient := agentRemoteConn.AgentClient()
+
+	csr, err := x509.CreateCertificateRequest(rand.Reader, &x509.CertificateRequest{}, key)
+	if err != nil {
+		log.Fatalf("fails to create CSR: %v", err)
```
amartinezfayo, line 126:
```suggestion
		log.Fatalf("failed to create CSR: %v", err)
```

test/integration/suites/node-attestation/README.md
```diff
@@ -0,0 +1,6 @@
...
+Since we don't actually need the real agent, everything just runs in the spire-server image
```
amartinezfayo, line 6: I guess that this can be removed now.

test/integration/suites/node-attestation/03-test-node-attestation
dfeldman: I will check

dfeldman: Ok, I think either way is equivalent. We want the test to fail IF the exit value is zero (since renewing after banning should not work) :) .

test/integration/suites/node-attestation/00-setup
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+echo ${ROOTDIR}
+"${ROOTDIR}/setup/x509pop/setup.sh" conf/server conf/ # Throw the agent cert in conf/, it is unused
```
MarcosDY, line 3: we are not using x509pop, you can remove it, or `maybe` update code to test an x509 attestation? just to veirfy another thing than jointoken?

test/integration/suites/node-attestation/03-test-node-attestation
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+# Test node attestation api
+jointoken=`docker-compose exec -u 1001 -T spire-server /opt/spire/conf/server/node-attestation -testStep jointoken`
```
MarcosDY, line 4: instead using a new client, why not to use regular server CLI? we need it only to create a token

test/integration/suites/node-attestation/03-test-node-attestation
```diff
@@ -0,0 +1,37 @@
...
+echo "Created Join Token" $jointoken
+
+svid1=`docker-compose exec -u 1001 -T spire-agent /opt/spire/conf/agent/node-attestation -testStep jointokenattest -tokenName $jointoken`
```
MarcosDY, line 7: can you add more steps to validate what happens when calling all those "agent specific" RPCs with a regular SVID?

test/integration/suites/node-attestation/03-test-node-attestation
```diff
@@ -0,0 +1,37 @@
...
+docker-compose exec -u 1001 -T spire-server /opt/spire/conf/server/node-attestation -testStep ban -tokenName ${jointoken} 
+if [[ $? -ne 0 ]];
+then
+	fail-now "Failed to do initial join token attestation"
+fi
+echo "Agent banned"
+
+if docker-compose exec -u 1001 -T spire-server /opt/spire/conf/server/node-attestation -testStep renew -certificate "${svid2}" 
+then
+	fail-now "OOOPS"
```
MarcosDY, line 30: "Expect agent to be banned"

test/integration/suites/node-attestation/03-test-node-attestation
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+#docker-compose exec -u 1001 -T spire-server /opt/spire/conf/server/node-attestation -testStep renew -certificate "${svid2}"
```
MarcosDY, line 33: I think we can add more steps if working we do attestion with x509 pop, for example we can ban and verify we has no longer access then, we can remove banned entry that will allow to attest it again, that is something we wanted when adding ban basically a way to prevent reattestation, but that is allowed once we delete banned entry, do you mind to add and example like that?

test/integration/suites/node-attestation/03-test-node-attestation
```diff
@@ -0,0 +1,37 @@
...
+#if [[ $? -eq 0 ]];
```
MarcosDY, line 34: remove commented step

test/integration/suites/node-attestation/03-test-node-attestation
MarcosDY: possible steps::
Attest agent
delete agent
reattest 
ban agent 
reatest (fails)
delete agent 
reatest (success)

test/integration/suites/node-attestation/conf/server.conf
```diff
@@ -0,0 +1,30 @@
...
+}
+
+plugins {
+	DataStore "sql" {
+		plugin_data {
+			database_type = "sqlite3"
+			connection_string = "/opt/spire/data/server/datastore.sqlite3"
+		}
+	}
+	NodeAttestor "x509pop" {
```
MarcosDY, line 22: mmm why join token works without

```
    NodeAttestor "join_token" {
        plugin_data {
        }
    }
```
may it fails? @amartinezfayo

test/integration/suites/node-attestation/conf/server.conf
```diff
@@ -0,0 +1,30 @@
+server {
```
MarcosDY, line 1: one of this server files is duplicated, can you remove and keep just one?

test/integration/setup/itclient/client.go
```diff
@@ -53,19 +57,62 @@ func New(ctx context.Context) *Client {
...
+		connection:   conn,
+		source:       source,
+	}
+}
 
-		connection: conn,
-		source:     source,
+func NewInsecure(ctx context.Context) *Client {
+	flag.Parse()
+	/* #nosec */
```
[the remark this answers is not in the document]
dfeldman: This would be better but it doesn't seem to work with gosec :(

test/integration/setup/itclient/client.go
```diff
@@ -53,19 +57,62 @@ func New(ctx context.Context) *Client {
...
+	tlsConfig := tls.Config{
+		GetClientCertificate: func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+			return &tls.Certificate{
+				Certificate: [][]byte{cert.Raw},
+				PrivateKey:  key,
+			}, nil
+		},
+		InsecureSkipVerify:    true,
+		VerifyPeerCertificate: nil,
```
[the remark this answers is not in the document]
dfeldman: You are right, it is the default ...

test/integration/setup/node-attestation/client.go
```diff
@@ -0,0 +1,160 @@
...
+	case "ban":
+		doBanStep(ctx)
+	case "renew":
+		doRenewStep(ctx)
+	default:
+		log.Fatalf("error: unknown test step\n")
+	}
+}
+
+func doJoinTokenStep(ctx context.Context) {
```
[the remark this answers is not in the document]
dfeldman: The reason was to fully execise the new V1 APIs... we rewrote CreateJoinToken

test/integration/suites/node-attestation/conf/server.conf
dfeldman: I don't think this is needed with the new implementation (Join Token attestation isn't implemented in a plugin, it's right in the API handler)

Maybe we should require it? But it is not required right now

amartinezfayo:
@dfeldman looks like DCO is missing, could you fix that?

test/integration/suites/node-attestation/00-setup
dfeldman: This is actually needed just because that script is used to bootstrap the cert
It would be cleaner to have a different script just for bootstrapping

PR #1786, opened 2020-08-11
adds mask to GetAuthorizedEntries call

kunzimariano:
This PR makes the Agent call `GetAuthorizedEntries` with a mask that only includes `SpiffeId, Selectors, FederatesWith and RevisionNumber`. By doing so removes `ParentId, DnsNames, Ttl, Admin, Downstream, and ExpiresAt`.

`ParentId` and `DnsNames` were already discussed https://github.com/spiffe/spire/issues/1361. Their removal provide a great reduction in data transmission. 

`Ttl, Admin, Downstream, and ExpiresAt` don't provide such a great benefit, but their usage seem unnecessary on the Agent side. Let me know if  this could carry unintended consequences that I'm not aware of.

One last thing worth mentioning. When running integration test, this is only exercised using the Agent with the experimental API enabled. For that reason I ran them locally with the experimental flag. If I'm not mistaken, the same happens with the tests when they run in the CI, making this change not fully exercised.


amartinezfayo:
>One last thing worth mentioning. When running integration test, this is only exercised using the Agent with the experimental API enabled. For that reason I ran them locally with the experimental flag. If I'm not mistaken, the same happens with the tests when they run in the CI, making this change not fully exercised.

We will unconditionally enable the new APIs on agent soon.

pkg/agent/client/util.go
```diff
@@ -32,22 +31,12 @@ func registrationEntryFromProto(e *types.Entry) (*common.RegistrationEntry, erro
 	if e.Id == "" {
 		return nil, fmt.Errorf("missing entry ID")
 	}
-	parentID, err := spiffeIDFromProto(e.ParentId)
```
amartinezfayo, line 35: I'm wondering if we should rename the `registrationEntryFromProto` function to reflect the fact that only the "essential" fields are returned. Otherwise could be confusing if in the future this function is called in other places expecting to map all the fields.

kunzimariano: Absolutely, although this is not a public function the name can be misleading. Does `slicedEntryFromProto` sound like a reasonable name?

amartinezfayo: Yes, I think that's a better name.

PR #1806, opened 2020-08-24
New downloading.md describes download options

sanderson042:
This new document has two purposes:
* To summarize the download and installation options for various platforms
* To serve as a landing page for download info on the new spiffe.io doc layout

Signed-off-by: Steve Anderson <steve.anderson@hpe.com>

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**


doc/downloading.md
```diff
@@ -0,0 +1,27 @@
+# Downloading SPIRE
+
+This page describes the options available for downloading SPIRE.
+
+## Docker Compose
```
amartinezfayo, line 5: The SPIRE 101 walkthrough is outdated (I filed #1815 to update it) and we don't have a docker-compose file anymore. In any case I think that the SPIRE 101 walkthrough relies on cloning the SPIRE repository first as the primary download method. I don't think that we should have a `Docker Compose` section here.

doc/downloading.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Linux Executables and Source
+
+* The SPIRE GitHub [releases](https://github.com/spiffe/spire/releases) page has download links and changelogs for each SPIRE release
+
+* The spiffe.io [Get SPIRE](https://spiffe.io/downloads/) page has additional download options and instructions for building SPIRE yourself
+
+* [Quickstart for Linux and MacOS X](https://spiffe.io/docs/latest/spire/installing/getting-started-linux-macos-x/) describes how to download and test a simple one-node installation of the SPIRE Server and Agent
+
+## MacOS Source
```
amartinezfayo, line 25: I'm not sure if this is the right title for this section, since it seems that what we want to describe is that we don't have pre-built SPIRE executables available for MacOS. So maybe `MacOS executables` would be a better choice?

doc/downloading.md
sanderson042: Hi Agustín - See my comments in 303552e and let me know what you think.

doc/running_spire.md
```diff
@@ -0,0 +1,27 @@
+# Options for Running SPIRE
```
amartinezfayo, line 1: Since there are many additional ways in which you can run SPIRE, I'm thinking that maybe this could be titled "Start using SPIRE" or something along those lines. Then the description below would be something like "This page describes the some options to get started with SPIRE."
What do you think?

amartinezfayo (commented):
Thank you @sanderson042!
I have one additional comment about the title. I think that it may make people think that these are the only ways to run SPIRE, while SPIRE can actually be run in a wider range of ways. I left a comment with a suggestion about that.

doc/running_spire.md
sanderson042: Sure, changes seems ok; see the new commit. I changed the file name to match. I also changed the Linux and MacOS sections to remove mentions of executables and source--I think those apply to both really, so those qualifiers aren't needed.

PR #1814, opened 2020-09-01
Fix fatal error when k8s registrar leader election enabled

JonathanO:
The LeaderElectionID used to have a default in older versions of controller-runtime, but now needs setting explicitly. Without it set the registrar will fatal error when reconcile/crd modes have leader election enabled. Populate it with a sensible value, based on the configured controller name for reconcile mode, or a fixed value for crd mode.

Also enable leader election in the integration test for reconcile, since it's a flag that's unlikely to get adhoc testing otherwise.


JonathanO:
@faisal-memon I've added the parameter for crd mode too. It has no impact if leader election is disabled. It just sets the name of the ConfigMap to use for leader election. The behaviour was changed in controller-runtime because the default used to be "controller", which led to collisions between different controllers if people hadn't altered it to make it unique to their controller.

Issue #1446, opened 2020-03-17
spiffe_id fields in SQL database tables have incorrect maximum length constraint


rturner3:
The `spiffe_id` field in the following SQL database tables resorts to the default maximum length used by gorm of 255 characters, a type of `varchar(255)`:

- `attested_node_entries`
- `node_resolver_map_entries`
- `registered_entries`

The SPIFFE Specification says that SPIFFE implementations MUST support SPIFFE URIs up to 2048 bytes in length. See [SPIFFE ID Section 2.3](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#23-maximum-spiffe-id-length) for reference.

See [gorm documentation](https://gorm.io/docs/models.html) of `size` struct tag for default size of `varchar` fields.

evan2645:
Thank you for reporting this @rturner3, it is definitely something we need to address.

Would be great to get a patch in for the next release, which is coming in the next 2-3 weeks.

Are there any compatibility concerns with this kind of migration?

rturner3:
To my knowledge, increasing the size of a `varchar` field in SQL shouldn't impact any existing data. And 2048 bytes is well within the range of maximum sizes allowed for `varchar` fields in some of the most popular SQL databases:

```
| Database   | Max varchar size (bytes) |
|------------|--------------------------|
| MySQL      |                   65,535 |
| PostgreSQL |               10,485,760 |
| sqlite     |            2,147,483,647 |
| Oracle     |                    4,000 |
```

References:
[MySQL 5.6 limits](https://dev.mysql.com/doc/refman/5.6/en/char.html)
[PostgreSQL limits](https://github.com/postgres/postgres/blob/fe604800681aed8228fbe67d2e175d6822c66ea2/src/include/access/htup_details.h#L578) (not documented explicitly anywhere, but is derived from this value in the code)
[SQLite limits](https://www.sqlite.org/limits.html)
[Oracle limits](https://docs.oracle.com/cd/B28359_01/server.111/b28320/limits001.htm#i287903)

rturner3:
[...]

PR #1859, opened 2020-09-22
Cache registration entry and Agent selectors in memory

rturner3:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
- Lookup procedure of authorized registration entries in SPIRE Server for a SPIRE Agent used in the following Server RPCs:
  - `GetAuthorizedEntries`
  - `BatchNewX509SVID`
  - `NewJWTSVID`

**Description of change**
When SPIRE Server evaluates the set of registration entries that an Agent is authorized to distribute identities for over the workload API, it no longer queries for this information from the datastore plugin. Now all registration entries and node selectors are cached in memory to optimize the lookup of the authorized entries for an Agent.

The in-memory cache is built in the context of a request to any of the RPCs which use the cache on a periodic basis. The cache refresh interval is currently set to 5 seconds, the same duration as the Agent sync interval. When a client of the entry cache attempts to access the cache, it compares the current time to the last time the cache was rebuilt, and if more than 5 seconds has elapsed, it will rebuild the cache.

This changes the consistency model of the propagation of registration entries in SPIRE from being strongly consistent to eventually consistent. Now there is a maximum 10 second propagation delay of registration entries to Agents (5 seconds for cache rebuild interval + 5 seconds for Agent sync interval).

This change greatly improves the average latency of the Agent sync with the Server when Agents manage a large set of registrations (>50). Previously, the Agent sync operation involved `>= 2*N` datastore queries, where `N` is the number of authorized registrations for an Agent.

With 50,000 Agents and 600 registrations, the cache consumes around 20 MB of memory, so introducing this cache does not significantly change the resource requirements for SPIRE Server.

Big thanks to @azdagron for authoring the core of this PR.

**Which issue this PR fixes**
#980


pkg/server/cache/entrycache/fullcache_ds.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"github.com/spiffe/spire/pkg/server/api"
+	"github.com/spiffe/spire/pkg/server/plugin/datastore"
+	"github.com/spiffe/spire/proto/spire/common"
+	"github.com/spiffe/spire/proto/spire/types"
+)
+
+const (
+	agentPageSize = 5000
+	entryPageSize = 5000
```
azdagron, line 17: Hmm, wonder if we should make these an order of magnitude smaller. SPIFFE IDs can technically be 2048 bytes a piece, which means if there were large registration entries, we could end up surpassing the 4MiB limit.

pkg/server/cache/entrycache/fullcache_test.go
```diff
@@ -0,0 +1,787 @@
...
+	// - agent/active - has a CertNotAfter that is still valid
+	// - agent/expired - has a CertNotAfter that expired
+	//
+	// agent/active maps to group/0 and group/1 based on selector subset matches.
+	// agent/expired maps to group/0 and group/2 based on selector subset matches.
+	//
+	// Normally, agent/expired should be authorized to receive group/0, workload/0, group/2, workload/2, and workload/4.
+	// However, the cache filters out all entries related to the expired agent other than ones shared with other Agents
+	// through node selector subset matching - in this case, just workload/0.
+	// In reality, an expired agent should not be able to request its authorized entries because endpoint security (mTLS)
```
azdagron, line 248: nit: mTLS will only verify the SVID validity during the handshake. Since agent connections live for several minutes, the authorization middleware takes care of ensuring that the SVID is still valid on each RPC request over the "long-lived" connection.

azdagron (commented):
Thanks for polishing this up @rturner3 ! Just a small comment about paging size and a comment nit...

pkg/server/cache/entrycache/fullcache_ds.go
rturner3: Good point. If we make the assumption that registration entries are generally expected to be smaller than 8 KB, then I think a page size of 500 would be workable.

That seems like a reasonable tradeoff between supporting large registration entries and supporting many registration entries, which would result in more paged queries with a lower page size.

rturner3: This reminds me... https://github.com/spiffe/spire/issues/1446 :)

pkg/server/cache/entrycache/fullcache_test.go
rturner3: That's true. I updated the wording to be more specific in that regard.

rturner3:
Rebased and squashed my contributions in the PR.

azdagron (approved):
Thanks @rturner3 !

elinesterov:
Hey folks, I just wanted to share some perf testing results we did with the fix. Special thanks to @yangmarcyang and @soundless, who did all the heavy lifting.

We see significant improvements in response time and resource usage especially when the DB gets close to a limit. Here are some graphs below ( more data to come with specific details of the env).

Node API

DB QPS

CPU Usage

azdagron:
That's good news @elinesterov. I do get a 403 trying to look at those images, but I'm happy to wait for the next round of data.

elinesterov:
@azdagron I knew that sharing imaged from google drive won't be easy :) I hope I found the right fix. It seems for the rest of the data is easier to package it up as gDoc.

azdagron:
Nice graphs! Any info on server memory usage with and without?

elinesterov:
Unfortunately, we didn't capture memory usage for spire only for DB :( I'll describe what we did in a separate ticket. There is definitely a huge room for improvement since the process we used is not scalable and hardly repeatable (from an infrastructure perspective).  We need to figure out how to use load/perf tools for gRPC to similate the agent traffic (might be along with some dummy attestor) instead of using real agents.

PR #1863, opened 2020-09-24
Use bytes instead of string in payload proto

marcosy:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Attestation payload proto file.

**Description of change**

The new API type `AttestationData` defined in `attestation.proto` file is using the `string` type for the payload instead of the `bytes` type that was used in the old API. This caused some external attestor plugins that worked well with SPIRE v0.10.1 to fail in v0.11.0 with the following error: 

```
error="failed to get SVID: error sending attestation request to SPIRE server: rpc error: code = Internal desc = grpc: error while marshaling: proto: field \"spire.types.AttestationData.Payload\" contains invalid UTF-8"
```
The root cause of the problem is that protobuf enforces UTF-8 encoding for string types.

This PR switches the payload to `bytes` type. Since string and byte are [wire compatible](https://developers.google.com/protocol-buffers/docs/encoding#structure), this change does not impacts compatibility.

**Which issue this PR fixes**

No known issue reported.


azdagron (approved):
Thanks and good catch, @marcosy !

PR #1871, opened 2020-09-29
Make it possible to set role name at authentication

hiyosi:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

UpstreamAuthority `vault` plugin.

**Description of change**

This PR adds the ability to identify a role when performing authentication.
The vault defaults to trying all certificate roles and returning any one that matches.
If set, Authenticate against only the named certificate role. 

c.f. https://www.vaultproject.io/api/auth/cert#parameters-7

**Which issue this PR fixes**

### DIFF
--- a/pkg/server/plugin/upstreamauthority/vault/vault_fake_test.go
+++ b/pkg/server/plugin/upstreamauthority/vault/vault_fake_test.go
@@ -30,6 +30,7 @@ pki_mount_point = "test-pki"
 ca_cert_path = "_test_data/keys/EC/root_cert.pem"
 cert_auth {
    cert_auth_mount_point = "test-cert-auth"
+   cert_auth_role_name = "test"
    client_cert_path = "_test_data/keys/EC/client_cert.pem"
    client_key_path  = "_test_data/keys/EC/client_key.pem"
 }`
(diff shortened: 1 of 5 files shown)


pkg/server/plugin/upstreamauthority/vault/vault_fake_test.go
```diff
@@ -92,8 +93,7 @@ approle_auth {
   "auth": {
     "client_token": "cf95f87d-f95b-47ff-b1f5-ba7bff850425",
     "policies": [
-      "web",
-      "stage"
+      "test"
```
amartinezfayo, line 96: I don't think that this is explicitly consumed and validated in the test. Does this make any difference?

amartinezfayo (commented):
Thank you @hiyosi for this!
I have one small comment about the test.

pkg/server/plugin/upstreamauthority/vault/vault_fake_test.go
hiyosi: You're right, the changes will not affect the results of the test.
I'll revert this.

hiyosi: 40b9f38

amartinezfayo (approved):
Thank you @hiyosi!

amartinezfayo:
@hiyosi could you please update the branch? Thanks!

hiyosi:
@amartinezfayo Thank you for your review.

amartinezfayo (approved):

PR #1884, opened 2020-09-30
Bump project version to v0.12.0

APTy:


azdagron (approved):
Thanks, @APTy! Looks like we missed an update to this after the 0.11.0 release :man_facepalming:

Issue #1895, opened 2020-10-06
BatchDeleteFederatedBundle RPC should support the RESTRICT, DISSOCIATE and DELETE modes


amartinezfayo:
The old registration API allowed to specify a `mode` (RESTRICT, DISSOCIATE, DELETE) for the `DeleteFederatedBundle` RPC in the `DeleteFederatedBundleRequest` message.
The new Bundle API always uses the RESTRICT mode in the `BatchDeleteFederatedBundle` RPC.
In order to not break backward compatibility, we need to restore the ability to specify a mode when calling the new `BatchDeleteFederatedBundle` RPC.

PR #1897, opened 2020-10-07
add mode to delete bundle RPC in bundle api

MarcosDY:
Add delete mode to Bundle API

**Which issue this PR fixes**
fixes: #1895


azdagron (commented):
My only concern here is the error messaging? Wonder if we can detect and return this condition from the datastore and return FAILED_PRECONDITION instead of UNKNOWN...
```
rpc error: code = Unknown desc = datastore-sql: cannot delete bundle; federated with 1 registration entries",
```

azdagron (approved):
\o/

PR #1942, opened 2020-10-27
Fix transient failures in mTLS integration tests

evan2645:
This commit fixes spurious test suite failure caused by some racing
components (socat and a proxy). Occasionally (and more frequently on
some platforms) socat is called to write to a socket that doesn't exist
yet. This condition didn't originally affect the outcome of the test,
however a recent bugfix that enforces `set -e` in the test framework
exposed it.

To fix it, move the socat call to a conditional.


azdagron (approved):
awesome \o/ thanks for the fix, @evan2645!

PR #1944, opened 2020-10-27
Remove pagination from hydration of entries cache

rturner3:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
In-memory registration entry cache, authorized registration entry sync between Agent and Server

**Description of change**
Pagination was previously used because there was a gRPC layer separating
SPIRE core from the sql datastore plugin. That gRPC boundary no longer
exists, so there are no longer concerns about query result exceeding the
maximum response message size of gRPC. It is more performant to do the
full table scan in one go rather than through sequential piecemeal
paginated requests.

### DIFF
--- a/pkg/server/cache/entrycache/fullcache_ds.go
+++ b/pkg/server/cache/entrycache/fullcache_ds.go
@@ -14,7 +14,6 @@ import (
 
 const (
 	agentPageSize = 500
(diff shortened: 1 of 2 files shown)


azdagron (approved):
\o/

PR #1973, opened 2020-11-23
Add `ExpiresAt` to `entry show` output

### LINKED ISSUE #1971: SPIRE Server `entry show` command doesn't display `entryExpiry` field
SPIRE Server `entry create` allows to set a field `entryExpiry`, but this field is not displayed by the command `entry show`.

* **Version**: 0.11.2
* **Subsystem**: server

maxlambrecht:

**Pull Request check list**

- [ ] Documentation updated?

**Description of change**
Adds `ExpiresAt` field to the `entry show` output when the field is other than zero. 

**Which issue this PR fixes**
Fixes #1971

### DIFF
--- a/cmd/spire-server/cli/entry/util.go
+++ b/cmd/spire-server/cli/entry/util.go
@@ -47,6 +47,10 @@ func printEntry(e *types.Entry, env *common_cli.Env) {
 		env.Printf("TTL           : %d\n", e.Ttl)
 	}
 
+	if e.ExpiresAt != 0 {
+		env.Printf("ExpiresAt     : %d\n", e.ExpiresAt)
+	}
+
 	for _, s := range e.Selectors {
 		env.Printf("Selector      : %s:%s\n", s.Type, s.Value)
 	}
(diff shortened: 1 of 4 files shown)


cmd/spire-server/cli/entry/util.go
```diff
@@ -47,6 +47,10 @@ func printEntry(e *types.Entry, env *common_cli.Env) {
 		env.Printf("TTL           : %d\n", e.Ttl)
 	}
 
+	if e.ExpiresAt != 0 {
+		env.Printf("ExpiresAt     : %d\n", e.ExpiresAt)
```
amartinezfayo, line 51: Since the output is meant to be consumed by humans, I think that we should show this formatted as a readable date. That would also be consistent with other places where we show expiration time, like in the `agent list` command.

```suggestion
		env.Printf("Expires At    : %s\n", time.Unix(e.ExpiresAt, 0))
```

amartinezfayo (commented):
Thanks for this @maxlambrecht!
I have one comment regarding the output format.

cmd/spire-server/cli/entry/util.go
amartinezfayo: To keep consistency, I'm thinking that maybe `Expiration time` is a more appropriate label?

amartinezfayo (approved):
🚀

Issue #1935, opened 2020-10-26
Feature Request: 'k8s:host-ip' selector


hiyosi:
* **Version**: 
* **Platform**: 
* **Subsystem**: Workload Attestor 'k8s' plugin

If the workload deal with the Admin SVID, we want to make it difficult to get the Admin SVID from another Node, even if the Agent SVID or Node Attestation Data is stolen.

One way to realize this is to restrict the nodes which the workload or the agent can launch.

#1273
I think what I'm thinking about is similar to this issue.
However it has not yet been discussed.

If the workload runs on K8s, currently the Workload Attestor 'k8s' plugin provides the k8s:node-name selector that references the .spec.nodeName of Pod Object as a way to restrict the Node.
But it seems to be a less recommended way of restricting nodes.
(See: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodename)
More common way, Taint/Toleration and Affinity/AntiAffinity might be used.
Also, K8s Node name is relatively easy to spoof, so if an Agent's SVID/Attestation Data is stolen, the Node name restriction can be easily breached.

So, as another way of restricting the nodes, I would like the k8s plugin to provide the `k8s:host-ip` selector.
The workload attestor 'k8s' plugin generates `k8s:host-ip` from Pod's `.status.hostIP` field.
In a limited environment where IP spoofing is difficult and RegistrationEntry can be managed at the IP level, this can be useful.
Ideally, it would be nice to be able to specify a range with CIDR or something like that, but it would be difficult with the current implementation of Workload Attestor.

hiyosi:
If this request is approved, I'll send a PR.

evan2645:
Hey @hiyosi, thank you for opening this and sorry for the delay in my response.

This issue has drawn my attention to the `k8s:node-name` selector you referenced... It's a pretty weak assertion coming from the agent - instead, we probably want to assert this through node attestation, especially if the purpose is to limit the exposure of a workload entry. Which brings me to `k8s:host-ip`

From what I can tell in this issue, the concern is primarily around mitigating the danger of having an admin SVID since these SVIDs are pretty powerful. To me, it seems like a big part of that consideration is limiting the agents that can request and issue an admin SVID, and it also looks like that is what we're trying to accomplish here... however, we can only control the scope of the agent authorization through the use of parent ID, and thus, node selectors.

[...]

PR #1979, opened 2020-11-26
Add k8s_psat:agent_node_ip selector

hiyosi:

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

NodeAttestor k8s_psat plugin

**Description of change**

Add `k8s_past:agent_node_ip` selector to k8s_psat plugin.

We often think that we want important workloads to work separate node from common workloads.
In such case, the SPIRE Server should only accepts the signing request  from specific Nodes.
Although we can use `k8s_psat_agent_node_name` for distinguish the Node's SPIFFE ID, in some case use Taint/Toleration or Affinity/AntiAffinity  instead of Pod's `.spec.nodeName` for node isolation.
Further, some environments, IP addresses is more fixed than Node names.

**Which issue this PR fixes**

related: https://github.com/spiffe/spire/issues/1935


evan2645 (approved):
Thank you for this @hiyosi

PR #1987, opened 2020-12-02
Create docker images from scratch

### LINKED ISSUE #1972: Production Hardened Docker Images
* **Version**: 0.11.2
* **Subsystem**: all

A production security product, should have as small of a threat surface as possible.  The gold standard for a secure docker image is a compiled executable running in a scratch image (ideally as non-root).  I believe that the default docker image should be a FROM scratch image.  If an alpine images is required for dev work, then an alternative should be available, but the default images should be as hardened as possible.

jkevlin:
Signed-off-by: jkevlin <jkevlin@ibm.com>

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

The Makefile was changed to allow for additional images
**Description of change**

Using scratch image when creating container images, require executables that are statically linked.
**Which issue this PR fixes**

fixes #1972


Makefile
```diff
@@ -268,20 +268,41 @@ go_ldflags := '${go_ldflags}'
 
 .PHONY: build
 
-build: tidy bin/spire-server bin/spire-agent bin/k8s-workload-registrar bin/oidc-discovery-provider
+build: tidy bin/spire-server bin/spire-agent bin/k8s-workload-registrar bin/oidc-discovery-provider bin/spire-server-static bin/spire-agent-static bin/k8s-workload-registrar-static bin/oidc-discovery-provider-static
```
rturner3, line 271: Would you mind excluding the static binaries from the `build` target? I don't see a use for them today outside of the scratch image.

I think a separate `build-static` target would be appropriate for these binaries.

Makefile
```diff
@@ -268,20 +268,41 @@ go_ldflags := '${go_ldflags}'
...
 $1: | go-check bin/
 	@echo Building $1...
-	$(E)$(go) build $$(go_flags) -ldflags $$(go_ldflags) -o $1 $2
+	$(E)$(go_path) go build $$(go_flags) -ldflags $$(go_ldflags) -o $1 $2
+
+endef
+define binary_rule_static
+.PHONY: $1
+$1: | go-check bin/
+	@echo Building $1... CGO_ENABLED=0 go build $$(go_flags) -ldflags $$(go_ldflags)  -o $1 $2
```
rturner3, line 283: Can you remove the build command after `Building $1...` here to be consistent with other build targets' output to stdout?

Makefile
```diff
@@ -268,20 +268,41 @@ go_ldflags := '${go_ldflags}'
...
+	$(E)$(go_path) CGO_ENABLED=0 go build $$(go_flags) -ldflags $$(go_ldflags) -o $1 $2
+
+endef
+# https://7thzero.com/blog/golang-w-sqlite3-docker-scratch-image
+define binary_rule_external_static
+.PHONY: $1
+$1: | go-check bin/
+	@echo Building $1... CGO_ENABLED=1 go build $$(go_flags) -ldflags '-linkmode external -extldflags "-static"' $$(go_ldflags) -o $1 $2
```
rturner3, line 291: Same here, please remove the build command from the `@echo`.

rturner3 (commented):
Thanks for the PR, @jkevlin!

We have some existing Makefile targets for the Docker images based on Alpine Linux: `images`, `*-image`. Since you are introducing new Docker image definitions here, I think it makes sense to add another set of targets for the scratch images as well. I think an `all`-style target like `static-images` and individual image targets such as `spire-server-static-image` would make sense.

Would you mind adding those to the PR?

test/integration/suites/k8s-scratch/conf/agent/spire-agent.yaml
```diff
@@ -0,0 +1,159 @@
...
+        }
+      }
+
+      WorkloadAttestor "k8s" {
+        plugin_data {
+          # Defaults to the secure kubelet port by default.
+          # Minikube does not have a cert in the cluster CA bundle that
+          # can authenticate the kubelet cert, so skip validation.
+          skip_kubelet_verification = true
+          token_path = "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
rturner3, line 77: Is this config parameter needed because of a difference between the Alpine and scratch containers (i.e. there is a symlink present in the Alpine container that is not present in the scratch one)?

Dockerfile.scratch
```diff
@@ -0,0 +1,48 @@
...
+RUN apk add --update gcc musl-dev
+ADD go.mod /spire/go.mod
+ADD proto/spire/go.mod /spire/proto/spire/go.mod
+RUN cd /spire && go mod download
+ADD . /spire
+WORKDIR /spire
+RUN make build-static
+
+# UPX is a free, portable, extendable, high-performance executable packer for several executable formats.
+RUN upx ./bin/spire-server-static \
```
rturner3, line 14: I would prefer we not introduce this tool in this PR. I would rather see either both the scratch and Alpine images using binaries compressed by this tool or neither of them. IMO introducing this `upx` tool should be a different PR.

What was the intent of using `upx` here? To reduce the size of the Docker images? I did notice in my local testing that the `spire-server-scratch` image was ~16.7 MB using `upx` and ~62.1 MB without, so I do think it's worth considering using this.

rturner3 (commented):
This is shaping up very nicely!

I think you will also need to add the `make scratch-images` command to the CI job configuration in the [publish images](https://github.com/spiffe/spire/blob/master/.travis.yml#L85) and [nightly integration tests](https://github.com/spiffe/spire/blob/master/.travis.yml#L97) stages since you've introduced an integration test that depends on the scratch images. @evan2645 did I miss any other CI configs related to CircleCI that might need to be updated?

Would you mind also adding some help docs to the Makefile for all the new targets you've introduced here in the `help` target?

test/integration/suites/k8s-scratch/conf/agent/spire-agent.yaml
jkevlin: Yes, alpine doesn't have the same symlink and have not figured if I can add a symlink to a scratch image.

Dockerfile.scratch
jkevlin: Ok, the intent was to reduce the size a component (mostly spire-agent) that potentially is going to be in use in 100's of containers.  It probably has no effect on memory, but makes a difference on startup time as the bytes have to move over the network or at least that was my thinking.

rturner3:
I just noticed that the `oidc-discovery-provider` has an image defined in `Dockerfile.scratch`, but no corresponding target in the Makefile. Could you add that target as well?

Just a couple other minor housekeeping things:

- I think your branch has a merge conflict with master.
- All but your first commit are missing the required Git sign-off message (added with the `--signoff` flag in the `git commit` command).

Probably the easiest way to fix this will be to squash all the commits once the changes are approved and rebase on top of master, then force push to your branch.

Makefile
```diff
@@ -604,7 +604,16 @@ $(protoc_gen_go_bin): | go-check
...
+	$(E)$(go_path) go build -o $(protoc_gen_go_bin) google.golang.org/protobuf/cmd/protoc-gen-go
+
+install-protoc-gen-go-grpc: $(protoc_gen_go_grpc_bin)
+
+$(protoc_gen_go_grpc_bin): | go-check
+	@echo "Installing protoc-gen-go-grpc $(protoc_gen_go_grpc_version)..."
+	$(E)rm -rf $(protoc_gen_go_grpc_base_dir)
+	$(E)mkdir -p $(protoc_gen_go_grpc_dir)
+	$(E)echo "module tools" > $(protoc_gen_go_grpc_dir)/go.mod
+	$(E)cd $(protoc_gen_go_grpc_dir) && GOBIN=$(protoc_gen_go_grpc_dir) $(go) get google.golang.org/grpc/cmd/protoc-gen-go-grpc@$(protoc_gen_go_grpc_version)
```
rturner3: I think this line didn't get merged properly. It looks like `$(go)` should be replaced with `$(E)$(go_path) go`.

Makefile
```diff
@@ -547,12 +613,12 @@ $(protoc_gen_go_grpc_bin): | go-check
 	$(E)rm -rf $(protoc_gen_go_grpc_base_dir)
 	$(E)mkdir -p $(protoc_gen_go_grpc_dir)
 	$(E)echo "module tools" > $(protoc_gen_go_grpc_dir)/go.mod
-	$(E)cd $(protoc_gen_go_grpc_dir) && GOBIN=$(protoc_gen_go_grpc_dir) $(go) get google.golang.org/grpc/cmd/protoc-gen-go-grpc@$(protoc_gen_go_grpc_version)
+	$(E)cd $(protoc_gen_go_grpc_dir) && GOBIN=$(protoc_gen_go_grpc_dir) $(go_path) go  get google.golang.org/grpc/cmd/protoc-gen-go-grpc@$(protoc_gen_go_grpc_version)
```
rturner3, line 616: nit: additional space between `go` and `get` can be removed

rturner3:
LGTM. If you could address the nit, squash the commits in your branch and include the signoff tag, and force push, I think we're good to go.

PR #2011, opened 2020-12-09
[#1930] Add security considerations section to token-based node attestors

### LINKED ISSUE #1930: Document security considerations for node attestors that leverage node-scoped tokens
Cloud-based node attestors that leverage node-scoped tokens all require TOFU semantics, and most of the tokens in question are relatively long-lived. These attestors should include a Security Considerations section in their documentation that explains the nature of this behavior, and the risks associated with their use.

At a minimum, the following node attestor documentations should be updated:
* aws_iid
* gcp_iid
* azure_msi

evan2645:
Node attestors which rely on instance identity tokens provided by a
cloud provider use TOFU semantics due to the fact that these tokens are
generally available to all code running on the node. This commit adds a
Security Considerations section to the reference documentation of these
attestors, describing the nature of this problem and the need for TOFU.

No other content was added or removed as part of this change, but
headings have been added to documents which previously lacked them

Fixes #1930 

### DIFF
--- a/doc/plugin_server_nodeattestor_aws_iid.md
+++ b/doc/plugin_server_nodeattestor_aws_iid.md
@@ -1,5 +1,4 @@
 # Server plugin: NodeAttestor "aws_iid"
-
 *Must be used in conjunction with the agent-side aws_iid plugin*
 
 The `aws_iid` plugin automatically attests instances using the AWS Instance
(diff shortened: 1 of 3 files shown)


azdagron (approved):
\o/ thanks, @evan2645.

PR #2012, opened 2020-12-10
Fix handling the Vault PKI certificate chain

hiyosi:
Signed-off-by: Tomoya Usami <tousami@zlab.co.jp>

**Pull Request check list**

- [x] Commit conforms to CONTRIBUTING.md?
- [x] Proper tests/regressions included?
- [ ] Documentation updated?

**Affected functionality**

UpstreamAuthority `spire` plugin

**Description of change**

This fixes that requiring a Vault PKI certificate when connecting to SPIRE Server if the VaultPKI secret engine is configured as an intermediate CA.
Before the fix, the X509CA certificate was not chained and all Upstream CAs, including the root, were treated as bundles.

**Which issue this PR fixes**

PR #2017, opened 2020-12-11
Fix bug that index out of range when no ca_chain is returned from the vault

hiyosi:

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

I'm very sorry.
This bug was born due to #2012.
Even though the comment(https://github.com/spiffe/spire/blob/master/pkg/server/plugin/upstreamauthority/vault/vault_client.go#L379) is mentioned that the chain will be empty if the Vault configured as the Root CA, I failed to take it.

**Which issue this PR fixes**

#2012


amartinezfayo (approved):
Thank you for the fix @hiyosi.
The assignment of `upstreamRootPEM` was the portion of the change that I spent most of the time in the original review, precisely to consider if there could be an index out of range there and I wrongly concluded that a check was not necessary. Thanks for the fix!

PR #2032, opened 2021-01-04
Report uptime as metrics for spire server and agent with config

### LINKED ISSUE #1231: Provide an uptime metric for SPIRE Server and SPIRE Agent
It would be useful to know how long instances of SPIRE Server and Agent have been alive in a given time period for the purposes of generating alerts like "SPIRE Agent has been down for 30 seconds of the last 1 minute period". Emitting an uptime metric makes it easier to reason about how well SPIRE is meeting a pre-determined SLA.

hixichen:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

- Called during bootstrap.
- Add uptimeReportInterval into agent and server config.

**Description of change**

Add a uptime metrics for spire-agent and spire-server. 
The goal here is to t indicate both Spire agent and Spire server are healthy. 
And, uptime metric will help to determine if the process is periodically crashing. 

This is different compared to the existing uptime in the debug API.

Test locally:
```
for run in {1..50}; do time make test; done
```

**Which issue this PR fixes**

fixes #1231


amartinezfayo:
Thank you @hixichen for this contribution.
Before going into the details of the code changes, I would like to make a note that the introduction of an uptime metric has been previously discussed in #1208, where the version metric was added and adding an uptime metric was considered to be useful for the future (captured in #1231).
I just want to make sure that the motivation for this change is driven by the need of knowing the uptime specifically, and that the version metric that is emitted when the server/agent starts is not enough for your needs.
Knowing a little more about the use case and the motivation for this in your context would be helpful.
Thanks again!

pkg/common/uptime/uptime_test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/spiffe/spire/test/fakes/fakemetrics"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUptime(t *testing.T) {
```
amoore877, line 12: nit, more accurate name would be `TestReportMetrics`

pkg/common/uptime/uptime_test.go
```diff
@@ -0,0 +1,39 @@
...
+			},
+		},
+	} {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(ctx)
+			defer cancel()
+			metrics := fakemetrics.New()
+			ReportMetrics(ctx, tt.reportInterval, metrics)
+			time.Sleep(tt.reportInterval)
```
amoore877, line 35: rather than sleep in a test, may be better to make use of use of a mock clock to be inserted into the reporting cadence

hixichen: updated with clock mock.

hixichen:
> Thank you @hixichen for this contribution.
> Before going into the details of the code changes, I would like to make a note that the introduction of an uptime metric has been previously discussed in #1208, where the version metric was added and adding an uptime metric was considered to be useful for the future (captured in #1231).
> I just want to make sure that the motivation for this change is driven by the need of knowing the uptime specifically, and that the version metric that is emitted when the server/agent starts is not enough for your needs.
> Knowing a little more about the use case and the motivation for this in your context would be helpful.
> Thanks again!

Hi,Agustín:
Great thanks for your reply. 
We are trying to use uptime to track long-term uptime since the version metric can be purged after some period of time.

hixichen:
@amartinezfayo  let me know if you have other comments. 

thanks.

pkg/agent/config.go
```diff
@@ -66,6 +66,9 @@ type Config struct {
 
 	// Telemetry provides the configuration for metrics exporting
 	Telemetry telemetry.FileConfig
+
+	// UptimeReportInterval the interval that spire agent reports its uptime.
+	UptimeReportInterval time.Duration
```
amartinezfayo, line 71: Do you think that it will be commonly needed to override the report interval or this is added just in case is needed? If there is not a specific use case at this point, I would be rather inclined to not have it. New settings means new possible misconfigurations, documentation changes and possible deprecation handling in the future. We can always add it in the future if we find a reason, but it's more difficult to remove it.

If we do think it's needed, then other changes in this PR are needed:
1) This setting needs to be parsed from the agent's config file in https://github.com/spiffe/spire/blob/master/cmd/spire-agent/cli/run/run.go, otherwise it's an unknown setting for the agent.
2) https://github.com/spiffe/spire/blob/master/doc/spire_agent.md needs to be updated to document this setting.
3) https://github.com/spiffe/spire/blob/master/conf/agent/agent_full.conf needs to be updated with this setting.

pkg/server/config.go
```diff
@@ -77,6 +77,9 @@ type Config struct {
 
 	// RateLimit holds rate limiting configurations.
 	RateLimit endpoints.RateLimitConfig
+
+	// UptimeReportInterval the interval that spire agent reports its uptime.
+	UptimeReportInterval time.Duration
```
amartinezfayo, line 82: Same comment as the agent side but for the server setting.

amartinezfayo (commented):
Thank you @hixichen for providing the details. I think that it makes sense to add this metric to cover that case.
My main concern is around the new agent/server config settings added. Please see my comments about that.

pkg/agent/config.go
hixichen: my personal opinion would be this should be `just in case` needed. 

@amoore877 any comment on this?

amoore877: Given that spire agents are generally widely deployed, at scale this amounts to a lot of metrics data. In the worst case, it would be nice to have this be configurable in case of needs to fine-tune (or turn off) to not overload an observability system in case the sane default is too aggressive.
Other cases may be different operators have different needs and standards for granularity of the metric.

For misconfiguration risk, we could add a sane minimum (10s, 20s?).

Would it make sense to instead make this an "experimental" configurable? There have been such experimental (or hidden) config in the past for features.

amoore877: It may also make sense for server and agent to be configured for different intervals.

amartinezfayo: We discussed this in today's maintainer's call and the consensus was that adding a new configurable is not needed. The introduction of this metric doesn't seem to really affect the total amount of metrics emitted and adding config settings for this looks very targeted for this case considering all the metrics that we already have.
We also discussed about the reporting interval and we feel that a 10 seconds interval seems a reasonable value to start with.

hixichen: @amartinezfayo  thanks for your detailed info. I updated the code and removed the configurable value.

amartinezfayo (commented):
I left a few nitpicks. We should be ready to go after that.
Thank you @hixichen for your patience!

pkg/common/uptime/uptime.go
```diff
@@ -1,9 +1,38 @@
...
+	"context"
+	"time"
 
-var start = time.Now()
+	"github.com/andres-erbsen/clock"
+	"github.com/spiffe/spire/pkg/common/telemetry"
+)
+
+// by default, report every 10 seconds.
+const _defaultReportInterval = time.Second * 10
```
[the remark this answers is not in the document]
hixichen: > I left a few nitpicks. We should be ready to go after that.
> Thank you @hixichen for your patience!

should do a better job from the first commit. Thanks for your help. 
Updated.

amartinezfayo (approved):
Thanks @hixichen!

hixichen:
Yay!
thanks. @amartinezfayo @amoore877

Issue #1751, opened 2020-07-25
Add integration test coverage for sql datastore plugin using read-only replica


rturner3:
There is no integration testing coverage of the `sql` datastore plugin's integration with read-only replicas (i.e. when `ro_connection_string` points to a secondary node in the cluster).

I started working on this by extending the existing `datastore-mysql` test suite to build an environment with a MySQL 5.7 cluster using group replication in a branch in my fork: https://github.com/rturner3/spire/tree/sql-replication-integration-tests

There are still some issues to sort out:

- Many of the the tests fail, most likely because they don't tolerate any replication delay. It may be useful to pass some replication delay configuration as a linker flag to the tests when they run against a cluster using replication.
- It would be great to have similar environments for other supported MySQL versions.
- Similarly, it would be nice to have test coverage of SPIRE SQL integration with PostgreSQL clusters using replication.

amartinezfayo:
Addressed in #2029 and #2035.

PR #2035, opened 2021-01-05
Add IT for Postgres replication 

MarcosDY:
Add integration test for Postgres using standby configurations, and using read only replica for queries

**Which issue this PR fixes**
Fixes Postgres side of #1751


amartinezfayo (approved):
Thanks @MarcosDY!

PR #2039, opened 2021-01-08
Implement an Upstream Authority backed by a GCP Certificate Authority

Jonpez2:
Signed-off-by: Jonathan Perry <jonathan@Perryss-iMac.local>

**Pull Request check list**

- [ ] Documentation updated?

I would appreciate guidance on how to write tests for this, as it kind of requires a functional GCP CA backend.

**Affected functionality**
Adds a new plugin for a GCP Certificate Authority-backed Upstream Authority

**Description of change**

**Which issue this PR fixes**


Jonpez2:
Another comment - on the slack channel, I got feedback from Andrew Harding that we need support for detecting when the root authority changes.  I was rooing around in the aws pca code to see what that might mean, and I'm afraid I don't get it.  Elucidation welcome please!

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+func getClient(ctx context.Context) (*pcaapi.CertificateAuthorityClient, error) {
+	creds, err := google.FindDefaultCredentials(ctx)
+
+	if err != nil {
+		return nil, err
+	}
+
+	pcaClient, err := pcaapi.NewCertificateAuthorityClient(ctx, option.WithCredentials(creds))
```
MarcosDY, line 112: in order to test it, do you tryed using hook + interface approach?

you can do something like:
```
type CertificateAuthorityClient interface {
       CreateCertificate(ctx context.Context, req *privatecapb.CreateCertificateRequest, opts ...gax.CallOption) (*privatecapb.Certificate, error)
}
--- 
type Plugin struct {
        c  *Config
        log hclog.Logger
 
        hook struct {
               getClient func(ctx context.Context) (CertificateAuthorityClient, error)
       }
       
func New() *Plugin {
       p := &Plugin{}
       p.hook.getClient = getClient

       return p
 }
 ---
 -func getClient(ctx context.Context) (*pcaapi.CertificateAuthorityClient, error) {
+func getClient(ctx context.Context) (CertificateAuthorityClient, error) {
---
func (p *Plugin) mintX509CA_(...
-       pcaClient, err := getClient(ctx)
+       pcaClient, err := p.hook.getClient(ctx)
```
then your test replaces hook with your own getClient and you have you implementation for `CertificateAuthorityClient`
makes it sense?

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+const (
+	// The name of the plugin
+	pluginName = "gcp_ca"
+	// The header and footer type for a PEM-encoded CSR
+	csrRequestType = "CERTIFICATE REQUEST"
+)
+
+// KeyUsage represents the set of actions that are valid for a given key. It's
+// a bitmap of the KeyUsage* constants.
+type KeyUsage int
```
azdagron, line 35: This type (and associated consts) seem unused....

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+	c  *Config
+
+	log hclog.Logger
+}
+
+// These are compile time assertions that the plugin matches the interfaces the
+// catalog requires to provide the plugin with a logger and host service
+// broker as well as the UpstreamAuthority itself.
+var _ catalog.NeedsLogger = (*Plugin)(nil)
+var _ catalog.NeedsLogger = (*Plugin)(nil)
```
azdagron, line 89: This line is duped. Seems like this came in from the "example" plugin code. If you are willing, would you mind removing the duplicate from there as well while you are at it?

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+
+func New() *Plugin {
+	return &Plugin{}
+}
+
+// SetLogger will be called by the catalog system to provide the plugin with
+// a logger when it is loaded. The logger is wired up to the SPIRE core
+// logger
+func (p *Plugin) SetLogger(log hclog.Logger) {
+	// TODO: store the logger for later use. If the plugin does not need to
```
azdagron, line 100: This comment from the "example" template can be removed, since you are certainly using the logger.

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+	// log, this method can be removed.
+	p.log = log
+}
+
+func getClient(ctx context.Context) (*pcaapi.CertificateAuthorityClient, error) {
```
azdagron, line 105: nitpick: We typically order the code such that the exported types and functions are at the top, and unexported are at the bottom.

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+	creds, err := google.FindDefaultCredentials(ctx)
+
```
azdagron, line 107: nitpick: remove this newline so that the error producing code and the error check are grouped together

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	pcaClient, err := pcaapi.NewCertificateAuthorityClient(ctx, option.WithCredentials(creds))
+
```
azdagron, line 113: nitpick: remove this newline so that the error producing code and the error check are grouped together

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return cert, nil
+}
+
+func (p *Plugin) mintX509CA_(log hclog.Logger, ctx context.Context, csr []byte, preferredTtl int32) (*upstreamauthority.MintX509CAResponse, error) {
```
azdagron, line 135: nitpick: we don't suffix "private" (i.e. unexported) struct methods with `_`

Also, `log` doesn't need to be passed. it is already available on `p`...

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+	csrParsed, err := x509.ParseCertificateRequest(csr)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse CSR: %v", err)
+	}
+
+	validity := time.Second * time.Duration(preferredTtl)
+	parentPath := "projects/" + p.c.Project + "/locations/" + p.c.Location + "/certificateAuthorities/" + p.c.CertName
```
azdagron, line 142: nitpick: the following is a little easier to read (at least for me :) )
```suggestion
	parentPath := path.Join("projects", p.c.Project. "locations", p.c.Location, "certificateAuthorities", p.c.CertName)
```

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+
+	log.Info("Minting X509 using certificate at " + parentPath + "; ttl " + validity.String())
```
azdagron, line 144: nitpick: I'd suggest making this more structured
```suggestion
	log.Info("Minting X509 intermediate CA ", "ca-certificate", parentPath, "ttl ", validity)
```

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+							},
+						),
+					},
+					SubjectConfig: &privatecapb.CertificateConfig_SubjectConfig{
+						Subject:    &subject,
+						CommonName: csrParsed.Subject.CommonName,
+					},
+					ReusableConfig: &privatecapb.ReusableConfigWrapper{
+						ConfigValues: &privatecapb.ReusableConfigWrapper_ReusableConfig{
+							ReusableConfig: fmt.Sprintf("projects/%s/locations/%s/reusableConfigs/%s", "privateca-data", p.c.Location, "subordinate-mtls-pathlen-0"),
```
azdagron, line 179: nitpick: there is a little less cognitive overhead understanding the final result here if you place fixed values directly into the format string
```suggestion
							ReusableConfig: fmt.Sprintf("projects/privateca-data/locations/%s/reusableConfigs/subordinate-mtls-pathlen-0", p.c.Location),
```

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	cresp, err := pcaClient.CreateCertificate(ctx, &createRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	cert, err := pemToX509(cresp.GetPemCertificate())
```
azdagron, line 197: There is a chance for code reuse here. We have a function for this already: https://github.com/spiffe/spire/blob/d618fa69a10a55b4de027a1c78102026ce6fb4b9/pkg/common/pemutil/certs.go#L12

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	certChain := make([]*x509.Certificate, len(cresp.PemCertificateChain))
+	for i, c := range cresp.PemCertificateChain {
+		certChain[i], err = pemToX509(c)
```
azdagron, line 204:
```suggestion
		certChain[i], err = pemutil.ParseCertificate([]byte(c))
```

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// The last certificate returned from the chain is the root.
```
azdagron, line 210: We should probably do some light validation that certChain has the expected (mininum) number of elements in it?

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+//
+// Implementation note:
+// The stream should be kept open in the face of transient errors
+// encountered while tracking changes to the upstream X.509 roots as SPIRE
+// core will not reopen a closed stream until the next X.509 CA rotation.
+func (p *Plugin) MintX509CA(request *upstreamauthority.MintX509CARequest, stream upstreamauthority.UpstreamAuthority_MintX509CAServer) error {
+	p.log.Info("Minting new X509")
+	ctx := stream.Context()
+	minted, err := p.mintX509CA_(p.log, ctx, request.Csr, request.PreferredTtl)
+
```
azdagron, line 242: nitpick: remove this newline so that the error producing code and the error check are grouped together

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	// Swap out the current configuration with the new configuration
+	p.setConfig(config)
+
+	return &plugin.ConfigureResponse{}, nil
+}
+
+func (p *Plugin) GetPluginInfo(ctx context.Context, req *plugin.GetPluginInfoRequest) (*plugin.GetPluginInfoResponse, error) {
+	// TODO: optionally fill out the plugin info. This is currently unused
```
azdagron, line 264: nitpick: remove this TODO from the example template

go.mod
```diff
@@ -62,13 +63,14 @@ require (
...
+	golang.org/x/sys v0.0.0-20201201145000-ef89a241ccb3
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
-	golang.org/x/tools v0.0.0-20200331025713-a30bf2db82d4
-	google.golang.org/api v0.29.0
-	google.golang.org/genproto v0.0.0-20200806141610-86f49bd18e98
-	google.golang.org/grpc v1.33.2
+	golang.org/x/tools v0.0.0-20201208233053-a543418bbed2
+	google.golang.org/api v0.36.0
+	google.golang.org/genproto v0.0.0-20201210142538-e3217bee35cc
+	google.golang.org/grpc v1.34.0
```
azdagron, line 73: Were these dependency bumps needed? I'm not opposed to them, but would like to understand why they were introduced.

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,284 @@
...
+
+	// The last certificate returned from the chain is the root.
+	upstreamRoot := certChain[len(certChain)-1]
+	bundle := x509util.DedupeCertificates([]*x509.Certificate{upstreamRoot})
+
+	derBundle := x509util.RawCertsFromCertificates(bundle)
+
+	// All else comprises the chain (including the issued certificate)
+	fullChain := []*x509.Certificate{cert}
+	fullChain = append(fullChain, certChain...)
```
azdagron, line 218: The `X509CaChain` should only contain the newly minted intermediate CA and then other intermediates necessary to chain it back to a root present in the `UpstreamX509Roots` field. My reading of this code is that the upstream root is present in `certChain` and would therefore be appended to `X509CaChain`, which is not correct. `X509CaChain` should not contain the root.

azdagron (commented):
Thanks for putting this together, @Jonpez2! Many of my comments were stylistic and nitpicky but there are a few important ones in there.

I also noticed that there were no unit tests accompanying this. Is that something you are comfortable putting together?

Regarding the "send root CA changes back over the stream" suggestion, that is necessary to propagate root CA changes in advance of them being used to sign intermediates. I haven't actually looked too closely at the GCP CA API here to know how easy this would be. If root rotation is not expected, then the stream can be closed after sending back the minted intermediate as it is now.

Jonpez2:
> Thanks for putting this together, @Jonpez2! Many of my comments were stylistic and nitpicky but there are a few important ones in there.

Yup, thank you for your input.

> 
> I also noticed that there were no unit tests accompanying this. Is that something you are comfortable putting together?

Very much so.  I was hoping for input, which you gave me, so I will get on it!

> 
> Regarding the "send root CA changes back over the stream" suggestion, that is necessary to propagate root CA changes in advance of them being used to sign intermediates. I haven't actually looked too closely at the GCP CA API here to know how easy this would be. If root rotation is not expected, then the stream can be closed after sending back the minted intermediate as it is now.

I'll get data out of GCP and do what I can.

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
Jonpez2: It does make total sense.  I'm going to see if GCP have an implementation so that we are testing reality.

Jonpez2:
Re updating the root - am I right in thinking that the k8s-workload-registrar needs to know about the root certs referenced here?  Is there already some way of updating those automatically?

azdagron:
I assume your talking about the k8s-workload-registrar getting an updated bundle so that it can continue to talk to the SPIRE server API over TCP+TLS? If so, the k8s-workload-registrar, when operating in this mode (i.e. talking to the TLS listener and not the UDS) uses the Workload API to obtain the admin X509-SVID and X.509 bundles it uses to handshake with the server. As such, it will receive updates to roots over the Workload API.

azdagron:
The more I think about it, the more I don't think this plugin needs to stream updates to roots back on the stream. SPIRE server prepares intermediates well in advance of using them. If there is a new root in GCP that is used to sign the intermediate, that root will make its way into the bundle and be propagated before the intermediate is activated. This is already the case for other upstream plugins, I'm not sure why I advocated you look into it. I'm sorry if for the wasted time and effort.

Jonpez2:
> I assume your talking about the k8s-workload-registrar getting an updated bundle so that it can continue to talk to the SPIRE server API over TCP+TLS? If so, the k8s-workload-registrar, when operating in this mode (i.e. talking to the TLS listener and not the UDS) uses the Workload API to obtain the admin X509-SVID and X.509 bundles it uses to handshake with the server. As such, it will receive updates to roots over the Workload API.

Ah OK, let me test that.  Thanks!

Jonpez2:
> The more I think about it, the more I don't think this plugin needs to stream updates to roots back on the stream. SPIRE server prepares intermediates well in advance of using them. If there is a new root in GCP that is used to sign the intermediate, that root will make its way into the bundle and be propagated before the intermediate is activated. This is already the case for other upstream plugins, I'm not sure why I advocated you look into it. I'm sorry for the wasted time and effort. Hopefully you weren't too far down that road.

Not at all.  Automtion of new certificate roots is always worth thinking about!

So I was wondering about the following: what if we start to specify the CA certs by labels - so we say that the chosen upstream CA cert should be 'whichever cert has this label and has the latest expiration', and then we put any other certs we find into the trust bundle.  Also you should be able to specify some other label to select extra certs to include in the trust bundle.  Then we would want a periodic sync up (once on startup, and say once per hour or so thereafter).

That would mean that certificate rotation would simply be a matter of adding a new cert with a longer expiration under the same label, and letting the prior one lapse.  Trivially automatable.

Does that seem sane?

Thanks!

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
Jonpez2: Fixed thanks

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
Jonpez2: Fixed thanks

go.mod
Jonpez2: I think they happened automagically when I typed something go-ish in.  Shall I revert them?

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
Jonpez2: Fixed thank you

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,332 @@
...
+
+	allCertRoots, err := loadCertificateAuthorities(ctx, pcaClient, p.c.RootSpec)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to load root CAs: %v.", err)
+	}
+	if len(allCertRoots) == 0 {
+		return nil, fmt.Errorf("No certificate authorities found with label pair %s:%s.", p.c.RootSpec.LabelKey, p.c.RootSpec.LabelValue)
+	}
+
+	// The certs get returned in issuance order, so just grab the last one and use that.
```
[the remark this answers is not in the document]
Jonpez2: I was thinking something more like the following:
- We always have 2 roots stamped with the appropriate label
- We specify some period p
- Every p, some process creates a new root as a duplicate of the latest root in the pair
- The process then disables the oldest root in the pair

Spire has a default TTL of its intermediate certs, and auto-rotates them every TTL/2.  That autorotation will requery CAS, find the latest of the two roots, and issue a cert chaining to that root.  The other root goes in the trust bundle.

No restarts needed!

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,332 @@
...
+	chosenCA := allCertRoots[len(allCertRoots)-1]
+	trustBundle := allCertRoots[:len(allCertRoots)-1]
+	if p.c.TrustBundleSpec.LabelKey != "" {
```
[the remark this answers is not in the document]
Jonpez2: The TrustBundleSpec actually just allows you to tell your SPIRE-managed workloads about other certs they should be able to reason about as roots.  The resultant trust bundle would allow for both the cert rotation and the 'other cert' trust.  Does that make sense?

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
```diff
@@ -0,0 +1,332 @@
...
+	p.log.Info("Minting X509 intermediate CA ", "ca-certificate", parentPath, "ttl ", validity)
+
+	subject := privatecapb.Subject{}
+	extractFirst := func(strings []string, into *string) {
+		if len(strings) > 0 {
+			*into = strings[0]
+		}
+	}
+
+	extractFirst(csrParsed.Subject.Organization, &subject.Organization)
```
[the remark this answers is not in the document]
Jonpez2: FWIW, in this code, the only possible source of the CSR is SPIRE itself, which we kinda gotta trust...

azdagron:
Hey @Jonpez2, just wanted to leave a quick note to apologize for the review latency. We're a little heads down preparing for the 1.0 release, but should be ready to jump on this and get it through in the next week or so. Thank you for your patience.

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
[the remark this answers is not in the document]
haydentherapper: If we're duplicating the latest root, is this only to handle the case of an expiring root? What about a compromised key, when immediate rotation is necessary?

During the autorotation, the latest of the two roots has not yet been distributed in a trust bundle to all agents. Should we instead be grabbing the first/oldest of the two roots, so that the latest root gets distributed to all agents before use? 

* Process `p` creates a new root
* `p` disables the oldest root
* On auto rotation, SPIRE starts issuing off the now-oldest root and includes the new root in the bundle.

This also handles the case of compromise - If the oldest is comprised, then the new root has already been distributed. If the new root is compromised, that's also fine, since we're issuing off the oldest root still. (If both are comprised, obviously this would require a full restart)

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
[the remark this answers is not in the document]
haydentherapper: Thanks, that makes sense. Are we planning to recommend using the `TrustBundleSpec` for rotation, or should users just use the rotation flow, using labeled CAs, that we've described above?

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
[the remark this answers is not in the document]
haydentherapper: Using the oldest also gives you the benefit of being able to test out the latest root, without it being in use yet.

Jonpez2:
> Hey @Jonpez2, just wanted to leave a quick note to apologize for the review latency. We're a little heads down preparing for the 1.0 release, but should be ready to jump on this and get it through in the next week or so. Thank you for your patience.

No problemo.  Thanks for the ping.

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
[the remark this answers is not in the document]
Jonpez2: I think just the rotation flow and labels.

pkg/server/plugin/upstreamauthority/gcpca/gcpca.go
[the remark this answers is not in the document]
Jonpez2: That makes sense to me.  Oldest root it is then.  Thank you!

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,341 @@
...
+	p.setConfig(config)
+
+	return &plugin.ConfigureResponse{}, nil
+}
+
+func (p *Plugin) GetPluginInfo(ctx context.Context, req *plugin.GetPluginInfoRequest) (*plugin.GetPluginInfoResponse, error) {
+	return &plugin.GetPluginInfoResponse{}, nil
+}
+
+func (p *Plugin) getConfig() (*Config, error) {
```
APTy, line 136: do we use `getConfig()` anywhere?

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,341 @@
...
+
+	pcaClient, err := pcaapi.NewCertificateAuthorityClient(ctx, option.WithCredentials(creds))
+	if err != nil {
+		return nil, err
+	}
+
+	return pcaClient, nil
+}
+
+func pemToX509(pemStr string) (*x509.Certificate, error) {
```
APTy, line 167: is this used anywhere either? looks like we may be using `pemutil` now

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	return cert, nil
+}
+
+func loadCertificateAuthorities(ctx context.Context, client *pcaapi.CertificateAuthorityClient, spec CertificateAuthoritySpec) ([]*privatecapb.CertificateAuthority, error) {
+	allCerts := make([]*privatecapb.CertificateAuthority, 0)
+	certIt := client.ListCertificateAuthorities(ctx, &privatecapb.ListCertificateAuthoritiesRequest{
+		Parent: spec.caParentPath(),
+		Filter: "labels." + spec.LabelKey + ":" + spec.LabelValue,
```
APTy, line 185: nit: a format string may be more readable:
```go
fmt.Sprintf("labels.%s:%s", spec.LabelKey, spec.LabelValue)
```

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,341 @@
...
+							},
+						),
+					},
+					SubjectConfig: &privatecapb.CertificateConfig_SubjectConfig{
+						Subject:    &subject,
+						CommonName: csrParsed.Subject.CommonName,
+					},
+					ReusableConfig: &privatecapb.ReusableConfigWrapper{
+						ConfigValues: &privatecapb.ReusableConfigWrapper_ReusableConfig{
+							ReusableConfig: fmt.Sprintf("projects/privateca-data/locations/%s/reusableConfigs/subordinate-mtls-pathlen-0", p.c.RootSpec.Location),
```
APTy, line 280: is this magic string documented anywhere? maybe we can make it a const and link to some docs on it?

APTy (commented):
Thanks for the PR @Jonpez2!

Sorry for a bit of a delay, but I've left some comments. Do you think you'll have time to add a page to the docs folder with usage/examples of the config specs? Also any unit tests that you have time for 😄

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
azdagron: It should be used instead of reaching into `p.c` directly. I've commented where that is taking place. Ideally this is called at the top of `MintX509CA` and then the returned configuration passed into subsequent helpers that satisfy the RPC.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	validity := time.Second * time.Duration(preferredTtl)
+
+	pcaClient, err := getClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	allCertRoots, err := loadCertificateAuthorities(ctx, pcaClient, p.c.RootSpec)
```
azdagron, line 219: `p.c` should not be used anywhere outside of getConfig and setConfig to properly handle concurrent calls to Configure.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,341 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to load root CAs: %v.", err)
+	}
+	if len(allCertRoots) == 0 {
+		return nil, fmt.Errorf("No certificate authorities found with label pair %s:%s.", p.c.RootSpec.LabelKey, p.c.RootSpec.LabelValue)
```
azdagron, line 224: `p.c` should not be used anywhere outside of getConfig and setConfig to properly handle concurrent calls to Configure.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,341 @@
...
+	// The certs get returned in issuance order, and we want the oldest, so just grab the first one and use that.
+	chosenCA := allCertRoots[0]
+
+	// The rest of them are still trusted
+	trustBundle := make([]*privatecapb.CertificateAuthority, 0)
+	if len(allCertRoots) > 1 {
+		trustBundle = append(trustBundle, allCertRoots[1:]...)
+	}
+
+	for _, spec := range p.c.TrustBundleSpecs {
```
azdagron, line 236: `p.c` should not be used anywhere outside of getConfig and setConfig to properly handle concurrent calls to Configure.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas_test.go
```diff
@@ -0,0 +1,58 @@
+package gcpcas
```
azdagron, line 1: This plugin could certainly use more test coverage. I'm happy to provide guidance on this if wanted.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/hashicorp/go-hclog"
+	"github.com/spiffe/spire/proto/spire/common/plugin"
+	"github.com/spiffe/spire/test/util"
+	"github.com/stretchr/testify/require"
+	// 	"github.com/spiffe/spire/pkg/server/plugin/upstreamauthority"
+)
+
+func TestGCPCAS(t *testing.T) {
+	ctx := context.Background()
+	fmt.Println("Hi")
```
azdagron, line 17: I realize this was probably just left in, but in general we shouldn't be outputting stuff from the `fmt` package in tests and should prefer `t.Logf` instead.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	p := New()
+	p.SetLogger(hclog.Default())
```
azdagron, line 20: For plugin testing we actually load the plugin using the catalog infrastructure, which wires up the logging and in general provides an easier interface to test against (since you can call the plugin like a normal gRPC client, which is particularly important for plugins with streaming interfaces). See `spiretest.LoadPlugin`.

Jonpez2:
Yes. I just need to make a bit of time to address the feedback.

dragcp:
On behalf of Jonpez2 and me, thank you for all your feedback. I have picked it up and incorporated them in #2172 

Please review there

azdagron:
#2172 supersedes this PR. Closing.

PR #2041, opened 2021-01-09
Fix file descriptor leak in Linux Agent UDS peer tracker

rturner3:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Agent UDS peer tracker on Linux

**Description of change**
The `/proc/<pid>/stat` file descriptor is kept open longer than it needs
to be. This can result in SPIRE Agent opening multiple file descriptors
for the same proc stat file over time. The file descriptors will be kept open
until the garbage collector runs.

Fix by just closing the stat file descriptor at the end of the function
where it is opened and used.


rturner3:
Closing in favor of @prashantv 's PR #2043 , we both leapt on the same problem and he came up with a cleaner solution :)

PR #2043, opened 2021-01-09
Fix file descriptor leak in linux peertracker

### LINKED ISSUE #2042: linux peertracker leaks file descriptors
ref https://github.com/spiffe/spire/blob/master/pkg/common/peertracker/tracker_linux.go#L152

statfd is never closed, the file is opened, read, and then the descriptor isn't used again. This leak is automatically cleaned up on GC as a runtime finalizer is set on files to automatically close them when the reference is cleaned up, but ideally the tracker would close the file immediately.

I can send a PR with the change.

prashantv:
- [x] Documentation updated?

**Affected functionality**
Agent Linux Peertracker

**Description of change**

statfd was never closed, so there was a file descriptor leak that
eventually got cleaned up on garbage collection due to a finalizer
on os.File.

Since the file is only opened to read the contents, the current logic
can be simplified by using `ioutil.ReadFile` which opens the file, reads
the contents, and closes the file.

**Which issue this PR fixes**
fixes #2042

### DIFF
--- a/pkg/common/peertracker/tracker_linux.go
+++ b/pkg/common/peertracker/tracker_linux.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


rturner3 (approved):
LGTM, thanks for contributing @prashantv !

azdagron (approved):

Issue #1921, opened 2020-10-16
Feature request: AWS KMS management of signing keys


derekbedwards:
**Feature request: AWS KMS management of signing keys**

**Background:**

My team is in the process of adopting SPIFFE and leveraging SPIRE to support workload identity as part of our platform’s zero trust strategy (we deal extensively with very sensitive information in a highly-regulated field.) We have successfully demonstrated a proof of concept of our infrastructure incorporating these components and are working closely with our information and application security teams to review and approve our patterns as we proceed toward leveraging SPIRE in production applications.

One concern that was raised during reviews related to SPIRE Server holding its own signing keys.

As a policy, we are required to manage signing keys through a mechanism such as AWS KMS or equivalent secure store as a mitigation against potential undiscovered vulnerabilities in containers, Kubernetes and associated infrastructure. The potential threat is that holding the signing keys locally presents an opportunity to spoof workload identities if the node on which SPIRE Server is running is ever compromised.

In order to help facilitate my team’s adoption of SPIRE and further enhance the security model for everyone, I would like to put a request to the community for the development of support for AWS KMS-managed signing keys.

Happy to provide any additional supporting information to facilitate this request.

kunzimariano:
I will try to explain how Initialization and rotation could work on a possible implementation.
  
It's important to remember that when we create a key in KMS the private part of it is never available to us. To do signing operations we need to refer it by its ID or an alias.
  
At a higher lever we will keep an in-memory map of the customer managed keys (CMK). Entries will be indexed by the identifier SPIRE uses, i.e. `x509-CA-A`, `x509-CA-B`, etc. As the value we will use an struct with the `public key`, the `key ID used by KMS`, and the `alias name`.
  
The reasoning behind the alias is to use it as a pointer to the current key, and move it to a new one once  we need to rotate.
  
##  Plugin Initialization
  
  
During the plugin initialization we will build the map based on the keys already present in KMS, if there is any. Keys without alias won't be taken into account.
  
The `KeyID` that SPIRE uses will be part of the alias, and will be extracted from it.
  
![image](https://user-images.githubusercontent.com/1418295/97484423-80384880-1937-11eb-9282-1861736ac8cb.png)
  
##  Key creation
    
The process will mainly be creating a CMK, and assigning an alias to it.
 
![image](https://user-images.githubusercontent.com/1418295/97484503-89291a00-1937-11eb-832b-f4c85888a1e0.png)
  
##  Key rotation
  
  
[...]

PR #2044, opened 2021-01-11
Keymanager plugin calls timeout

### LINKED ISSUE #1963: Enumerate all places where timeouts are necessary when calling the server keymanager plugin
Based on what was discussed on https://github.com/spiffe/spire/issues/1941 we need to list all the places where we could potentially add timeouts, and whether those are the best places to do it or not. 

We should also reach an agreement regarding the upper bound limit.

kunzimariano:
In preparation for a [new keymanager plugin](https://github.com/spiffe/spire/issues/1921) that depends on an external resource, and based on what was presented [here](https://github.com/spiffe/spire/issues/1963). This PR adds timeouts to keymanager plugin calls, as an upper bound of  30 seconds. 

--
Fixes #1963


evan2645 (approved):
Thank you for this @kunzimariano!

PR #2055, opened 2021-01-15
Update upgrading.md for 1.0.0 release

### LINKED ISSUE #2007: Update upgrading.md for 1.0.0 release

amartinezfayo:
This commit updates upgrading.md to be aware of 1.x.x releases and the exception made for 0.12.

Fixes #2007.


doc/upgrading.md
```diff
@@ -2,21 +2,29 @@
...
+* Newest SPIRE Server instance is at 1.0.3
+* Other SPIRE Server instances are supported at 1.0.x and 0.12.x
+
+Example 2:
+* Newest SPIRE Server instance is at 1.2.3
+* Other SPIRE Server instances are supported at 1.2.x and 1.1.x
 
 ### SPIRE Agent Compatibility
-SPIRE Agents must not be newer than the oldest SPIRE Server that they communicate with, and may be up to one minor version older.
+SPIRE Agents must not be newer than the oldest SPIRE Server that they communicate with, and may be up to one minor version older. As an exception, versions 0.12.x are compatible with 1.0.x versions (as long as the oldest SPIRE Server is at 1.0.x).
```
evan2645, line 19: I'm finding this edit a little confusing. I think it would be clearer if we stated "agent versions ... are compatible with server versions ...".

Is the comment in parenthesis at the end strictly true? I'm having trouble imagining the scenario

evan2645 (commented):
Looking great @amartinezfayo! One small question about the agent/server compatibility stuff

doc/upgrading.md
amartinezfayo: Now that I think it better, it doesn't make a lot of sense to add that comment in parenthesis.
I'll rephrase the edit removing it.

PR #2065, opened 2021-01-21
Return an error when having an expired SVID when rotating the agent SVID

### LINKED ISSUE #2061: Agent will never be stopped by SVID expired
* **Version**: 0.12.0
* **Platform**:  any
* **Subsystem**: Agent

If an Agent is unable to rotate its own certificate by the deadline, it will be required to re-attest.

However, I found out that Agent never shutting down in v0.12.0.
If the SVID has expired,  It never receive expected error from server.  Agent will receive a `context deadline exceeded` error and repeats to connect due to the error.

```
 time="2021-01-21T01:11:21Z" level=debug msg="Rotating agent SVID" subsystem_name=manager
 time="2021-01-21T01:11:51Z" level=error msg="Could not rotate agent SVID" error="failed to dial dns:///spire-server:8081: context deadline exceeded: connection closed" subsystem_name=manager
deadline exceeded: connection closed" subsystem_name=manager
```

In v0.11.x, the manager process handled shutting down by referring to the SVID.
https://github.com/spiffe/spire/blob/v0.11.2/pkg/agent/svid/rotator.go#L68

In 0.12.0, the manager process handles shutting down by referring to the error messages.
https://github.com/spiffe/spire/blob/v0.12.0/pkg/agent/svid/rotator.go#L70

amartinezfayo:
Agents should shutdown when having an expired SVID during agent SVID rotation.
The current behavior continues trying to rotate using an expired SVID, which will never succeed without going through re-attestation.
This is a regression introduced in v0.12.0. Previous versions would shutdown properly.

Fixes #2061.


azdagron (approved):
\o/

Issue #2068, opened 2021-01-22
Serverless support: agent changes to identify the entries that indicate externally stored identities


amartinezfayo:
This issue tracks the work needed to be done in SPIRE Agent in order to add a capability in the cache manager to be able to identify the registration entries that indicate that the issued identity must be stored externally.

This issue depends on #2067.

amartinezfayo:
Done in #2088.

PR #2088, opened 2021-01-29
Serverless support: agent changes to identify the entries that indicate externally stored identities

MarcosDY:
Update cache to push any update on entries with "store_svid" to a buffered pipe that is consumed for `svidstore` services and stored svids using an specific plugin

**Which issue this PR fixes**
#2068


proto/spire/agent/svidstore/svidstore.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
```
MarcosDY, line 1: svidstore.proto is part of PR #2076 and must be merged before this PR

proto/spire/common/common.proto
```diff
@@ -83,6 +83,8 @@ message RegistrationEntry {
     repeated string dns_names = 10;
     /** Revision number is bumped every time the entry is updated */
     int64 revision_number = 11;
+    /** Determines if the issued identity is exportable to a store */
+    bool store_svid = 12;
```
MarcosDY, line 87: changes on common protos are done to be used for agent but no changes is applied to server side, changes to server will go in a separated PR

pkg/agent/svid/store/service.go
```diff
@@ -0,0 +1,134 @@
...
+	return &service{
+		c: &c,
+	}
+}
+
+type service struct {
+	c *Config
+
+	hooks struct {
+		// test hook used to verify put is done
```
amartinezfayo, line 40:
```suggestion
		// test hook used to synchronize the completion of the request to put an SVID in an external store
```

pkg/agent/svid/store/service.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if _, err := p.c.SVIDStore.PutX509SVID(ctx, req); err != nil {
+		log.WithError(err).Errorf("Failed to store X509-SVID")
+		return
+	}
+
+	log.Debug("X509-SVID stored successfully")
+}
+
+// triggerStoredHook is only used for unit tests to verify put finished
```
amartinezfayo, line 95:
```suggestion
// triggerStoredHook is only used for unit tests to synchronize the completion of the request to put an SVID in an external store
```

pkg/agent/manager/cache/cache.go
```diff
@@ -114,6 +117,9 @@ type Cache struct {
 
 	// bundles holds the trust bundles, keyed by trust domain id (i.e. "spiffe://domain.test")
 	bundles map[string]*bundleutil.Bundle
+
+	// buffered pipes for 'exportables' SVIDs
```
amartinezfayo, line 121:
```suggestion
	// buffered pipes for SVIDs that must be stored through an SVIDStore plugin
```

pkg/agent/manager/cache/cache.go
```diff
@@ -386,20 +393,77 @@ func (c *Cache) UpdateSVIDs(update *UpdateSVIDs) {
...
 			telemetry.Entry:    record.entry.EntryId,
 			telemetry.SPIFFEID: record.entry.SpiffeId,
 		})
 		log.Debug("SVID updated")
 
+		// TODO: may we allow subscribe to notifications when SVID is exportable?
+		// at the same time may we add a filter to avoid using this entries for attestation?
+		if record.entry.StoreSvid {
+			if err := c.storeSVID(record); err != nil {
+				log.WithError(err).Error("failed to export SVID")
```
amartinezfayo, line 406:
```suggestion
				log.WithError(err).Error("failed to store SVID")
```

pkg/agent/manager/config.go
```diff
@@ -33,6 +34,9 @@ type Config struct {
 
 	// Clk is the clock the manager will use to get time
 	Clk clock.Clock
+
+	// Buffered pipes used to push all 'exportables' SVIDs
```
amartinezfayo, line 38:
```suggestion
	// Buffered pipes used to push all SVIDs that must be stored through an SVIDStore plugin
```

proto/spire/common/common.proto
```diff
@@ -83,6 +83,8 @@ message RegistrationEntry {
     repeated string dns_names = 10;
     /** Revision number is bumped every time the entry is updated */
     int64 revision_number = 11;
+    /** Determines if the issued identity is exportable to a store */
```
amartinezfayo, line 86:
```suggestion
    /** Determines if the issued SVID must be stored through an SVIDStore plugin */
```

proto/spire/types/entry.proto
```diff
@@ -50,6 +50,9 @@ message Entry {
 
     // Revision number is bumped every time the entry is updated
     int64 revision_number = 11;
+
+    // Determines if the issued identity is exportable to a store
```
amartinezfayo, line 54:
```suggestion
    // Determines if the issued SVID must be stored through an SVIDStore plugin
```

pkg/agent/manager/cache/cache.go
```diff
@@ -386,20 +393,77 @@ func (c *Cache) UpdateSVIDs(update *UpdateSVIDs) {
...
+
+func getPluginName(selectors []*common.Selector) (string, error) {
+	if len(selectors) == 0 {
+		return "", errors.New("no selectors found")
+	}
+
+	name := selectors[0].Type
+	for _, s := range selectors {
+		if name != s.Type {
+			return "", errors.New("selector contains multiple types")
```
amartinezfayo, line 461: I think that it would be good to include in the error the expected type and the type found.

amartinezfayo (commented):
Thank you @MarcosDY! I'm excited to see this coming!
The first thing that I wanted to comment on is around the `pipe_size` configurable that this PR introduces. Please see my considerations below.
I also added a bunch of suggestions around the use of "store" vs "export", which I guess may be debatable.

cmd/spire-agent/cli/run/run.go
```diff
@@ -43,6 +43,9 @@ const (
 	defaultLogLevel          = "INFO"
 	defaultDefaultSVIDName   = "default"
 	defaultDefaultBundleName = "ROOTCA"
+
+	// Default SVID store pipe size
+	defaultPipeSize = 1000
```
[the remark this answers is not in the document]
MarcosDY: I moved to 5000.

cmd/spire-agent/cli/run/run.go
```diff
@@ -62,6 +65,7 @@ type agentConfig struct {
 	LogFile           string    `hcl:"log_file"`
 	LogFormat         string    `hcl:"log_format"`
 	LogLevel          string    `hcl:"log_level"`
+	PipeSize          int       `hcl:"pipe_size"`
```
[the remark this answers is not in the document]
MarcosDY: I think adding this config add some power than we'll lose if we choose to have a default, we are reserving memory when we create bigs pipes, and we dont know how much SVIDs will be marked as exportable, that is something that can be different for each case.
and we'll have a pipe 'for each' configured SVIDStore, so we may have double amount of memory without uses.

pkg/agent/manager/pipe/pipe.go
```diff
@@ -0,0 +1,105 @@
+// TODO: ADD unit tests
+package pipe
```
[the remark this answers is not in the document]
MarcosDY: storepipe? storesvidpipe? do you have another ideas about names?

pkg/agent/manager/cache/cache.go
MarcosDY: mmm not sure about that because the guy that is creating it will contain that list, and we are not collecting all types we just fails at first one that is different.
if another are ok about returning that as part of error, are you ok about refactor this code to generate a map with all types and return is as part of error when len(map) > 1 ?

cmd/spire-agent/cli/run/run.go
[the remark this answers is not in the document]
evan2645: I agree with @amartinezfayo ... in the long run, "nitty gritty" configurables like this make SPIRE harder to use for everyone, and provide benefit for only a very small number of people who know how to use it correctly and when. What is pipe size? Don't I always want a big pipe? :)

pkg/agent/agent.go
```diff
@@ -32,6 +37,22 @@ import (
...
 
+type bufferedPipe struct {
+	store catalog.SVIDStores
+	in    pipe.In
+	out   pipe.Out
+}
+
+type bufferedPipes map[string]bufferedPipe
+
+func (p bufferedPipes) pipeIns() map[string]pipe.In {
```
evan2645, line 48: I'm not sure if this stuff is here because we are taking pipe size in as a configurable, but I'd really prefer to keep `agent.go` as clean as possible, and tuck this stuff away somewhere else. If this PR is merged as-is, >17% of the `agent.go` code will be handling SVIDStore stuff... compare that with node attestation which is <6%

pkg/agent/manager/pipe/pipe.go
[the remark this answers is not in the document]
evan2645: What happens if a pipe is full?

I can't help but wonder if there are other ways to solve this problem - there's quite a bit of code present for managing these pipes

cmd/spire-agent/cli/run/run.go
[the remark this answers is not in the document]
MarcosDY: Right now it is 5000 (default)

pkg/agent/agent.go
MarcosDY: that code is to be able to create 1 Store service for each configured store svid, I'll move it out here.

pkg/agent/manager/pipe/pipe.go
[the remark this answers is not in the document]
MarcosDY: thanks to your question, I found a bug I introduced when I was doing some tests... 
Right now pipe will lock (because chan) when limit reached, so no more svids will be pushed until we consume an element using `out`.

As you it is something that worries me... another thing we can do is put more logic on cache to push svids from there... Instead using buffered channels.
I avoid doing that because we'll adding a lot of logic there and I dont think we really want to make cache to use plugins in order to store SVIDs in some plattforms.

PR #2110, opened 2021-02-11
Fix Vault token auth bug

hiyosi:

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

UpstreamAuthority 'vault' plugin

**Description of change**

The Vault token will  be set correctly in the Token Auth Method.

In v0.12.0, we get an error as below even if the token is given via plugin config or environment variable.

```
level=error msg="Fatal run error" error="rpc error: code = Unknown desc = failed to prepare authenticated client: token lookup failed: Error making API request.\n\nURL: GET http://<vault_addr>/v1/auth/token/lookup-self\nCode: 400. Errors:\n\n* missing client token"
```

**Which issue this PR fixes**

No Issues or PRs.
https://spiffe.slack.com/archives/CBNCC2V17/p1613059475073700

### DIFF
--- a/pkg/server/plugin/upstreamauthority/vault/vault_client_test.go
+++ b/pkg/server/plugin/upstreamauthority/vault/vault_client_test.go
@@ -163,29 +163,43 @@ func (vcs *VaultClientSuite) Test_NewAuthenticatedClient_TokenAuth() {
 	vcs.fakeVaultServer.LookupSelfResponseCode = 200
 	for _, c := range []struct {
 		name      string
+		token     string
 		response  []byte
 		reusable  bool
 		namespace string
+		err       string
 	}{
 		{
 			name:     "Token Authentication success / Token never expire",
+			token:    "test-token",
 			response: []byte(testLookupSelfResponseNeverExpire),
 			reusable: true,
 		},
 		{
 			name:     "Token Authentication success / Token is renewable",
+			token:    "test-token",
 			response: []byte(testLookupSelfResponse),
 			reusable: true,
 		},
 		{
 			name:     "Token Authentication success / Token is not renewable",
+			token:    "test-token",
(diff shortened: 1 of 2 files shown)


pkg/server/plugin/upstreamauthority/vault/vault_client_test.go
```diff
@@ -163,30 +163,44 @@ func (vcs *VaultClientSuite) Test_NewAuthenticatedClient_TokenAuth() {
 	vcs.fakeVaultServer.LookupSelfResponseCode = 200
 	for _, c := range []struct {
 		name      string
+		token     string
 		response  []byte
 		reusable  bool
 		namespace string
+		wantError bool
```
amartinezfayo, line 170: I think that in general I prefer to have an `err string` here as an expected error that can be used to match with the error obtained.  Test cases that don't expect errors would have an empty string. Then you can do something like
```
			if c.err != "" {
				vcs.Require().Equal(err.Error(), c.err)
			} ...
```
in the assertions to check for the error. I think that allows us to catch some unexpected behavior if the error thrown is different than what we expect.

amartinezfayo (commented):
Thank you @hiyosi for this fix and the patience for the review!
I have a small comment about the test. Looking great otherwise.

pkg/server/plugin/upstreamauthority/vault/vault_client_test.go
hiyosi: Thanks for review!  
Update: 5e0fd1f

amartinezfayo (approved):
Thank you @hiyosi!
Could you please update the branch so we can merge this?

hiyosi:
@amartinezfayo  Done rebasing the branch with master branch.  Thanks!

PR #2112, opened 2021-02-15
Serverless support: AWS plugin

### LINKED ISSUE #2070: Serverless support: AWS plugin
Implement a plugin that provides support to store issued identities in AWS Secrets Manager.

MarcosDY:
Implements AWS Secrets Manager plugin, this plugin will create a secret if no secrets is found with provided `secretname`
If a secret is found it will put a secret value with marshaled X509SVIDResponse.

No changes on Agent catalog are applied because it depends on another PR.

**Which issue this PR fixes**
fixes #2070


proto/spire/agent/svidstore/svidstore.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
```
MarcosDY, line 1: Proto is from another PR.

pkg/agent/plugin/svidstore/aws/aws.go
```diff
@@ -0,0 +1,240 @@
...
+// PutX509SVID puts the specified X509-SVID in the configured AWS Secrets Manager
+func (p *SecretsManagerPlugin) PutX509SVID(_ context.Context, req *svidstore.PutX509SVIDRequest) (*svidstore.PutX509SVIDResponse, error) {
+	opt, err := secretFromSelectors(req)
+	if err != nil {
+		return nil, err
+	}
+
+	regions := p.config.Regions
+	// TODO: may we allow default regions to be replaced for selectors?
+	if len(regions) == 0 {
```
MarcosDY, line 130: it is an important TODO, what I want to define here is what we really want to do with regions, on actual code we Put SVIDs in secrets from all `configured` regions from config file, so we are not allowing another regions or put into a single one of those configured regions so, so I'm thinking in different alternatives:

- Keep at it is now.
- Allow to set regions from selectors we may be able to have a selector like: `aws_secretsmanager:region:us-west-1:us-east-1 ` and we can put on those regions. (as default we push to all configured regions or selector regions)
- Get regions from selectors, but it MUST be a subset  of configured regions.

PR #2122, opened 2021-02-22
pkg/server/plugin/upstreamauthority/disk: remove double import

alrs:
The `disk` package was twice-importing `testutil`.


amartinezfayo (approved):
Thank you @alrs!

PR #2128, opened 2021-03-02
Add count subcommands to server cli agent, bundle, and entry commands

bri365:
Allow users to get only the total number of agents, bundles, or entries

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Spire server CLI agent, bundle, and entry command

**Description of change**
Add count commands to allow users to get just the total number of objects from the server cli commands

**Which issue this PR fixes**
part of [2129](https://github.com/spiffe/spire/issues/2129)


amartinezfayo:
Hi @bri365, thank you for this contribution!
I'm curious to know what's the main use case that you see for this. Thanks!

bri365:
Thanks @amartinezfayo, I added this to support large scale performance testing. When loading or working with 100,000 or more agents, I needed a simple way to verify they were all present.

MarcosDY:
you have something similar on debug endpoint where you can get that information too:
https://github.com/spiffe/spire/blob/master/pkg/server/api/debug/v1/service.go#L81

bri365:
In large environments, do we imagine users may wish to see what they might be asking for in terms of totals before issuing an agent list command to a server with 500,000 agents? I can use other means for testing if we don't see value in exposing this for users.

pkg/server/api/agent/v1/service_test.go
```diff
@@ -117,6 +117,93 @@ var (
...
+			spiretest.AssertLogs(t, test.logHook.AllEntries(), tt.expectLogs)
+			if tt.err != "" {
+				spiretest.RequireGRPCStatusContains(t, err, tt.code, tt.err)
+				require.Nil(t, resp)
+				return
+			}
+
+			require.NoError(t, err)
+			require.NotNil(t, resp)
+			require.Equal(t, resp.Count, tt.count)
```
MarcosDY, line 202: instead of only count, can you compare response protos? just to know we are getting response we want in all fields.
you can do that using `spiretest.AssertProtoEqual(t, tt.resp, resp`
at the same time, in golang require.Equal/assert.Equal expects (t, EXPECTED, ACTUAL)

This same comment is for all count RPCs

pkg/server/api/bundle/v1/service_test.go
```diff
@@ -1072,6 +1072,52 @@ func TestListFederatedBundles(t *testing.T) {
...
+			count: 1,
+		},
+		{
+			name:  "2 bundles",
+			count: 2,
+		},
+		{
+			name:  "3 bundles",
+			count: 3,
+		},
```
MarcosDY, line 1102: can you add a test case for ds error?

pkg/server/api/entry/v1/service_test.go
```diff
@@ -37,6 +37,68 @@ var (
...
+		{
+			name:  "3 entries",
+			count: 3,
+		},
+		{
+			name:    "ds error",
+			err:     "failed to count entries: ds error",
+			code:    codes.Internal,
+			logMsg:  "Failed to count entries",
+			dsError: errors.New("ds error"),
```
MarcosDY, line 70: ds will returns an status.Error

pkg/server/api/entry/v1/service_test.go
```diff
@@ -37,6 +37,68 @@ var (
...
+		},
+	} {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			ds := fakedatastore.New(t)
+			test := setupServiceTest(t, ds)
+			defer test.Cleanup()
+
+			for i := 0; i < int(tt.count); i++ {
+				entry, err := test.ds.CreateRegistrationEntry(ctx, &datastore.CreateRegistrationEntryRequest{
```
MarcosDY, line 80: `entry` is not used you can just do `_, err := test.ds.....`

pkg/server/api/entry/v1/service_test.go
```diff
@@ -37,6 +37,68 @@ var (
...
+					},
+				})
+				require.NoError(t, err)
+				require.NotNil(t, entry)
+			}
+
+			resp, err := test.client.CountEntries(context.Background(), &entrypb.CountEntriesRequest{})
+			require.NoError(t, err)
+			require.NotNil(t, resp)
+			require.Equal(t, tt.count, resp.Count)
```
MarcosDY, line 97: can you assert response instead of only Count? (with that you can remove require.NotNil() on previous line

test/integration/setup/adminclient/client.go
```diff
@@ -368,6 +371,19 @@ func batchSetFederatedBundle(ctx context.Context, c *itclient.Client) error {
 	return nil
 }
 
+func countFederatedBundles(ctx context.Context, c *itclient.Client) error {
```
MarcosDY, line 374: in all places it is called countBundles, may we call the same here?

test/integration/suites/spire-server-cli/04-bootstrap-agents
```diff
@@ -5,9 +5,9 @@ docker-compose exec -T spire-server \
     /opt/spire/bin/spire-server bundle show > conf/agent/bootstrap.crt
 
 # Set conf files for each agent
-cp -R conf/agent conf/agent-1
-cp -R conf/agent conf/agent-2
-cp -R conf/agent conf/agent-3
+cp -R conf/agent/ conf/agent-1
```
MarcosDY, line 8: is `/` required?

MarcosDY:
As a general comment, can you create issues for adding the same filters that regular lists for each of count?

test/integration/suites/spire-server-cli/04-bootstrap-agents
bri365: yes it appears to be, at least on Big Sur release. Didn't work without it

cmd/spire-server/cli/agent/count.go
```diff
@@ -0,0 +1,54 @@
...
+	agentClient := serverClient.NewAgentClient()
+	countResponse, err := agentClient.CountAgents(ctx, &agent.CountAgentsRequest{})
+	if err != nil {
+		return err
+	}
+
+	count := int(countResponse.Count)
+	msg := fmt.Sprintf("%d attested ", count)
+	msg = util.Pluralizer(msg, "agent", "agents", count)
+	env.Printf(msg + "\n")
```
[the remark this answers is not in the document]
bri365: yep; leftover from previous edit

pkg/server/api/entry/v1/service_test.go
bri365: It's currently checked for not nil; however the check for no error is probably sufficient

test/integration/setup/adminclient/client.go
bri365: Thanks; cut/paste error

amartinezfayo (commented):
Thank you so much @bri365 for this contribution!
I have two comments:
1) We shouldn't add the deprecated `-registrationUDSPath` argument. These are new sub-commands so we don't need a deprecation path to transit from `-registrationUDSPath` to `-socketPath`. We can safely introduce `-socketPath` since no existing users are relying on these sub-commands.
2) Looks like the changes needed in cmd/spire-server/cli/cli.go to effectively add the commands are missing, so the new commands will not be exposed. I'm not sure if this is intentionally left to a next PR.

bri365:
@amartinezfayo thank you for the feedback. 
Regarding deprecation of -registrationUDSPath, that is handled in the root cli command (cmd/spire-server/cli/run/run.go) and not the count subcommand.
I added the count subcommands to cli.go and integration tests to ensure them as well.

amartinezfayo:
> Regarding deprecation of -registrationUDSPath, that is handled in the root cli command (cmd/spire-server/cli/run/run.go) and not the count subcommand.

Got it! I was confused by the code in tests. Thanks!

azdagron (approved):
Awesome. This is looking good, @bri365 ! Thanks for the contribution.

PR #2135, opened 2021-03-05
Update supported versions in SECURITY.md

rturner3:
SPIRE only supports the most recent major version and the previous major version for security patches. The security policy currently specifies exact version ranges of major releases that are supported.

<=0.10.x is no longer supported after the release of 0.12.0. Updating the supported versions in SECURITY.md to reflect the current state.


amartinezfayo (approved):
LGTM!

PR #2142, opened 2021-03-09
Expose a configuration to disable rate limiting on JWT signing and X509 signing

### LINKED ISSUE #2140: Expose a configuration to disable rate limiting on JWT signing and X509 signing
We have a configurable to disable rate limiting on node attestation.
It is desirable to also be able to disable rate limiting on JWT and X509 signing since that may be needed in certain deployments.

amartinezfayo:
Add a config setting to be able to disable rate limiting on X509 signing and JWT signing.

Fixes #2140.


azdagron (approved):
\o/

PR #2155, opened 2021-03-15
Fix spiffe ID equality

SirNexus:
Signed-off-by: SirNexus <ai.carson@f5.com>

**Pull Request check list**

- [] Documentation updated?

**Affected functionality**
k8s-workload-registrar crd mode update method

**Description of change**
I experienced an issue using crd mode of the k8s registrar where on each restart, new svids would be generated for all workloads in my cluster. I submitted a question in the spiffe slack [here](https://spiffe.slack.com/archives/CBNCC2V17/p1615581868106500) and eventually realized that the issue lied in the fact that in crd mode, the k8s-workload-registrar was updating entries for every workload due to a bug in the equality (pointer comparison rather than value).

**Which issue this PR fixes**
n/a


azdagron (approved):
Thanks for this fix, @SirNexus !

PR #2159, opened 2021-03-17
AttestAgent: Include the attestor type in the error returned when could not find the node attestor type

amartinezfayo:
Improve the error returned to include the attestor type in the error returned when could not find the node attestor type.


azdagron (approved):
:rocket:

PR #2169, opened 2021-03-24
Configurable cache reload interval

### LINKED ISSUE #2147: Configurable entry cache reload interval
Hello there, we have deployed SPIRE v0.12.0 in our clusters. In some large ones, typically there are about 300K agents running. After attestation, each agent has ~10 selectors. So, the table node_resolver_map_entries has about 3M records. In our setup, we observe that the metric spire_server_entry_cache_reload_elapsed_time is between 10 seconds and 14 seconds.

Currently the reload interval is hardcoded as 5 seconds in SPIRE, which seems too frequent for our case. So, is it possible to make it configurable?

yangmarcyang:
Change-Id: I98493b94043febbaf7437c5ff2f20ca6fc3ec680
Signed-off-by: Marc Yang <marc.yang@bytedance.com>

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
This makes the in-memory cache reload interval configurable.

**Description of change**
Add cache_reload_interval option into spire-server configuration. The default value is 5s.

**Which issue this PR fixes**
fixes #2147


azdagron (commented):
Thanks for opening this @yangmarcyang. I apologize for the latency in the review.

This is looking OK as a short term solution. I've opened up #2182 to track work towards a better solution that doesn't require operator configuration or tuning.

Considering this is hopefully a short-term band-aid to the problem, I think we should move this configurable into the `experimental` section like so:
```
server {
    ...
    experimental {
        cache_reload_interval = ...
    }
    ...
}
```

In addition, we need to provide documentation that 1) clearly state that it is experimental and subject to change or removal, 2) provide details on why someone would want to change this configurable (i.e. mitigate high database load for extra large deployments), and 3) warn of the implications of increasing this interval (i.e. slower propogation of new or updated entries to agents).

yangmarcyang:
Thanks @azdagron for the comment. I have moved it into experimental section, and added your words into the documentation because you already explained the pros & cons clearly :)

azdagron (approved):
Great! Thanks, @yangmarcyang.

Issue #1573, opened 2020-05-21
Support additional roots in upstream CA bundles


mcpherrinm:
I'm using the AWS PCA upstream authority.

One issue with the current plugin for me is that I can't add additional CAs to the bundle of roots. 
Unfortunately there's no way to add them together right now:  The current PCA plugin only returns the root which signed the intermediate.

AWS PCA is per-region, so I'd like to have at least two PCAs in different regions. 
As well, rotating to a new root would require a period where both the new and old roots are trusted.

One option would be to add a bundle_file_path to the PCA plugin for the upstream bundle which works the same as the disk CA does today.

Or this could be a feature supported directly by SPIRE, instead of requiring re-implementation in each plugin.  It is unclear if this is required in other cases though:  This is possibly just a shortcoming of PCA in that the upstream can't provide additional roots.

mcpherrinm:
https://github.com/spiffe/spire/pull/1574 is a rough sketch of how it might be implemented in the PCA plugin

mcpherrinm:
This might actually be solved by #1550, by just using the administrative AppendBundle.

evan2645:
> This might actually be solved by #1550, by just using the administrative AppendBundle.

I think this is the right way forward in the long term. What wasn't clear to me was whether or not it's valuable to have a static on-disk file that can be configured (vs having to make an interaction). My hunch is that if something is dropping a file, it can call a command whenever the resource changes... either way, I've gone ahead and picked up #1574 in the interest of getting a solution in.

I'd be interested to hear if the existing functionality exposed by AppendBundle is good enough for you. Of course, we'll need to add it to the CLI utilities.

evan2645:
#1574 and #1550 have both been addressed. An API is now available for adding certificates to the bundle, and the aws_pca upstream authority plugin now supports `supplemental_bundle_path` configurable.

Closing this as fixed

PR #2172, opened 2021-03-25
Implement upstream authority plugin for GCP

dragcp:
Signed-off-by: ramand.dragcp <dheepak.gcp@gmail.com>

**Pull Request check list**

- [*] Commit conforms to CONTRIBUTING.md?
- [*] Proper tests/regressions included?
- [*] Documentation updated?

**Affected functionality**
Adds a new plugin for GCP Certificate Authority Service backed upstream authority

**Description of change**
Thank you for reviewing the PR #2039 by @Jonpez2. I am continuing that PR here. I believe I have incorporated all of the feedback

A few notes:
* Changes to go.sum and go.mod happened automatically as part of "make build" 
* We pick all the CAs in GCP CAS that are in enabled state and also matches the specified label key/value pair. The one with the earliest expiration is used to create and sign intermediate CAs
* I have tried to included relevant documentation links for most of the APIs and types of GCP CAS. 

**Which issue this PR fixes**


azdagron:
Thank you for pushing the feature forward, @dragcp. I'll close out #2039 in favor of this.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,393 @@
...
+func (p *Plugin) GetPluginInfo(ctx context.Context, req *plugin.GetPluginInfoRequest) (*plugin.GetPluginInfoResponse, error) {
+	return &plugin.GetPluginInfoResponse{}, nil
+}
+
+func (p *Plugin) getConfig() (*Config, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if p.c == nil {
+		return nil, makeError(codes.Internal, "not configured")
```
amartinezfayo, line 207: This is a failed precondition rather than an internal failure.
```suggestion
		return nil, makeError(codes.FailedPrecondition, "not configured")
```

amartinezfayo (commented):
Thank you @dragcp! I have some initial comments/suggestions.

doc/plugin_server_upstreamauthority_gcp_cas.md
```diff
@@ -0,0 +1,44 @@
...
+# Considerations
+This plugin relies on GCP Certificate Authority Service which is currently in Beta. So please do not use this plugin in production
+
+# Configuration
+The plugin accepts the following configuration options:
+
+| Configuration                 | Description                                                       |
+| ----------------------------- | ----------------------------------------------------------------- |
+| root_cert_spec.project_name   | Project in GCP that has the root CA certificate                   |
+| root_cert_spec.region_name    | The name of the region within GCP                                 |
```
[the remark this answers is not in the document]
dragcp: Sorry @amartinezfayo 

This configuration table is correct. We do need root_cert_spec. It is the "sample  configuration" section which is incorrect. The correct version is there in gcpcas.go as well as in gcpcas_test.go

I will correct and update the sample configuration in this doc to be:
```
root_cert_spec {
     project_name = "MyProject"
     region_name = "us-central1"
     label_key = "myapp-identity-root"
     label_value = "true"
}
trust_bundle_cert_spec = [
    {
         project_name = "DifferentProject"
         region_name = "us-east1"
         label_key = "differentapp-identity-root"
         label_value = "true"
    }
]
```

With that being the case, is this Configuration table reasonable? Or is there a better way to denote / describe it?

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,393 @@
...
+	RootSpec         CertificateAuthoritySpec   `hcl:"root_cert_spec,block"`
+	TrustBundleSpecs []CertificateAuthoritySpec `hcl:"trust_bundle_cert_spec,block"`
+}
+
+type CAClient interface {
+	CreateCertificate(ctx context.Context, req *privatecapb.CreateCertificateRequest) (*privatecapb.Certificate, error)
+	LoadCertificateAuthorities(ctx context.Context, spec CertificateAuthoritySpec) ([]*privatecapb.CertificateAuthority, error)
+}
+
+type gcpCAClient struct {
```
azdagron, line 70: nit: we typically group unexported structs, methods, and functions below exported ones

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,393 @@
...
+	pcaClient *pcaapi.CertificateAuthorityClient
+}
+
+func (client *gcpCAClient) CreateCertificate(ctx context.Context, req *privatecapb.CreateCertificateRequest) (*privatecapb.Certificate, error) {
+	return client.pcaClient.CreateCertificate(ctx, req)
+}
+func (client *gcpCAClient) LoadCertificateAuthorities(ctx context.Context, spec CertificateAuthoritySpec) ([]*privatecapb.CertificateAuthority, error) {
+	// https://pkg.go.dev/cloud.google.com/go/security/privateca/apiv1beta1#CertificateAuthorityClient.ListCertificateAuthorities
+	allCerts := make([]*privatecapb.CertificateAuthority, 0)
```
azdagron, line 79: nit: Unless you are specifying a length or capacity, or need to disambiguate between a nil and empty slice for some reason, we prefer the following syntax for declaration:
```suggestion
	var allCerts []*privatecapb.CertificateAuthority
```
`make` usage requires extra scrutiny by code readers to ensure the parameters are correct so it's better to avoid it if there isn't a compelling reason like the ones listed above.

doc/plugin_server_upstreamauthority_gcp_cas.md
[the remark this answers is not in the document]
amartinezfayo: I see now that there is a hierarchy, I missed that. I think that it's good to have the table but maybe split them in two: one for `root_cert_spec` and the other one for `trust_bundle_cert_spec` configs, and make it explicit. That way we can avoid to have the prefix notation that may be confusing.
Having a configuration example in addition to that is also good IMO.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,393 @@
...
+
+func (p *Plugin) Configure(ctx context.Context, req *plugin.ConfigureRequest) (*plugin.ConfigureResponse, error) {
+	// Parse HCL config payload into config struct
+	config := new(Config)
+	if err := hcl.Decode(config, req.Configuration); err != nil {
+		return nil, makeError(codes.InvalidArgument, "unable to decode configuration: %v", err)
+	}
+
+	// Swap out the current configuration with the new configuration
+	p.setConfig(config)
```
azdagron, line 193: We should be validating the configuration setting (or replacing) here. At a minimum we need to ensure the required fields are present and where it is possible, ensure the values are within the proper range or have the proper shape.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,393 @@
...
+	return p.c, nil
+}
+
+func (p *Plugin) setConfig(c *Config) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.c = c
+}
+
+func getClient(ctx context.Context) (CAClient, error) {
```
azdagron, line 219: nit: the Plugin methods should be grouped together in exported then unexported order. This standalone function should be moved after the Plugin implementation.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	return &gcpCAClient{pcaClient}, nil
+}
+
+func (p *Plugin) mintX509CA(ctx context.Context, csr []byte, preferredTTL int32) (*upstreamauthority.MintX509CAResponse, error) {
+	p.log.Debug("Request to GCP_CAS to mint new X509")
+	csrParsed, err := x509.ParseCertificateRequest(csr)
+	if err != nil {
+		return nil, makeError(codes.Internal, "unable to parse CSR: %v", err)
```
azdagron, line 237: Probably want InvalidArgument here since this would imply SPIRE gave the plugin a bad value.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,393 @@
...
+		return nil, makeError(codes.Internal, "no certificate authorities found with label pair %q:%q", rootSpec.LabelKey, rootSpec.LabelValue)
+	}
+
+	// The certs get returned in expiry order, and we want the one that is expiring the earliest
+	// so just grab the first one and use that.
+	chosenCA := allCertRoots[0]
+
+	// All of the CAs that are eligible for signing are still trusted
+	trustBundle := make([]*privatecapb.CertificateAuthority, 0, len(allCertRoots))
+	trustBundle = append(trustBundle, allCertRoots...)
```
azdagron, line 266: nit:
```suggestion
	trustBundle := append([]*privatecapb.CertificateAuthority(nil), allCertRoots...)
```

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	extractFirst(csrParsed.Subject.Organization, &subject.Organization)
+	extractFirst(csrParsed.Subject.OrganizationalUnit, &subject.OrganizationalUnit)
+	extractFirst(csrParsed.Subject.Locality, &subject.Locality)
+	extractFirst(csrParsed.Subject.Province, &subject.Province)
+
+	// https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/security/privateca/v1beta1#SubjectAltNames
+	san := privatecapb.SubjectAltNames{}
+	uris := make([]string, 0)
```
azdagron, line 294:
```suggestion
	var uris []string
```

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -0,0 +1,393 @@
...
+	fullChain := []*x509.Certificate{cert}
+	fullChain = append(fullChain, certChain[:len(certChain)-1]...)
+
+	derChain := x509util.RawCertsFromCertificates(fullChain)
+
+	// Then we append all the extra cert roots we loaded, plus anything we loaded from the
+	// trust bundle spec
+	rootBundle := make([]*x509.Certificate, 0, len(trustBundle))
+	for _, c := range trustBundle {
+		// Note. We are intentionally retrieving the immediate CAs from the GCP CAS that
```
azdagron, line 369: Does this mean that we can end up with intermediates in the trust bundle that don't chain up to a root also in the trust bundle? This can be very problematic for downstream software, which does not always verify such chains successfully by default (e.g. OpenSSL).

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas_test.go
```diff
@@ -0,0 +1,175 @@
...
+		ca.PemCaCertificates = make([]string, 0)
+		for _, caentry := range x509CA {
+			ca.PemCaCertificates = append(ca.PemCaCertificates, string(pemutil.EncodeCertificate(caentry)))
+		}
+		allCerts = append(allCerts, ca)
+	}
+	return allCerts, nil
+}
+
+func TestGcpCAS(t *testing.T) {
```
azdagron, line 106: nit: prefer the test methods to be at the top and all of the scaffolding (e.g. `fakeClient`) to be below. This follows our general guidance of exported methods/types sorting before unexported methods/types.

azdagron (commented):
Thanks for putting the final touches on this @dragcp. Things are looking pretty good. I had one big question around the inclusion of intermediates being included in the trust bundle that don't terminate with a root in the trust bundle that I think we need to figure out.

Also, it looks like the use of `cloud.google.com/go` has pulled in a bump to the gRPC and protobuf dependencies. I think it would be best if those dependencies were bumped in their own PR (that includes the changes to all of the regenerated code). I'm happy to put that PR up if you'd rather.

doc/plugin_server_upstreamauthority_gcp_cas.md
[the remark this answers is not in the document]
dragcp: Thank you

I will incorporate all of your feedback, squash, rebase and repush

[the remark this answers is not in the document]
azdagron: If you wouldn't mind just pushing new commits on top, it's easier to see the diff of what has changed that way (reduces review burden). A rebase and squash at the end after it has all been sorted out would be great.

dragcp:
> Also, it looks like the use of `cloud.google.com/go` has pulled in a bump to the gRPC and protobuf dependencies. I think it would be best if those dependencies were bumped in their own PR (that includes the changes to all of the regenerated code). I'm happy to put that PR up if you'd rather.

Do I need to do "go get cloud.google.com/go", "make all" and "make generate-check" and commit whatever has changed to a new PR?

azdagron:
We can scope it more finely to just gRPC and protobuf with:
```
$ go get google.golang.org/grpc@v1.36.0

$ go get google.golang.org/protobuf/proto@v1.26.0

$ go mod tidy

$ make generate
```

dragcp:
Thank you. I have raised a PR at #2175 to upgrade the dependencies separately

azdagron:
#2175 has been merged.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
dragcp: The scenario that I have in mind is:

Lets say that there are three departments of the same organization in google cloud project. The three departments have their own CA - say caX, caY and caZ respectively. And they might all have the same root CA - caM

If each department is their own trust domain ( say domain X, Y and Z ) and trust domain X and Y needs to trust each other but not domain Z

If we have caM in the trust bundle, then wouldn't it allow any department to accept any SVID generated from any department?

I was hoping that if we can tag with appropriate labels, then the trust bundle will only contain caX and caY rather than caM

Or is it a precondition that each trust domain should have an exclusive root CA? That would also eliminate the above problem

doc/plugin_server_upstreamauthority_gcp_cas.md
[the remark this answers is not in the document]
dragcp: Pushing new commits on top is easier for me. I was under the wrong impression that squash and rebase was necessary

Let me stop squashing

azdagron:
I just merged a big change related to UpstreamAuthority plugin interfaces that will cause a some conflicts here. If you don't mind, I can merge in master and push a commit to fix it all up. Happy to clean up my mess. Please let me know if you are ok with that.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/spiffe/spire/pkg/common/x509util"
+	"github.com/spiffe/spire/pkg/server/plugin/upstreamauthority"
+	"github.com/spiffe/spire/proto/spire/common/plugin"
+	"github.com/spiffe/spire/test/spiretest"
+	"github.com/stretchr/testify/require"
+	privatecapb "google.golang.org/genproto/googleapis/cloud/security/privateca/v1beta1"
+)
+
+func generateCert(cn string, issuer *x509.Certificate, issuerKey crypto.PrivateKey) (*x509.Certificate, crypto.PrivateKey, error) {
+	priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
MarcosDY, line 26: why not to use keys we have on fixture? https://github.com/spiffe/spire/blob/master/test/testkey/new.go#L41

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas_test.go
```diff
@@ -0,0 +1,175 @@
...
+            project_name = "proj1"
+            region_name = "us-central1"
+            label_key = "somelable"
+            label_value = "somevalue"
+        }
+    ]
+    `})
+	require.NoError(t, err)
+
+	priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
MarcosDY, line 151: may we use fixture here?

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
azdagron: Outside of verification woes by stacks that use OpenSSL (and probably others), I think including intermediates in the trust bundle is dangerous. It makes it much easier to accidentally open up trust in unintended ways. In this case, what you REALLY trust is the root. Without some way of constraining that intermediate, its hard to anchor trust in it. What you don't want is workloads in two trust domains with intermediates in the trust bundle to accidentally trust each other because they got their hands on the root somewhere (say from the system pool) and included it in their verification chain. That opens up doors for one trust domain to impersonate workloads in the other trust domain.

dragcp:
> 
> I just merged a big change related to UpstreamAuthority plugin interfaces that will cause a some conflicts here. If you don't mind, I can merge in master and push a commit to fix it all up. Happy to clean up my mess. Please let me know if you are ok with that.

Sorry @azdagron. I did not notice your comment till now. Thank you. Please feel free to do so

I have addressed the review comments locally. After you push the changes, I will commit on top of that

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
evan2645: > Or is it a precondition that each trust domain should have an exclusive root CA? That would also eliminate the above problem

It's not strictly required, but is strongly encouraged and generally assumed

I agree with everything @azdagron said. SPIRE used to include intermediates in the bundle (and exclude the real roots) to address the use case you're describing, but we found that in practice it didn't work well, wasn't very safe, and led to a lot of confusion. Removing certs from the bundle is hard from a compat standpoint - adding them is of course easy :)

I think people that find themselves in the circumstance you describe... which I might summarize as "We all have to have intermediates that share a root due to corporate policy, but we don't _really_ trust each other" ... should lean on name constraints to solve their problems.

doc/plugin_server_upstreamauthority_gcp_cas.md
```diff
@@ -0,0 +1,44 @@
...
+| root_cert_spec.label_key      | Label key - value pair is used to filter and select the relevant certificate  |
+| root_cert_spec.label_value    | Label key - value pair is used to filter and select the relevant certificate  |
+| trusted_root_spec             | ( Optional ) Array of entries containing project_name, region_name, label_key and label_value as described in root_cert_spec  |
+
+
+##Sample configuration:
```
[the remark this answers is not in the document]
dragcp: I have added it to server_full_conf

doc/plugin_server_upstreamauthority_gcp_cas.md
```diff
@@ -0,0 +1,44 @@
+# Server plugin: UpstreamAuthority "gcp_cas"
+
+The `gcp_cas` plugin uses Certificate Authority from Google Cloud Platform ( GCP )'s Certificate Authority Service ( CAS ) 
+to generate intermediate signing certificates for SPIRE Server.
+
+# Considerations
+This plugin relies on GCP Certificate Authority Service which is currently in Beta. So please do not use this plugin in production
```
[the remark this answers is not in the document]
dragcp: I have added a warning message

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
dragcp: I have switched the order

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
dragcp: I am validating it. And added a few tests for that

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
dragcp: Updated

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
dragcp: Makes sense. Thank you. Switched to root rather than intermediate

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas_test.go
dragcp: I am using it now

azdagron:
Thanks, @dragcp. I'll take a look.

azdagron:
I've given it a lot of thought. I don't think we want the functionality provided by `trust_bundle_spec`. It makes it too easy to do the wrong thing. When something is included in the bundle, from the perspective of the workload, other participants in the trust domain, _and even those federating with that trust domain_, that root is considered authoritative for that trust domain.

In other words, the trust bundle is _not_ the appropriate way to distribute other roots to workloads. There are other mechanisms for that.

The `trust_bundle_spec` is used to grab roots for other CAs presumably unrelated to the signing of SPIRE intermediates. I can't think of a good use case where these bundles should be considered authoritative for the trust domain but aren't actually used to sign SPIRE intermediates.

Maybe I'm missing something here?

azdagron:
I've just merged #2180. Consequently, you'll need to touch up the import for the upstreamauthority definitions.

Old import:
`upstreamauthorityv0 "github.com/spiffe/spire/proto/spire/server/upstreamauthority/v0"`
New import:
`upstreamauthorityv0 "github.com/spiffe/spire/proto/spire/plugin/server/upstreamauthority/v0"`

If you'd rather me touch it up and push a commit, I'm happy to oblige. Just let me know.

dragcp:
> I've given it a lot of thought. I don't think we want the functionality provided by `trust_bundle_spec`. It makes it too easy to do the wrong thing. When something is included in the bundle, from the perspective of the workload, other participants in the trust domain, _and even those federating with that trust domain_, that root is considered authoritative for that trust domain.
> 
> In other words, the trust bundle is _not_ the appropriate way to distribute other roots to workloads. There are other mechanisms for that.
> 
> The `trust_bundle_spec` is used to grab roots for other CAs presumably unrelated to the signing of SPIRE intermediates. I can't think of a good use case where these bundles should be considered authoritative for the trust domain but aren't actually used to sign SPIRE intermediates.
> 
> Maybe I'm missing something here?

Wouldn't it be the same as "supplemental_bundle_path" in aws_pca plugin? The only difference is that this is coming from GCP CAS instead of local disk
https://github.com/spiffe/spire/blob/master/doc/plugin_server_upstreamauthority_aws_pca.md

The associated ticket has a use case listed for two sets of root CA in a single trust domain where we sign with one set but the trust bundle has both:
https://github.com/spiffe/spire/issues/1573

azdagron:
Can you help me understand maybe a concrete use case that requires the functionality provided by trust bundle spec? I've dug into the previous PR when it was introduced and I'm not convinced there wasn't a misunderstanding there on how the trust bundle is built and maintained through the preparation and activation intervals, based on the responses from the UpstreamAuthority.

I'm not opposed to it per-se, I just need to understand the concrete scenario it is solving. If the use case is theoretical, I'd probably advocate for leaving it for now. It's much easier to add something like this later than it is to deprecate it out if it turns out to be the wrong thing.

dragcp:
We have an application composed of multiple services talking to each other. We might also have the same service running concurrently with different versions of the software ( say RC vs Prod ).  Any service can call any other service including RC version calling Production version. We want each service to authorize the call based on which service is calling and whether that service is running non-prod version or prod version

We could model this as single trust domain. There are a set of root CAs for prod and set of root CAs for non-prod. Then we can have a set of spire servers that sign with only non-prod CAs and another set of spire servers which sign only with prod CAs. The trust bundle in all cases will contain both the prod and non-prod root CAs. 

In this case, we will have two sets of information. (1) the SPIFFE Id identifies the workload. Say spiffe://mytrustdomain.org/serviceA. (2) the rootCA can tell us whether it was a prod version of the software or non-prod version of the software

So after a service to service call has performed mTLS for authentication, it can traverse the CA chain to identify whether it was signed by prod CA or non-prod CA and use that information to take authorization decision

Extending it further, we could have few more intermediate certificates with parent as non-prod CA for finer grained authorization ( than just a binary prod vs non-prod )

azdagron:
Thank you for taking the time to provide that use case. Some questions:

> Then we can have a set of spire servers that sign with only non-prod CAs and another set of spire servers which sign only with prod CAs
Do these SPIRE servers share a database (i.e. configured to use the same datastore)? If they have a shared database, then there is no need for each SPIRE server to return anything but the exact upstream root that signed its intermediate, since upstream roots are appended by each SPIRE server to the trust bundle. The datastore layer is the canonical source of the trust bundle for the trust domain. The roots returned from the UpstreamAuthority plugin are _not _ treated as the definitive set of authorities, but are instead appended to the existing roots present in bundle.

> In this case, we will have two sets of information. (1) the SPIFFE Id identifies the workload ... (2) the rootCA can tell us whether it was a prod version of the software or non-prod version of the software
An alternative here would be to model the prod and non-prod services as having distinct SPIFFE IDs (e.g. spiffe://mytrustdomain.org/prod/serviceA v.s. spiffe://mytrustdomain.org/prod/serviceB). One complication of using the rootCA as an indicator of whether it is prod or non-prod is that each 1) workload needs to understand what the set of prod CAs is, and that understanding needs to survive CA rotation. 2) each workload needs to perform this awkward inspection of the root CA to determine prod v.s. non-prod.

azdagron:
Furthermore, let's say you decide later on that you need to federate with other trust domains. Now you've likely pushed this prod/non-prod authorization semantic onto the other trust domain.

dragcp:
Thank you

Let me discuss this with @Jonpez2 and get back to you tomorrow

azdagron:
Awesome. I appreciate your willingness to work through these things. We're very excited about this plugin.

dragcp:
Could you please help with one data point:

Is there any functional difference between two different spiffe servers running with isolated independent databases ( say sqllite ) vs a single shared database?

Is it that:
If the upstream authority plugins on two different servers are using two different root CAs and only the signing CA's root is returned as "upstreamX509Roots", then the trust bundle received by different workloads will not have both the CAs?

azdagron:
Assuming each server has an independent CA, whether self-signed or a distinct upstream CA, then these are effectively two trust domains that happen to share the same name.

That's one difference. There are others, like each server will have its own set of registration entries, view of attested agents, etc.

dragcp:
Thank you

azdagron:
Another alternative: use a distinct trust domain for prod v.s. non-prod and federate the two.

dragcp:
Thank you for all the suggestions. I have dropped trust_bundle_cert_spec and also merged in from the master.

azdagron (approved):
Thank you for this, @dragcp and @Jonpez2!

Issue #2206, opened 2021-04-12
Metrics are only emitted few seconds on Prometheus


inajob:
I noticed all metrics are deleted after few seconds on Prometheus.
The cause is go-metics default expiration. https://github.com/armon/go-metrics/blob/8ca742bd545671bedae063d0672cb3c8da7d4437/prometheus/prometheus.go#L23

It is good for reducing metrics, but I think this feature should only activate for some specific time series.

For example, I want to collect the version of SPIRE server.
`spire_server_started` seems to bo good metric for this purpose, but this metric only emits for few seconds after server started.

I think this is not expected behavior, and not only `spire_server_started` but almost all metrics should be emitted forever.

* **Version**: 0.12.0
* **Platform**: collect metrics from Prometheus 2.23.0
* **Subsystem**: -

rturner3:
Hi @inajob , thanks for reporting this. I don't see any use cases as of now for stopping reporting of a stat, so changing the default behavior of the Prometheus metrics sink to not expire metrics makes sense to me.

I can push a quick PR for this.

rturner3:
Hey @inajob , before trying to change this default behavior, I wanted to understand a bit more what kinds of issues you're running into when the metrics are not emitted during every reporting cycle.

Also is this behavior problematic in your consumption of all metrics or just gauges like `spire_server_started`? I see that the metrics expiration configuration in `armon/go-metrics` is something that applies to all metric types. If the issue is specific to gauges, we may be able to work around this by periodically emitting the gauge values in SPIRE.

I am not as familiar about whether this metrics "expiration" behavior is a normal producer pattern for Prometheus, so if you have more experience using Prometheus, I would appreciate your input.

inajob:
Thank you for PR!

I think it is not only the gauge values problem. It seems to be ok that you let expiration of all metric type forever.

---

I noteiced below two problems, but it seems to be less impact than expiration of metrics.

In case of metric that has various labels, this change may cause combinatorial explosion problem.
But according to [this telemetry document](https://github.com/spiffe/spire/blob/main/doc/telemetry.md), there seems to be no metrics which cause problem.

And this change affects the behavior of count type metrics. Before this change, count type metrics reset every expiration. But after this change, these accumulate even if there are no events.

jalev:
I would argue only to only consider expiring things if people start complaining about it. Too much telemetry is better than spotty telemetry. 

[...]

PR #2239, opened 2021-04-21
Do not expire Prometheus metrics by default

rturner3:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Prometheus metrics sink

**Description of change**
The `armon/go-metrics` package by default drops emission of Prometheus metrics after 60 seconds. This behavior is undesirable as the default because it breaks dashboard and alert queries. SPIRE metrics in general are meant to be continually emitted and not unregistered, so switching the default behavior to not drop metrics after 60 seconds.

**Which issue this PR fixes**
#2206


evan2645 (approved):
I wondered if this was the right thing to do or not... if it could mislead operators into thinking that something is still being emitted when it is not. I had a look at the gauges we emit, and convinced myself that continuing to emit the last set value is the right thing to do in all cases.

Issue #1207, opened 2019-10-21
audit logging registrations / bundles


amoore877:
It may be helpful to be able to configure a specific logging on SPIRE Server that is purely for CRUD operations on registrations and trust bundles. This could be at the Server endpoint layer, and for every such call of interest we log information such as caller (as able to determine from gRPC call), operation attempted (delete registration, update bundle...), details of operation (specific registration(s) looks for, selection filter criteria...), and success / failure status of the operation

Having such information in a separate log file, or at least with a very distinct log field (eg. "audit"), would greatly help by

- standardizing how such information is recorded
- allowing easier automated observation into such information
- allowing storing of this specific information as a longer-lasting archive than regular logs, for post-incident investigation

amoore877:
node operations (agent eviction) should also be audited

MarcosDY:
I think that adding a new field like `type=audit` will do the work,
Sample:
```
spire-server_1  | time="2021-03-01T13:36:29Z" level=info msg="Entry created" caller-addr="172.20.0.3:48894" caller-id="spiffe://domain.test/admin" entry_id=bbde054d-3ec6-417d-a37b-dbe36a29f0e0 method=BatchCreateEntry service=entry.v1.Entry spiffe_id="spiffe://domain.test/bar" subsystem_name=api type=audit
```
`caller-addr and caller-id are already in api v2, for calls done from outside.`
 
I think separating api logs (for CRUDs) into different logs will make things harder when configuring (and debuging), but with a field we can easily search for those logs.
Looking for CRUDs logs I figure out that not all RPCs has logs for success calls, and some are `debug` and others are `info`. We may make all fo them `info` and make sure to have logs for each CRUD.

About node operations are you suggesting adding adit to for Attest/Renew too?

`All changes I'm suggesting are  for api v2, because the old API is deprecated.`

MarcosDY:
About `audit` field, in api v2 we have `service` (service=entry.v1.Entry), `subsystem`  (subsystem_name=api), and method (method=BatchCreateEntry) so with that it is possible to filter what we need without new fileld

amoore877:
> I think separating api logs (for CRUDs) into different logs will make things harder when configuring (and debuging), but with a field we can easily search for those logs.
that seems more maintainable to me too

> About node operations are you suggesting adding adit to for Attest/Renew too?
I think that might be helpful. It's not what I was concerned about when writing this issue, but I could see others finding value in it.

[...]

PR #2297, opened 2021-05-17
Add audit log to bundle service

MarcosDY:
Implements audit logs, add it to bundle endpoints to use it as a scaffolding before updating another endpoints

Solves part of #1207


pkg/common/peertracker/uds_bsd.go
```diff
@@ -12,9 +15,71 @@ func getCallerInfo(fd uintptr) (CallerInfo, error) {
...
+		return 0, fmt.Errorf("failed GIDs lookup: no GIDs for process")
+	case 1:
+		return gids[0], nil
+	default:
+		return gids[1], nil
+	}
+}
+
+func getAddr(proc *process.Process) (string, error) {
+	path, err := proc.Exe()
```
rturner3, line 79: This seems to be unimplemented in gopsutil: https://github.com/shirou/gopsutil/blob/76779af9093add061eca53ec76ae4116acbf8a4c/process/process_freebsd.go#L67

Which seems strange, since this information should be available as part of `ki_args`: https://github.com/freebsd/freebsd-src/blob/d0ea5e467f3c44909667c5ee90c3d26653fb6687/sys/sys/user.h#L124

rturner3 (commented):
I may not get a chance to look through the rest of the PR until tomorrow, so leaving some early feedback based on my reading of the BSD code added in the `peertracker` package.

pkg/common/peertracker/uds_bsd.go
```diff
@@ -12,9 +15,71 @@ func getCallerInfo(fd uintptr) (CallerInfo, error) {
...
+		return 0, fmt.Errorf("failed GIDs lookup: %v", err)
+	}
+
+	switch len(gids) {
+	case 0:
+		return 0, fmt.Errorf("failed GIDs lookup: no GIDs for process")
+	case 1:
+		return gids[0], nil
+	default:
+		return gids[1], nil
```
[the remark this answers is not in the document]
MarcosDY: I run a a bundle show to display what it gets and that is getting groupID:

```
 Audit log                                     caller-gid=20 caller-path=/Users/marcosyacob/repositories/spire/bin/spire-server caller-uid=501 method=GetBundle request-id=4e340138-92d5-419d-a491-2303f630ecc5 service=bundle.v1.Bundle status=success subsystem_name=api trust_domain=example.org type=audit
```
```
~/ id -u 
501
```
```
~/ id -g 
20
```

What I'm not sure is if must return `all` user groups or just the first one

pkg/server/api/bundle/v1/service.go
```diff
@@ -518,3 +569,50 @@ func applyBundleMask(b *types.Bundle, mask *types.BundleMask) {
...
+			fields[k] = v
+		}
+	}
+	return fields
+}
+
+func fieldsFromJwtAuthoritiesProto(jwtAuthorities []*types.JWTKey) logrus.Fields {
+	fields := logrus.Fields{}
+	for i, jwtAuthoriry := range jwtAuthorities {
+		fields[fmt.Sprintf("jwt_authority_expires_at_%d", i)] = jwtAuthoriry.ExpiresAt
```
rturner3, line 603: nit: could we use `.` rather than `_` as the separator between the property name and the index? e.g. `jwt_authority_expires_at.0`, `jwt_authority_expires_at.1`, etc. I think it more clearly distinguishes the fact that this field is part of a flattened list of values

pkg/server/api/bundle/v1/service.go
```diff
@@ -518,3 +569,50 @@ func applyBundleMask(b *types.Bundle, mask *types.BundleMask) {
...
+		for k, v := range fieldsFromX509AuthoritiesProto(proto.X509Authorities) {
+			fields[k] = v
+		}
+	}
+	return fields
+}
+
+func fieldsFromJwtAuthoritiesProto(jwtAuthorities []*types.JWTKey) logrus.Fields {
+	fields := logrus.Fields{}
+	for i, jwtAuthoriry := range jwtAuthorities {
```
rturner3, line 602:
```suggestion
	for i, jwtAuthority := range jwtAuthorities {
```

pkg/server/api/bundle/v1/service.go
```diff
@@ -518,3 +569,50 @@ func applyBundleMask(b *types.Bundle, mask *types.BundleMask) {
...
+	if inputMask == nil || inputMask.X509Authorities {
+		for k, v := range fieldsFromX509AuthoritiesProto(proto.X509Authorities) {
+			fields[k] = v
+		}
+	}
+	return fields
+}
+
+func fieldsFromJwtAuthoritiesProto(jwtAuthorities []*types.JWTKey) logrus.Fields {
+	fields := logrus.Fields{}
```
rturner3, line 601:
```suggestion
	fields := make(logrus.Fields, 0, len(jwtAuthorities))
```

pkg/server/api/bundle/v1/service.go
```diff
@@ -518,3 +569,50 @@ func applyBundleMask(b *types.Bundle, mask *types.BundleMask) {
...
+		fields[fmt.Sprintf("jwt_authority_expires_at_%d", i)] = jwtAuthoriry.ExpiresAt
+		fields[fmt.Sprintf("jwt_authority_key_id_%d", i)] = jwtAuthoriry.KeyId
+		fields[fmt.Sprintf("jwt_authority_public_key_%d", i)] = api.HashByte(jwtAuthoriry.PublicKey)
+	}
+
+	return fields
+}
+
+func fieldsFromX509AuthoritiesProto(x509Authorities []*types.X509Certificate) logrus.Fields {
+	fields := logrus.Fields{}
```
rturner3, line 612:
```suggestion
	fields := make(logrus.Fields, 0, len(x509Authorities))
```

pkg/server/api/bundle/v1/service.go
```diff
@@ -518,3 +569,50 @@ func applyBundleMask(b *types.Bundle, mask *types.BundleMask) {
...
+	for i, x509Authority := range x509Authorities {
+		fields[fmt.Sprintf("x509_authorities_asn1_%d", i)] = api.HashByte(x509Authority.Asn1)
```
rturner3, line 614: As the value here isn't actually the ASN.1 representation of the X.509 authority, could we use a more precise key name? Maybe `x509_authorities_asn1_sha256.%d`?

pkg/server/api/audit/audit.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (l *log) AddFields(fields logrus.Fields) {
+	for key, value := range fields {
+		l.fields[key] = value
+	}
+}
+
+func (l *log) Emit(fields logrus.Fields) {
+	l.log.WithFields(l.fields).WithFields(fields).Info("Audit log")
```
rturner3, line 40: Given how frequently this will be emitted and the fact that we already have the `type` field set to `audit`, is there any benefit to logging this "Audit log" message? It seems like it is redundant data.

Since this audit logger is specific to the API layer, maybe something like "API accessed" would be more appropriate?

rturner3 (commented):
Hey @MarcosDY , thanks for getting this started! I left a few comments about the interface definition and implementation.

The other thing I noticed throughout the PR was that log field keys are defined directly in the places where the log field values are set. In general, log field keys should be constants defined in `pkg/common/telemetry/names.go` to match current logger conventions.

pkg/server/api/audit/audit.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (l *log) EmitError(err error) {
+	fields := fieldsFromError(err)
+	l.log.WithFields(l.fields).WithFields(fields).Info("Audit log")
+}
+
+func (l *log) EmitBatch(s *types.Status, fields logrus.Fields) {
```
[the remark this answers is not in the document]
rturner3: Oops, I think I forgot to update this comment later on. I think my other suggestion of separate APIs would be a cleaner pattern - one API which accepts fields and one which doesn't.

pkg/common/peertracker/uds_bsd.go
```diff
@@ -3,6 +3,9 @@
 package peertracker
```
evan2645, line 3: The peertracker package is all about accepting connections from a colocated process, and ensuring that the thing on the other end of the connection is always the same thing (and that the connection does not get handed off). Since these additions are not directly related to that task, I'd much prefer to see this logic live somewhere else.

For example, we don't know that the process package operations for UID and GID lookup here are safe from PID race. Instead, we want peertracker to handle the PID race problem, and do the UID and GID lookups separately while we are being protected by peertracker e.g.:
```
watcher, _ := peertracker.WatcherFromContext(ctx)

uid, gid, _ := doUidGidLookup(watcher.PID())

if !watcher.IsAlive() {
    // Something is wrong
}
```

pkg/server/api/audit/audit.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+func (l *log) Emit(fields logrus.Fields) {
+	l.log.WithFields(l.fields).WithFields(fields).Info("Audit log")
+}
+
+func (l *log) EmitError(err error) {
+	fields := fieldsFromError(err)
+	l.log.WithFields(l.fields).WithFields(fields).Info("Audit log")
```
evan2645, line 45: Should the message string that we emit with these log lines be defined as a constant?

I also wonder what is the best thing to put here ... We already have the type=audit field attached ... can or should the message be something more useful?

pkg/server/api/bundle/v1/service.go
```diff
@@ -162,6 +191,10 @@ func (s *Service) PublishJWTAuthority(ctx context.Context, req *bundlev1.Publish
 
 // ListFederatedBundles returns an optionally paginated list of federated bundles.
 func (s *Service) ListFederatedBundles(ctx context.Context, req *bundlev1.ListFederatedBundlesRequest) (*bundlev1.ListFederatedBundlesResponse, error) {
+	rpccontext.AddRPCAuditFields(ctx, logrus.Fields{
+		"page_size":  req.PageSize,
+		"page_token": req.PageToken,
```
evan2645, line 196: Is this information useful to an auditor? I think they are only significant if you know the exact size and ordering of the data at the time the request was issued

pkg/server/api/middleware/authorization.go
```diff
@@ -47,6 +48,13 @@ func (m *authorizationMiddleware) Preprocess(ctx context.Context, methodName str
 	if id, ok := rpccontext.CallerID(ctx); ok {
 		fields[telemetry.CallerID] = id.String()
 	}
+	// Add request ID to logger, it simplify debugging when calling batch endpints
+	requestID, err := uuid.NewV4()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to create request ID: %v", err)
+	}
+	fields["request-id"] = requestID.String()
```
evan2645, line 56: I think most of the fields we define today are snake_case ?

pkg/server/endpoints/endpoints.go
```diff
@@ -202,7 +203,7 @@ func (e *Endpoints) createUDSServer(unaryInterceptor grpc.UnaryServerInterceptor
 	return grpc.NewServer(
 		grpc.UnaryInterceptor(unaryInterceptor),
 		grpc.StreamInterceptor(streamInterceptor),
-		grpc.Creds(auth.UntrackedUDSCredentials()))
+		grpc.Creds(peertracker.NewCredentials()))
```
evan2645, line 206: Hmm I wonder if there are compatibility concerns here. If a SPIRE Server is running in a k8s pod without `hostPid` set, is the CLI still usable?

evan2645 (commented):
Thanks for your work on this @MarcosDY!

I'll leave this PR to @rturner3 and @azdagron to review, I had just a few high level comments

pkg/server/api/audit/audit_test.go
```diff
@@ -0,0 +1,283 @@
...
+						"type":   "audit",
+						"emit":   "test",
+						"a":      "1",
+						"b":      "2",
+					},
+				},
+			},
+		},
+	} {
+		tt := tt
```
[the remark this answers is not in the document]
MarcosDY: You are right it is no required, some time ago some lints fails because we was calling a variable from a for inside a function, and it suggest to capture var before, (I think it was golint), 
So I get used to always add it, but for regular `t.Run()` it is not required

pkg/server/api/audit/audit.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/spiffe/spire-api-sdk/proto/spire/api/types"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type Log interface {
+	AddFields(logrus.Fields)
+	Emit(logrus.Fields)
+	EmitBatch(*types.Status, logrus.Fields)
```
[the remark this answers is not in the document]
MarcosDY: It is called batch to reference it is used in "batch" endpoints, those endpoints can return multiple status.

So for now we implemented:
Emit: allows emit a single audit log (regular endpoints)
EmitBatch: allows emit multiple `types status` on a single RPC.
EmitError: emits audit logs when a regular endpoint fails.

Not sure about renaming to Audit, what do you think about renames to :
Emit -> Audit
EmitBatch -> AuditTypeStatus
EmitError -> AuditError

pkg/server/api/audit/audit.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/spiffe/spire-api-sdk/proto/spire/api/types"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type Log interface {
+	AddFields(logrus.Fields)
+	Emit(logrus.Fields)
```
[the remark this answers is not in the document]
MarcosDY: mmm when I start implementing this new naming I got into:

rpccontext functions, and I'm worried about names there:
```
func EmitRPCAudit(ctx context.Context, fields logrus.Fields) {
	AuditLog(ctx).AuditWithFields(fields)
}

func EmitRPCAuditError(ctx context.Context, err error) {
	AuditLog(ctx).AuditWithError(err)
}

func EmitBatchRPCAudit(ctx context.Context, s *types.Status, fields logrus.Fields) {
	AuditLog(ctx).AuditWithTypesStatus(s, fields)
}
```

And when calling : `rpccontext.EmitRPCAudit(ctx, logrus.Fields{})`

Do you think we must maintain Emit for rpccontext functions?

or we can rename to : 
```
AuditRPC()
AuditRPCWithFields(logrus.Fields)
AuditRPCWithTypesStatus(logrus.Fields, types.Status)
AuditRPCWithError(err error)
```

pkg/server/api/bundle/v1/service.go
MarcosDY: it must be 3*len(), we add 3 per each jwtAuthority

pkg/server/api/audit/audit.go
MarcosDY: I'm not sure what another thing to add as message, as you I think we already have `type` we can make each function to set a message, or we can set a 'SUCCESS' / 'FAILURE' based on if error is present

pkg/server/api/bundle/v1/service.go
MarcosDY: that is the only information in request, I dont think it is useful either.
about size it is for response, and ordering is defined from datastore, 
what do you think about removing page_size and token, but add response size?
something like:
```
fields["respone_size"] = len(resp.Bundles)
```
or maybe we can keep token and add next token:
```
fields["page_token"] = req.PageToken
fields["respone_size"] = len(resp.Bundles)
fields["next_page_token"] = resp.NextPageToken()
```

pkg/server/endpoints/endpoints.go
MarcosDY: that is exactly what worry me, and why audit log must be optional, 
but I didnt found another options to get IDs, 
What do you think about allowing different levels of audit logs?
- Disable: no audit log is provided (default)
- Enable: enable audit log with full capabilities
- Restricted: add audit log but disable UDS tracking 

With restricted feauture we can allow adding audit logs with TLS tracking but not with UDS tracking, in that way we allow some audit features for cases where allowing `hostPid` is not an option

pkg/server/api/bundle/v1/service.go
```diff
@@ -518,3 +569,50 @@ func applyBundleMask(b *types.Bundle, mask *types.BundleMask) {
...
+	}
+	return fields
+}
+
+func fieldsFromJwtAuthoritiesProto(jwtAuthorities []*types.JWTKey) logrus.Fields {
+	fields := make(logrus.Fields, 3*len(jwtAuthorities))
+	for i, jwtAuthority := range jwtAuthorities {
+		fields[fmt.Sprintf("jwt_authority_expires_at.%d", i)] = jwtAuthority.ExpiresAt
+		fields[fmt.Sprintf("jwt_authority_key_id.%d", i)] = jwtAuthority.KeyId
+		fields[fmt.Sprintf("jwt_authority_public_key.%d", i)] = api.HashByte(jwtAuthority.PublicKey)
```
rturner3, line 605: nit: use constants from telemetry package rather than redefining them here

rturner3: Also for the public key, can we call the log field key `jwt_authority_public_key_sha256` so it's more obvious that it's a hash?

pkg/server/api/bundle/v1/service.go
```diff
@@ -518,3 +569,50 @@ func applyBundleMask(b *types.Bundle, mask *types.BundleMask) {
...
+	}
+
+	return fields
+}
+
+func fieldsFromX509AuthoritiesProto(x509Authorities []*types.X509Certificate) logrus.Fields {
+	fields := make(logrus.Fields, len(x509Authorities))
+	for i, x509Authority := range x509Authorities {
+		fields[fmt.Sprintf("x509_authorities_asn1.%d", i)] = api.HashByte(x509Authority.Asn1)
```
rturner3, line 614: 1) Can we call this `x509_authorities_asn1_sha256` to indicate that this is a hash?
2) Can we use a constant from the telemetry package?

pkg/server/api/middleware/audit.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/spiffe/spire/pkg/common/telemetry"
+	"github.com/spiffe/spire/pkg/server/api/audit"
+	"github.com/spiffe/spire/pkg/server/api/rpccontext"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func WithAuditLog(udsTrackerEnabled bool) Middleware {
+	return auditLogMiddleware{
+		udsTrackerEnabled: udsTrackerEnabled,
```
rturner3, line 18: The maintainers had a discussion on this today and decided that for now we'd like to see the auditing feature either be fully enabled or not enabled at all. Allowing this "half-enabled" auditing mode for the TCP endpoint only feels kind of counterintuitive because the auditing of certain requests becomes dependent on how the Server is configured.

Since audit logging is disabled by default, I think we should be ok from a compatibility perspective right now.

pkg/server/api/rpccontext/audit.go
```diff
@@ -0,0 +1,50 @@
...
+
+func AuditRPCWithTypesStatus(ctx context.Context, s *types.Status, fields logrus.Fields) {
+	if auditLog, ok := AuditLog(ctx); ok {
+		auditLog.AuditWithTypesStatus(fields, s)
+	}
+}
+
+func AuditLog(ctx context.Context) (audit.Logger, bool) {
+	auditLog, ok := ctx.Value(auditLogKey{}).(audit.Logger)
+	return auditLog, ok
```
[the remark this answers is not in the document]
MarcosDY: that will not work, because type assertions on returns statements results into a single output (same that happens with maps)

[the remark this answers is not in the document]
rturner3: Oh yeah, you're totally right. Please disregard my comment

pkg/server/api/middleware/audit.go
rturner3: I don't think this flag is needed anymore?

MarcosDY: it is for unit tests, to avoid reading tracker when it will not be possible/desirable on tests

rturner3: I see, that makes sense.

pkg/server/api/bundle/v1/service.go
rturner3: Echoing what @evan2645 said, I'm not sure any of these request parameters are really all that useful from an auditing perspective. There are no restrictions over what data is exposed over the `ListFederatedBundles` API, e.g. client `A` sees bundles `1` and `2` whereas client `B` sees bundles `3` and `4` in the response. So if this RPC is called, the assumption would be that the client received all the bundles.

Whether the client intentionally ignores any of the bundles during its handling of the paginated results (i.e. there are 3 pages worth of data, but the client only requests pages `1` and `2`) doesn't seem necessary to log when we may not be able to receive the same data from the server when replaying the request(s) with the audited request parameters.

pkg/server/endpoints/endpoints.go
rturner3: I made this comment already in a different place in the PR, but for posterity, we settled on only allowing the first two options @MarcosDY mentioned - either disabled (default behavior) or fully enabled for both TCP and UDS endpoints, but no "half-enabled" mode where TCP endpoint auditing is enabled, but UDS endpoint auditing is not.

pkg/common/telemetry/names.go
```diff
@@ -227,6 +251,12 @@ const (
 	// Nonce tags some nonce for communication
 	Nonce = "nonce"
 
+	// PageSize tags a page size used on pagination
+	PageSize = "page_size"
+
+	// PageToken tags token used on pagination
+	PageToken = "page_token"
+
```
rturner3, line 259: These constants can be removed now since there is no reference to them anymore.

rturner3 (approved):
Looks great, @MarcosDY! 🔍 📁 🥳

PR #2315, opened 2021-05-21
Replace golint to revive

hiyosi:

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

linters

**Description of change**

Use `revive` instead of `golint`
Since `golint` has been deprecated(the `golint` repository has already been archived), I think it is better to replace. 
`revive` is mentioned by the golangci-lint doc.

https://golangci-lint.run/usage/linters/

What do you think about this code style? Is it acceptable?

**Which issue this PR fixes**

https://github.com/spiffe/spire/pull/2198
@MarcosDY may also be working on same as this. Please comment if you have any concerns.


MarcosDY:
Looks like it is deprecated from 1.41, do you tried to upgrade golanglint to that version?

.golangci.yml
```diff
@@ -28,6 +28,6 @@ linters:
     - gocritic
 
 linters-settings:
-  golint:
+  revive:
     # minimal confidence for issues, default is 0.8
-    min-confidence: 0.0
+    confidence: 0.0
```
MarcosDY, line 33: may we add `ignoreGeneratedHeader = true`?

hiyosi:
> Looks like it is deprecated from 1.41, do you tried to upgrade golanglint to that version?

oh, you are right.

Once I tried to upgrade the version, I find that there is a bug.
https://github.com/golangci/golangci-lint/pull/2005

I guess it's better to wait for the bug to be fixed.

azdagron (approved):
Thanks, @hiyosi !

PR #2319, opened 2021-05-21
make sure to not use suites on TestConfigure 

MarcosDY:
Make sure suite is not used to configure plugins when running TestConfigure,  on gcp and azure NodeAttestors


azdagron (approved):
\o/

PR #2327, opened 2021-05-25
Fail early in FetchJWTSVID when SPIFFE ID(s) are invalid

rturner3:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
`FetchJWTSVID` argument processing, logging

**Description of change**
If the client of `FetchJWTSVID` sets the `spiffe_id` field in the `JWTSVIDRequest` and the value is an invalid SPIFFE ID, FetchJWTSVID does not reject the request until after it has done the expensive workload attestation check. Such requests will always be rejected due to an invalid argument, so process the `spiffe_id` request parameter earlier to avoid unnecessary expensive operations later on.

Similarly, in the more unlikely case that there is a matching entry for the client in the Agent cache that does not have a valid SPIFFE ID, fail the entire request before trying to fetch a JWT-SVID for any of the matching entries.

This change also removes the high cardinality SPIFFE ID value from the log message emitted when any of the SPIFFE IDs of the JWT-SVIDs to be fetched are invalid. Instead, include this as a log field so that the message is a constant string. This is more in line with logging practices in SPIRE where dynamic values are only added as field values in the log statement rather than as formatted parameters in the message.

Also include the SPIFFE ID as a log field when the fetch of the JWT-SVID from the cache fails to give clearer indication of the failure.


azdagron (approved):
Thanks, @rturner3!

PR #2331, opened 2021-05-26
Add MatchAny and Superset filters

MarcosDY:
Add datastore changes to support filtering Attested nodes and Registration entries using MatchAny and Superset filters on BySelectors.
Add MatchAny and Superset filters when filtering Registration entries ByFederatesWith.

**Which issue this PR fixes**
fixes partially #2141


azdagron (approved):
Thanks, @MarcosDY!

PR #2333, opened 2021-05-27
fix data consistency naming and misc cleanup

bri365:

Correct remaining incorrect naming from last datastore interface refactor
Correct a few miscellaneous items


azdagron (approved):
\o/

PR #2337, opened 2021-05-29
elevate datastore from plugin directory

### LINKED ISSUE #2330: Elevate datastore to pkg/server
Consider moving datastore to pkg/server as a next step in removing plugin artifacts from the datastore and elevating it to a core component.

This makes more clear the nature of the datastore interface as a core server component and opens the way for future pluggable lower level components. For consideration and feedback, I propose renaming plugin/datastore to plugin/sqlstore to reflect the SQL only nature of that implementation. Leaving it in the plugin directory also simplifies testing that relies on existing plugin suite infrastructure.

The resulting directory structure would look like the following:

pkg/server/datastore/ as the datastore interface definition and facade layer
pkg/server/plugin/sqlstore/ as the SQL implementation of the datastore interface

bri365:
In the interest of continuing the datastore evolution from plugin to core component, this PR moves datastore up a level and creates a datastore facade for the SQL implementation. This work also facilitates prototyping alternative datastore implementations like a KV store under the datastore facade.

Specifically this PR:
- Elevates datastore from plugin to pkg/server as a core component
- Renames SQL datastore plugin to sqlstore

For consistency, the sqlstore/sql plugin logs as datastore-sql

Note: the initial commit created a SQLStore struct and repository in the plugin/sqlstore directory. These did not appear to add real value as the SQL datastore is not intended to be pluggable, so the facade is removed in the second commit.

fixes #2330


azdagron:
Thanks for opening this @bri365. Sorry it took a bit to provide feedback on the naming. In terms of the change, I think that the rename from `pkg/server/plugin/datastore` to `pkg/server/datastore` seems just fine. I'm also fine with the rename from `sql` to `sqlstore`, however I think I'd still rather that package be a child of `datastore` instead of the plugin directory.

I think in the future, if we open up pluggable data storage, the package layout would look something like this:
`pkg/server/datastore` <-- datastore interface definition
`pkg/server/datastore/sqlstore` <--- concrete datastore interface implementation using old-style sql backend
`pkg/server/datastore/kvstore` <--- concrete datastore interface implementation using new-style the kvstore plugin type
`pkg/server/plugin/kvstore` <--- KV store interface with version facades
`pkg/server/plugin/kvstore/whatever` <--- KV store plugin built-in 

In any case, `sqlstore` doesn't belong under `pkg/server/plugin`... that directory is intended for packages for plugin types (e.g. keymanager)

bri365:
I left sqlstore under the plugin directory because I believed the plugin test structure for integration tests required it. I will double check that assumption

azdagron:
Yeah, should should be fine. The sql integration tests just work off of the go package path. Doesn't really matter where they live.

bri365:
ye[p; running integration tests now. I'm thinking of keeping the sqlstore/sql filestructure under datastore so kvstore/etcd would live alongside that - make sense?

azdagron:
>  I'm thinking of keeping the sqlstore/sql filestructure under datastore so kvstore/etcd would live 

:cow: :razor: 

I'm not sure about that one. What other packages would live under `sqlstore`?

I imagine kvstore/etcd would be under the plugin directory:
`pkg/server/plugin/kvstore/etcd`
and the kvstore facade under:
`pkg/server/plugin/kvstore`
and the datastore implementation using the kvstore facade (and therefore backed by any plugin):
`pkg/server/datastore/kvstore`

The `sql` package (or `sqltore`) is a datastore implementation using over SQL, and IMO belongs under:
`pkg/server/datastore/sql`

bri365:
Fair enough. sql or sqlstore implement datastore, so the extra level is unneeded.

pkg/server/catalog/catalog.go
```diff
@@ -59,8 +59,10 @@ type Config struct {
 	HealthChecker    health.Checker
 }
 
+type datastoreRepository struct{ datastore.Repository }
+
 type Repository struct {
-	datastore.Repository
+	datastoreRepository
```
azdagron, line 65: what's the purpose of this change?

bri365:
Consistency and paving the way for KV to avoid future merge collisions 

Brian

azdagron:
Sorry about the merge conflict. I merged Marcos' filtering additions :) If you wouldn't mind resolving, I think we're ready to merge afterwards.

bri365:
LOL! One of us was going to have to deal with that...

azdagron (approved):
\o/

PR #2345, opened 2021-06-04
Add Cure53 Audit Results to README

evan2645:
We were holding back on publishing this report as we wanted to publish a
blog post along with its release, however the post is yet to be written
and it's unclear when we'll have the time to do it. Rather than further
delay the publishing of this report, which was kindly sponsored by the
CNCF, we should make it available in the absence of a blog post (which
can always come later).


azdagron (approved):
I concur. Thanks for putting this up. And thank you Cure53 and CNCF for the productive audit!

PR #2356, opened 2021-06-09
Add ability to authenticate with Kubernetes auth method for vault plugin

### LINKED ISSUE #2348: Vault UpstreamAuthority: Support kubernetes authentication to Vault
Feature request: The Vault UpstreamAuthority plugin should support using kubernetes authentication, specifically with a PSAT. The benefits of PSATs over long-term stable credentials are well documented in the PSAT node attestor plugin, it should be supported for Vault auth too.

hiyosi:
Signed-off-by: Tomoya Usami <usami.tomoya@gmail.com>

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

**Description of change**

Add ability to authenticate with Kubernetes auth method for vault plugin

**Which issue this PR fixes**

This PR fixes #2348


hiyosi:
Setup Vault
```
$ vault auth enable -path kubernetes-test kubernetes
$ vault write auth/kubernetes-test/config issuer="https://kubernetes" token_reviewer_jwt="${VAULT_SA_TOKEN}" kubernetes_host="https://kubernetes" kubernetes_ca_cert=@/tmp/ca.pem
$ vault write auth/kubernetes-test/role/my-role bound_service_account_names=spire-server bound_service_account_namespaces=spire policies=spire 

```

Vault logs
````
2021-06-09T08:14:22.630Z [INFO]  core: successfully enabled credential backend: type=kubernetes path=kubernetes-test/
...
2021-06-09T08:15:40.150Z [DEBUG] identity: creating a new entity: alias="id:"34d2d3c1-cc8c-fd11-2e3c-25549db2ff47"  canonical_id:"e7c1de0f-d713-36fc-6fb1-d0ec980264bc"  mount_type:"kubernetes"  mount_accessor:"auth_kubernetes_e9dae46a"  mount_path:"auth/kubernetes-test/"  metadata:{key:"service_account_name"  value:"spire-server"}  metadata:{key:"service_account_namespace"  value:"spire"}  metadata:{key:"service_account_secret_name"  value:""}  metadata:{key:"service_account_uid"  value:"6db91896-b3ab-41cb-afce-717bd0401772"}  name:"6db91896-b3ab-41cb-afce-717bd0401772"  creation_time:{seconds:1623226540  nanos:150603106}  last_update_time:{seconds:1623226540  nanos:150603106}  namespace_id:"root""
```

pkg/server/plugin/upstreamauthority/vault/vault_client.go
```diff
@@ -30,6 +31,8 @@ const (
 	defaultCertMountPoint    = "cert"
 	defaultPKIMountPoint     = "pki"
 	defaultAppRoleMountPoint = "approle"
+	defaultK8sMountPoint     = "kubernetes"
+	defaultK8sTokenPath      = "/var/run/secrets/kubernetes.io/serviceaccount/token" // #nosec G101
```
amartinezfayo, line 35: Setting a default token path that is not a projected service account token will effectively make a choice that negatively affects the security posture, as the default service account token is unbounded.
I suggest that we should not allow leaving blank the token path assuming the default (non projected) service account token.
I recommend that instead we should require setting an explicit token path, so the choice about the token used is a conscious decision.

pkg/server/plugin/upstreamauthority/vault/vault_client.go
```diff
@@ -115,6 +126,8 @@ func NewClientConfig(cp *ClientParams, logger hclog.Logger) (*ClientConfig, erro
 	defaultParams := &ClientParams{
 		CertAuthMountPoint:    defaultCertMountPoint,
 		AppRoleAuthMountPoint: defaultAppRoleMountPoint,
+		K8sAuthMountPoint:     defaultK8sMountPoint,
+		K8sAuthTokenPath:      defaultK8sTokenPath,
```
amartinezfayo, line 130: Please look at the comment about assuming a default token path.

amartinezfayo (commented):
Thank you @hiyosi for this! It is looking great!
I have some comments. My main concern is around assuming a default (non projected) service account token.

pkg/server/plugin/upstreamauthority/vault/vault_client.go
hiyosi: That makes sense.

However, I intended as follows.
The BoundServiceAccountToken feature has been enabled (as beta) since K8s v1.21.0. It use the Projected Service Account Token as defaults and replaces the legacy Service Account Token.
Therefore, you can get the token generated by the projected service account token at the above path.

Should we take care for the Bound Service Account Token as defaults a little later?

amartinezfayo: I think that we can evaluate that when this feature is GA (planned for v1.22), considering how many users may be running pre v1.21 deployments by that time.
My suggestion is that at that moment we can consider to introduce the default value and document this.

hiyosi: I agree with your suggestion.

pkg/server/plugin/upstreamauthority/vault/vault_client.go
hiyosi: fa226e1

PR #2369, opened 2021-06-21
Stop returning Bundle on unsuccessful FetchJWTSVID 

### LINKED ISSUE #2326: FetchJWTSVID does not return values for "svids" field when caller has identity, but not one requested with "spiffe_id" request field
### Background
SPIFFE requires that `FetchJWTSVID` return a [non-empty](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Workload_API.md#621-fetchjwtsvid) `svids` field in the `JWTSVIDResponse` message for all successful requests.

### Description of issue

SPIRE Agent's implementation of `FetchJWTSVID` does not meet this part of the specification in the case when the caller sets the `spiffe_id` request field to a non-default value, and the caller has a registered identity based on its selectors, but it doesn't have an identity with the `spiffe_id` from the request. The identity that the caller is registered for must have a common selector with another registration that the caller is not authorized for.

### Example repro
1. Create an entry that matches no process, but has a common selector with the calling process (Unix user named `rturner3` in this example):
```
$ spire-server entry create -parentID 'spiffe://example.org/agent/foo' -spiffeID 'spiffe://example.org/rturner3' -selector 'unix:user:rturner3' -selector 'unix:group:nobody'
Entry ID         : 45291f21-e852-42e2-a0c9-986e1dee5854
SPIFFE ID        : spiffe://example.org/rturner3
Parent ID        : spiffe://example.org/agent/foo
Revision         : 0
TTL              : default
Selector         : unix:group:nobody
Selector         : unix:user:rturner3
```

2. Wait 10 seconds for the Agent to pick up the new entry
3. Request a JWT-SVID with the `sub` claim `spiffe://example.org/rturner3` and expect the `PermissionDenied` status code:
```
$ spire-agent api fetch jwt -audience 'spiffe://example.org/bob' -spiffeID 'spiffe://example.org/rturner3'
rpc error: code = PermissionDenied desc = no identity issued
```

4. Create an entry that matches the caller process and has a common selector with the first registration (`unix:user:rturner3`), but has a different SPIFFE ID from the first registration:
```
$ spire-server entry create -parentID 'spiffe://example.org/agent/foo' -spiffeID 'spiffe://example.org/ryan' -selector 'unix:user:rturner3'
Entry ID         : e8c0de26-003f-4eab-a624-8c68cc84835a
SPIFFE ID        : spiffe://example.org/ryan
Parent ID        : spiffe://example.org/agent/foo
Revision         : 0
TTL              : default
Selector         : unix:user:rturner3
```

5. Wait 10 seconds for the Agent to pick up the new entry
6. Request a JWT-SVID with the `sub` claim `spiffe://example.org/rturner3` again, and expect PermissionDenied again, but this time we get back a response with no SVIDs:
```
$ spire-agent api fetch jwt -audience 'spiffe://example.org/bob' -spiffeID 'spiffe://example.org/rturner3'
bundle(spiffe://example.org):
{
    "keys": [
        {
            ...
        },
       ...
    ]
}
```

MarcosDY:
stop returning a success response when calling FetchJWTSVID when no issued identity

**Which issue this PR fixes**
fixes #2326


azdagron (approved):
\o/, thanks @MarcosDY

PR #2374, opened 2021-06-23
Add a ban command to SPIRE Server agent CLI

### LINKED ISSUE #2373: Add ban agent to spire-server CLI
A feature was added on refactor to SPIRE Server API, that allows Agents to be marked as banned, banned agents must not be able to connect to spire until they are removed and re-attested
But this feature was not added to SPIRE Server CLI, 
This client CMD must allow users to mark Agents as banned.

MarcosDY:
Add a `spire-server agent ban` that calls SPIRE Server's Agent API  to ban an agent.

**Which issue this PR fixes**
fixes #2373


cmd/spire-server/cli/agent/agent_test.go
```diff
@@ -51,6 +59,67 @@ func (s *agentTest) afterTest(t *testing.T) {
 	t.Logf("STDERR:\n%s", s.stderr.String())
 }
 
+func TestBanHelp(t *testing.T) {
+	test := setupTest(t, agent.NewBanCommandWithEnv)
+
+	test.client.Help()
+	require.Equal(t, `Usage of agent ban:
+  -registrationUDSPath string
```
amartinezfayo, line 67: We don't need to support this flag, since this command is being introduced now and there are no compatibility concerns.

amartinezfayo (commented):
Thank you for this @MarcosDY!

cmd/spire-server/cli/agent/agent_test.go
MarcosDY: it is inherited by `https://github.com/spiffe/spire/blob/main/cmd/spire-server/util/util.go#L133` that is common to all our cmds, an option is to reimplement it without that dependency?

amartinezfayo: In that case I think that it's ok to leave it as is.

amartinezfayo (approved):
Thanks for this @MarcosDY!

PR #2384, opened 2021-06-25
Log a warning when the `aws_kms` KeyManager plugin is loaded

evan2645:
The `aws_kms` KeyManager plugin does not currently support ACME. Users
who configure both ACME-based bundle endpoint server _and_ the `aws_kms`
KeyManager will find that federation simply does not work. This is a
fairly major breakage of a SPIRE core feature, and it is being tracked
in #2302.

Unfortunately, we have already released the `aws_kms` plugin, and we are
anticipating a 1.0 release in less than two weeks. The situation
presents a terrible user experience, and I think we have to do
_something_. Particularly bad is the experience of a user who is already
using `aws_kms` plugin and now wants to configure ACME-based federation.

I see three options (perhaps there are more):
* Fix the bug
* Revoke support for the `aws_kms` plugin
* Warn whenever someone uses the `aws_kms` plugin

The first option is seemingly off the table as the author of the plugin
is not currently working on the fix, and SPIRE mainatainers are heads
down getting ready for release. We hope to pick a 1.0 release candidate
_today_.

The second option presents the best user experience for anyone not
already using `aws_kms`, but also the worst user experience for anyone
who is already running it.

I chose the third option, which a good experience for anyone already
running `aws_kms` (since presumably they are not using ACME ...), and a
mediocre experience for anyone who is not already running it but will
soon.

Any votes for option 2, or any takers for option 1?


kunzimariano:
@evan2645: 

@marcosy will be the taker of option 1. I will review the potential changes too.

azdagron (approved):
Thanks, @evan2645!

@marcosy @kunzimariano, thanks for raising your hand to address the issues with the AWSKMS key manager. W're going to take this for now.  If you could back this warning out with the PR that addresses the AWSKMS issues, that would be swell.

Issue #2218, opened 2021-04-14
Enhance AWS node attestor server plugin to validate EC2 instances across multiple AWS account.


sushil-prasad:
We are trying to set up the spire infrastructure in our production environment. The production environment has multiple AWS accounts. We want to deploy a spire server in one of the accounts and attest the EC2 hosts in all other AWS accounts using AWS aws_iid node attestor. 

When both the spire server and the spire agent runs in the same AWS account, the spire server is able to attest to the EC2 host. However when spire server and spire agent are in different AWS account, the node attestation fails with this message:
`ERRO[0000] agent crashed                                 error=“failed to get SVID: error getting attestation response from SPIRE server: rpc error: code = Internal desc = failed to attest: rpc error: code = Unknown desc = aws-iid: attempted attestation but an error occurred querying AWS via describe-instances: InvalidInstanceID.NotFound:`

sushil-prasad:
One potential solution for this problem could be to allow providing a **role** name as part of AWS aws_iid node attestor plugin configuration. The spire server will perform AssumeRole in the AWS account presented by the AWS aws_iid document. It is assumed that same role name is present in all AWS accounts. If no role name is provided as part of plugin configuration, the spire server will continue with the current implementation.

walmav:
If we limit the selectors to be based on the fields of the IID and since we are able to verify the IID document through AWS DSA public certificate, wonder if we can get away from the need for AWS specific credentials on the SPIRE server.

georgi-lozev:
I like @walmav's idea. It will make the workflow similar to the one implemented in the `gcp_iit` plugin. The plugin could then get a list of allowed accounts and accept verified IIDs only from those.
 
Looking into the [docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html) I see there are 3 types of signatures available. 
Do you know what is the difference between them and are there any security implications from using one or the other?

sushil-prasad:
I was digging into the AWS IID. When I run `curl  http://169.254.169.254/latest/dynamic/instance-identity/document` on the EC2 instance, I see following output
{
  “accountId” : “153436698118",
  “architecture” : “x86_64",
  “availabilityZone” : “us-west-2b”,
  “billingProducts” : null,
  “devpayProductCodes” : null,
  “marketplaceProductCodes” : null,
  “imageId” : “ami-0e81828bfafcdddbd”,
  “instanceId” : “i-06309d51c78628654”,
[...]

PR #2387, opened 2021-06-29
Added assume role functionality to AWS server node Attestor ( #2218)

sushil-prasad:
If **assume_role_arn_template** parameter is set in the AWS server node Attestor plugin, the spire server will assume the role as specified by the **assume_role_arn_template** after expanding the template. Currently only template variable available is **AccountID** from the AWS IID document sent by the spire agent to the spire server.


pkg/server/plugin/nodeattestor/aws/client.go
```diff
@@ -78,17 +82,37 @@ func (cc *clientsCache) getClient(region string) (Client, error) {
 		return nil, status.Error(codes.FailedPrecondition, "not configured")
 	}
 
-	client, err := cc.newClient(cc.config, region)
+	var asssumeRoleArn bytes.Buffer
+	if cc.config.AssumeRoleArnTemplate != "" {
+		tmpl, err := template.New("assume-role-arn").Parse(cc.config.AssumeRoleArnTemplate)
+		if err != nil {
+			return nil, err
```
MarcosDY, line 89: all errors returned by plugins must be grpc status, can you return:
```
status.Errorf(codes.Internal, "failed to create tempalte: %v", err)
```
Same comment for all returned errors

pkg/server/plugin/nodeattestor/aws/client.go
```diff
@@ -78,17 +82,37 @@ func (cc *clientsCache) getClient(region string) (Client, error) {
...
+		}
+
+		data := struct {
+			AccountID string
+		}{
+			AccountID: accountID,
+		}
+
+		err = tmpl.Execute(&asssumeRoleArn, data)
+		if err != nil {
```
MarcosDY, line 99: NIT: 
```
		if err := tmpl.Execute(&asssumeRoleArn, data); err != nil {
```

pkg/server/plugin/nodeattestor/aws/client.go
```diff
@@ -78,17 +82,37 @@ func (cc *clientsCache) getClient(region string) (Client, error) {
...
+			return nil, err
+		}
+	}
+
+	client, err := cc.newClient(cc.config, region, asssumeRoleArn.String())
+
```
MarcosDY, line 105: NIT: remove line

pkg/server/plugin/nodeattestor/aws/session.go
```diff
@@ -33,13 +38,28 @@ func (cfg *SessionConfig) Validate(defaultAccessKeyID, defaultSecretAccessKey st
...
+	if asssumeRoleArn != "" {
+		staticsess, err := session.NewSession(&aws.Config{Credentials: awsConf.Credentials})
+		if err != nil {
+			return nil, err
+		}
+
+		awsConf.Credentials = credentials.NewCredentials(&stscreds.AssumeRoleProvider{
+			Client:   sts.New(staticsess),
+			RoleARN:  asssumeRoleArn,
+			Duration: 15 * time.Minute,
```
MarcosDY, line 60: why 15 minutes? is it based in something?

doc/plugin_server_nodeattestor_aws_iid.md
```diff
@@ -27,6 +28,18 @@ A sample configuration:
...
+
+In the following configuration,
+```
+    NodeAttestor "aws_iid" {
+        plugin_data {
+            assume_role_arn_template = "arn:aws:iam::{{.AccountID}}:role/spire-server-delegate"
+        }
+    }
+```
+assuming AWS IID dcoument sent from the spire agent contains `accountId : 12345678`, the spire server will assume "arn:aws:iam::12345678:role/spire-server-delegate" role before making any AWS call for the node attestation. If your use case does not require dynamic AccountID in the assume role, you could hardcode the AccountID in the assume_role_arn_template.
```
MarcosDY, line 41: s/dcoument/document/g

pkg/server/plugin/nodeattestor/aws/session.go
sushil-prasad: I was using the same value as used here - https://github.com/spiffe/spire/blob/main/pkg/server/plugin/upstreamauthority/awspca/pca_client.go#L39.

As per AWS AssumeRole documentation (https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html), **The value specified can can range from 900 seconds (15 minutes) up to the maximum session duration that is set for the role. The maximum session duration setting can have a value from 1 hour to 12 hours.** 
Since this duration is not configurable, in order to make this work across all scenarios, using the lowest value does seems ok to me. 
As per go library documentation, the credentials are cached and refreshed automatically on expiration.

pkg/server/plugin/nodeattestor/aws/client.go
sushil-prasad: fixed

pkg/server/plugin/nodeattestor/aws/client.go
sushil-prasad: fixed

doc/plugin_server_nodeattestor_aws_iid.md
sushil-prasad: fixed

doc/plugin_server_nodeattestor_aws_iid.md
```diff
@@ -27,6 +28,18 @@ A sample configuration:
     }
 ```
 
+If `assume_role_arn_template` is set, the spire server will assume the role as specified by the `assume_role_arn_template` after expanding the template. Currently only template variable available is `AccountID` from the AWS IID document sent by the spire agent to the spire server.
+
+In the following configuration,
+```
+    NodeAttestor "aws_iid" {
+        plugin_data {
+            assume_role_arn_template = "arn:aws:iam::{{.AccountID}}:role/spire-server-delegate"
```
evan2645, line 37: It seems to me like the example here is going to be the 99%+ case, particularly when considering that the only template variable is account ID. Given that, I'd really prefer to see a simple role name configured here, and we can populate the rest of the ARN as needed given the account ID e.g. `assume_role = "spire-server-delegate"`. It is easier to configure, easier to document. We can always add a template configurable in the future if/when we find the use cases and reliable sources for extra template data

pkg/server/plugin/nodeattestor/aws/client.go
```diff
@@ -78,17 +82,35 @@ func (cc *clientsCache) getClient(region string) (Client, error) {
...
 
-	cc.clients[region] = client
+	cc.clients[cacheKey] = client
 	return client, nil
 }
 
-func newClient(config *SessionConfig, region string) (Client, error) {
-	sess, err := newAWSSession(config.AccessKeyID, config.SecretAccessKey, region)
+func newClient(config *SessionConfig, region string, asssumeRoleARN string) (Client, error) {
+	sess, err := newAWSSession(config.AccessKeyID, config.SecretAccessKey, region, asssumeRoleARN)
```
evan2645, line 113: What happens if we assume a role arn that we already have? E.g. attesting an agent in the same AWS account as the SPIRE Server?

evan2645 (commented):
Thanks for your patience @sushil-prasad! Let me know what you think about the comment I left on the shape of the config

pkg/server/plugin/nodeattestor/aws/client.go
sushil-prasad: I am pretty sure you will need to give additional permission to the current role to assume the same role. I will test this scenario and add the comment

evan2645: Ok great, thank you! If we can easily know the account that the code is running in, maybe it makes sense to detect this condition and not assume any role? Whatever you think makes the most sense is ok with me, just want to make sure that the case is handled well

doc/plugin_server_nodeattestor_aws_iid.md
sushil-prasad: I am ok with the suggestion. I will change the plugin config to use
**assume_role = "spire-server-delegate"**

which will be internally translated to 
**arn:aws:iam::{{.AccountID}}:role/spire-server-delegate**

pkg/server/plugin/nodeattestor/aws/client.go
sushil-prasad: If both the spire server and the spire agent is running the same account and **assume_role** parameter is set, you see following error which make sense to me. (spire-server is the role name)

> ERRO[0034] Nodeattestor(aws_iid): failed to describe instance: AccessDenied: User: arn:aws:sts::<account-id>:assumed-role/spire-server/<aws-instance-id> is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::<account-id>:role/spire-server

I am leaning towards not making the special condition for the detecting the same account. Here are some considerations:
- Right now the spire server does not know its own account. We may need to make additional AWS calls to know the current account id
- If the use case does not need multiple accounts, you do not need to configure assume_role.
- If the use case needs multiple accounts, you will need to configure other accounts with additional permissions. In that case why not do the same on all accounts and keep the AWS configuration consistent across all accounts.

doc/plugin_server_nodeattestor_aws_iid.md
sushil-prasad: Updated the code to use assume_role instead of assume_role_arn_template

pkg/server/plugin/nodeattestor/aws/client.go
evan2645: Fair enough. I think the docs capture it. We can always add something later if we need to.

evan2645 (approved):
Thanks for this @sushil-prasad!

PR #2395, opened 2021-07-02
Back out AuditLogEnabled server configurable

rturner3:
The audit log feature is still under development and should not be accessible to users until it is completed. This change is motivated by the desire to keep the main branch of SPIRE release-able, such that maintainers do not need to manipulate a set of cherry-picked commits to craft patch releases.


azdagron (approved):
Thanks, @rturner3 !

PR #2399, opened 2021-07-07
Add error log when authentication fails on spire server middleware

MarcosDY:
Current server's middleware implementation does not emit logs when authentication fails, but it return error to caller, 

So add a log line that provides some information about caller and error


azdagron (commented):
Thanks, @MarcosDY! One small comment...

azdagron (approved):
\o/

PR #2400, opened 2021-07-07
Added metrics configuration options to filter out published metrics and labels

sushil-prasad:
This is an optional setting in the spire telemetry configuration section. This will allow user to filter out published metrics based on
- AllowedLabels
- BlockedLabels
- AllowedPrefixes
- BlockedPrefixes


pkg/common/telemetry/config.go
```diff
@@ -11,7 +11,15 @@ type MetricsConfig struct {
...
 
+type MetricsConfigOptions struct {
+	AllowedPrefixes []string // A list of metric prefixes to allow, with '.' as the separator
+	BlockedPrefixes []string // A list of metric prefixes to block, with '.' as the separator
+	AllowedLabels   []string // A list of metric labels to allow, with '.' as the separator
+	BlockedLabels   []string // A list of metric labels to block, with '.' as the separator
+}
+
 type FileConfig struct {
+	Options    MetricsConfigOptions
```
evan2645, line 22: It's a little bit more work, but I think the config experience would be better if this struct was embedded in all the sink configurations, rather than adding it as a top-level config struct. Reason being that currently, all top level blocks are actual sinks. This would be the only one that's not.

We could then update the docs to have a new section `#### Universal configuration options` in which we document these, and state that they apply to all sinks.

evan2645 (commented):
thanks so much for this contribution @sushil-prasad!

I left a note about the shape of the configuration. Otherwise, this is pretty straightforward change. Please let us know what you think

pkg/common/telemetry/config.go
sushil-prasad: make sense. I will make the changes

sushil-prasad: I tried to make the suggested change and it seems difficult. 

The problem is that there is one metrics for multiple sinks. Some of the sinks are also array. If we push down the these options to the sink definitions, we may run into potential conflicting definitions of metrics options. Currently these options are processed independent of the underlying sink. So it is best to keep these options outside the sink definition in order to avoid the conflicting definitions.

So we have two options for implementation:
1> Keep the options parallel to sync definitions (this is the current implementation)
2> Move the options out of the telemetry object and make it parallel to the telemetry in the main config file. This implementation also does not look very good in terms of shape of config file.

Let me know what you think about this.

evan2645: > The problem is that there is one metrics for multiple sinks. Some of the sinks are also array. If we push down the these options to the sink definitions, we may run into potential conflicting definitions of metrics options. Currently these options are processed independent of the underlying sink. So it is best to keep these options outside the sink definition in order to avoid the conflicting definitions.

Hmm ok yes, that makes sense... Thanks for trying, sorry to have sent you down that road.

How about a third option, which is have the configurables as fields on `FileConfig` (i.e. top level to telemetry config rather than in an `Options` block)? That way they apply at the right level, and have a much different shape than the sink configs

sushil-prasad: Do you mean this

sushil-prasad: Do you mean the following structure 

type FileConfig struct {
         Prometheus *PrometheusConfig    `hcl:"Prometheus"`
	DogStatsd  []DogStatsdConfig    `hcl:"DogStatsd"`
	Statsd     []StatsdConfig       `hcl:"Statsd"`
	M3         []M3Config           `hcl:"M3"`
	InMem      *InMem               `hcl:"InMem"`

	AllowedPrefixes []string `hcl:"AllowedPrefixes"` // A list of metric prefixes to allow, with '.' as the separator
         BlockedPrefixes []string `hcl:"BlockedPrefixes"` // A list of metric prefixes to block, with '.' as the separator
         AllowedLabels   []string `hcl:"AllowedLabels"`   // A list of metric labels to allow, with '.' as the separator
         BlockedLabels   []string `hcl:"BlockedLabels"`   // A list of metric labels to block, with '.' as the separator

         UnusedKeys []string `hcl:",unusedKeys"`
}

And the corresponding configuration file will look like this
telemetry {
    DogStatsd = [
        { address = "127.0.0.1:8126" },
    ]
    AllowedPrefixes=["AllowedPrefixes"]
    BlockedPrefixes=["BlockedPrefixes"]
    AllowedLabels=["AllowedLabels"]
    BlockedLabels=["BlockedLabels”]
}

This shape seems reasonable to me. @evan2645

evan2645: Yes that's exactly what I meant :) thank you!!

sushil-prasad: cool. I have pushed the code changes @evan2645

evan2645 (approved):

Issue #1975, opened 2020-11-24
[RFC] Regit


prasadborole1:
**Background:**
SPIRE implements the concept of “admin” workloads. A workload presenting an admin SVID is able to act broadly across the entire registration API. In an ecosystem where multiple registrants are creating/deleting registrations, it results in registrants being able to affect registrations created by other registrants. Therefore, we need a way to limit the access of registration APIs to subsets of SPIFFE IDs by the registrant. More background discussion can be found in the github issue ​[#716](https://github.com/spiffe/spire/issues/716)​.

**Possible solutions:**
1. **Plug-in-based approach** (Registration authorizer):
The registration authorizer plug-in would be invoked in the path of create, update and
delete registration RPCs. This will allow plug-in to get details of requestor identity, action to be performed and subject of the action. It will return a boolean response denoting whether requestor identity is authorized to perform action on registration with specific SPIFFE ID present in the RPC request.
We currently have a plug-in “Notifier” which gets invoked after each bundle update call. The registration authorizer plug-in would behave in a similar way. Since both of these plug-ins are intended to be used for very different purposes, I’m proposing to create a new one instead of overloading existing “Notifier” plug-in.
This approach offers more flexibility as users can potentially implement their own authorization model depending on different use cases or can integrate with any other external authorization solution.
By default we will have a Noop plug-in which will always return true to ensure backward compatibility.

2. **OPA-based config**:
In this approach, we make changes to SPIRE core to include OPA-based config which will define admin registration authorization. The basic implementation can be found ​here​.
This approach would make authorization a first class citizen of SPIRE which brings both advantages and challenges. All the SPIRE users will get authorization by default but it means arriving at consensus on the basic permission model so that most customer use cases are
 addressed. Also, there will be challenges around compatibility when modifying the permission model while the feature is settling in the production environment.
Sample Implementation of Registration Authorizer plug-in:
A sample permission model for registration RPCs in the plug-in could look like:
Actor​ ​:​ ​Role, Role​ ​:​ ​[{
Actions: ​[], Resources: ​[], Namespaces: ​[],
​}, ​{
​}]
Example:
spiffe:​//example.ca/foo/workload:​ ​foobarRegistrant, foobarRegistrant: ​[​ ​{
Actions: ​["CREATE",​ ​"UPDATE",​ ​"DELETE"]
Resources: ​[“REGISTREATIONS”],
Namespaces: ​["spiffe://example.ca/foo",​ ​"spiffe://example.ca/bar"]
[...]

PR #2416, opened 2021-07-13
Add OPA policy for SPIRE API authorization

lumjjb:
**Pull Request check list**

- [x] Commit conforms to CONTRIBUTING.md? 🤞 
- [ ] Documentation updated? Want some feedback on this!
 
**Affected functionality**

Changes SPIRE authorization mechanics to use OPA instead. Allows extensibility of authorization policies to be written in rego, allowing more granular policy decisions. This can be done by adding the appropriate rego files and JSON data via SPIRE server config. If unconfigured, defaults to the standard SPIRE authorization policy.

**Description of change**

Changed the middleware authorization preprocessing to call out to OPA instead, part of this makes the old authorization code obsolete since it calls out to OPA for the decisions, thus removes the old authz code. The main changes of this PR are the server authorization middleware, the server endpoint as well as the `common/policy`

**Which issue this PR fixes**

This fixes https://github.com/spiffe/spire/issues/1975


lumjjb:
@APTy can I add the DCO for your commit on your behalf? :)

azdagron:
Looks like there are bugs in the downstream authorization logic.

> 2021/07/13 16:47:58 RPC "NewDownstreamX509CA": failed to call NewDownstreamX509CA: rpc error: code = Internal desc = caller is not a downstream workload

lumjjb:
Thanks @azdagron , i think i see what the issue is now... that error message is from https://github.com/spiffe/spire/blob/main/pkg/server/api/svid/v1/service.go#L321-L324

Which seems to gather the downstream entries from the context... I think this was implicitly done in the old authorization code, I just added code to do that in the authorization. So 🤞 it solves it..

I have some thoughts on this, I'd like to get your feedback... but all condition on whether this is the actual cause of the failure.... (so we can wait for the CI first before discussing further, if not we can ignore the following :D, and probably comes back to me going back to the drawing board...)

However, I do have a question on that, if this is the explicit behavior of the authorization middleware, then perhaps I should add some tests to check for this as well... However, this gets a bit puzzling because there is a reliance on the call to authorization... Which in the case of an authorization policy of "allow all", would not result in this context being set... This is true of the old implementation as well. Which makes me think whether this is something that should be addressed more generally as a preprocessor? or not explicitly relying on `rpccontext.CallerDownstreamEntries(ctx)`, and checking if it doesn't have any entries explicitly set...

lumjjb:
Updated with docs

pkg/common/policy/policy_examples.go
```diff
@@ -0,0 +1,349 @@
+package policy
+
+const defaultPermissionData = `
```
azdagron, line 3: It's a little weird for server-specific defaults to live under `common`. I wonder if there shouldn't be a `pkg/server/policy` package that has a `NewEngineOrDefault()` method that either constructs an engine with the provided config or returns an engine with the default policy loaded. That would also get rid of my concern about NewEngine returning `nil`.

pkg/common/policy/policy_examples.go
```diff
@@ -0,0 +1,349 @@
...
+{
+    "apis": [
```
azdagron, line 5: These defaults are now listed in three spots (MD, conf/server/permissions.json, and here). I think we should try and consolidate them into one spot. One solution is to create the `pkg/server/policy` package, include the default files in that location, and embed them into the binary using the `go:embed` directive (https://pkg.go.dev/embed). The markdown could then reference that file directly and we only have one spot to update things.

As I mentioned elsewhere, I don't think the "developer" conf under conf/ should configure the policy engine since most users will probably use the default.

pkg/common/policy/policy.go
```diff
@@ -0,0 +1,170 @@
...
+		return result, err
+	}
+
+	// TODO(tjulian): figure this out
+	if len(rs) == 0 || len(rs[0].Expressions) == 0 {
+		return result, errors.New("policy: no matching policies found")
+	}
+
+	exp := rs[0].Expressions[0]
+	resultMap := exp.Value.(map[string]interface{})
```
azdagron, line 141: Is the value guaranteed to be of this type? If not, we should use the comma-ok pattern on the type coercion to prevent a panic and return an error instead.

pkg/common/policy/policy.go
```diff
@@ -0,0 +1,170 @@
...
+
+	var ok bool
+	result.Allow, ok = resultMap["allow"].(bool)
```
azdagron, line 144: nit: `"allow"` and friends should probably be constants.

pkg/common/policy/policy.go
```diff
@@ -0,0 +1,170 @@
...
+	if !ok {
+		return result, errors.New("policy: result did not contain \"allow\" bool value")
+	}
+
+	result.AllowIfAdmin, ok = resultMap["allow_if_admin"].(bool)
+	if !ok {
+		return result, errors.New("policy: result did not contain \"allow_if_admin\" bool value")
+	}
```
azdagron, line 152: This repetition could be cleaned up by defining a function-local helper:
```suggestion
	getBoolValue := func(name string) (bool, error) {
		value, ok := resultMap[name].(bool)
		if !ok {
			return false, fmt.Errorf("policy: result did not contain %q bool value", name)
		}
		return value, nil
	}

	if result.AllowIfAdmin, err = getBoolValue(allowIfAdminKey); err != nil {
		return Result{}, err
	}
```

pkg/common/policy/policy.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func (e *Engine) testPolicy() error {
+	ctx := context.Background()
+	for _, i := range sampleInputs {
+		var inp Input
+		err := json.Unmarshal([]byte(i), &inp)
+		if err != nil {
+			return err
+		}
```
azdagron, line 118: nit:
```suggestion
		if err := json.Unmarshal([]byte(i), &inp); err != nil {
			return err
		}
```

pkg/common/policy/policy.go
```diff
@@ -0,0 +1,170 @@
...
+
+		_, err = e.Eval(ctx, inp)
+		if err != nil {
+			return errors.New("policy initialization: issue testing configured policy")
+		}
```
azdagron, line 123: nit:
```suggestion
		if _, err := e.Eval(ctx, inp); err != nil {
			return fmt.Errorf("policy is misconfigured: %v", err)
		}
```

pkg/common/policy/policy.go
```diff
@@ -0,0 +1,170 @@
...
+	AllowIfDownstream bool `json:"allow_if_downstream"`
+	AllowIfAgent      bool `json:"allow_if_agent"`
+}
+
+// NewEngine returns a new policy engine. Or nil if no
+// config is provided.
+func NewEngine(cfg *EngineConfig) (*Engine, error) {
+	if cfg == nil || cfg.FileProvider == nil {
+		return nil, nil
+	}
```
azdagron, line 57: Returning `nil` from a function named `NewEngine` is a bit weird IMO. I'd rather see this function not called if config is empty. Also, IMO, if the config is provided but doesn't have a valid provider set (i.e. FileProvider) that should probably be a failure instead of a silent ignore?

pkg/common/policy/policy.go
```diff
@@ -0,0 +1,170 @@
+package policy
```
azdagron, line 1: This package seems fairly server-centric, i.e. in the inputs and outputs and tests that are performed. I wonder if some of it belongs in a server policy package. Since we don't have any agent rego plans yet, maybe this should all be defined under server and then the common code refactored out later if/when needed....

azdagron (commented):
Thanks for opening this, @lumjjb ! I haven't completely reviewed all of the added documentation, but this should be enough to keep you busy while I wrap it up :)

pkg/common/policy/policy_examples.go
lumjjb: I am all for using embed.. is there any concern with the toolchain, i think `embed` is a fairly new golang feature - just checking if this is an issue for broader use cases or CI

>As I mentioned elsewhere, I don't think the "developer" conf under conf/ should configure the policy engine since most users will probably use the default.

yea agree - can change this to not set it and let it use the default.

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,387 @@
...
+
+If the policy engine configuration is not set, it defaults to the default SPIRE
+authorization policy, which has the following ruleset as expressed in
+this [rego policy](/conf/server/policy.rego) and [permissions data
+binding](/conf/server/permissions.json). These encode the following behavior:
+
+
+| API | Authorized if (OR) |
+| ------------- | ----------- |
+| /spire.api.server.svid.v1.SVID/MintX509SVID | local or admin|
```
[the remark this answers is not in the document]
lumjjb: fixed!

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,387 @@
...
+| /spire.api.server.agent.v1.Agent/RenewAgent | agent |
+| /spire.api.server.agent.v1.Agent/CreateJoinToken | local, admin |
+| /grpc.health.v1.Health/Check | local |
+| /grpc.health.v1.Health/Watch | local |
+
+If multiple are specified in the authorized column, it is an disjunction of the
+listed options. Where the options are:
+- local: local caller through UNIX socket
+- admin: caller has a SPIFFE ID with an "Admin" flag
+- downstream: caller is a downstream SPIFFE ID
```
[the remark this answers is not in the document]
lumjjb: fixed in commit

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,387 @@
...
+- `data`: the databinding from the databinding file (permissions file)
+
+| input field   | Description | Example |
+| ------------- | ----------- | ------- |
+| caller        | The SPIFFE ID (if available) of the caller | spiffe://example.org/workload1 |
+| full_method   | The full method name of the API call | /spire.api.server.svid.v1.SVID/MintJWTSVID |
+| req           | The API call request body | { "filter": {} } |
+
+The request (`req`) is the marshalled JSON object from the [SPIRE
+api sdk](https://github.com/spiffe/spire-api-sdk/).
```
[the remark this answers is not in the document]
lumjjb: Added text for this in table as well as sentence

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,387 @@
...
+
+## Permissions file (databinding)
+
+The permissoins file consists of a JSON blob which represents the data that is
```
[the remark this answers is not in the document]
lumjjb: fixed

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,387 @@
...
+can be used in the policy in rego to say that if the input's full method is
+equal to one of the objects defined in the `apis` fields' `full_method` 
+sub-field, then `allow` should be set to true.
+```
+allow = true {
+    input.full_method == data.apis[_].full_method
+}
+```
+
+### Default configurations
```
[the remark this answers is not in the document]
lumjjb: Agreed, added link to file paths instead.

lumjjb:
@azdagron all the comments should be addressed.

azdagron:
Awesome, thanks @lumjjb. I'll try and get to a review either tonight or tomorrow morning.

pkg/server/endpoints/middleware.go
```diff
@@ -41,56 +42,6 @@ func Middleware(log logrus.FieldLogger, metrics telemetry.Metrics, ds datastore.
...
-	any := middleware.AuthorizeAny()
-	local := middleware.AuthorizeLocal()
-	agent := middleware.AuthorizeAgent(agentAuthorizer)
-	downstream := middleware.AuthorizeDownstream(entryFetcher)
-	admin := middleware.AuthorizeAdmin(entryFetcher)
-
-	localOrAdmin := middleware.AuthorizeAnyOf(local, admin)
-	localOrAdminOrAgent := middleware.AuthorizeAnyOf(local, admin, agent)
-
-	return map[string]middleware.Authorizer{
```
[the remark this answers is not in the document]
lumjjb: Yes - these are covered by https://github.com/spiffe/spire/blob/main/pkg/server/endpoints/endpoints_test.go#L380-L445

Squashed out a few bugs while developing with these tests :)

pkg/server/endpoints/endpoints.go
```diff
@@ -118,6 +120,14 @@ func New(ctx context.Context, c Config) (*Endpoints, error) {
...
 	}
 
+	pe := c.PolicyEngine
+	if pe == nil {
+		pe, err = policy.DefaultAuthPolicy()
+		if err != nil {
+			return nil, err
+		}
+	}
+
```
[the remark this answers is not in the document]
lumjjb: changes made.

pkg/server/policy/policy.go
```diff
@@ -0,0 +1,190 @@
...
+	AllowIfDownstream bool `json:"allow_if_downstream"`
+	AllowIfAgent      bool `json:"allow_if_agent"`
+}
+
+// NewEngineOrDefault returns a new policy engine. Or if no
+// config is provided, provides the default policy
+func NewEngineOrDefault(cfg *EngineConfig) (*Engine, error) {
+	if cfg == nil || cfg.FileProvider == nil {
+		return DefaultAuthPolicy()
+	}
```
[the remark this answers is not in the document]
lumjjb: changes made.

pkg/server/policy/policy.go
```diff
@@ -0,0 +1,190 @@
...
+
+		if _, err := e.Eval(ctx, inp); err != nil {
+			return fmt.Errorf("policy is misconfigured: %w", err)
+		}
+	}
+	return nil
+}
+
+// Eval determines whether access should be allowed on a resource.
+func (e *Engine) Eval(ctx context.Context, input Input) (result Result, err error) {
```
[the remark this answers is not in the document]
lumjjb: fixed

pkg/server/policy/policy.go
```diff
@@ -0,0 +1,190 @@
...
+	rs, err := e.rego.Rego(rego.Input(input)).Eval(ctx)
+	if err != nil {
+		return result, err
+	}
+
+	// TODO(tjulian): figure this out
```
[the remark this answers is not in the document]
APTy: Heh, my guess would be that the `Eval()` call returns a slice of [rego.Result](https://pkg.go.dev/github.com/open-policy-agent/opa/rego#Result) objects, each of which contains a slice of ExpressionValues.

So in this implementation, we're only looking at the **first** result and the **first** expression value within that result, which means we're potentially throwing out extra information that could be useful for policy evaluation or logging purposes. Maybe there could be "deny" style rules to be grabbed, or different reasons for "allow", I can't remember exactly.

So it could be worth looking into what the rest of those result/expressions could contain, or leave that as a (more detailed) TODO for later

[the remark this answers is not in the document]
lumjjb: Ah i see, since then we've introduced a contract between the policy and the authorization component, specifying specific variables that must be set, and only those variables are queried. These are listed [here](https://github.com/lumjjb/spire/blob/policy_opa/doc/authorization_policy_engine.md#rego-policy). I think this addresses mostly the concerns from this TODO, and like you said we can add on to this list later on.

lumjjb:
@azdagron Ready for another round!

pkg/server/policy/validate.go
```diff
@@ -0,0 +1,144 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+// validatePolicy runs a few sample inputs with the policy just to make sure/
+// it doesn't throw any errors
+func (e *Engine) validatePolicy() error {
```
azdagron, line 11: This should have the context threaded through from NewEngineFromRego.

lumjjb: Ah! One slipped through the cracks. Fixed!

pkg/server/server.go
```diff
@@ -122,7 +123,12 @@ func (s *Server) run(ctx context.Context) (err error) {
 		return err
 	}
 
-	endpointsServer, err := s.newEndpointsServer(ctx, cat, svidRotator, serverCA, metrics, caManager)
+	policyEngine, err := policy.NewEngineOrDefault(ctx, s.config.PolicyEngineConfig)
+	if err != nil {
+		return err
```
amartinezfayo, line 128: We should format this error to provide an indication of the context, otherwise it may be difficult to guess that the error happened while creating a new authorization policy engine.

pkg/server/policy/policy.go
```diff
@@ -0,0 +1,170 @@
+package policy
```
amartinezfayo, line 1: I'm wondering if the directory name `policy` is specific enough, thinking that we may end up having other kind of _policies_ in the future that may have a collision with this. Maybe `authpolicy` sounds better to me. @azdagron Thoughts?

pkg/server/policy/policy.go
```diff
@@ -0,0 +1,170 @@
...
+func NewEngineOrDefault(ctx context.Context, cfg *EngineConfig) (*Engine, error) {
+	if cfg == nil {
+		return DefaultAuthPolicy(ctx)
+	}
+	return NewEngine(ctx, cfg)
+}
+
+// NewEngine returns a new policy engine. Or nil if no
+// config is provided.
+func NewEngine(ctx context.Context, cfg *EngineConfig) (*Engine, error) {
```
amartinezfayo, line 69: This function is not exercised in the tests. I think that would be good to have it covered.

pkg/server/policy/policy_test.go
```diff
@@ -0,0 +1,262 @@
...
+			require.Nil(t, err, "failed to unmarshal data JSON")
+
+			ctxIn := context.Background()
+			store := inmem.NewFromObject(json)
+			ctx := context.Background()
+
+			pe, err := policy.NewEngineFromRego(ctx, tt.rego, store)
+			require.Nil(t, err, "failed to create policy engine")
+
+			res, err := pe.Eval(ctxIn, tt.input)
```
amartinezfayo, line 223: Looks like the test cases cover only happy paths when executing `Eval`. I think that it would be good to have coverage of failures also.

amartinezfayo (commented):
Thank you @lumjjb for this contribution! It's exciting to see this coming!
I have some comments.

pkg/server/policy/policy.go
azdagron: authpolicy seems nice :)

lumjjb:
Thanks for the review @amartinezfayo ! Got tied up with some stuff but will look at this tomorrow!

doc/spire_server.md
```diff
@@ -68,6 +68,7 @@ This may be useful for templating configuration files, for example across differ
 | `ratelimit`                 | Rate limiting configurations, usually used when the server is behind a load balancer (see below)  |                                                                |
 | `socket_path`               | Path to bind the SPIRE Server API socket to                                                       | /tmp/spire-server/private/api.sock                             |
 | `trust_domain`              | The trust domain that this server belongs to (should be no more than 255 characters)              |                                                                |
+| `policy_engine`             | The policy engine used to determine authorization decision                                        | default SPIRE authorization policy                             |
```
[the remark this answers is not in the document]
lumjjb: Agreed on the comments here! Will make these changes... only 1 question on one comment..

> server_full.conf needs to be updated with the new configs.

I think @azdagron had another comment from before on this that we may want to default to not setting it - thus using the default policy? This was to prevent replication of the policy in multiple files. Thoughts on the direction with this?

doc/spire_server.md
```diff
@@ -84,6 +85,17 @@ This may be useful for templating configuration files, for example across differ
 | `attestation`               | Whether or not to rate limit node attestation. If true, node attestation is rate limited to one attempt per second per IP address. | true |
 | `signing`                   | Whether or not to rate limit JWT and X509 signing. If true, JWT and X509 signing are rate limited to 500 requests per second per IP address (separately). | true |
 
+| policy_engine               | Description                    | Default        |
+|:----------------------------|--------------------------------|----------------|
+| `file_provider`             | File to retrieve OPA rego policy for authorization. |      |
```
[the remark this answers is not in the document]
lumjjb: I think it's possible that this may also point to an external OPA endpoint, so there's that possibility that instead of a file it would point to a server with some connection configuration.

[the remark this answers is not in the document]
lumjjb: not sure about other alternatives to OPA, i think there is a realm of possibility but not sure what that would be today... Am in favor of adding an OPA specific config in this case.

pkg/server/policy/policy.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	e := &Engine{
+		rego: pr,
+	}
+
+	// Test policy with some simple calls to ensure that the
+	// policy can be evaluated properly.
+	if err := e.validatePolicy(ctx); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
lumjjb: fixed

pkg/server/server.go
lumjjb: fixed

pkg/server/api/middleware/authorization_opa.go
```diff
@@ -0,0 +1,130 @@
...
+		if res.AllowIfDownstream {
+			if ctx, ok := isDownstream(ctx, entries); ok {
+				return ctx, true, nil
+			}
+		}
+	}
+
+	if res.AllowIfAgent {
+		ctx, isAgent, err := isAgent(ctx, agentAuthorizer)
+		if err == nil && isAgent {
```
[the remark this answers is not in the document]
lumjjb: Yea good catch.

Just to provide some context for when I was writing this.. 

`isAgent` may return an error if it can't figure out if its an agent, but if an authorization policy says please allow "if agent" or "if local" then this call to `isAgent` would error.. so the order of operations is what really "saves" it here. 

Maybe... the way to address this is changing `isAgent` such that it will not return an error if caller does not have an `x509`?

doc/spire_server.md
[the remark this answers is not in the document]
lumjjb: current updated this to look like:

```
auth_policy_engine {
    opa_file_provider {
        ...
    }
}
```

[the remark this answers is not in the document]
amartinezfayo: >I think @azdagron had another comment from before on this that we may want to default to not setting it - thus using the default policy? This was to prevent replication of the policy in multiple files. Thoughts on the direction with this?

@lumjjb `server_full.conf` is for documentation purposes only. It has all the possible config settings with comments so users can look at the different configurables in SPIRE. It doesn't look like the comment from @azdagron applies to that, right?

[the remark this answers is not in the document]
lumjjb: ah i see - yea i got mixed up. Yea that works.

doc/spire_server.md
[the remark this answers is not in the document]
lumjjb: renamed to `opa_file_provider`

pkg/server/policy/policy.go
lumjjb: Yup - created tests for `NewEngineFromConfigOrDefault`, and made this function private, since it is just a helper. Should encapsulate all the edge cases for `NewEngine`.

pkg/server/policy/policy_test.go
lumjjb: Yup - the way its set up, it's designed so that once a policy engine is successfully created, it should by really hard for `Eval` to actually return an error (I tried a lot of things, including divide by zeros....).. it seems like mainly err returned would be because of a system failure.. This is because of the `vallidatePolicy` function catches a lot of the errors that can happen during policy instantiation, so i've added the case where Eval would fail in the instantiation tests here instead:

https://github.com/lumjjb/spire/blob/policy_opa/pkg/server/authpolicy/policy_test.go#L418-L427

lumjjb:
@amartinezfayo I think I should've addressed all of the comments.. 🤞 . Thanks for the great comments!

doc/spire_server.md
```diff
@@ -68,6 +68,7 @@ This may be useful for templating configuration files, for example across differ
 | `ratelimit`                 | Rate limiting configurations, usually used when the server is behind a load balancer (see below)  |                                                                |
 | `socket_path`               | Path to bind the SPIRE Server API socket to                                                       | /tmp/spire-server/private/api.sock                             |
 | `trust_domain`              | The trust domain that this server belongs to (should be no more than 255 characters)              |                                                                |
+| `auth_policy_engine`        | The [auth policy engine](doc/authorization_policy_engine.md) used for authorization decision      | default SPIRE authorization policy                             |
```
amartinezfayo, line 71: This document is already in `/doc` so a relative reference to `doc` will link to /doc/doc.

```suggestion
| `auth_policy_engine`        | The [auth policy engine](/doc/authorization_policy_engine.md) used for authorization decisions      | default SPIRE authorization policy                             |
```

amartinezfayo (commented):
Thanks @lumjjb! One more small comment.

lumjjb:
Committed the suggestion! let me know if you'd like me to rebase the PR

Benardi:
Minor detail, but the documentation at `doc/authorization_policy_engine.md` still points to `pkg/server/policy/policy.rego` and  `pkg/server/policy/permissions.json` instead of  `pkg/server/authpolicy/policy.rego` and `pkg/server/authpolicy/permissions.json`

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,224 @@
...
+    admin_entries := { entry | entry := input.req.entries[_]; entry.admin == true}
+    count(admin_entries) == 0
+}
+
+check_entry_create_admin_flag {
+    input.full_method != "/spire.api.server.entry.v1.Entry/BatchCreateEntry"
+}
+```
+This sets `check_entry_create_admin_flag` to true if the full method is not for
+entry creation or if it is, that there are no entires that contain the admin
```
amoore877, line 213: typo: `entires` in a few places

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,224 @@
...
+| allow_local       | if true, sets result.allow_if_local to true | |
+| allow_admin       | if true, sets result.allow_if_admin to true | |
+| allow_downstream  | if true, sets result.allow_if_downstream to true | |
+| allow_agent       | if true, sets result.allow_if_agent to true | |
+
+# Extending the policy
+
+This section contains examples of how the authorization policy can be extended.
+
+## Example 1: Entry creation namespacing restrictions
```
amoore877, line 141: `Entry creation namespacing restrictions` is a great example to have; I think for the complexity of authorization policies it would be helpful to also add examples for:
- a "subdepartment" permissions split. Say `schedulers/finance` is able to create `/finance` and also `schedulers/finance/EMEA` is able to create `/finance/EMEA`.
- ^Are we able to also create such that `schedulers/finance` can still create `/finance` but is excluded from creating `/finance/EMEA`, or is that not possible?
- having a delete or update example may also be helpful, just so examples aren't restricted to single API

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,224 @@
...
+}
+```
+
+The fields of the result are the following:
+- `allow`: a boolean that if true, will authorize the call
+- `allow_if_local`: a boolean that if true, will authorize the call only if the
+  caller is a local UNIX socket call
+- `allow_if_admin`: a boolean that if true, will authorize the call only if the
+  caller is a SPIFFE ID with the Admin flag set
+- `allow_if_downstream`: a boolean that if true, will authorize the call
```
evan2645, line 60: I think we're missing `allow_if_agent` ?

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,224 @@
...
+result = {
+  "allow": true/false,
+  "allow_if_admin": true/false,
+  "allow_if_local": true/false,
+  "allow_if_downstream": true/false,
+  "allow_if_agent": true/false,
+}
+```
+
+The fields of the result are the following:
```
evan2645, line 54: What happens if more than one of the fields is true? What if both `allow` and `allow_if_agent` are true?

evan2645 (commented):
This is great, thank you so much @lumjjb! We're way overdue for this

I had a bunch of thoughts about the shape and naming of the configs, but I'm happy to circle back around in a follow-up PR with a proposal so as not to block on it here. I think the only _real_ thing is the docs question about edge case on the response object. Also missing the `allow_if_agent` bit in the docs ... other than that, lgtm

conf/server/server_full.conf
```diff
@@ -102,6 +102,19 @@ server {
...
+    # auth_policy_engine: The auth policy engine used for authorization decision
+    # For more details, refer to doc/authorization_policy_engine.md
+    # auth_policy_engine {
+    #     # Indicates that OPA provided through local files be used to 
+    #     # instantiate the auth policy engine
+    #     opa_file_provider {
+    #         # Path to the policy rego file
+    #         policy_path = "./conf/server/policy.rego"
+    #         # Path to the permissions JSON databindings
+    #         permissions_path = "./conf/server/permissions.json"
```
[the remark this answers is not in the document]
lumjjb: Yea - I agree very much on this - initial use of this was only for permissions, or at least the common usecase is to encapsulate permissions. Yea I think i will comb through and change it to `policy_data_path`.

conf/server/server_full.conf
```diff
@@ -102,6 +102,19 @@ server {
     # Format of logs, <text|json>. Default: text.
     # log_format = "text"
 
+    # auth_policy_engine: The auth policy engine used for authorization decision
+    # For more details, refer to doc/authorization_policy_engine.md
+    # auth_policy_engine {
+    #     # Indicates that OPA provided through local files be used to 
+    #     # instantiate the auth policy engine
+    #     opa_file_provider {
```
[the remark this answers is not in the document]
lumjjb: >The fact that we are using OPA internally feels like a little bit of an implementation detail... is it "just" rego evaluation, or are there OPA-specific things here? Would it make sense for the name of this provider to be something about rego instead?

I think rego is really only tied to only OPA right now, the evaluation semantics are still all OPA specific, i.e. handling of errors, execution order, etc. So I think in this case "opa" may be a more direct translation of what's happening since it encapsulates both the static and dynamic semantics. 

>It would be nice if the name was a bit shorter too ... the provider suffix is probably redundant, perhaps something like rego_file or opa_file
>I imagine that if we had rego_file, then we may in the future have something like local_opa etc?

Yea, I really like `local_opa` even now, since most likely the next extension to this is `remote_opa` where policy can be defined in a remote server. What do you think about this?

doc/authorization_policy_engine.md
lumjjb: good catch - fixed

doc/authorization_policy_engine.md
lumjjb: ah yea - forgot to include this, copying this from the conversation in the issue.

conf/server/server_full.conf
```diff
@@ -102,6 +102,19 @@ server {
     # Format of logs, <text|json>. Default: text.
     # log_format = "text"
 
+    # auth_policy_engine: The auth policy engine used for authorization decision
+    # For more details, refer to doc/authorization_policy_engine.md
+    # auth_policy_engine {
```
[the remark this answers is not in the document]
lumjjb: Addressing the other comment together with this... Maybe in that case, perhaps we can just call it

Hmm... only thing that comes to mind here is around when one OPA policy engine fails to evaluate, then there's a fallback policy engine. i.e. remote OPA policy server is down, default back to local OPA policy. 

`auth_opa_policy_engine` and properties of it would be `local` 

```
auth_opa_policy_engine = {
  local {
    policy_path = xyz
    policy_data_path = xyz
  }
}
```

doc/authorization_policy_engine.md
lumjjb: good idea - let me add some other examples that cover these.. 

>^Are we able to also create such that schedulers/finance can still create /finance but is excluded from creating /finance/EMEA, or is that not possible?

it should be possible.. it would just a bit more complicated to express. It may be a difficult example to follow, let me mock up some examples so we can take a look

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+The rego policy can then be updated to check for this, an example of an allow
+clause would look like the following. Note that it is important to check to see
+how this fits in with the other parts of the rego policy.
+
+```
+# Any allow check
+allow = true {
+    check_entry_create_admin_flag
```
Benardi, line 202: Shouldn't it be `check_entry_create_namespace`?

I'm met with the following error if I use this line as it stands, which I surmise might arise from `check_entry_create_namespace` not being defined:

```bash
O[0000] Fatal run error                               error="unable to obtain authpolicy engine: 1 error occurred: spire.rego:97: rego_unsafe_var_error: var check_entry_create_admin_flag is unsafe"
ERRO[0000] Server crashed                                error="unable to obtain authpolicy engine: 1 error occurred: spire.rego:97: rego_unsafe_var_error: var check_entry_create_admin_flag is unsafe"
```

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,235 @@
...
+| allow_local       | if true, sets result.allow_if_local to true | |
+| allow_admin       | if true, sets result.allow_if_admin to true | |
+| allow_downstream  | if true, sets result.allow_if_downstream to true | |
+| allow_agent       | if true, sets result.allow_if_agent to true | |
+
+# Extending the policy
+
+This section contains examples of how the authorization policy can be extended.
+
+## Example 1: Entry creation namespacing restrictions
```
Benardi, line 152: I've tried to follow this example. I've built the server and agent using `make build` after editing `policy.rego` and `permissions.json`. 

As a test, I've done the following:
*  I've created an agent with the following SPIFFE ID:  **spiffe://example.org/schedulers/finance**
* I then created the following entry
  + SPIFFE ID **//example.org/schedulers/hr**
   + Parent ID **spiffe://example.org/schedulers/finance**

I managed to create the entry and even receive the corresponding SVID from the Workload API (via [ghostunnel](https://github.com/ghostunnel/ghostunnel)).
Shouldn't this have failed?

lumjjb: Hey Benardi! Let's chat on slack see whether we can figure out what the issue is - or whether my OPA code has a bug

doc/authorization_policy_engine.md
lumjjb: yes - you are right, i copy-pasted and forgot to change 😓

doc/authorization_policy_engine.md
Benardi: Sure thing.

Benardi: So far, we've established that as I was using `spire-server entry create`, the call to BatchCreateEntry API must have been marked as local, and thus namespacing was bypassed.

I still have to surmise a way to make a non-local call to the BatchCreateEntry API to discern whether the namespacing will be enforced.

lumjjb:
@evan2645 since there's a couple of comments around the config, going to summarize the different comments and replies in 1 place for easier discussion, I am in agreement with all of them and want to propose a new config format for discussion. Let me know what you think about the new proposed config/comments and I can make the changes.


- If we are accommodating future support for other authorization engines, would a  plugin would be more appropriate - possibly in the future?
  - Yep agreed, if we get more, agree that plugin would be more appropriate. No current foreseeable plans to have other authorization engines for now.
- can there be more than one policy engine set? The shape of this config makes it feel that way. It's hard for me to imagine the uses cases for multiple policy engines, and what the expected behavior might be if we get different answers etc ... should we try to shape the config so that it's clear that there can be only one?
  - Yea - agree on having multiple policy engines isnt that useful. The only scenario that i see this useful for example is in OPA if we have a remote and a local option, then we can. set some behavior where if remote fails to query, fall back to a fail secure local OPA policy.
- The fact that we are using OPA internally feels like a little bit of an implementation detail... is it "just" rego evaluation, or are there OPA-specific things here? 
  - Rego is pretty much tied to OPA right now, but OPA encapsulates the runtime semantics as well (i.e. how the rego is parsed and how errors are handled), so i think would be more precise to classify it as "opa". (and the fact that it takes in a data binding as well)
- It would be nice if the name was a bit shorter too ... the provider suffix is probably redundant, perhaps something like rego_file or opa_file.I imagine that if we had rego_file, then we may in the future have something like local_opa etc?
  - Changing this to local
- data bindings could be many things right? Not just permissions?
  - Yup - changed to policy_data_path

## ** Proposed config ** - that hopefully addresses all the comments
```
auth_opa_policy_engine = {
  local {
    rego_path = xyz
    policy_data_path = xyz
  }
}
```

doc/authorization_policy_engine.md
lumjjb: Doc updated with some more examples!

lumjjb:
@amoore877 @azdagron @evan2645 @amartinezfayo Ok! I think i've addressed all the comments! 🤞

azdagron:
The bit I was worried about (in terms of behavior regression) has been addressed. I'll leave it up to @evan2645 to comment on the shape of the configuration :)

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,374 @@
...
+
+# Extending the policy
+
+This section contains examples of how the authorization policy can be extended.
+
+## OPA Warning
+
+It is important when implementing custom policies that one understands the 
+evaluation semantics and details of OPA rego. An example of subtleties of OPA
+rego policy is the evaluation of a variable is taken as an logical OR of all
```
amoore877, line 157: nit (please don't bother if no other changes being made): `as an logical OR` -> `as a logical OR`

doc/authorization_policy_engine.md
```diff
@@ -0,0 +1,374 @@
...
+how this fits in with the other parts of the rego policy.
+
+```
+# Any allow check
+allow = true {
+    check_entry_create_namespace
+}
+```
+
+## Example 1b: Sub-department namespacing with exlcusions
```
amoore877, line 220: Thanks for adding!

pkg/server/api/middleware/authorization.go
```diff
@@ -59,18 +51,18 @@ func (m *authorizationMiddleware) Preprocess(ctx context.Context, methodName str
...
 		rpccontext.Logger(ctx).WithError(err).Error("Failed to authenticate caller")
 		return nil, err
 	}
-	return authorizedCtx, nil
+	if allow {
+		return ctx, nil
+	}
+
+	deniedErr := status.Errorf(codes.PermissionDenied, "Authorization denied for method %s", methodName)
+	rpccontext.Logger(ctx).WithError(deniedErr).Error("Failed to authenticate caller")
```
amoore877, line 64: "Failed to authenticate caller" is handled higher up in the function, where we got some error in the opa auth. I think this error log should be "Failed to authorize caller", or maybe better "Caller is unauthorized" since it's not really a failure of the service but instead an explicit rejection.

pkg/server/api/middleware/authorization.go
```diff
@@ -59,18 +51,18 @@ func (m *authorizationMiddleware) Preprocess(ctx context.Context, methodName str
...
 	if err != nil {
 		rpccontext.Logger(ctx).WithError(err).Error("Failed to authenticate caller")
 		return nil, err
 	}
-	return authorizedCtx, nil
+	if allow {
+		return ctx, nil
+	}
+
+	deniedErr := status.Errorf(codes.PermissionDenied, "Authorization denied for method %s", methodName)
```
amoore877, line 63: This entire effort is a fantastic security addition. Has there been discussion for future work on enabling triage of auth failures? In particular, I am thinking of the Batch APIs. On rejection, the caller (and SPIRE operators) only seem to know that at least one of the entries in the Batch was denied, but not which one. Depending on complexity of policies being set, it could be a bit of work to determine denial reason. If the caller was a workload, it is also cumbersome to get the entire list of entries printed out.

amoore877:
Awesome work :)

pkg/server/api/middleware/authorization.go
lumjjb: yea, that's a great idea!!! :). Having error messages being propagated from the OPA engine would be super useful. I'm still unsure what the best way to do this is with OPA. Let me ask some of the OPA folks to see if they have anything that can do this in the pipeline or if they have some ideas of how this can be done today.

pkg/server/api/middleware/authorization.go
lumjjb: i'm not sure if i understood your comment correctly. i've updated it in https://github.com/spiffe/spire/pull/2416/commits/bdf9ad4736eec98efa955e2d778b3e2b6da7a348, if you can take a quick look.

`Failed to authenticate caller` is what is being passed back today if the authentication middleware denies.

amoore877: The new wording addresses my comment.

The previous logic had a general error for authorization, which could mean either the auth itself failed (could be authn or authz failure) or there was no "error" but instead the caller is rejected but authz. So originally it may have somewhat made sense to bundle all that into a single error return.

With this change, we now explicitly know if there was an error, or if there was no error and the call is authorized or not. An error now is actually a real error, and the true/false boolean is an explicit authz accept or deny return. So for the flag check, there is no "failure" error to log, but instead "you're not authorized". Does that make sense?

lumjjb: Ah yes! Gotcha, thanks for providing more detail!

pkg/server/api/middleware/authorization.go
evan2645: In general, reason of denial is a pretty big deal, and as complexity grows this gets more important as @amoore877 is noting here. This is another general issue with adopting opa/rego in this fashion, so if there's not an existing solution I imagine that there is a community need for one.

lumjjb: There's an example of it here https://play.openpolicyagent.org/p/giimmNjfrj

it does look like the policies need to be instantiated a certain way - like a deny list rather than an allow list , will have to look into this a bit more.

evan2645 (approved):
thanks for this contribution @lumjjb ... it is hugely impactful!

There's a merge conflict, but the comments I raised previously have been fully addressed. Can't wait to see what people cook up using this :P

doc/authorization_policy_engine.md
PhilCorney: Hey Benardi,
Did you find a way to test calling the Entry API through Ghostunnel?
I'm looking to test something simliar. I've got the [tunnel](https://github.com/ghostunnel/ghostunnel/tree/master/docs/spiffe-workload-api-demo/unix) working was going to configure the front-end-agent to call the server tcp connection, but unsure how to form the request.
Many thanks
Phil

Issue #2280, opened 2021-05-12
Add Spiffe ID templating for k8s-workload-registrar


edwarnicke:
Currently the [k8s-workload-registrar](https://github.com/spiffe/spire/tree/main/support/k8s/k8s-workload-registrar) allows for [creation of Spiffe IDs using CRDs](https://github.com/spiffe/spire/tree/main/support/k8s/k8s-workload-registrar#spiffe-id-custom-resources)

It would be incredibly helpful to be able to have k8s-workload-registrar to allow creating of SpiffeIDs for workloads dynamically using *templates* defined by a CRD:

```yaml
apiVersion: spiffeid.spiffe.io/v1beta1
kind: SpiffeID
metadata:
  name: my-spiffe-id
  namespace: my-namespace
spec:
  spiffeId: spiffe://example.org/{{selector k8s:sa}}/{{selector k8s:pod-uid}}
  parentId: spiffe://example.org/spire/server
```

lumjjb:
Hi @edwarnicke! @evan2645 and @dfeldman mentioned that we may have some overlap in what we're trying to do with the k8s workload registrar!

@mrsabath and I have also been planning on doing something similar to this using the k8s workload registrar, with a little bit of a different angle. We were initially thinking of writing an operator where you would define a schema, together with the source of where to get that information (and required selectors that come with them). 

We wrote a short draft document below to illustrate the idea. It would be great if there's an opportunity to collaborate!

https://github.com/IBM/trusted-service-identity/pull/95/files

evan2645:
This would be awesome :-D

There are plans to eventually move away from k8s-workload-registrar, so I don't know if spending a huge amount of effort there will be worth the time, however I think this is a pretty lightweight mechanism that would be straightforward to implement and consume (and thus worth the effort).

I think the biggest thing to determine is where we want this configurable to live. Does it go only in CRDs? Or in labels/annotations? Or can it be configured directly on the registrar, and be used as the default ID? Or... maybe all those things? :)

Would be great to know if any of these shapes help your cause @lumjjb, and also @azdagron has expressed some thoughts here that I think he will share as soon as he gets a moment

lumjjb:
Yea - I think this will definitely help. +1 on being able to do this in labels and annotations! That way I think it could be easier to do mutating webhook integrations for pod creation.

azdagron:
[...]

PR #2417, opened 2021-07-14
Template for customizing SVID format

mrsabath:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Created a template to customize the format of SVID. As a result, the administrator is able to define the format of the SVID that is using the values available in the Pod object (namespace, pod name, UID etc), and arbitrary values that might be used for describing non-Kubernetes metadata (region, cluster-name, service provider etc). This PR provides a single format that is cluster scoped. It is advisable to use the same format across multiple clusters reporting to a single SPIRE server. 

**Description of change**

This change is optional and it preserves the current SVID format that can be provided via label or annotation. When no `identity_template` is defined, existing format `"ns/%s/sa/%s", pod.Namespace, pod.Spec.ServiceAccountName` is still used as default. If the `identity_template` is provided, it overrides the default format. This change is applying only to `crd` mode. 
The custom format might use arguments obtained from the Pod element, and also arbitrary values provided in `context` map. 

**Which issue this PR fixes**

#2280


support/k8s/k8s-workload-registrar/README.md
```diff
@@ -55,6 +55,8 @@ The following configuration directives are specific to `"crd"` mode:
 | `webhook_enabled`          | bool    | optional | Enable a validating webhook to ensure CRDs are properly fomatted and there are no duplicates. Only needed if manually creating entries | `false` |
 | `webhook_cert_dir`         | string  | optional | Directory for certificates when enabling validating webhook. The certificate and key must be named tls.crt and tls.key. | `"/run/spire/serving-certs"` |
 | `webhook_port`             | int     | optional | The port to use for the validating webhook. | `9443` |
+| `identity_template`        | string  | optional | The template for custom SVID format |  |
+| `context`                  | map[string]string | optional | The map of key/value pairs of arbitrary string parameters for specified by `identity_template` | |
```
edwarnicke, line 59: OK... context here is super super smart :)

edwarnicke:
Is this one identity template at cluster scope?  Or does it subscope to namespace and/or namespace+service account?

mrsabath:
@edwarnicke this is a first version, cluster scoped. We will later namespace scope it, with service-accounts as selectors

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -183,10 +201,45 @@ func (r *PodReconciler) podSpiffeID(pod *corev1.Pod) string {
...
 
 func (r *PodReconciler) podParentID(nodeName string) string {
 	return makeID(r.c.TrustDomain, "k8s-workload-registrar/%s/node/%s", r.c.Cluster, nodeName)
 }
+
+func (r *PodReconciler) getIdentityTemplate(pod *corev1.Pod) string {
+	tpl := r.c.IdentityTemplate
+	tmpl, err := template.New("IdentityTemplate").Parse(tpl)
+	if err != nil {
+		r.c.Log.WithError(err).Errorf("Error parsing the template (%v)", tpl)
```
MarcosDY, line 222: looks like to me that if template can not be created it must fails here, since tmpl will be nil and panic on line 240

MarcosDY: what to you think about `Errorf("Error parsing the template %q", tpl`?

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -183,10 +201,45 @@ func (r *PodReconciler) podSpiffeID(pod *corev1.Pod) string {
...
+		PodNodeNameLabel:         pod.Spec.NodeName,
+	}
+
+	templateMaps := IdentityMaps{
+		Context: r.c.Context,
+		Pod:     podInfo,
+	}
+	var svid bytes.Buffer
+	err = tmpl.Execute(&svid, templateMaps)
+	if err != nil {
```
MarcosDY, line 241: in case no error is returned what happens in the client of this function that will get an empty svid?

edwarnicke:
> @edwarnicke this is a first version, cluster scoped. We will later namespace scope it, with service-accounts as selectors

As long as you have clear line of site in your head to that... I'm all for walk before you run :)

support/k8s/k8s-workload-registrar/config.go
```diff
@@ -85,6 +87,15 @@ func (c *CommonMode) ParseConfig(hclConfig string) error {
 	if c.DisabledNamespaces == nil {
 		c.DisabledNamespaces = defaultDisabledNamespaces()
 	}
+	if c.IdentityTemplate == "" && c.Context != nil {
+		return errs.New("context defined without identity_template")
+	}
+	// IdentityTemplate represents the format after the trust domain and as such, it must not begin with spiffe://, // or /
+	if strings.HasPrefix(c.IdentityTemplate, "spiffe://") ||
+		strings.HasPrefix(c.IdentityTemplate, "//") ||
+		strings.HasPrefix(c.IdentityTemplate, "/") {
```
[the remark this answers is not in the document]
mrsabath: duh, of course! Thank you!

support/k8s/k8s-workload-registrar/config.go
```diff
@@ -85,6 +87,15 @@ func (c *CommonMode) ParseConfig(hclConfig string) error {
 	if c.DisabledNamespaces == nil {
 		c.DisabledNamespaces = defaultDisabledNamespaces()
 	}
+	if c.IdentityTemplate == "" && c.Context != nil {
```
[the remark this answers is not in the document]
mrsabath: Actually, if the template uses only Pod elements, then Context is not necessary. So I will check the template for `Context` substring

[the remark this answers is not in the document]
mrsabath: But a good point about orphaned Context

support/k8s/k8s-workload-registrar/config_test.go
```diff
@@ -110,6 +110,51 @@ func TestLoadMode(t *testing.T) {
 				InsecureSkipClientVerification: true,
 			},
 		},
+		{
+			name: "identity_template",
```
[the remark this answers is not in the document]
mrsabath: Good idea, thank you.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
mrsabath: Sounds good. Fixed!

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
mrsabath: Empty SVID means `ignore`. I fixed the error flow.

support/k8s/k8s-workload-registrar/config.go
```diff
@@ -36,18 +36,20 @@ type Mode interface {
...
+	ServerSocketPath   string            `hcl:"server_socket_path"`
+	AgentSocketPath    string            `hcl:"agent_socket_path"`
+	ServerAddress      string            `hcl:"server_address"`
+	Cluster            string            `hcl:"cluster"`
+	PodLabel           string            `hcl:"pod_label"`
+	PodAnnotation      string            `hcl:"pod_annotation"`
+	Mode               string            `hcl:"mode"`
+	DisabledNamespaces []string          `hcl:"disabled_namespaces"`
+	IdentityTemplate   string            `hcl:"identity_template"`
+	Context            map[string]string `hcl:"context"`
```
faisal-memon, line 52: Should these be defined in config_crd.go because they are specific to the crd mode?

support/k8s/k8s-workload-registrar/config.go
```diff
@@ -85,6 +87,19 @@ func (c *CommonMode) ParseConfig(hclConfig string) error {
 	if c.DisabledNamespaces == nil {
 		c.DisabledNamespaces = defaultDisabledNamespaces()
 	}
+	// eliminate orphaned context
```
faisal-memon, line 90: Need to update the logic on lines 81-83 of the new version so that only of PodLabel, PodAnnotation, or IdentityTemplate is set.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -161,32 +184,85 @@ func (r *PodReconciler) updateorCreatePodEntry(ctx context.Context, pod *corev1.
...
 
 	// the controller has not been configured with a pod label or a pod annotation.
+	if r.c.IdentityTemplate != "" {
+		// create an entry using provided identity template.
+		svid, err := r.getIdentityTemplate(pod)
+		if err != nil {
+			return "", err
+		}
+		return makeID(r.c.TrustDomain, svid), nil
+	}
```
faisal-memon, line 216: I think we need to gate this similar to the PodAnnotation/PodLabel. ie if the pod has label `spiffe.io/spiffeid` then we generate the the entry using the `IdentityTemplate`. Without the gate it'll generate an entry for every pod in the system.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -161,32 +184,85 @@ func (r *PodReconciler) updateorCreatePodEntry(ctx context.Context, pod *corev1.
...
 	return makeID(r.c.TrustDomain, "k8s-workload-registrar/%s/node/%s", r.c.Cluster, nodeName)
 }
+
+func (r *PodReconciler) getIdentityTemplate(pod *corev1.Pod) (string, error) {
+	tpl := r.c.IdentityTemplate
+	tmpl, err := template.New("IdentityTemplate").Parse(tpl)
+	if err != nil {
+		r.c.Log.WithError(err).Errorf("error parsing the template %q", tpl)
+		return "", err
+	}
```
faisal-memon, line 233: Is it possible to do this once, in `NewPodReconciler()` since the template format won't change?

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
edwarnicke: > I think we need to gate this similar to the PodAnnotation/PodLabel. ie if the pod has label `spiffe.io/spiffeid` then we generate the the entry using the `IdentityTemplate`. Without the gate it'll generate an entry for every pod in the system.

This gets back to my original point about Scope.  I may want this for every Pod in the cluster (I guarantee there will be folks who want that).  I may want this for every Pod in my Namespace, or Namespace+ServiceAccount.  I may want to pick and choose which Pods I want to use this template for.

What I would really really not want is to have to annotate *every* Pod *everytime* if my desired outcome is 'all in cluster/namespace/serviceaccount'...

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
edwarnicke: Won't change over what timescale?  I'm all for caching things where possible... but presuming they *never* change seems to require some demonstration :)

faisal-memon: The `k8s-workload-registrar` isn't dynamically configurable, so if you want to change any config you have to restart the container. SPIRE server has the same behavior.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
faisal-memon: i see how that can be a drag. how about a configurable `IdentityTemplateLabel`? If its empty generate the entry for every pod in the cluster. If its non-empty restrict to only Pods with the label.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -161,32 +184,85 @@ func (r *PodReconciler) updateorCreatePodEntry(ctx context.Context, pod *corev1.
...
+
+	// Create the IdentityMaps struct, with maps, one for Pod and one for Context:
+	podInfo := map[string]string{
+		PodServiceAccountIDLabel: pod.Spec.ServiceAccountName,
+		NamespaceIDLabel:         pod.Namespace,
+		PodNameIDLabel:           pod.Name,
+		PodUIDLabel:              string(pod.UID),
+		PodHostnameLabel:         pod.Spec.Hostname,
+		PodNodeNameLabel:         pod.Spec.NodeName,
+	}
```
faisal-memon, line 243: Ive always seen templates use a standard struct that holds the desired values. Using the struct you can have different types, ie `types.UID`, rather than just all `string`. Why did you use map?

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
edwarnicke: @faisal-memon OK.. that seems like a big problem... are you telling me I have to basically reboot my ability to change issue identities whenever I want to reconfigure?  What if a Pod is scheduled while it's rebooting?  How do I operationalize that in a reasonable way?

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
edwarnicke: @faisal-memon Again... that takes things back to global... so I can either have identities issued to *every* Pod in my cluster (sometimes desireable, but not always) or I can require every Pod spec for every app to be altered (sometimes but not always doable).  The root issue isn't 'annotations' vs 'labels'... its 'change every Pod' vs selection of Pods to which things should be applied.

K8s (and all of Cloud Native) tends heavily away from 'enumeration' (which is what change per Pod is) and towards selection.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
faisal-memon: Unfortunately yes. In that case we will get the reconcile event when the k8s-workload-registrar comes back up.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
faisal-memon: What do you suggest?

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
edwarnicke: @faisal-memon I don't see that as deployable :(  Not by anyone... outage on re-configure isn't going to fly in any real cluster :(

faisal-memon: There is no outage, everything that was running before will continue to run. New pods deployed while the re-configure is happening will see a delay to get their certs issued. It would be a good enhancement to allow for dynamic reconfig, but out of scope for this PR.

edwarnicke: @faisal-memon Ah... got it... so degradation, not outage.  Totally agree about it being a different scope to allow dynamic reconfig, not this PR.

support/k8s/k8s-workload-registrar/config.go
mrsabath: If we want to support this feature only on CRD mode, as it was showcased here, then yes.

support/k8s/k8s-workload-registrar/config.go
mrsabath: agree

evan2645:
@edwarnicke I haven't reviewed this PR yet, but I wanted to quickly drop by to let you know that we are in the early stages of designing a new CRD-centric k8s controller for SPIRE that will directly address your use case ... we will share it as soon as it's in a share-able state.

As for the way that k8s-workload-registrar is today, I think supporting the ns/sa scoping in this codebase will be a little bit clunky. I guess then the question comes down to priority - how soon do we want or need it, and do we want to make the effort to support this in current k8s-workload-registrar, or wait for the new thing?

edwarnicke:
> I guess then the question comes down to priority - how soon do we want or need it, and do we want to make the effort to support this in current k8s-workload-registrar, or wait for the new thing?

I get the clunky thing... lets get the good done so we can move onto the better :)

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
mrsabath: So per agreement during our contributors call, we will create an `IdentityTemplateLabel` parameter in the registrar configuration(CRD mode only). All the pods with this label set to `true` would get the identity defined in the custom template.  If the value of this parameter is empty, the template format would apply to all the pods. 
As a next step, we would focus on scoping it on the level of the namespace, service-accounts, or both.

support/k8s/k8s-workload-registrar/config.go
mrsabath: I moved this check into config_crd.go since it's CRD relevant only

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
mrsabath: done

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
mrsabath: done

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -51,6 +55,41 @@ func (c *CRDMode) ParseConfig(hclConfig string) error {
...
+
+	// IdentityTemplate represents the format following the trust domain and as such, it must not begin with spiffe://, // or /
+	if strings.HasPrefix(c.IdentityTemplate, "spiffe://") ||
+		strings.HasPrefix(c.IdentityTemplate, "/") {
+		return errs.New("identity_template cannot start with spiffe:// or /")
+	}
+
+	if c.CommonMode.Mode == modeCRD {
+		if c.IdentityTemplate == "" && c.CommonMode.PodAnnotation == "" && c.CommonMode.PodLabel == "" {
+			return errs.New("in crd mode, only one workload registration mode is allowed: identity_template, pod_annotation or pod_label")
```
lumjjb, line 89: this seems to want to check the existence of 2 or more, the conditional doesn't seem to do that, it only checks for existence of all 3

support/k8s/k8s-workload-registrar/config_crd_test.go
```diff
@@ -0,0 +1,349 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	testMinimalCRDConfig = `
+		trust_domain = "TRUSTDOMAIN"
+		cluster = "CLUSTER"
+		server_socket_path = "SOCKETPATH"
+		mode = "crd"
+	`
```
lumjjb, line 18: nit: line break here so its easier to see where the multiline indicators start/end.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -161,32 +205,89 @@ func (r *PodReconciler) updateorCreatePodEntry(ctx context.Context, pod *corev1.
...
-	// create an entry based on the service account.
-	return makeID(r.c.TrustDomain, "ns/%s/sa/%s", pod.Namespace, pod.Spec.ServiceAccountName)
+	if r.c.IdentityTemplate != "" {
+		// create an entry using provided identity template.
+		svid, err := r.getIdentityTemplate(pod)
+		if err != nil {
+			return "", err
+		}
+
+		if r.c.IdentityTemplateLabel != "" {
```
lumjjb, line 237: probably better to have this check early on so there's a quick exit without doing any extra work if there isn't a need to

lumjjb: It may be nicer to check for when to return `"",nil` and then otherwise return `makeID`, reducing 2 places calling to `makeID` to just 1 place, this may make the flow a bit easier to digest.

support/k8s/k8s-workload-registrar/README.md
faisal-memon, line 59: update readme with new configurable

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -51,6 +55,41 @@ func (c *CRDMode) ParseConfig(hclConfig string) error {
 		c.WebhookPort = defaultWebhookPort
 	}
 
+	if c.IdentityTemplate != "" && c.CommonMode.PodAnnotation != "" {
+		return errs.New("workload registration configuration is incorrect, can't specify both identity_template and pod_annotation")
+	}
+
+	if c.IdentityTemplate != "" && c.CommonMode.PodLabel != "" {
+		return errs.New("workload registration configuration is incorrect, can't specify both identity_template and pod_label")
+	}
```
faisal-memon, line 64:
```suggestion
	if c.IdentityTemplate != "" && (c.CommonMode.PodAnnotation != "" || c.CommonMode.PodLabel != "") {
		return errs.New("workload registration configuration is incorrect, can only use one of identity_template, pod_annotation, or pod_label")
	}
```

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -51,6 +55,41 @@ func (c *CRDMode) ParseConfig(hclConfig string) error {
...
+	if strings.HasPrefix(c.IdentityTemplate, "spiffe://") ||
+		strings.HasPrefix(c.IdentityTemplate, "/") {
+		return errs.New("identity_template cannot start with spiffe:// or /")
+	}
+
+	if c.CommonMode.Mode == modeCRD {
+		if c.IdentityTemplate == "" && c.CommonMode.PodAnnotation == "" && c.CommonMode.PodLabel == "" {
+			return errs.New("in crd mode, only one workload registration mode is allowed: identity_template, pod_annotation or pod_label")
+		}
+	}
```
faisal-memon, line 91: I think this check might be covered by the one i made a suggestion on above.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -16,46 +16,87 @@ limitations under the License.
...
+	Ctx                   context.Context
+	DisabledNamespaces    []string
+	Log                   logrus.FieldLogger
+	PodLabel              string
+	PodAnnotation         string
+	Scheme                *runtime.Scheme
+	TrustDomain           string
+	IdentityTemplate      string
+	Context               map[string]string
+	IdentityTemplateLabel string
```
faisal-memon, line 60: nit: Move `IdentityTemplateLabel` to be under `IdentityTemplate` to match the struct in config_crd.go

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -16,46 +16,87 @@ limitations under the License.
...
+}
+
+// PodInfo is created for every processed Pod and it holds pod specific information
+type PodInfo struct {
+	PodServiceAccountIDLabel string
+	NamespaceIDLabel         string
+	PodNameIDLabel           string
+	PodUIDLabel              string
```
faisal-memon, line 68:
```suggestion
	PodUIDLabel              types.UID
```

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -16,46 +16,87 @@ limitations under the License.
...
+	identityTempl *template.Template
 }
 
 // NewPodReconciler creates a new PodReconciler object
-func NewPodReconciler(config PodReconcilerConfig) *PodReconciler {
-	return &PodReconciler{
-		Client: config.Client,
-		c:      config,
+func NewPodReconciler(config PodReconcilerConfig) (*PodReconciler, error) {
+
```
faisal-memon, line 88: nit: extra whitespace

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -16,46 +16,87 @@ limitations under the License.
...
+	tpl := config.IdentityTemplate
```
faisal-memon, line 89: Can we just directly use `config.IdentityTemplate` instead of having the extra `tpl` variable.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -161,32 +205,89 @@ func (r *PodReconciler) updateorCreatePodEntry(ctx context.Context, pod *corev1.
...
 }
+
+func (r *PodReconciler) getIdentityTemplate(pod *corev1.Pod) (string, error) {
+
+	// Create the IdentityMaps struct, with maps, one for Pod and one for Context:
+	podInfo := PodInfo{
+		PodServiceAccountIDLabel: pod.Spec.ServiceAccountName,
+		NamespaceIDLabel:         pod.Namespace,
+		PodNameIDLabel:           pod.Name,
+		PodUIDLabel:              string(pod.UID),
```
faisal-memon, line 265:
```suggestion
		PodUIDLabel:              pod.UID,
```

support/k8s/k8s-workload-registrar/config_crd.go
mrsabath: Actually, this one makes sure at least one of the three is selected. I changed a bit the error message to make it less ambiguous.

mrsabath:
Thank you @faisal-memon, @edwarnicke, @lumjjb,  and @MarcosDY for your comments. I think all of them have been addressed now.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller_test.go
```diff
@@ -142,6 +148,212 @@ func (s *PodControllerTestSuite) TestPodLabel() {
...
 }
 
+// TestIdentityTemplate checks the various formats of the SPIFFE ID provided via IdentityTemplate.
+func (s *PodControllerTestSuite) TestIdentityTemplate() {
+	tests := []struct {
+		identityTemplate      string
+		identityTemplateLabel string
+		labelValue            string
+		context               map[string]string
+		expectedSVID          string
```
rturner3, line 158:
```suggestion
		expectedSpiffeIDPath          string
```

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller_test.go
```diff
@@ -142,6 +148,212 @@ func (s *PodControllerTestSuite) TestPodLabel() {
...
+
+		// this section is testing identity template formatting:
+		{
+			identityTemplate: DefaultTemplate,
+			expectedSVID:     fmt.Sprintf("ns/%s/sa/%s", PodNamespace, PodServiceAccount),
+			spiffeIDCount:    1,
+		},
+		{
+			identityTemplate: "ns/{{.Pod.Namespace}}/sa/{{.Pod.ServiceAccount}}",
+			expectedSVID:     "ns/" + PodNamespace + "/sa/" + PodServiceAccount,
```
rturner3, line 241: nit: since all the other tests are building the string with `fmt.Sprintf()`, can we do the same here just to be consistent?

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller_test.go
```diff
@@ -142,6 +148,212 @@ func (s *PodControllerTestSuite) TestPodLabel() {
...
+			s.Require().NoError(err)
+			continue
+		}
+		s.Require().NoError(err)
+
+		labelSelector := labels.Set(map[string]string{
+			"podUid": string(pod.ObjectMeta.UID),
+		})
+
+		// Verify that exactly 1 SPIFFE ID  resource was created for this pod
```
rturner3, line 328: spacing nit
```suggestion
		// Verify that exactly 1 SPIFFE ID resource was created for this pod
```

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -161,32 +202,91 @@ func (r *PodReconciler) updateorCreatePodEntry(ctx context.Context, pod *corev1.
...
+		return "", err
+	}
+	return makeID(r.c.TrustDomain, svid), nil
 }
 
 func (r *PodReconciler) podParentID(nodeName string) string {
 	return makeID(r.c.TrustDomain, "k8s-workload-registrar/%s/node/%s", r.c.Cluster, nodeName)
 }
+
+func (r *PodReconciler) getIdentityTemplate(pod *corev1.Pod) (string, error) {
```
rturner3, line 257: The name `getIdentityTemplate()` is a bit inaccurate here because the result is the actual SPIFFE ID path generated from the template + pod + context.
```suggestion
func (r *PodReconciler) generateSpiffeIDPath(pod *corev1.Pod) (string, error) {
```

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller_test.go
```diff
@@ -142,6 +148,212 @@ func (s *PodControllerTestSuite) TestPodLabel() {
...
+
+		_, err = s.r.Reconcile(req)
+		if err != nil && test.err != "" {
+			s.Require().Error(err)
+			s.Require().Contains(err.Error(), test.err)
+			err = s.k8sClient.Delete(s.ctx, &pod)
+			s.Require().NoError(err)
+			continue
+		}
+		s.Require().NoError(err)
```
rturner3, line 322: This looks a little funky to me, why are we doing essentially the same call and assertions twice in a row? Are we relying on some timing of the fake K8s API server? If so, how do we know that two executions is the magic number of times we need to call `s.r.Reconcile()`?

I am also not sure about the condition `if err != nil && test.err != ""`. What happens if `test.err != "" && err == nil`? Wouldn't we want to fail the test since we expected an error but didn't get one?

rturner3 (commented):
Hi @mrsabath, thanks for your patience on this PR! This feature seems really valuable, and this is a great start.

I left a few comments about how we could potentially avoid silently accepting some misconfiguration of the registrar as well as some feedback about how to improve the docs.

support/k8s/k8s-workload-registrar/config_crd.go
faisal-memon: i see now. I think we want to preserve the service account based registration for backwards compatibility.

faisal-memon: we can do that by setting the default value of IdentityTemplate to match what the service account based registration would do.

mrsabath: I agree with you @faisal-memon . What do you think if we assign the default, service account-based IdentityTemplate in Config_crd, when none of the other options is set (label, annotation) and/or when `identity_template=""`? 

Typically, `identity_template_label` would probably be missing as well, so we don't risk creating SPIFFEID for every pod in the cluster. All pods get SPIFFEID only if `identity_template_label=""`

support/k8s/k8s-workload-registrar/README.md
```diff
@@ -52,9 +52,12 @@ The following configuration directives are specific to `"crd"` mode:
...
 | `pod_controller`           | bool    | optional | Enable auto generation of SVIDs for new pods that are created | `true` |
-| `webhook_enabled`          | bool    | optional | Enable a validating webhook to ensure CRDs are properly fomatted and there are no duplicates. Only needed if manually creating entries | `false` |
+| `webhook_enabled`          | bool    | optional | Enable a validating webhook to ensure CRDs are properly formatted and there are no duplicates. Only needed if manually creating entries | `false` |
 | `webhook_cert_dir`         | string  | optional | Directory for certificates when enabling validating webhook. The certificate and key must be named tls.crt and tls.key. | `"/run/spire/serving-certs"` |
 | `webhook_port`             | int     | optional | The port to use for the validating webhook. | `9443` |
+| `identity_template`        | string  | optional | The template for custom [Identity Template Based Workload Registration](#identity-template-based-workload-registration) |  |
+| `identity_template_label`  | string  | optional | Pod label for selecting pods that get SVID defined by `identity_template` format. If not set, applies to all the pods |  |
```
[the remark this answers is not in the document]
mrsabath: Small caveat @rturner3 , right now, when `identity_template_label= "" `,  it applies to all the pads. When `identity_template_label` is not specified, no pods get the SPIFFE IDs. This relates to my earlier response to a comment from @faisal-memon. If no pod_label nor pod_annotation is provided, should we default to the service account-based IdentityTemplate, and apply it to ALL containers in the cluster? I think this is what happens today prior to this PR.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -33,29 +37,64 @@ import (
...
 
 // NewPodReconciler creates a new PodReconciler object
-func NewPodReconciler(config PodReconcilerConfig) *PodReconciler {
-	return &PodReconciler{
-		Client: config.Client,
-		c:      config,
+func NewPodReconciler(config PodReconcilerConfig) (*PodReconciler, error) {
+	tmpl, err := template.New("IdentityTemplate").Parse(config.IdentityTemplate)
+	if err != nil {
+		config.Log.WithError(err).Errorf("error parsing the template %q", config.IdentityTemplate)
```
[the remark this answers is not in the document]
mrsabath: Thank you, good suggestion.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller_test.go
mrsabath: So, there are two, different issues here. 
One is the "duplicate" reconcile calls. This is based on the original code from here: https://github.com/spiffe/spire/blob/main/support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller_test.go#L153-L157
I was initially puzzled as well, until I realized there are two different reconcile calls, for different objects, doing two different things:
* p.Reconcile()
* s.r.Reconcile()
so were should be good here.

The second part of your comment, regarding the `test.err`. The expected error can happen in several places of the flow depending on the nature of the error, and there is no easy way to indicate the exact location where the error should happen, without rewriting the entire testing logic, and that would create unnecessary complexity here. 

If you still believe this should be re-written, I can will go over this test flow in detail.

support/k8s/k8s-workload-registrar/config_crd.go
faisal-memon: sounds good to me.

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -51,6 +56,27 @@ func (c *CRDMode) ParseConfig(hclConfig string) error {
 		c.WebhookPort = defaultWebhookPort
 	}
 
+	if c.IdentityTemplate != "" && (c.CommonMode.PodAnnotation != "" || c.CommonMode.PodLabel != "") {
+		return errs.New("workload registration configuration is incorrect, can only use one of identity_template, pod_annotation, or pod_label")
+	}
+
+	// if no identity format set, use the default identity template
+	if c.CommonMode.Mode == modeCRD && c.IdentityTemplate == "" && c.CommonMode.PodAnnotation == "" && c.CommonMode.PodLabel == "" {
```
faisal-memon, line 64:
```suggestion
	if c.Mode == modeCRD && c.IdentityTemplate == "" && c.PodAnnotation == "" && c.PodLabel == "" {
```

faisal-memon:
```suggestion
	if c.IdentityTemplate == "" && c.CommonMode.PodAnnotation == "" && c.CommonMode.PodLabel == "" {
```

faisal-memon, line 64:
```suggestion
	if  c.IdentityTemplate == "" && c.PodAnnotation == "" && c.PodLabel == "" {
```

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -51,6 +56,27 @@ func (c *CRDMode) ParseConfig(hclConfig string) error {
 		c.WebhookPort = defaultWebhookPort
 	}
 
+	if c.IdentityTemplate != "" && (c.CommonMode.PodAnnotation != "" || c.CommonMode.PodLabel != "") {
```
faisal-memon, line 59:
```suggestion
	if c.IdentityTemplate != "" && (c.PodAnnotation != "" || c.PodLabel != "") {
```

support/k8s/k8s-workload-registrar/README.md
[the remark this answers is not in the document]
rturner3: I'm not sure I followed the first part of your comment about the two scenarios where `identity_template_label = ""` vs. when `identity_template_label` isn't set. Aren't those the same scenario? The config property is defined as a `string` in the `CRDMode` struct, so if unset, it should default to `""`.

I do see your point about backcompat now though regarding falling back to the service-account based path convention. Thanks for calling it out. With that in mind, I think your documentation text for `identity_template_label` actually looks good.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller_test.go
rturner3: Oh, thanks for pointing out the difference in `Reconcile()` calls, I missed that detail. One instance where the Go convention of terse variable names can bite you 😄 

> The expected error can happen in several places of the flow depending on the nature of the error, and there is no easy way to indicate the exact location where the error should happen

Would you mind elaborating on this a bit? Are the errors not deterministically returned by the same method from test execution to test execution?

Assuming the errors returned by each method are deterministic, a common way to handle this would be to have separate test template variables for errors coming from different methods, e.g. `expectPodReconcileErr`, `expectSpiffeIDReconcileErr`, etc. It looks like there are only four error cases being tested, so I would be curious to hear if there are any other challenges with adopting an approach like that.

support/k8s/k8s-workload-registrar/README.md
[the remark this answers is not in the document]
rturner3: Ah, I see you took this suggestion in the latest patch. Sorry to lead you astray there...would you mind reverting back to the previous text?

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -51,6 +55,37 @@ func (c *CRDMode) ParseConfig(hclConfig string) error {
...
+	// eliminate reference to non-existing context
+	// strip out the blanc space first
+	if c.Context == nil && c.IdentityTemplate != "" && strings.Contains(strings.ReplaceAll(c.IdentityTemplate, " ",""), "{{.Context.") {
+		return errs.New("identity_template references non-existing context")
+	}
+
+	// IdentityTemplate represents the format following the trust domain and as such, it must not begin with spiffe://, // or /
+	if strings.HasPrefix(c.IdentityTemplate, "spiffe://") ||
+		strings.HasPrefix(c.IdentityTemplate, "/") {
+		return errs.New("identity_template cannot start with spiffe:// or /")
```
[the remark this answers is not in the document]
rturner3: Just a couple follow-ups to this:
1) The method I proposed to detect invalid SPIFFE ID paths caused by a bad identity template + context should catch the two specific cases being checked here, so I think we can safely remove this if statement.
2) I spoke with @azdagron offline, and he mentioned that we won't be able to take the latest go-spiffe v2 in the SPIRE repo for some time. For now, we can get by with using the analogue logic in `github.com/spiffe/spire/pkg/common/idutil` that already exists in SPIRE (the logic I referenced in go-spiffe is being backported from SPIRE itself). Here is an example of how you would implement the check (can be added in `NewPodReconciler()`:
```go
var sb strings.Builder
if err := templ.Execute(&sb, data); err != nil {
    // handle err    
}

testSpiffeIDPath := sb.String()
testSpiffeID := fmt.Sprintf("spiffe://testdomain/%s", testSpiffeIDPath)
if err := idutil.CheckIDStringNormalization(testSpiffeID); err != nil {
    // something is wrong with the template which is resulting in invalid SPIFFE ID paths
}
```

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -161,32 +223,78 @@ func (r *PodReconciler) updateorCreatePodEntry(ctx context.Context, pod *corev1.
...
+			return "", nil
 		}
-		return ""
 	}
 
-	// the controller has not been configured with a pod label or a pod annotation.
-	// create an entry based on the service account.
-	return makeID(r.c.TrustDomain, "ns/%s/sa/%s", pod.Namespace, pod.Spec.ServiceAccountName)
+	// create an entry using provided identity template.
+	svid, err := r.generateSpiffeIDPath(pod)
```
rturner3, line 267:
```suggestion
	spiffeIDPath, err := r.generateSpiffeIDPath(pod)
```

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -161,32 +223,78 @@ func (r *PodReconciler) updateorCreatePodEntry(ctx context.Context, pod *corev1.
...
+	// the controller has not been configured with a pod label or a pod annotation, so identity_template must be set
+	// as it is enforced in config_crd
+	if r.c.IdentityTemplate == "" {
+		// this is just a sanity check
+		return "", nil
+	}
+
+	// if identity_template_label is not provided, all pods get the template formatted SVID
+	if r.c.IdentityTemplateLabel != "" {
+		// if identity_template_label is provided, only pods that have this label=`true` get the template formatted SVID
```
rturner3, line 256:
```suggestion
		// if identity_template_label is provided, only pods that have this label=`true` get the template formatted SPIFFE ID
```

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -161,32 +223,78 @@ func (r *PodReconciler) updateorCreatePodEntry(ctx context.Context, pod *corev1.
...
+	}
+
+	// the controller has not been configured with a pod label or a pod annotation, so identity_template must be set
+	// as it is enforced in config_crd
+	if r.c.IdentityTemplate == "" {
+		// this is just a sanity check
+		return "", nil
+	}
+
+	// if identity_template_label is not provided, all pods get the template formatted SVID
```
rturner3, line 254:
```suggestion
	// if identity_template_label is not provided, all pods get the template formatted SPIFFE ID
```

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -51,6 +56,27 @@ func (c *CRDMode) ParseConfig(hclConfig string) error {
 		c.WebhookPort = defaultWebhookPort
 	}
 
+	if c.IdentityTemplate != "" && (c.CommonMode.PodAnnotation != "" || c.CommonMode.PodLabel != "") {
+		return errs.New("workload registration configuration is incorrect, can only use one of identity_template, pod_annotation, or pod_label")
+	}
+
+	// if no identity format set, use the default identity template
```
faisal-memon, line 63: nit: The comments should begin with capital letter, end with period.

mrsabath:
Thank you @faisal-memon, @edwarnicke, @lumjjb,  and @MarcosDY for your comments. I think all of them have been addressed now.

mrsabath:
Thank you @rturner3 for the nice suggestion to use `github.com/spiffe/spire/pkg/common/idutil` to validate the SPIFFE path. This actually helped with several tests, so I simplified, reshuffled, and reduced some older tests.

support/k8s/k8s-workload-registrar/README.md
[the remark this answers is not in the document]
mrsabath: Actually, your suggestion was correct, so I updated the documentation as needed. I believe the current `README.md` version is the one we want. Is there something else here that I am missing?

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller_test.go
mrsabath: I was able to simplify the pod_controller test after applying the new test that you suggested. No need to create different error types anymore.

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -51,6 +55,19 @@ func (c *CRDMode) ParseConfig(hclConfig string) error {
 		c.WebhookPort = defaultWebhookPort
 	}
 
+	if c.IdentityTemplate != "" && (c.PodAnnotation != "" || c.PodLabel != "") {
+		return errs.New("workload registration configuration is incorrect, can only use one of identity_template, pod_annotation, or pod_label")
+	}
+
+	// If no identity format set, use the default identity template.
+	if c.Mode == modeCRD && c.IdentityTemplate == "" && c.PodAnnotation == "" && c.PodLabel == "" {
```
faisal-memon, line 63: i don't think you need `c.Mode == modeCRD` because we won't get to this point if it was some other mode configured.

support/k8s/k8s-workload-registrar/README.md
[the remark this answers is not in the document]
rturner3: I think I mistook how this config was being used. For some reason I got it into my head that `identity_template_label` was used even when `identity_template` is unset, but I see that is not the case. Looks like we're good here!

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -51,6 +55,19 @@ func (c *CRDMode) ParseConfig(hclConfig string) error {
 		c.WebhookPort = defaultWebhookPort
 	}
 
+	if c.IdentityTemplate != "" && (c.PodAnnotation != "" || c.PodLabel != "") {
+		return errs.New("workload registration configuration is incorrect, can only use one of identity_template, pod_annotation, or pod_label")
+	}
+
+	// If no identity format set, use the default identity template.
+	if c.IdentityTemplate == "" && c.PodAnnotation == "" && c.PodLabel == "" {
```
rturner3, line 63: I think we can remove this case now that the same thing is being done in `NewPodReconciler()`

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -33,29 +37,106 @@ import (
...
+	}
+	testSpiffeIDPath := sb.String()
+	testSpiffeID := fmt.Sprintf("spiffe://testdomain/%s", testSpiffeIDPath)
+	if err := idutil.CheckIDStringNormalization(testSpiffeID); err != nil {
+		// The format of the template is incorrect and it is resulting in invalid SPIFFE ID paths.
+		config.Log.WithError(err).WithFields(logrus.Fields{
+			"identity_template": config.IdentityTemplate,
+			"context":           config.Context,
+		}).Error("Error validating the identity template components")
+		return &PodReconciler{}, err
```
rturner3, line 132: I think it would aid in debugging if we wrap the error from `idutil` to give it a little more context and make it easier to trace down the error to this code path.

```suggestion
		return &PodReconciler{}, fmt.Errorf("validating identity template: %w", err)
```

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -33,29 +37,106 @@ import (
...
+		return &PodReconciler{
+			Client: config.Client,
+			c:      config,
+		}, nil
+	}
+
+	tmpl, err := template.New("IdentityTemplate").Parse(config.IdentityTemplate)
+	if err != nil {
+		config.Log.WithError(err).WithField("identity_template", config.IdentityTemplate).Error("error parsing identity template")
+		return &PodReconciler{}, err
```
rturner3, line 102: Wrapping the error will help with traceability back to this code path
```suggestion
		return &PodReconciler{}, fmt.Errorf("parsing identity template: %w", err)
```

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -33,29 +37,106 @@ import (
...
+			NodeName:       "xx",
+		},
+	}
+	var sb strings.Builder
+	if err := tmpl.Execute(&sb, templateMaps); err != nil {
+		config.Log.WithError(err).WithFields(logrus.Fields{
+			"identity_template": config.IdentityTemplate,
+			"context":           config.Context,
+		}).Error("Error executing the identity template")
+		return &PodReconciler{}, err
```
rturner3, line 122:
```suggestion
		return &PodReconciler{}, fmt.Errorf("executing identity template: %w", err)
```

rturner3 (commented):
Thanks for the updates, @mrsabath!

I just left a few small comments about improving traceability between error messages and initialization code paths. I think there is probably also some opportunity for refactoring `PodReconciler` since it is getting a little complex to reason about with it supporting three different ID modes (i.e. lots of `if` statements throughout methods), but I think that's out of scope for this initial PR. Once the latest comments have been addressed, I think we're pretty much good to go on this from my perspective.

mrsabath:
Thank you @rturner3 for your suggestion! This makes the debugging easier and improves testing readability. 
What are the next steps to complete this PR?

rturner3 (dismissed):
> Thank you @rturner3 for your suggestion! This makes the debugging easier and improves testing readability.
> What are the next steps to complete this PR?

It LGTM! We can merge the PR later this week once the v1.0.1 release is completed.

rturner3 (approved):

amartinezfayo (approved):

PR #2439, opened 2021-08-03
Fix broken link in MAINTAINERS.md

rturner3:
There is a link to the upgrade integration test README.md that results
in a 404 when clicking it in GitHub because the suffix of the file is
the incorrect case.


amartinezfayo (approved):
Thanks @rturner3!

PR #2477, opened 2021-08-31
Fix nil pointer dereference when allow_unsafe_ids is configured (#2476)

### LINKED ISSUE #2476: Bug: nil pointer dereference when allow_unsafe_ids is configured
Hello there, nil pointer dereference issue is encountered when allow_unsafe_ids is configured in server.conf. The root cause is that in `NewServerConfig` of `cmd/spire-server/cli/run/run.go`, `sc.Log` is called before its initialization.

It can be fixed by simply moving the logger initialization up. I will create a PR for that.

yangmarcyang:
Change-Id: I9200600b251e2f51be7277104112222f92964265
Signed-off-by: Marc Yang <marc.yang@bytedance.com>

fixes #2476

### DIFF
--- a/cmd/spire-server/cli/run/run.go
+++ b/cmd/spire-server/cli/run/run.go
@@ -356,13 +356,6 @@ func NewServerConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


azdagron (approved):
Thanks for this fix! 🤦

PR #2530, opened 2021-09-14
Upgrade grpc-go to latest

rturner3:
A unit test was depending on an exact string to be present in an error
returned by grpc-go. The text in that error condition changed with the
latest release. Change test to use grpc-go error status handlers and
check for the status code rather than the exact error text, which is not
guaranteed to be stable across releases.


azdagron (approved):
\o/, thanks @rturner3. I have a PR almost done that removes these old api/ packages in favor of go-spiffe v2. I guess I should be pushing that up soon :)

PR #2537, opened 2021-09-17
Upgrade all dependencies that do not introduce compilation errors

rturner3:
Bumps all dependencies that do not cause unit tests to fail. The remaining dependencies that are not on the latest version as reported by `go-mod-outdated` are:
- k8s.io dependencies (changes some API(s) that require code changes in k8s-workload-registrar)
- HCL, which breaks some tests for unclear reasons (latest 1.x releases have no release notes in GitHub)
- github.com/go-logr/logr which has to be upgraded in step with k8s.io libraries due to a transitive dependency

Replaces usage of deprecated AWS KMS field `types.KeyMetadata.CustomerMasterKeySpec` parameter with identical parallel parameter `types.KeyMetadata.KeySpec` to satisfy the linter. See [deprecated note](https://github.com/aws/aws-sdk-go-v2/blob/0023eb21e8b4a1ea4b5d89574c1351240739a2a2/service/kms/types/types.go#L266-L271) in source code for reference.

Also fixes an `azure_msi` `NodeResolver` plugin test to not rely on MSI for authentication. Interestingly, this test only failed on macOS, even though it doesn't appear to be a platform-specific function.


azdagron (approved):
Looks good! The multiple `require` directives was a result of some tidying with go1.17, but when I retidy locally I can't get it to split again /shrug.

Let's ship this for now, it shouldn't cause a problem.

PR #2595, opened 2021-10-14
Add gcloud secrets manager plugin

### LINKED ISSUE #2071: Serverless support: Google Cloud plugin
Implement a plugin that provides support to store issued identities in Google Secret Manager.

MarcosDY:
Add gcloud_secretsmanager plugin to SVIDStore catalog

**Which issue this PR fixes**
fixes #2071


pkg/agent/plugin/svidstore/gcloudsecretmanager/gcloud.go
```diff
@@ -0,0 +1,246 @@
...
+	}); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to delete secret: %v", err)
+	}
+
+	p.log.With("secret_name", opt.secretName()).Debug("Secret deleted")
+	return &svidstorev1.DeleteX509SVIDResponse{}, nil
+}
+
+// getSecret gets secret from GCloud and validtes if it has `spire-svid` label, nil if not found
+func getSecret(ctx context.Context, client secretsClient, secretName string) (*secretmanagerpb.Secret, error) {
```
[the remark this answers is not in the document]
evan2645: > another option is to add a selector that allow us to create a new tag with some information we must validate before update/delete a secret. what do you think about that approach?

I was wondering if we should set trust domain name or something like that as part of the tag value we already use

> secrets manager api didnt allow us to assign a policy on creation 'but' we can use SetIamPolicy, that will requiere some more information on selectors

Yes I think we will need something like this. The policy is a big JSON blob though, it would be much better if we could reference some pre-existing policy like the way the AWS plugin references a pre-existing key id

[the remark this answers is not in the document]
MarcosDY: sounds good to me, so 2 changes:

- spire-svid = td.IDString()
-  add new selector that allow us to get iam policy (optional?) and call SetIAMPolicy when we create/update an SVID.

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
```diff
@@ -0,0 +1,286 @@
...
+		return nil, status.Errorf(codes.InvalidArgument, "unknown configurations detected: %s", strings.Join(config.UnusedKeys, ","))
+	}
+
+	client, err := p.hooks.newClient(ctx, config.ServiceAccountFile)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to create secretmanager client: %v", err)
+	}
+
+	// gcp secret manager does not allow ".", hash td as label
+	tdHash := sha1.Sum([]byte(req.CoreConfiguration.TrustDomain)) //nolint: gosec // We use sha1 to hash TD in 128b to avoid label restrictions
```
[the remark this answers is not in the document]
amartinezfayo: The limitations of the label strings are really unfortunate.
Having a hash is not ideal in terms of user experience, since it's not understandable by humans looking at the secrets. I've been thinking about creative ways to work around this, but the limits in the characters set + the length limitation (63 characters) are really hard constraints to make this something readable by humans (trust domain names may have a length of 255).

@evan2645 any thoughts on this?

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
```diff
@@ -0,0 +1,286 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to add secret version: %v", err)
+	}
+
+	p.log.With("state", resp.State).With("name", resp.Name).Debug("Secret payload updated")
+
+	return &svidstorev1.PutX509SVIDResponse{}, nil
+}
+
+// DeleteX509SVID deletes a Secret in the configured Google Cloud Secrets manager
```
amartinezfayo, line 182:
```suggestion
// DeleteX509SVID deletes a secret in the configured Google Cloud Secret Manager
```

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
```diff
@@ -0,0 +1,286 @@
...
+		Etag: secret.Etag,
+	}); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to delete secret: %v", err)
+	}
+
+	p.log.With("secret_name", opt.secretName()).Debug("Secret deleted")
+	return &svidstorev1.DeleteX509SVIDResponse{}, nil
+}
+
+// getSecret gets secret from Google Cloud and validates if it has `spire-svid` label, nil if not found
```
amartinezfayo, line 210: This comment should be updated to reflect the new logic.

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
```diff
@@ -0,0 +1,286 @@
...
+func getSecret(ctx context.Context, client secretManagerClient, secretName string, td string) (*secretmanagerpb.Secret, error) {
```
amartinezfayo, line 211: We should use the tdHash variable name here, to avoid confusions.

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
```diff
@@ -0,0 +1,286 @@
...
+	secret, err := client.GetSecret(ctx, &secretmanagerpb.GetSecretRequest{
+		Name: secretName,
+	})
+	switch status.Code(err) {
+	case codes.OK:
+		// Verify that secret contains "spire-svid" label and it is enabled
+		if ok := validateLabels(secret.Labels, td); !ok {
+			return nil, status.Errorf(codes.InvalidArgument, "secret does not contain the 'spire-svid' label for trust domain %q", td)
```
amartinezfayo, line 219: This error should be updated. Maybe the error should be more opaque in terms of the logic behind this validation and should only mention that the secret is not managed by this SPIRE deployment, and leave the details to the logging.
We may log the label that was expected and the label that was found. I think that would be useful for operators troubleshooting problems.

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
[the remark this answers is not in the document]
evan2645: > Having a hash is not ideal in terms of user experience, since it's not understandable by humans looking at the secrets.

I worried about the same

>  I've been thinking about creative ways to work around this, but the limits in the characters set + the length limitation (63 characters) are really hard constraints to make this something readable by humans (trust domain names may have a length of 255).

same :)

I don't have a better solution. If we do character substitution, we'd have to substitute with a character that's not otherwise allowed in the trust domain name. And to address the length problem, we could truncate ... I'm not sure how real the risk of collision is here ... but then again, it's certainly less risky than not scoping the label to trust domain name.

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
MarcosDY: mmm not sure about logs here, we are not providing another logs for errors from anothers plugins we rely on errors for that

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
```diff
@@ -0,0 +1,290 @@
...
+					Bindings: []*iam.Binding{
+						{
+							Role:    opt.roleName,
+							Members: []string{opt.serviceAccount},
+						},
+					},
+				},
+			})
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to set IAM policy to secret: %v", err)
```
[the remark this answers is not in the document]
MarcosDY: ```Thinking further on this .. what if someone changes the selectors on the entry,```
every time someone change a selector current "secret" will be removed before creating a new one.
It was done this way to avoid adding loginc to know when we must update a "secret" and when it is possible or not, so just create a new one with the same name after old secret was removed

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
```diff
@@ -0,0 +1,336 @@
...
+			},
+		})
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to create secret: %v", err)
+		}
+		p.log.With("secret_name", secret.Name).Debug("Secret created")
+	}
+
+	if opt.roleName != "" && opt.serviceAccount != "" {
+		if !secretFound {
```
evan2645, line 139: I think there might be some ways to simplify this logic a bit .. maybe something like this?

```suggestion
                outdatedPolicy, err := p.shouldSetPolicy(ctx, secret, opt)
                if err != nil {
                     return nil, err
                }
                
                if !secretFound || outdatedPolicy {
                     p.setIamPolicy ...
```

evan2645 (commented):
Thanks Marcos - two more comments on the recent changes ... I don't see any problems though, this is pretty much ready to go I think.

amartinezfayo (approved):
Thanks @MarcosDY! This looks good to me. @evan2645 may want to look at the latest changes also.

evan2645 (approved):
wonderful, thank you @MarcosDY!!

PR #2596, opened 2021-10-14
refactor awssecretsmanager unit tests

MarcosDY:
refactor awssecretsmanager unit tests to use Put and delete interface instead of plugin


rturner3 (approved):
LGTM!

PR #2625, opened 2021-11-16
Fix incorrect description of vault configuration

hiyosi:

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

The document of `vault` upstream authority plugin.

**Description of change**

The spire-server only uses `sign-intermediate` endpoint, it is not affected by the pki role configurations.
In my understanding, the pki role is for issuing leaf certificates. 
Actually, The spire-server can start without pki role.

**Which issue this PR fixes**


azdagron (approved):
\o/

PR #2643, opened 2021-11-29
k8s-workload-registrar: dns_name_templates support

### LINKED ISSUE #2354: Convert SPIFFE ID to DNS name
* **Version**: all
* **Platform**: all
* **Subsystem**: k8s-workload-registrar

Older software, like NGINX, doesn't speak SPIFFE ID. This causes issues during the mTLS workload. When you visit a website using a browser, the browser verifies the certificate returned has a DNS name that matches what was entered by the user. NGINX does a similar check on the client side when validating server certificates. Since our identity system is based on SPIFFE ID, not hostnames, it would be useful to have this check be based on SPIFFE ID. ie the certificate as the correct spiffe ID for the service the client is trying to consume.

One way to do that is to translate the SPIFFE ID into a DNS name.  For example using a similar templating format as proposed in https://github.com/spiffe/spire/issues/2280, we would have a configurable like:

`spiffeid_dns_name = {{serviceaccount}}.{{namespace}}.svc.cluster.local`

Which would a DNS name using that format to the entry. NGINX can then use this DNS name to verify identity of the server.

sjberman:
Signed-off-by: Saylor Berman <s.berman@f5.com>

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
k8s-workload-registrar CRD mode DNS names

**Description of change**
Added a new configuration field `dns_name_templates` to the CRD mode of the k8s-workload-registrar. This field accepts a list of template strings to add to the dnsNames list of the SpiffeID CRD. Same template values apply to this field as do the `identity_template` field.

**Which issue this PR fixes**
Fixes #2354


support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -71,8 +73,21 @@ func (c *CRDMode) ParseConfig(hclConfig string) error {
 	}
 
 	// Eliminate reference to the non-existing context (strip out the blank space first).
```
faisal-memon, line 75: Does this comment need to be updated?

sjberman: Yeah this could probably be made clearer.

faisal-memon:
I think we need a check similar to what's done for the `identity_template` here: https://github.com/spiffe/spire/pull/2643/files#diff-a9f46f58f991248695f26bd9ea0ffc3d74f976f9dd3c4730d0873305868778f4R105-R133

It expands the template using some dummy values to ensure the result is a valid spiffe id. For DNS name templates we should do the same and validate using [x509util.ValidateDNS()](https://github.com/spiffe/spire/blob/main/pkg/common/x509util/dns.go#L13)

sjberman:
@faisal-memon Even without implementing that, I was seeing invalid DNS errors when I provided bad templates.

sjberman:
Spoke offline with @faisal-memon, added earlier validation.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -316,3 +364,33 @@ func (r *PodReconciler) generateSpiffeIDPath(pod *corev1.Pod) (string, error) {
...
+	}
+
+	for i, dnsNameTemplate := range r.dnsNameTemplates {
+		var dnsNameBuilder strings.Builder
+		if err := dnsNameTemplate.Execute(&dnsNameBuilder, templateMaps); err != nil {
+			r.c.Log.WithError(err).WithFields(logrus.Fields{
+				"dns_name_template":      r.c.DNSNameTemplates[i],
+				"dns_name_template_maps": templateMaps,
+			}).Error("Error executing the dns name template")
+			return nil, err
```
evan2645, line 390: In what cases might this fail? Do we want to fail the entire SVID here or just skip over the bad name?

support/k8s/k8s-workload-registrar/mode-crd/controllers/endpoint_controller_test.go
```diff
@@ -125,6 +125,6 @@ func (s *EndpointControllerTestSuite) TestAddDNSName() {
 	})
 	s.Require().NoError(err)
 	s.Require().Len(spiffeIDList.Items, 1)
-	s.Require().Len(spiffeIDList.Items[0].Spec.DnsNames, 2)
-	s.Require().Equal("test-endpoint.default.svc", spiffeIDList.Items[0].Spec.DnsNames[1])
```
evan2645, line 129: What causes these lines to change?

evan2645 (commented):
This is looking ok to me for the most part .. @faisal-memon did you have any further thoughts?

support/k8s/k8s-workload-registrar/config_crd.go
```diff
@@ -40,6 +41,7 @@ type CRDMode struct {
 	WebhookServiceName    string            `hcl:"webhook_service_name"`
 	IdentityTemplate      string            `hcl:"identity_template"`
 	IdentityTemplateLabel string            `hcl:"identity_template_label"`
+	DNSNameTemplates      *[]string         `hcl:"dns_name_templates"`
```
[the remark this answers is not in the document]
sjberman: The intention here is to be able to discern between a user-supplied empty list, and a nil value (meaning a user didn't specify anything).

There may be a use case where a user sets "add_svc_dns_name" to true, but doesn't want any other dns names, and therefore gives the registrar an empty dns list.

support/k8s/k8s-workload-registrar/mode-crd/controllers/endpoint_controller_test.go
sjberman: Previously the pod name was already included in the dnsNames list when this test ran. Now it gets added at a different point, and isn't available for this test.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
sjberman: It may not ever fail in this case since we've already done the initial validation when creating the reconciler. I can just `continue` if that would be preferred (still log an error though).

support/k8s/k8s-workload-registrar/config_crd.go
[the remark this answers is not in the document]
evan2645: > There may be a use case where a user sets "add_svc_dns_name" to true, but doesn't want any other dns names, and therefore gives the registrar an empty dns list.

Aaah I see. Thank you for the clarification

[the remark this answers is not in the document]
faisal-memon: so if a user goes `dns_name_templates=""` thats the user-supplied empty list?

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
```diff
@@ -79,28 +81,32 @@ type IdentityMaps struct {
...
+
+	return &PodReconciler{
+		Client:           config.Client,
+		c:                config,
+		identityTempl:    identityTempl,
+		dnsNameTemplates: dnsNameTemplates,
+	}, nil
+}
+
+func validateTemplates(config PodReconcilerConfig) (*template.Template, []*template.Template, error) {
```
faisal-memon, line 108: nit: This function also parses the template, so maybe a better name would be `validateAndParseTemplates()`? Or split the validate and parse parts into separate functions.

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
evan2645: I thought a little more about it and feel that perhaps it's better to just fail entirely, the signal is stronger that way.

evan2645 (approved):
lgtm. thank you very much for the contribution @sjberman!

support/k8s/k8s-workload-registrar/config_crd.go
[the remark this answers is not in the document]
evan2645: > so if a user goes dns_name_templates="" thats the user-supplied empty list?

I think it would be `dns_name_templates = []` ?

support/k8s/k8s-workload-registrar/mode-crd/controllers/pod_controller.go
evan2645: I agree .. splitting this up would be great, I didn't comment on it but the function is a little large and hard to grok

faisal-memon:
> This is looking ok to me for the most part .. @faisal-memon did you have any further thoughts?

looks good to me, just the minor comment about the validate function.

evan2645:
I left an approval, but would also like to get @faisal-memon sign off

I also approved the integration test run

support/k8s/k8s-workload-registrar/config_crd.go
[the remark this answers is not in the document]
sjberman: @evan2645 is correct, it would be an empty list.

faisal-memon:
> I left an approval, but would also like to get @faisal-memon sign off
> 
> I also approved the integration test run

Thanks @sjberman, splitting that function up makes it alot more readable. Good to go from my end.

evan2645 (approved):

PR #2661, opened 2021-12-15
Bump deps for 1.1.2 release

evan2645:
This commit bumps the deps for SPIRE API and Plugin interfaces such that
they depend on the release versions.


amartinezfayo (approved):
🎉

PR #2679, opened 2022-01-12
Update supported versions matrix

evan2645:
When this matrix was installed, we lamented that it would quickly fall
out of date and that it should really be automated. It was eventually
merged without automation under the argument that something here is
better than nothing and if it had to be automated it wouldn't get done
right away.

I submit this update under the same argument.


azdagron (approved):
😆

PR #2703, opened 2022-01-28
1152/reopen log file

### LINKED ISSUE #1152: pkg/common/log doesn't support logrotate
Currently, if you log to a file, `pkg/common/log` opens it once and assumes that FD will always be valid.

if you use a system such as l[ogrotate(8)](https://linux.die.net/man/8/logrotate) (as is suggested in the logrus README) you must use the `copytruncate` option, which is racy and can drop log entries.

Ideally, `pkg/common/log` would support re-opening its file when it receives a signal (HUP, USR1, and USR2 are popular options). This would enable a system like `logrotate(8)` to safely rotate logs.

zmt:
**Pull Request check list**

- [x] Documentation updated

**Affected functionality**
Support logrotate without lossy copytruncate directive.

**Description of change**
Add capability to reopen log_file upon receipt of SIGUSR2 so logrotate can trigger it via postrotate script.

Fixes #1152.

### DIFF
--- a/cmd/spire-agent/cli/run/run.go
+++ b/cmd/spire-agent/cli/run/run.go
@@ -55,11 +55,12 @@ type Config struct {
 }
 
 type agentConfig struct {
-	DataDir                       string    `hcl:"data_dir"`
-	AdminSocketPath               string    `hcl:"admin_socket_path"`
-	InsecureBootstrap             bool      `hcl:"insecure_bootstrap"`
-	JoinToken                     string    `hcl:"join_token"`
-	LogFile                       string    `hcl:"log_file"`
+	DataDir                       string `hcl:"data_dir"`
+	AdminSocketPath               string `hcl:"admin_socket_path"`
+	InsecureBootstrap             bool   `hcl:"insecure_bootstrap"`
+	JoinToken                     string `hcl:"join_token"`
+	LogFile                       string `hcl:"log_file"`
+	LogReopener                   log.ReopenableWriteCloser
 	LogFormat                     string    `hcl:"log_format"`
 	LogLevel                      string    `hcl:"log_level"`
 	SDS                           sdsConfig `hcl:"sds"`
@@ -190,6 +191,9 @@ func (cmd *Command) Run(args []string) int {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	util.SignalListener(ctx, cancel)
+	if c.LogReopener != nil {
+		go log.ReopenOnSignal(ctx, c.LogReopener)
+	}
 
 	err = a.Run(ctx)
 	if err != nil {
@@ -363,7 +367,15 @@ func NewAgentConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool)
 	logOptions = append(logOptions,
 		log.WithLevel(c.Agent.LogLevel),
 		log.WithFormat(c.Agent.LogFormat),
-		log.WithOutputFile(c.Agent.LogFile))
+	)
+	if c.Agent.LogFile != "" {
+		reopenableFile, err := log.NewReopenableFile(c.Agent.LogFile)
+		if err != nil {
+			return nil, err
+		}
+		logOptions = append(logOptions, log.WithReopenableOutputFile(reopenableFile))
+		ac.LogReopener = reopenableFile
+	}
 
 	logger, err := log.NewLogger(logOptions...)
 	if err != nil {
--- a/cmd/spire-server/cli/run/run.go
+++ b/cmd/spire-server/cli/run/run.go
@@ -78,11 +78,12 @@ type serverConfig struct {
 	JWTIssuer       string             `hcl:"jwt_issuer"`
 	JWTKeyType      string             `hcl:"jwt_key_type"`
 	LogFile         string             `hcl:"log_file"`
-	LogLevel        string             `hcl:"log_level"`
-	LogFormat       string             `hcl:"log_format"`
-	RateLimit       rateLimitConfig    `hcl:"ratelimit"`
-	SocketPath      string             `hcl:"socket_path"`
-	TrustDomain     string             `hcl:"trust_domain"`
+	LogReopener     log.ReopenableWriteCloser
+	LogLevel        string          `hcl:"log_level"`
+	LogFormat       string          `hcl:"log_format"`
+	RateLimit       rateLimitConfig `hcl:"ratelimit"`
+	SocketPath      string          `hcl:"socket_path"`
+	TrustDomain     string          `hcl:"trust_domain"`
 
 	ConfigPath string
 	ExpandEnv  bool
@@ -229,6 +230,12 @@ func (cmd *Command) Run(args []string) int {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	util.SignalListener(ctx, cancel)
+	if c.LogReopener != nil {
+		go func() {
+			c.Log.Info("spawning reopener")
+			log.ReopenOnSignal(ctx, c.LogReopener)
+		}()
+	}
 
 	err = s.Run(ctx)
 	if err != nil {
@@ -337,7 +344,15 @@ func NewServerConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool
 	logOptions = append(logOptions,
 		log.WithLevel(c.Server.LogLevel),
 		log.WithFormat(c.Server.LogFormat),
-		log.WithOutputFile(c.Server.LogFile))
+	)
+	if c.Server.LogFile != "" {
+		reopenableFile, err := log.NewReopenableFile(c.Server.LogFile)
+		if err != nil {
+			return nil, err
+		}
+		logOptions = append(logOptions, log.WithReopenableOutputFile(reopenableFile))
+		sc.LogReopener = reopenableFile
+	}
 
 	logger, err := log.NewLogger(logOptions...)
 	if err != nil {
--- a/conf/agent/agent_full.conf
+++ b/conf/agent/agent_full.conf
@@ -14,6 +14,25 @@ agent {
     # join_token = ""
 
     # log_file: File to write logs to.
+    #
+    # If set, spire-agent will spawn a handler to reopen the file upon receipt
+    # of SIGUSR2 to support log rotation. To use logrotate without lossy
+    # copytruncate option, users MUST add a postrotate script to the logrotate
+    # configuration to send the SIGUSR2 signal to the spire-agent process.
+    #
+    # Example logrotate.conf:
+    #
+    # /path/to/spire-agent.log {
+    #     compress
+    #     delaycompress
+    #     maxsize 100M
+    #     missingok
+    #     rotate 7
+    #     postrotate
+    #         killall -USR2 spire-agent
+    #     endscript
+    # }
+    #
     # log_file = ""
 
     # log_format: Format of logs, <text|json>. Default: text.
--- a/conf/server/server_full.conf
+++ b/conf/server/server_full.conf
@@ -100,6 +100,25 @@ server {
     # jwt_issuer = ""
 
     # log_file: File to write logs to
+    #
+    # If set, spire-server will spawn a handler to reopen the file upon receipt
+    # of SIGUSR2 to support log rotation. To use logrotate without lossy
+    # copytruncate option, users MUST add a postrotate script to the logrotate
+    # configuration to send the SIGUSR2 signal to the spire-server process.
+    #
+    # Example logrotate.conf:
+    #
+    # /path/to/spire-server.log {
+    #     compress
+    #     delaycompress
+    #     maxsize 100M
+    #     missingok
+    #     rotate 7
+    #     postrotate
+    #         killall -USR2 spire-server
+    #     endscript
+    # }
+    #
     # log_file = ""
 
     # log_level: Sets the logging level <DEBUG|INFO|WARN|ERROR>. Default: INFO.
--- a/pkg/common/log/options.go
+++ b/pkg/common/log/options.go
@@ -17,6 +17,9 @@ const (
 // An Option can change the Logger to apply desired configuration in NewLogger
 type Option func(*Logger) error
 
+// WithOutputFile does not support logrotate without copytruncate.
+//
+// Deprecated: Use WithReopenableOutputFile instead.
 func WithOutputFile(file string) Option {
 	return func(logger *Logger) error {
 		if file == "" {
@@ -37,6 +40,24 @@ func WithOutputFile(file string) Option {
 		}
 
 		logger.Closer = fd
+		return nil
+	}
+}
+
+// WithReopenableOutputFile uses ReopenableFile to support handling a signal
+// to rotate log files (e.g. from a logrotate postrotate script).
+func WithReopenableOutputFile(reopenableFile *ReopenableFile) Option {
+	return func(logger *Logger) error {
+		logger.SetOutput(reopenableFile)
+
+		// If, for some reason, there's another closer set, close it first.
+		if logger.Closer != nil {
+			if err := logger.Closer.Close(); err != nil {
+				return err
+			}
+		}
+
+		logger.Closer = reopenableFile
 		return nil
 	}
 }
--- a/pkg/common/log/reopen.go
+++ b/pkg/common/log/reopen.go
@@ -1,0 +1,95 @@
+package log
+
+import (
+	"context"
+	"io"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+)
+
+const _reopenSignal = syscall.SIGUSR2
+
+var _ ReopenableWriteCloser = (*ReopenableFile)(nil)
+
+// Reopener inspired by https://github.com/client9/reopen
+type Reopener interface {
+	Reopen() error
+}
+
+type ReopenableWriteCloser interface {
+	Reopener
+	io.WriteCloser
+}
+
+type ReopenableFile struct {
+	name string
+	f    *os.File
+	mu   sync.Mutex
+}
+
+func NewReopenableFile(name string) (*ReopenableFile, error) {
+	file, err := os.OpenFile(name, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0640)
+	if err != nil {
+		return nil, err
+	}
+	return &ReopenableFile{
+		name: name,
+		f:    file,
+	}, nil
+}
+
+func (r *ReopenableFile) Reopen() error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.f != nil {
+		if err := r.f.Close(); err != nil {
+			return err
+		}
+		r.f = nil
+	}
+	newFile, err := os.OpenFile(r.name, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0640)
+	if err != nil {
+		r.f = nil
+		return err
+	}
+	r.f = newFile
+	return nil
+}
+
+func (r *ReopenableFile) Write(p []byte) (n int, err error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	return r.f.Write(p)
+}
+
+func (r *ReopenableFile) Close() error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	return r.f.Close()
+}
+
+func ReopenOnSignal(ctx context.Context, rwc ReopenableWriteCloser) {
+	signalCh := make(chan os.Signal, 1)
+	signal.Notify(signalCh, _reopenSignal)
+	reopenOnSignal(ctx, rwc, signalCh)
+}
+
+func reopenOnSignal(
+	ctx context.Context,
+	rwc ReopenableWriteCloser,
+	signalCh chan os.Signal,
+) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-signalCh:
+			rwc.Reopen()
+		}
+	}
+}
--- a/pkg/common/log/reopen_test.go
+++ b/pkg/common/log/reopen_test.go
@@ -1,0 +1,74 @@
+package log
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestReopenOnSignal(t *testing.T) {
+	const (
+		_tempDir         = "spirelogrotatetest"
+		_testLogFileName = "test.log"
+		_rotatedSuffix   = "rotated"
+		_firstMsg        = "a message"
+		_secondMsg       = "another message"
+	)
+	dir, err := os.MkdirTemp("", _tempDir)
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	logFileName := filepath.Join(dir, _testLogFileName)
+	rotatedLogFileName := logFileName + "." + _rotatedSuffix
+	rf, err := NewReopenableFile(logFileName)
+	require.NoError(t, err)
+
+	fsInfo, err := rf.f.Stat()
+	require.NoError(t, err)
+	assert.Equal(t, int64(0), fsInfo.Size(), "%s should be empty", fsInfo.Name())
+
+	logger, err := NewLogger(WithReopenableOutputFile(rf))
+	require.NoError(t, err)
+
+	logger.Warning(_firstMsg)
+
+	fsInfo, err = rf.f.Stat()
+	require.NoError(t, err)
+	initialLogSize := fsInfo.Size()
+	assert.NotEqual(t, int64(0), fsInfo.Size(), "%s should not be empty", fsInfo.Name())
+
+	signalCh := make(chan os.Signal, 1)
+	// prevent test from hanging until test runner timeout when failing
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
+	defer cancel()
+
+	renamedCh := make(chan struct{})
+	go func() {
+		// emulate logrotate
+		os.Rename(logFileName, rotatedLogFileName)
+		signalCh <- _reopenSignal
+		close(renamedCh)
+	}()
+	reopenOnSignal(ctx, rf, signalCh)
+	<-renamedCh
+	fsInfo, err = rf.f.Stat()
+	require.NoError(t, err)
+	assert.Equal(t, int64(0), fsInfo.Size(), "%s should be empty again", fsInfo.Name())
+
+	rotatedLog, err := os.Open(rotatedLogFileName)
+	require.NoError(t, err)
+	fsInfo, err = rotatedLog.Stat()
+	require.NoError(t, err)
+	assert.Equal(t, initialLogSize, fsInfo.Size(), "%s should be same size as before rename", fsInfo.Name())
+
+	logger.Warning(_secondMsg)
+	fsInfo, err = rf.f.Stat()
+	require.NoError(t, err)
+	assert.NotEqual(t, int64(0), fsInfo.Size(), "%s should not be empty", fsInfo.Name())
+	assert.NotEqual(t, initialLogSize, fsInfo.Size(), "%s should not be same size as initial file", fsInfo.Name())
+}
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -9,6 +9,7 @@ import (
 	"github.com/spiffe/go-spiffe/v2/spiffeid"
 	"github.com/spiffe/spire/pkg/common/catalog"
 	"github.com/spiffe/spire/pkg/common/health"
+	"github.com/spiffe/spire/pkg/common/log"
 	"github.com/spiffe/spire/pkg/common/telemetry"
 )
 
@@ -41,6 +42,9 @@ type Config struct {
 	PluginConfigs catalog.HCLPluginConfigMap
 
 	Log logrus.FieldLogger
+
+	// LogReopener facilitates handling a signal to rotate log file.
+	LogReopener log.ReopenableWriteCloser
 
 	// Address of SPIRE server
 	ServerAddress string
--- a/pkg/common/log/log_test.go
+++ b/pkg/common/log/log_test.go
@@ -67,3 +67,40 @@ func TestOutputFile(t *testing.T) {
 		}
 	}
 }
+
+// Make sure writing to reopenable logfile behaves identically to static file.
+func TestReopenableOutputFile(t *testing.T) {
+	msg := "This should get written"
+
+	for _, format := range []string{DefaultFormat, TextFormat, JSONFormat} {
+		f, err := os.CreateTemp("", "testoutputfile")
+		require.NoError(t, err)
+		tmpfile := f.Name()
+		defer os.Remove(tmpfile)
+
+		reopenableFile, err := NewReopenableFile(f.Name())
+		require.NoError(t, err)
+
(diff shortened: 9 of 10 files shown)


zmt:
I am debating whether I should add a 4th commit on top to remove `log.WithOutputFile` altogether. I would appreciate an opinion from reviewers.

zmt:
I turned off my linter because I was getting too many "should have a comment or be unexported" warnings...

Fixing lint.

cmd/spire-server/cli/run/run.go
```diff
@@ -229,6 +230,12 @@ func (cmd *Command) Run(args []string) int {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	util.SignalListener(ctx, cancel)
+	if c.LogReopener != nil {
+		go func() {
+			c.Log.Info("spawning reopener")
```
amoore877, line 235: why no `c.Log.Info("spawning reopener")` in agent code?

cmd/spire-server/cli/run/run.go
```diff
@@ -337,7 +344,15 @@ func NewServerConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool
...
-		log.WithOutputFile(c.Server.LogFile))
+	)
+	if c.Server.LogFile != "" {
+		reopenableFile, err := log.NewReopenableFile(c.Server.LogFile)
+		if err != nil {
+			return nil, err
+		}
+		logOptions = append(logOptions, log.WithReopenableOutputFile(reopenableFile))
+		sc.LogReopener = reopenableFile
+	}
```
amoore877, line 355: Agent and Server code seem nearly identical.
Is there a place we could extract the common functionality?

conf/agent/agent_full.conf
```diff
@@ -14,6 +14,25 @@ agent {
...
     # log_file: File to write logs to.
+    #
+    # If set, spire-agent will spawn a handler to reopen the file upon receipt
+    # of SIGUSR2 to support log rotation. To use logrotate without lossy
+    # copytruncate option, users MUST add a postrotate script to the logrotate
+    # configuration to send the SIGUSR2 signal to the spire-agent process.
+    #
+    # Example logrotate.conf:
+    #
+    # /path/to/spire-agent.log {
```
amoore877, line 25: I'm not familiar with the logrotate config, but is this the absolute minimal config we can put in the example while still showing the necessary 2 lines? Want to make sure the necessary usage is totally clear

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+type ReopenableFile struct {
+	name string
+	f    *os.File
+	mu   sync.Mutex
+}
+
+func NewReopenableFile(name string) (*ReopenableFile, error) {
+	file, err := os.OpenFile(name, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0640)
```
amoore877, line 33: `os.O_APPEND|os.O_CREATE|os.O_WRONLY` and `0640` are used in more than one place. Should we make them constants for the file?

cmd/spire-server/cli/run/run.go
zmt: Good catch. I will add it.

cmd/spire-server/cli/run/run.go
zmt: Nearly identical, yes, but to extract it to a helper, I think it would need to accept the config under construction as args, allowing one or the other to be nil or as `interface{}`. Neither of those options seems better than similar copypasta as I've done here.

pkg/common/log/reopen.go
zmt: Sure. Good suggestion.

conf/agent/agent_full.conf
zmt: I suppose I could reduce it to:
```
/path/to/spire-agent.log {
	rotate 7
	postrotate
		killall -USR2 spire-agent
	endscript
}
```

zmt: might want to include the `compress` and `delaycompress` as recommendations, but `rotate` and `postrotate` are the minimum config. to use with this.

conf/server/server_full.conf
```diff
@@ -100,6 +100,21 @@ server {
...
+    # of SIGUSR2 to support log rotation. To use logrotate without lossy
+    # copytruncate option, users MUST add a postrotate script to the logrotate
+    # configuration to send the SIGUSR2 signal to the spire-server process.
+    #
+    # Minimal example of logrotate.conf:
+    #
+    # /path/to/spire-agent.log {
+    #     rotate 7
+    #     postrotate
+    #         killall -USR2 spire-agent
```
zmt, line 114: should be spire-server - oops

zmt:
I was partway through migrating `k8s-workload-registrar` to `ReopenableFile` logging when I noticed that I should also update `oidc-discovery-provider`. The `k8s-workload-registrar` would be a little more complicated than server and agent because of the various modes of operation, but doable. I do not see a good place to spawn a `ReopenOnSignal` goroutine in `oidc-discovery-proviider` so punting on those for now and undeprecating `WithOutputFile`.

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,106 @@
...
+	defer r.mu.Unlock()
+
+	return r.f.Close()
+}
+
+// Name implements part of os.FileInfo without needing a lock on the
+// underlying file.
+func (r *ReopenableFile) Name() string {
+	return r.name
+}
```
zmt, line 84: Maybe I should add a `Stat()` method which locks & returns `r.f.Stat()`?

zmt:
> I am debating whether I should add a 4th commit on top to remove `log.WithOutputFile` altogether. I would appreciate an opinion from reviewers.
I was partway through migrating `k8s-workload-registrar` to `ReopenableFile` logging when I noticed that I should also update `oidc-discovery-provider`. The `k8s-workload-registrar` would be a little more complicated than server and agent because of the various modes of operation, but doable. I do not see a good place to spawn a `ReopenOnSignal` goroutine in `oidc-discovery-proviider` so punting on those for now and undeprecating `WithOutputFile`.

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,106 @@
...
+	rwc ReopenableWriteCloser,
+	signalCh chan os.Signal,
+) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-signalCh:
+			// satisfy errcheck
+			_ = rwc.Reopen()
```
rturner3, line 103: If we fail to reopen the logger, we can assume that SPIRE will fail to write to the log file in the future. Some other code depending on `ReopenableFile` to be in a functional state will likely panic once it calls `Write()`. This could lead to some confusing and hard to debug errors because we've lost the root cause of why the logger is inoperable.

When we get an error here, I think we would want to preemptively signal the error back to the parent goroutine and trigger a termination of the process.

pkg/common/log/reopen_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+func TestReopenOnSignal(t *testing.T) {
+	const (
+		_tempDir         = "spirelogrotatetest"
+		_testLogFileName = "test.log"
+		_rotatedSuffix   = "rotated"
+		_firstMsg        = "a message"
+		_secondMsg       = "another message"
+	)
+	dir, err := os.MkdirTemp("", _tempDir)
```
rturner3, line 22: We have a utility called `spiretest.TempDir()` that simplifies the creation of temporary directories

pkg/common/log/reopen_test.go
```diff
@@ -0,0 +1,75 @@
...
+	logger.Warning(_firstMsg)
+
+	fsInfo, err = rf.f.Stat()
+	require.NoError(t, err)
+	initialLogSize := fsInfo.Size()
+	assert.NotEqual(t, int64(0), fsInfo.Size(), "%s should not be empty", fsInfo.Name())
+
+	signalCh := make(chan os.Signal, 1)
+	// prevent test from hanging until test runner timeout when failing
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
```
rturner3, line 47: We generally steer away from introducing test-specific timeouts. I would prefer if we remove this since we do not want to rely on any test timing other than the test passing in the timeframe allowed by `go test` for the package.

pkg/common/log/reopen_test.go
```diff
@@ -0,0 +1,75 @@
...
+	require.NoError(t, err)
+	fsInfo, err = rotatedLog.Stat()
+	require.NoError(t, err)
+	assert.Equal(t, initialLogSize, fsInfo.Size(), "%s should be same size as before rename", fsInfo.Name())
+
+	logger.Warning(_secondMsg)
+	fsInfo, err = rf.f.Stat()
+	require.NoError(t, err)
+	assert.NotEqual(t, int64(0), fsInfo.Size(), "%s should not be empty", fsInfo.Name())
+	assert.NotEqual(t, initialLogSize, fsInfo.Size(), "%s should not be same size as initial file", fsInfo.Name())
```
rturner3, line 74: Perhaps we can also compare `ModTime()` with the first file just before we rotate the file to rule out the case where the messages written to the first and second file result in the same number of bytes being written to the first file and rotated file.

pkg/common/log/reopen.go
zmt: > I don't think we really need it since logrus accepts a generic io.Writer for its output stream. I'm not seeing a use case for Name() either other than for tests. The test is already reaching into the f *os.File field for Stat() though, so I think it may be cleaner to just rely directly on r.f for Name() as well.

I missed that detail. Agreed. I only added this to satisfy that test to get the PR up.

cmd/spire-agent/cli/run/run.go
```diff
@@ -190,6 +191,12 @@ func (cmd *Command) Run(args []string) int {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	util.SignalListener(ctx, cancel)
+	if c.LogReopener != nil {
+		go func() {
+			c.Log.Info("spawning logfile reopener")
```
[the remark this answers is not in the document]
zmt: Ack.

cmd/spire-agent/cli/run/run.go
```diff
@@ -55,11 +55,12 @@ type Config struct {
...
-	AdminSocketPath               string    `hcl:"admin_socket_path"`
-	InsecureBootstrap             bool      `hcl:"insecure_bootstrap"`
-	JoinToken                     string    `hcl:"join_token"`
-	LogFile                       string    `hcl:"log_file"`
+	DataDir                       string `hcl:"data_dir"`
+	AdminSocketPath               string `hcl:"admin_socket_path"`
+	InsecureBootstrap             bool   `hcl:"insecure_bootstrap"`
+	JoinToken                     string `hcl:"join_token"`
+	LogFile                       string `hcl:"log_file"`
+	LogReopener                   log.ReopenableWriteCloser
```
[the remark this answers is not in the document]
zmt: Absolutely. I was concerned about that exact point. My first instinct was to make it unexported, but I didn't see an extant example. I went with grouping near the relevant `LogFile` config. I figured it would elicit some response with guidance.

pkg/common/log/options.go
```diff
@@ -17,6 +17,9 @@ const (
 // An Option can change the Logger to apply desired configuration in NewLogger
 type Option func(*Logger) error
 
+// WithOutputFile requires lossy copytruncate directive in logrotate.
+//
+// Used by oidc-discovery-provider and k8s-workload-registrar.
```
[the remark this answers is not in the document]
zmt: Ack. Will do.

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,106 @@
...
+	newFile, err := os.OpenFile(r.name, _fileFlags, _fileMode)
+	if err != nil {
+		r.f = nil
+		return err
+	}
+	r.f = newFile
+	return nil
+}
+
+func (r *ReopenableFile) Write(p []byte) (n int, err error) {
```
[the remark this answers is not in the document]
zmt: Copied from the stdlib usage of `p` for consistency vs. sense.

pkg/common/log/reopen.go
zmt: > If we fail to reopen the logger, we can assume that SPIRE will fail to write to the log file in the future.

I was presuming it would continue writing to the old file descriptor (for the now rotated file). I guess that's not a safe assumption - it's just the behavior when the signal handling fails in test. If the reopen fails in production runtime, most likely the filesystem is read-only or otherwise broken in some way.

> When we get an error here, I think we would want to preemptively signal the error back to the parent goroutine and trigger a termination of the process.

I was hoping to avoid plumbing through an `errCh` from `main`.

pkg/common/log/reopen_test.go
zmt: Ack.

pkg/common/log/reopen_test.go
zmt: Ack.

pkg/common/log/reopen_test.go
zmt: I like this idea.

conf/server/server_full.conf
```diff
@@ -100,6 +100,21 @@ server {
...
+    # of SIGUSR2 to support log rotation. To use logrotate without lossy
+    # copytruncate option, users MUST add a postrotate script to the logrotate
+    # configuration to send the SIGUSR2 signal to the spire-server process.
+    #
+    # Minimal example of logrotate.conf:
+    #
+    # /path/to/spire-agent.log {
+    #     rotate 7
+    #     postrotate
+    #         killall -USR2 spire-server
```
amoore877, line 114: example for server file has `killall spire-server` but log file is `path/to/spire-agent.log`

zmt: Thank you.

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,110 @@
...
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+)
+
+const (
+	_fileFlags    = os.O_APPEND | os.O_CREATE | os.O_WRONLY
+	_fileMode     = 0640
+	_reopenSignal = syscall.SIGUSR2
```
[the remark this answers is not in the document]
azdagron: Also, `syscall.SIGUSR2` (and in general, this functionality) isn't available on windows. We've just undergone an effort to get us building under windows, so we'll need to make this functionality a no-op on windows.

pkg/common/log/reopen_test.go
```diff
@@ -0,0 +1,115 @@
...
+	signalCh := make(chan os.Signal, 1)
+
+	renamedCh := make(chan struct{})
+	go func() {
+		// emulate logrotate
+		err = os.Rename(logFileName, rotatedLogFileName)
+		require.NoError(t, err)
+		signalCh <- _reopenSignal
+		// explicitly cancel so test continues
+		cancel()
```
azdagron, line 55: Since signalCh is buffered, I'm not sure we have a guarantee about which channel `reopenOnSignal` will select on (i.e. the signal channel or the context). As such, this seems racy or prone to break with underlying changes to the scheduler.

I wonder if a more stable way of ensuring that the reopen is processed completely is to wrap the `rwc` in a `Reopener` that calls `Reopen` on the wrapped `Reopener` and then cancels the context.

pkg/common/log/reopen_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+	logger.Warning(_firstMsg)
+
+	fsInfo, err = rwc.f.Stat()
+	require.NoError(t, err)
+	initialLogSize := fsInfo.Size()
+	initialLogModTime := fsInfo.ModTime()
+	assert.NotEqual(t, int64(0), fsInfo.Size(), "%s should not be empty", fsInfo.Name())
+
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
zmt: I cycled on including that or not q few times. I will add it back.

pkg/common/log/reopen_test.go
zmt: I was struggling with it a bit here. I did high count under race detector and this appears stable as written.

> prone to break with underlying changes to the scheduler

True.

I will try your wrapper suggestion.

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// Name implements part of os.FileInfo without needing a lock on the
+// underlying file.
+func (r *ReopenableFile) Name() string {
+	return r.name
+}
+
+// ReopenOnSignal returns a function compatible with RunTasks.
+func ReopenOnSignal(rwc ReopenableWriteCloser) func(context.Context) error {
```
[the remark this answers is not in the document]
zmt: Yes, technically. That might be cleaner in test, but I think it obscures the intent.

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,110 @@
...
+		f:    file,
+	}, nil
+}
+
+func (r *ReopenableFile) Reopen() error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.f != nil {
+		if err := r.f.Close(); err != nil {
```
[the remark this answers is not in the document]
zmt: Good point on open before close. I agree that's better.

pkg/common/log/reopen.go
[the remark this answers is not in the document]
zmt: Ack on the style nit - will fix - I'm coming from a different school.

OK. We can noop for windows. Maybe it should spit out a warning in the log  that it cannot rotate when `log_file` is set under windows? Also, who runs `logrotate` on windows?

pkg/common/log/reopen.go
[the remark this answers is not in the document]
azdagron: I'm not sure I follow. This is a function that reopens something that can be reopened on receipt of a signal. It doesn't care about writing and closing. Making the change also has the added benefit of removing functions from test fakes that aren't ever called. Otherwise, I don't see a point to the `Reopener` interface if we aren't going to use it anywhere.

pkg/common/log/reopen.go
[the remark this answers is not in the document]
azdagron: I don't think folks running on windows will have an expectation of signal based log rotation, so I'd be ok leaving out the warning.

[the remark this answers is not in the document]
zmt: Ack.

pkg/common/log/reopen.go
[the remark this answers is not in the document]
zmt: Thanks for your additional clarification. Yeah, I think I was over-indexed on aligning with the production use where this takes a `*ReopenableFile` (which happens to be a `ReopenableWriteCloser`). This is exactly the site which justifies the `Reopener` as a standalone interface (which I created instinctually). To require a `ReopenableWriteCloser` here means I only got that design 1/2 correct. ❤️

pkg/common/log/reopen_test.go
zmt: That works a charm.

pkg/common/log/reopen.go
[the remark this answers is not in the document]
zmt: Per further discussion, implementing no op for windows will involve build tags and alternate code because `syscall.SIGUSR2` is not defined for windows. I will need to find or make an environment where I can confirm it builds under windows.

azdagron:
Hi @zmt! How are things progressing getting a Windows env up an running? No rush of course but if you need one of us to push up a commit to get this building under windows, we're happy to do it. LMK!

loveyana:
@zmt I am a bit worried about the introduction of a third-party tool similar to linux logrotate in the way log rotate is supported. 

Why not consider using a solution similar to https://github.com/natefinch/lumberjack for log rotate? I think it also supports cross-platform.

Similar to the following code:
```golang
import (
	log "github.com/sirupsen/logrus"
	"time"
	"github.com/natefinch/lumberjack"
	"io"
)
func main(){
        LogFilePath := "/tmp/log/logfile.log"
	lumberjackLogrotate := &lumberjack.Logger{
		Filename:   LogFilePath,
		MaxSize:    5,  // Max megabytes before log is rotated
		MaxBackups: 90, // Max number of old log files to keep
		MaxAge:     60, // Max number of days to retain log files
		Compress:   true,
	}

	log.SetFormatter(&log.TextFormatter{FullTimestamp: true, TimestampFormat: time.RFC1123Z})
	logMultiWriter := io.MultiWriter(os.Stdout, lumberjackLogrotate)
	log.SetOutput(logMultiWriter)
}
```

pkg/common/log/reopen.go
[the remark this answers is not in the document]
MarcosDY: @zmt, I'm happy to help with the Windows stuff. May I add a commit on top of this branch with that?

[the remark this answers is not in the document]
zmt: yes - thanks - merged your PR to my PR just now

zmt:
> @zmt I am a bit worried about the introduction of a third-party tool similar to linux logrotate in the way log rotate is supported.
> 
We're not introducing a new dependency per se. In our current deployment, we already use logrotate and without this change, we are forced to use the `copytruncate` directive which is known to be lossy. I believe most SPIRE deployments do not log to files, but rather log to stdout/stderr and let journald or similar handle it on *nix systems. This change will only apply to configurations logging to a file and only change behavior on *nix to reopen the file descriptor when signaled. The process signaling to reopen could be anything, but we need this behavior to prevent potential loss of logs with logrotate `copytruncate` directive. Someone formerly on my team originally filed the #1152 issue over 2 years ago. We just now have time and priority to get to it.

> Why not consider using a solution similar to https://github.com/natefinch/lumberjack for log rotate? I think it also supports cross-platform.
> 
I think that is a question for the maintainers, but from a sysadmin perspective, I'm more in favor of standard log handling mechanisms like journald, syslog, etc. on *nix systems. Those can be more easily fed into existing log aggregation infrastructure at large installations.

zmt:
> Hi @zmt! How are things progressing getting a Windows env up an running? No rush of course but if you need one of us to push up a commit to get this building under windows, we're happy to do it. LMK!

It's not - buried in other work. Thankfully @MarcosDY fixed it up for me!

azdagron:
> I'm more in favor of standard log handling mechanisms like journald, syslog, etc. on *nix systems.

I think I'd agree with you here, @zmt.

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,91 @@
...
+	}, nil
+}
+
+func (r *ReopenableFile) Reopen() error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	newFile, err := os.OpenFile(r.name, fileFlags, fileMode)
+	if err != nil {
+		r.f = nil
```
azdagron, line 50: I don't think we want to clear this on failure to open... shouldn't we continue trying to log to the existing file?

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+
+	if r.f != nil {
+		if err := r.f.Close(); err != nil {
+			return err
```
azdagron, line 56: Do we want to tear down the server if we can't close the file descriptor for the old log file? I wonder if we shouldn't just kick off a goroutine that logs a warning...

pkg/common/log/reopen.go
azdagron: Wonder if we shouldn't kick off a goroutine to log a warning that we couldn't rotate due to failure to open the file...

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,91 @@
...
+	r.f = newFile
+	return nil
+}
+
+func (r *ReopenableFile) Write(b []byte) (n int, err error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.f == nil {
+		return 0, fmt.Errorf("%s is nil", r.name)
```
azdagron, line 70:
```suggestion
		return 0, fmt.Errorf("%s log file is not open", r.name)
```

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return r.f.Write(b)
+}
+
+func (r *ReopenableFile) Close() error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.f == nil {
+		return fmt.Errorf("%s is nil", r.name)
```
azdagron, line 81:
```suggestion
		return fmt.Errorf("%s log file has already been closed", r.name)
```

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,91 @@
...
+
+	r.f = newFile
+	return nil
+}
+
+func (r *ReopenableFile) Write(b []byte) (n int, err error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.f == nil {
```
azdagron, line 69: If we take my suggestion above, then r.f should never be nil.

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return r.f.Write(b)
+}
+
+func (r *ReopenableFile) Close() error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.f == nil {
```
azdagron, line 80: If we take my suggestion above, then r.f should never be nil.

pkg/common/log/reopen.go
zmt: > Wonder if we shouldn't kick off a goroutine to log a warning that we couldn't rotate due to failure to open the file...

Where would we log to if there is no log file - stderr?

pkg/common/log/reopen.go
zmt: Again, where would this warning be logged?

pkg/common/log/reopen.go
zmt: > I don't think we want to clear this on failure to open... shouldn't we continue trying to log to the existing file?
Yes. Seems obvious when you put it that way.

> Wonder if we shouldn't kick off a goroutine
I think we shouldn't. See my response in the other comment thread.

pkg/common/log/reopen.go
zmt: If we can't close the old log file, it will leak file descriptors. If the close fails, we may still be able to write to it. If we kick off a goroutine, we could compound the fd leak with a goroutine leak.

azdagron: We're waxing a bit theoretical here, but, as far as I understand, under no circumstances is it safe to try and use the file to write with (or even retry the close) after a failure to close. Even if the failure is EINTR, posix says it should remain open, but not all platforms behave that way (e.g. linux). Because of these reasons, Go will "close" the file. Subsequent calls to Write or Close will fail.

Kicking off a goroutine that logs shouldn't result in a leak unless the logging call blocks, which would mean other logging calls are blocking and the server would be in a terrible state already.

In any case, I don't think any of these circumstances really benefits from killing the server. I'm ok if we don't try and log the failure to close, but I still feel that we shouldn't tear down the server if that happens.

pkg/common/log/reopen.go
azdagron: What is your concern kicking off a goroutine that tries to log the failure to open? If logging is working, it is a short lived goroutine. If logging isn't working (and is blocking for whatever reason), this one goroutine will be the least of the server's problems.

zmt: I'm not totally opposed to making the attempt to log to old file descriptor in a goroutine, but do we need to do that in a goroutine? I implemented that inline.

azdagron: The problem there is that the output won't be formatted correctly e.g. with a timestamp, json encoded, etc.

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,89 @@
...
+
+func NewReopenableFile(name string) (*ReopenableFile, error) {
+	file, err := os.OpenFile(name, fileFlags, fileMode)
+	if err != nil {
+		return nil, err
+	}
+	return &ReopenableFile{
+		name:        name,
+		f:           file,
+		unsafeClose: file.Close,
```
azdagron, line 47: As written, the unsafeClose function will only ever close the originally opened file and won't result in subsequent log files being closed in later rotations. If we want to continue to hook this functionality, then this I'd suggest we change this to:

```
func closeFile(f *os.File) error {
    return f.Close()
}

type ReopenableFile struct {
    ...
    closeFunc func(*os.File) error
}

func NewReopenableFile(name string) (*ReopenableFile, error) {
    ...
    return &ReopenableFile{
        closeFunc: closeFile,
    }
}
```

Then tests can replace the closeFunc on the returned *ReopenableFile to inject the error.

zmt: Good catch. I thought I had solved the test injection. So close (pun intended).

pkg/common/log/reopen.go
```diff
@@ -0,0 +1,89 @@
...
+func (r *ReopenableFile) Reopen(logger *Logger) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	newFile, err := os.OpenFile(r.name, fileFlags, fileMode)
+	if err != nil {
+		reopenErr := fmt.Errorf("unable to reopen %s: %w", r.name, err)
+		// best effort to log error to old file descriptor
+		go logger.Error(reopenErr)
+		return reopenErr
```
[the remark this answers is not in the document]
zmt: I'm tempted to try the "Add suggestion to batch" functionality here, but I will also need to revert the `Reopen` signature change so I'll just fix it all up in the morning.

pkg/common/log/reopen_posix.go
```diff
@@ -0,0 +1,41 @@
...
+	reopener Reopener,
+	signalCh chan os.Signal,
+) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-signalCh:
+			if err := reopener.Reopen(logger); err != nil {
+				return err
```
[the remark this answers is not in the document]
zmt: That is cleaner.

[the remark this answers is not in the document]
zmt: Hmm. I need to learn the logrus spy in `spiretest` for this change. Also, given the 2 flaky tests I found on the other PR which appear to be races between that spy and `logrus`, I hope it doesn't make the error test flaky.

pkg/common/log/fakes_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"testing"
+)
+
+var (
+	_ ReopenableWriteCloser = (*fakeReopenableFile)(nil)
+)
+
+type fakeReopenableFile struct {
+	rf        *ReopenableFile
+	t         *testing.T
```
zmt, line 15: nit to self: put `t` first

pkg/common/log/reopen_windows.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// ReopenOnSignal returns a noop function compatible with RunTasks since
+// windows does not have signals as on *nix.
+func ReopenOnSignal(logger *Logger, reopener Reopener) func(context.Context) error {
+	return func(ctx context.Context) error {
+		<-ctx.Done()
+		return nil
+	}
+}
```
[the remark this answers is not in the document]
zmt: Or add something later after this lands?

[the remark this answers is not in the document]
zmt: Still have a windows lint issue (in that it timed out).

azdagron (approved):
Thanks, @zmt !

PR #2715, opened 2022-01-31
Fix linter errors related to import aliases

rturner3:
In files where the protobuf (spire-api-sdk) and API packages were imported for a given API set, the alias conventions `<type>v1` and `<type>` were used, respectively.


azdagron (commented):
In general, this change departs from the convention we have everywhere else (e.g. the server API implementation). Whatever we choose, we should be consistent.

Currently, in the service implementation test files, `<service>v1` is the named import for the API sdk (which happens to be their actually package name), and `<service>` is used for the service implementation package.

Other places (i.e. endpoints packages) don't import the API definitions. In these places, we have used `<service>v1` to name the import of the service implementation.

Although we haven't been consistent, we should be avoiding `_` in named imports, according to naming conventions we follow from [Effective Go](https://go.dev/doc/effective_go#package-names).

I'm ok with a holistic change to the way we name these packages and their respect imports within SPIRE. The API however, is following API guideline conventions for the names so we should keep the API SDK as-is. We can however adopt whatever convention we want for how we name those imports within SPIRE.

I will say that as much as possible, we should strive to keep these files trivially "go imports"-able. It's a pain to manually mess with imports.

rturner3:
> In general, this change departs from the convention we have everywhere else (e.g. the server API implementation). Whatever we choose, we should be consistent.
> 
> Currently, in the service implementation test files, `<service>v1` is the named import for the API sdk (which happens to be their actually package name), and `<service>` is used for the service implementation package.
> 
> Other places (i.e. endpoints packages) don't import the API definitions. In these places, we have used `<service>v1` to name the import of the service implementation.
> 
> Although we haven't been consistent, we should be avoiding `_` in named imports, according to naming conventions we follow from [Effective Go](https://go.dev/doc/effective_go#package-names).
> 
> I'm ok with a holistic change to the way we name these packages and their respect imports within SPIRE. The API however, is following API guideline conventions for the names so we should keep the API SDK as-is. We can however adopt whatever convention we want for how we name those imports within SPIRE.
> 
> I will say that as much as possible, we should strive to keep these files trivially "go imports"-able. It's a pain to manually mess with imports.

Sounds good. I have no intention to change the existing import alias conventions. I adapted the PR to match the conventions used in other places in the project.

azdagron (approved):
Thanks, @rturner3 !

PR #2756, opened 2022-02-08
Configure Dependabot to not raise PRs for AWS SDK v1 patch versions

rturner3:
The AWS SDK has near daily releases, and most releases do not affect our usage. Tell dependabot to chill a bit and only care about minor and major versions.


azdagron (approved):
awesome

PR #2765, opened 2022-02-09
Consolidate awssecretsmanager svidstore plugin to use only AWS SDK v2

rturner3:
The plugin is using the APIs from the v2 SDK, but also using some utilities from the v1 SDK. There are equivalent utilities in the v2 SDK, so use those instead so that we don't unnecessarily mix v1 and v2 dependencies.


azdagron (approved):
Superb.

PR #2885, opened 2022-03-24
Adding downstream support to CRD

### LINKED ISSUE #2886: Add support for downstream entries in the Spiffe ID CRD
* **Version**: Current
* **Platform**: Kubernetes
* **Subsystem**: CRD Registrar

While using the CRD resource as a means of registering secondary (nested) servers in other Kubernetes clusters, I found that there is not currently support for adding downstream identities via the CRD. I think this is a valid use case for the CRD.

noxora:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

This PR impacts the Spiffe ID CRD resource and the controller in charge of reconciling the Spiffe ID resource

**Description of change**

This PR adds a field to the CRD which allows the user to mark resource as a downstream entry. 

**Which issue this PR fixes**

fixes #2886


doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -41,3 +41,17 @@ A sample configuration:
...
+
+Some useful values are:
+
+| Value                 | Description                                                  |
+| ----------------------| ------------------------------------------------------------ |
+| .PluginName           | The name of the plugin                                       |
+| .Fingerprint          | The SHA1 fingerprint of the agent's x509 certificate         |
+| .TrustDomain          | The configured trust domain                                  |
+| .Subject.CommonName   | The common name field of the agent's x509 certificate        |
+| .Subject.SerialNumber | The serial number field of the agent's x509 certificate      |
```
faisal-memon, line 57: Is this section relevant to this PR?

faisal-memon:
@noxora Can you update the README and add a section on using the new field? https://github.com/spiffe/spire/tree/main/support/k8s/k8s-workload-registrar/mode-crd

doc/plugin_server_nodeattestor_x509pop.md
noxora: This isn't. This is from https://github.com/spiffe/spire/pull/2851 and some confusion from me. I haven't used forks in a bit and am confused because this documentation change should already be merged into main per the linked PR.

noxora:
> @noxora Can you update the README and add a section on using the new field? Or just a simple update to the SPIFFE ID Custom Resource Example section? https://github.com/spiffe/spire/blob/main/support/k8s/k8s-workload-registrar/mode-crd/README.md

@faisal-memon I've gone ahead and added to the example, I don't think it's complicated enough to warrant its own section, since it's just adding a boolean field to the CRD, but it's a good idea to clarify that it's both optional and should default to false

Outside of that, I wanted to test my changes in an environment and make sure that there isn't a nil pointer exception because I'm worried about support/k8s/k8s-workload-registrar/mode-crd/api/spiffeid/v1beta1/spiffeid_types.go being nil and then attempting to set nil when creating the spiffeID

doc/plugin_server_nodeattestor_x509pop.md
faisal-memon: Did you pull this branch while still on the other branch?

noxora: I did, I fixed it

faisal-memon:
> 
> Outside of that, I wanted to test my changes in an environment and make sure that there isn't a nil pointer exception because I'm worried about support/k8s/k8s-workload-registrar/mode-crd/api/spiffeid/v1beta1/spiffeid_types.go being nil and then attempting to set nil when creating the spiffeID

Are you worried the new field will be nil?

faisal-memon:
The changes look good for manually created entries. Do we need a mechanism for auto-generated entries to have `downstream` set?

noxora:
Yes, I was worried it would be nil. I don't think this change would impact auto-registration unless that needed to set downstream in the yaml. My other concerns were that there might be other behavior that needs to be updated that I hadn't addressed, since I don't know the entire system very well, just what I've read. 

By manually created entries, do you mean those created with yaml files or those entered through a terminal on the server?

azdagron (approved):
\o/, thanks @noxora

PR #2929, opened 2022-04-08
Update example configuration for Windows workload attestor

amartinezfayo:
Now that #2914 is fixed (thanks @azdagron!) we can update the example config of the `Windows` workload attestor to not have the `plugin_data` section, since it's not needed.


MarcosDY (approved):
LGTM!!!

PR #2934, opened 2022-04-10
Remove dead link to C++ library

georgevanburgh:
I'm assuming this library moved orgs? If not, and this is a different library to the original, it might make sense to remove this link entirely.


amartinezfayo:
Thank you @georgevanburgh for pointing this!
While the library under the HewlettPackard org is a lot more complete than the original, it's still under development and is not an officially maintained SPIFFE library (it may be in the future). For this reason, I think that it would be better to just remove the link to the archived library.
This text also links to https://spiffe.io/docs/latest/deploying/libraries/ which mentions the library from HPE and has a note indicating the status of the library. So users are able to reach the library through that link.

azdagron (approved):
\o/, thanks @georgevanburgh !

PR #2960, opened 2022-04-19
Windows support: update UpstreamAuthority `spire` plugin to work with Named Pipes

### LINKED ISSUE #2822: Windows support: update UpstreamAuthority `spire` server plugin to work with Named Pipes transport
Update the UpstreamAuthority server `spire` plugin to work with Named Pipes transport for the configured Workload API endpoint.

amartinezfayo:
Update the UpstreamAuthority `spire` plugin to also work with a configured Workload API endpoint exposed over named pipes on Windows.

Fixes #2822.


doc/plugin_server_upstreamauthority_spire.md
```diff
@@ -8,13 +8,14 @@ In the case of X509-SVID, this is easily achieved because of the chaining semant
...
-| ----------------------- | ---------------------------------------------------------------------------- |
-| server_address          | IP address or DNS name of the upstream SPIRE server in the same trust domain |
-| server_port             | Port number of the upstream SPIRE server in the same trust domain            |
-| workload_api_socket     | Path to the Workload API socket (e.g. the SPIRE Agent API socket)                  |
+| Configuration                | Description                                                                  |
+| ---------------------------- | ---------------------------------------------------------------------------- |
+| server_address               | IP address or DNS name of the upstream SPIRE server in the same trust domain |
+| server_port                  | Port number of the upstream SPIRE server in the same trust domain            |
+| workload_api_socket          | Path to the Workload API socket (Unix only; e.g. the SPIRE Agent API socket) |
+| workload_api_named_pipe_name | Pipe name of the Workload API named pipe (Windows only; e.g. pipe name of the SPIRE Agent API named pipe) |
```
evan2645, line 16: Hmm .. we signal that windows support is experimental in SPIRE core and we require experimental config options to be set ... but here they are not experimental and we don't have an experimental section. Someone looking at these docs may be led to believe that windows is fully supported, and be confused when it doesn't just work (at which point they learn they have to flip on experimental configs in core to proceed).

We should probably signal here too that this is experimental (somehow). Maybe just stating it in the docs and config example, or maybe creating a new experimental section or putting experimental in the name?

evan2645 (commented):
Just a docs related comment ... LGTM otherwise

doc/plugin_server_upstreamauthority_spire.md
amartinezfayo: Thanks for pointing this. I think that what makes more sense is to have an experimental section in the plugin config with this setting. That should be the most clear way to point that the support of all the features in Windows is experimental. I'm updating the PR with this change.

evan2645 (approved):
Thank you @amartinezfayo!

PR #2961, opened 2022-04-19
Windows support: do not panic if system procedures are not found

amartinezfayo:
SPIRE should not panic if a system procedures is not found.
This PR updates agent code to check the existence of the procedures before calling them.


azdagron (approved):
Good catch!

PR #2962, opened 2022-04-19
Windows support: fail to configure the WorkloadAttestor `unix` plugin

amartinezfayo:
The WorkloadAttestor `unix` plugin is not supported on Windows. Fail to configure it when running on Windows.


azdagron (approved):
\o/

PR #3015, opened 2022-05-01
Update OIDC Discovery Provider release configs

amartinezfayo:
The SPIRE OIDC Discovery Provider release configurations had some settings that have been deprecated and removed.
This updates the release configs so they are up to date to be used as a template.


azdagron (approved):
\o/

PR #3018, opened 2022-05-02
Windows support: make the Workload API listener publicly available

amartinezfayo:
The security descriptor associated with the named pipe listener for the Workload API had the default security descriptor.
The ACLs in the default security descriptor for a named pipe grant full control to the LocalSystem account, administrators, and the creator owner. And they only grant read access to members of the Everyone group and the anonymous account.

Everyone should have read, write and execute permissions over the Workload API. This PR updates the configuration of the Workload API listener to set that.


azdagron (approved):
Good catch.

PR #3019, opened 2022-05-03
Add windows binaries to release

MarcosDY:
Push windows artifacts when building release


azdagron (approved):
Thanks so much @MarcosDY !

PR #3070, opened 2022-05-16
Update Azure SDK packages to v1.0.0

rturner3:
This updates all the Azure SDK packages to v1.0.0. They all need to be updated together since there are breaking API changes in v1.0.0 that cause compile-time errors if not all updated in lockstep.


azdagron (approved):
Thanks for shepherding these updates, @rturner3 !

PR #3100, opened 2022-05-23
Support SHA-256 in Windows workload attestor plugin

### LINKED ISSUE #2980: Windows support: add a `sha256` selector to the "windows" workload attestor plugin
Similarly to the "unix" workload attestor, there is a need to have a `sha256` selector in the "windows" workload attestor that can attest the SHA256 digest of the workload binary.

nweedon-u:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Adding capabilities to the Windows workload attestor.

**Description of change**
Add the following to the Windows workload attestor to bring it into feature parity with the Unix workload attestor:
- `discover_workload_path` configuration option
- `workload_size_limit` configuration option
- `windows:path` workload path selector
- `windows:sha256` workload path selector

**Which issue this PR fixes**
fixes #2980


pkg/agent/plugin/workloadattestor/windows/windows_windows_test.go
```diff
@@ -104,6 +110,43 @@ func TestAttest(t *testing.T) {
 			},
 			expectCode: codes.OK,
 		},
+		{
+			name: "successful getting path and hashing process binary",
```
MarcosDY, line 114: can you add  the next test cases:
- failed to get binary path
- failed to hash binary
- binary exceeds limit size

pkg/agent/plugin/workloadattestor/windows/windows_windows_test.go
```diff
@@ -42,6 +44,10 @@ var (
 )
 
 func TestAttest(t *testing.T) {
```
MarcosDY, line 46: Can you add a TestConfigure(t *testing.T)? I think it will only have 2 test cases: `sucess to decode configuration`, `failed to decode`

pkg/common/util/hash_test.go
```diff
@@ -0,0 +1,40 @@
...
+	hash, err := GetSHA256Digest(path, 100)
+	require.NoError(t, err)
+	require.Equal(t, "1307990e6ba5ca145eb35e99182a9bec46531bc54ddf656a602c780fa0240dee", hash)
+}
+
+func Test_GetSHA256Digest_AboveLimit(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "file")
+	require.NoError(t, ioutil.WriteFile(path, []byte("some data"), 0600))
+	hash, err := GetSHA256Digest(path, 5)
+	require.Error(t, err)
```
MarcosDY, line 31: instead of require.Error, can you use: `spiretest.RequireErrorContains()` so we are sure we hit the error we wanted?

MarcosDY (commented):
Great job!!! 
I just added some comments

nweedon-u:
Thanks for the review @MarcosDY! I have updated the PR with your suggestions. :)

MarcosDY (commented):
looks great!!! just some minor comments

nweedon-u:
> looks great!!! just some minor comments

Thanks! Should be good to go. :)

MarcosDY (approved):
LGTM!

pkg/agent/plugin/workloadattestor/windows/windows_windows.go
```diff
@@ -60,12 +82,28 @@ func (p *Plugin) Attest(ctx context.Context, req *workloadattestorv1.AttestReque
 		selectorValues = addSelectorValueIfNotEmpty(selectorValues, "group_name", group)
 	}
 
+	// obtaining the workload process path and digest are behind a config flag
```
amartinezfayo, line 85: I think that this is true for the unix attestor but I'm not sure that there is an extra permissions requirement in Windows (i.e. the QueryFullProcessImageName function needs that the handle must be created with the PROCESS_QUERY_INFORMATION or PROCESS_QUERY_LIMITED_INFORMATION access right, which we have).

I still think that this should be behind a config flag due to the cost of calculating the SHA256 digest, that may not be needed by everyone. But looks like this comment needs to be updated.

PR #3159, opened 2022-06-10
Add Sigstore on workload attestor for SPIRE

willallves:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**


willallves:
Add sigstore attestor
thanks to:
- @rodrigolc
- @joaoguazzelli
- @marcusvtms
- @otaviolcs3
- @tjamir
- @matheusdefarias

doc/plugin_agent_workloadattestor_windows.md
```diff
@@ -19,26 +14,6 @@ It does so by opening an access token associated with the workload process. The
 | `windows:group_name:se_group_enabled:true` | The group name of an enabled group associated with the access token from the workload process (e.g. `windows:group_name:se_group_enabled:true:computer-or-domain\mygroup`) |
 | `windows:group_name:se_group_enabled:false` | The group name of a not enabled group associated with the access token from the workload process (e.g. `windows:group_name:se_group_enabled:false:computer-or-domain\mygroup`) |
 
-Workload path enabled selectors (available when configured with `discover_workload_path = true`):
```
amartinezfayo, line 22: This shouldn't be removed

doc/plugin_agent_nodeattestor_k8s_psat.md
```diff
@@ -50,7 +50,7 @@ volumeMounts:
       name: spire-agent
 ```
 
-A full example of this attestor is provided in [the SPIRE examples repository](https://github.com/spiffe/spire-examples/tree/main/examples/k8s/simple_psat).
```
amartinezfayo, line 53: This shouldn't be changed.

doc/plugin_agent_nodeattestor_k8s_sat.md
```diff
@@ -21,7 +21,7 @@ The main configuration accepts the following values:
 | `cluster`       | Name of the cluster. It must correspond to a cluster configured in the server plugin. |
 | `token_path`      | Path to the service account token on disk | "/var/run/secrets/kubernetes.io/serviceaccount/token" |
 
-The token path defaults to the default location kubernetes uses to place the token and should not need to be overridden in most cases.
```
amartinezfayo, line 24: This shouldn't be changed.

doc/plugin_agent_workloadattestor_windows.md
```diff
@@ -19,26 +14,6 @@ It does so by opening an access token associated with the workload process. The
...
-calculating a sha256 for large workload binaries, causing a denial-of-service.
-Defenses against this are:
-
-- disabling calculation entirely by setting `workload_size_limit` to a negative value
-- use `workload_size_limit` to enforce a limit on the binary size the
-  plugin is willing to hash. However, the same attack could be performed by spawning a
-  bunch of processes under the limit.
-  The workload API does not yet support rate limiting, but when it does, this attack can
-  be mitigated by using rate limiting in conjunction with non-negative `workload_size_limit`.
-
```
amartinezfayo, line 41: This shouldn't be removed.

doc/plugin_server_nodeattestor_k8s_psat.md
```diff
@@ -76,4 +76,4 @@ This plugin generates the following selectors:
 The node and pod selectors are only provided for label keys in the `allowed_node_label_keys` and `allowed_pod_label_keys` configurables.
 
 
-A full example of this attestor is provided in [the SPIRE examples repository](https://github.com/spiffe/spire-examples/tree/main/examples/k8s/simple_psat)
```
amartinezfayo, line 79: This shouldn't be changed.

doc/scaling_spire.md
```diff
@@ -61,9 +61,9 @@ Another use case is SPIFFE interoperability between organizations, such as betwe
 
 These multiple trust domain and interoperability use cases both require a well-defined, interoperable method for a Workload in one trust domain to authenticate a Workload in a different trust domain. Trust between the different trust domains is established by first authenticating the respective bundle endpoint, followed by retrieval of the foreign trust domain bundle via the authenticated endpoint.
 
-For additional detail on how this is achieved, refer to the following SPIFFE spec that describes the mechanism: https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Trust_Domain_and_Bundle.md#5-spiffe-bundle-endpoint
```
amartinezfayo, line 64: This shouldn't be changed.

doc/scaling_spire.md
```diff
@@ -84,7 +84,7 @@ SPIRE has a feature to programmatically authenticate on behalf of identified wor
 The SPIRE OIDC Discovery Provider retrieves a WebPKI certificate using the ACME protocol, which it uses to secure an endpoint that serves an OIDC compatible JWKS bundle and a standard OIDC discovery document. The remote OIDC authenticated service needs then to be configured to locate the endpoint and qualify the WebPKI service.   Once this configuration is in place, the remote system’s IAM policies and roles can be set to map to specific SPIFFE IDs.  The workload, in turn, will talk to the OIDC-authenticated system by sending a JWT-SVID.  The target system then fetches a JWKS from the pre-defined URI which is served by the OIDC Discovery Provider.  The target system uses the JWKS file to validate the  JWT-SVID, and if the SPIFFE ID contained within the JWT-SVID is authorized to access the requested resource, it serves the request.  The workload is then able to access the foreign remote service without possessing any credentials provided by it.
 
 For a configuration reference on the OIDC Discovery Provider, see:
-https://github.com/spiffe/spire/tree/main/support/oidc-discovery-provider
```
amartinezfayo, line 87: This shouldn't be changed.

amartinezfayo (commented):
Thank you @willallves for opening this PR!
I'm just starting with the review. I noticed that updates in go.mod / go.sum are not included in the PR. Could you please add them? Also, there are changes that I don't think that are intended to be part of the PR. I pointed them.

I see other PRs that have been opened that depend on this PR. Ideally, PRs that depend on other should be opened after the dependent PRs are merged because they will most likely need to be updated to reflect feedback.

matheusdefarias (dismissed):

doc/plugin_agent_workloadattestor_unix.md
```diff
@@ -28,8 +28,6 @@ General selectors:
 | `unix:supplementary_gid`   | **Currently only supported on linux:** The supplementary group ID of the workload (e.g. `unix:supplementary_gid:2000`)         |
 | `unix:supplementary_group` | **Currently only supported on linux:** The supplementary group name of the workload (e.g. `unix:supplementary_group:www-data`) |
 
-Workload path enabled selectors (available when configured with `discover_workload_path = true`):
-
```
amartinezfayo, line 32: This shouldn't be removed.

pkg/agent/plugin/workloadattestor/k8s/k8s_posix.go
```diff
@@ -199,8 +224,17 @@ func (p *Plugin) Attest(ctx context.Context, req *workloadattestorv1.AttestReque
 			status, lookup := lookUpContainerInPod(containerID, item.Status)
 			switch lookup {
 			case containerInPod:
+				selectors := getSelectorValuesFromPodInfo(&item, status)
+				log.Debug("Attemping to get signature info from image", status.Name)
+				sigstoreSelectors, err := p.sigstore.AttestContainerSignatures(ctx, status)
```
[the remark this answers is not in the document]
amartinezfayo: @willallves something that we discussed in the last maintainer's sync is that it would be prudent to have this feature enabled under an experimental configuration, making this an [experimental feature](https://github.com/spiffe/spire/blob/main/doc/upgrading.md#experimental-features).
The main motivation for this is that we don't really know how the changes to support the validation of images signatures through sigstore would work with existing deployments in the real world.
Could you please have the `sigstore {...}` config under an `experimental {...}` section?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+				verifyFunction: nil,
+				fetchImageManifestFunction: func(ref name.Reference, options ...remote.Option) (*remote.Descriptor, error) {
+					return &remote.Descriptor{
+						Manifest: []byte(`sometext`),
+					}, nil
+				},
+				skippedImages: nil,
+			},
+			args: args{
+				ref: func(d name.Digest, err error) name.Digest { return d }(name.NewDigest("example.com/sampleimage@sha256:5fb2054478353fd8d514056d1745b3a9eef066deadda4b90967af7ca65ce6505")),
```
MarcosDY, line 1101: it is confusing to create a function on each arg, why not to have a struct that implements this interface and we set values there?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+			},
+			want:    false,
+			wantErr: true,
+		},
+		{
```
MarcosDY, line 1121: can you add a test case to hash manifest?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+				verifyFunction:             tt.fields.verifyFunction,
+				skippedImages:              tt.fields.skippedImages,
+				fetchImageManifestFunction: tt.fields.fetchImageManifestFunction,
+			}
+			got, err := sigstore.ValidateImage(tt.args.ref)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("sigstoreImpl.ValidateImage() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if got != tt.want {
```
MarcosDY, line 1151: can you use require.Equal here?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			sigstore := &sigstoreImpl{
+				subjectAllowList: tt.fields.subjectAllowList,
+			}
+			sigstore.AddAllowedSubject(tt.args.subject)
+			if !reflect.DeepEqual(sigstore.subjectAllowList, tt.want) {
```
MarcosDY, line 1245: can you use require.Equals?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+				t.Errorf("sigstore.subjectAllowList = %v, want %v", sigstore.subjectAllowList, tt.want)
+			}
+		})
+	}
+}
+
+func TestSigstoreimpl_ClearAllowedSubjects(t *testing.T) {
```
MarcosDY, line 1252: same comment for all tests that has a single test case.

It is an overkill to use test table driving for a single test case, you must use a single t.Run instead

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
```diff
@@ -0,0 +1,426 @@
...
+		if sigSelectors != nil {
+			selectors = append(selectors, *sigSelectors)
+		}
+	}
+	return selectors
+}
+
+// SelectorValuesFromSignature extracts selectors from a signature.
+// returns a list of selectors.
+func (s *sigstoreImpl) SelectorValuesFromSignature(signature oci.Signature, containerID string) *SelectorsFromSignatures {
```
MarcosDY, line 163: I'm worried about all logs here, we create several logs and keep function to continue, why it is done this way? may we make function to fail instead?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+
+func TestSigstoreimpl_SelectorValuesFromSignature(t *testing.T) {
+	type fields struct {
+		allowListEnabled bool
+		subjectAllowList map[string]bool
+	}
+	type args struct {
+		signature oci.Signature
+	}
+	tests := []struct {
```
MarcosDY, line 1344: this function is creating several logs in case of failures but this test is not capturing them.
I think we need to capture errors to know what it is happening here

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			sigstore := &sigstoreImpl{
+				allowListEnabled: tt.fields.allowListEnabled,
+				subjectAllowList: tt.fields.subjectAllowList,
+				logger:           hclog.Default(),
+			}
+			if got := sigstore.SelectorValuesFromSignature(tt.args.signature, tt.containerID); !reflect.DeepEqual(got, tt.want) {
```
MarcosDY, line 1492: can you use require.Equal?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+				t.Errorf("sigstoreImpl.SelectorValuesFromSignature() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func Test_getBundleSignatureContent(t *testing.T) {
```
MarcosDY, line 1499: private functions must not be tested directly, and use functions that call it instead

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+				return
+			}
+			if got != tt.want {
+				t.Errorf("getBundleSignatureContent() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestSigstoreimpl_AttestContainerSignatures(t *testing.T) {
```
MarcosDY, line 1606: can you test logs generated where to know what is happening inside this function?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+		name    string
+		fields  fields
+		status  corev1.ContainerStatus
+		want    []string
+		wantErr bool
+	}{
+		{
+			name: "Attest image with signature",
+			fields: fields{
+				verifyFunction: func(context context.Context, ref name.Reference, co *cosign.CheckOpts) ([]oci.Signature, bool, error) {
```
MarcosDY, line 1635: can you verify received parameters? we need to know what we are sending to cosign libraries

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+								Payload: bundle.RekorPayload{
+									Body:           "ewogICJzcGVjIjogewogICAgInNpZ25hdHVyZSI6IHsKICAgICAgImNvbnRlbnQiOiAiTUVVQ0lRQ3llbThHY3Iwc1BGTVA3ZlRYYXpDTjU3TmNONStNanhKdzlPbzB4MmVNK0FJZ2RnQlA5NkJPMVRlL05kYmpIYlVlYjBCVXllNmRlUmdWdFFFdjVObzVzbUE9IgogICAgfQogIH0KfQ==",
+									LogID:          "samplelogID",
+									IntegratedTime: 12345,
+								},
+							},
+						},
+					}, true, nil
+				},
+				fetchImageManifestFunction: func(ref name.Reference, options ...remote.Option) (*remote.Descriptor, error) {
```
MarcosDY, line 1649: can you verify received parameters? we need to know what we are sending to cosign libraries

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+
+		return co
+	}
+
+	tests := []struct {
+		name    string
+		fields  fields
+		status  corev1.ContainerStatus
+		want    []string
+		wantErr bool
```
MarcosDY, line 1630: we must validate error message to know why it is failing

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+				sigstorecache:              NewCache(maximumAmountCache),
+				checkOptsFunction:          emptyCheckOptsFunction,
+				logger:                     hclog.Default(),
+			}
+			got, err := sigstore.AttestContainerSignatures(context.Background(), &tt.status)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("sigstoreImpl.AttestContainerSignatures() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
MarcosDY, line 1728: avoid using DeepEqual

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+	}
+	type args struct {
+		rekorURL string
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		want    url.URL
+		wantErr bool
```
MarcosDY, line 1747: we must validate error message

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			sigstore := &sigstoreImpl{
+				rekorURL: tt.fields.rekorURL,
+			}
+			if err := sigstore.SetRekorURL(tt.args.rekorURL); (err != nil) != tt.wantErr {
+				t.Errorf("sigstoreImpl.SetRekorURL() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if !reflect.DeepEqual(sigstore.rekorURL, tt.want) {
```
MarcosDY, line 1822: avoid using DeepEqual

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"testing"
+)
+
+var (
+	selectors1 = Item{
+		Key: "signature1",
+		Value: []SelectorsFromSignatures{
+			{
+
+				Subject:        "spirex1@example.com",
```
MarcosDY, line 16:
```suggestion
				Subject:        "spirex1@example.com",
```

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,209 @@
...
+			},
+		},
+	}
+
+	selectors2 = Item{
+		Key: "signature2",
+		Value: []SelectorsFromSignatures{
+			{
+
+				Subject:        "spirex2@example.com",
```
MarcosDY, line 29:
```suggestion
			{
				Subject:        "spirex2@example.com",
```

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,209 @@
...
+				Content:        "content4",
+				LogID:          "4444444444444444",
+				IntegratedTime: "4444444444444444",
+			},
+		},
+	}
+)
+
+func TestNewCache(t *testing.T) {
+	tests := []struct {
```
MarcosDY, line 63: test table driving is an overkill for a single test case, use a single t.Run instead

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,209 @@
...
+			name:         "Existing entry",
+			want:         &selectors1,
+			key:          selectors1.Key,
+			errorMessage: "An existing items key's should return the existing item",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := cacheInstance.GetSignature(tt.key); !reflect.DeepEqual(got, tt.want) {
```
MarcosDY, line 128: avoid using DeepEqual

MarcosDY, line 128: this test validates that the return but not what happen with items, we must test all what is happening, if not we may miss some errors

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	tests := []struct {
+		name       string
+		item       *Item
+		wantLength int
+		wantKey    string
+		wantValue  *Item
+	}{
+		{
```
MarcosDY, line 153: this progresive approach using the same cache to test different test cases can result in unexpected behavior, 
all unit test must not depends of previous cases.

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,209 @@
...
+			wantKey:    selectors3.Key,
+			wantValue:  &selectors3Updated,
+		},
+	}
+
+	putKeys := 0
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cacheInstance.PutSignature(*tt.item)
+			putKeys++
```
MarcosDY, line 194: this counter must not be used, you can rely on test case name

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,209 @@
...
+			gotLen := cacheInstance.items.Len()
+			if gotLen != tt.wantLength {
+				t.Errorf("Item count should be %v after putting %v keys", tt.wantLength, putKeys)
```
MarcosDY, line 197: what is the actual lenght?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+			gotItem, present := m[tt.wantKey]
```
MarcosDY, line 199: what happens with push front it is not tested in this unit test

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,209 @@
...
+			if !present {
+				t.Errorf("Key put but not found: %v", tt.wantKey)
+			}
+
+			if !reflect.DeepEqual(gotItem.item, tt.wantValue) {
```
MarcosDY, line 204: avoud using deepEqual

pkg/agent/plugin/workloadattestor/k8s/k8s_posix.go
```diff
@@ -199,8 +239,19 @@ func (p *Plugin) Attest(ctx context.Context, req *workloadattestorv1.AttestReque
 			status, lookup := lookUpContainerInPod(containerID, item.Status)
 			switch lookup {
 			case containerInPod:
+				selectors := getSelectorValuesFromPodInfo(&item, status)
+				if p.config.EnableSigstore {
+					log.Debug("Attemping to get signature info for container", telemetry.ContainerName, status.Name)
+					sigstoreSelectors, err := p.sigstore.AttestContainerSignatures(ctx, status)
+					if err != nil {
+						log.Error("Error retrieving signature payload: ", "error", err)
```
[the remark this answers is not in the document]
amartinezfayo: Although we have a few exceptions, we have traditionally followed an "all or nothing" approach, where we return an error if there is a failure collecting one of the selectors. Otherwise, it's difficult to assert why the selector was not included. Was it omitted because there was an error or because it was not discovered? 
Particularly in this case, since the user explicitly enabled the sigstore functionality to get to this point, I would suggest that we fail if there is an error retrieving the signature payload.

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,1827 @@
+package sigstore
```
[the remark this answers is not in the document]
joaoguazzelli: as we discussed in slack it will be kept as sigstore because of the hooks

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
joaoguazzelli: as we talked in slack this is no longer necessary since the function is now returning errors

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,444 @@
...
+
+func TestNewCache(t *testing.T) {
+	want := &cacheImpl{
+		size:     3,
+		items:    list.New(),
+		mutex:    sync.RWMutex{},
+		itemsMap: make(map[string]MapItem),
+	}
+	got := NewCache(3)
+	require.Equal(t, want, got, "NewCache() = %v, want %v", got, want)
```
MarcosDY, line 84: same comment that in the another test class, why you required this message? is not regular message enough? regular message will display a rich diff error, and it will already have test name.
the very same comment for all cases

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,444 @@
...
+		itemsMap: m,
+	}
+
+	tests := []struct {
+		name         string
+		want         *Item
+		key          string
+		errorMessage string
+	}{
+		{
```
MarcosDY, line 113: this test case does not include a success scenario and more important how it is affecting the list.

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,444 @@
...
+
+	tests := []struct {
+		name       string
+		item       *Item
+		wantLength int
+		wantKey    string
+		wantValue  *Item
+	}{
+		{
+			name:       "Put first element",
```
MarcosDY, line 152: this progressive approach can result on unexpected results on test cases, each test case must not affect the next one.

So I think you will need to create a new cache on each test case, and provide that cache with the values you need to have before it runs

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,444 @@
...
+			wantValue:  &selectors3Updated,
+		},
+	}
+
+	for _, tt := range tests {
+		cacheInstance.PutSignature(*tt.item)
+		gotLen := cacheInstance.items.Len()
+		if gotLen != tt.wantLength {
+			t.Errorf("Item count should be %v in test case %q", tt.wantLength, tt.name)
+		}
```
MarcosDY, line 193:
```suggestion
                 require.Len(t, cacheInstance.items, tt.WantLength)
```

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,444 @@
...
+		gotItem, present := m[tt.wantKey]
+		if !present {
+			t.Errorf("Key put but not found: %v", tt.wantKey)
+		}
```
MarcosDY, line 197: require.True(t, present, "key not found")

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
```diff
@@ -0,0 +1,444 @@
...
+		require.Equal(t, gotItem.item, tt.wantValue, "Value different than expected. \nGot: %v \nWant:%v", gotItem.item, tt.wantValue)
+	}
+}
+
+func TestCacheimpl_CheckOverflowAndUpdates(t *testing.T) {
```
MarcosDY, line 202: this approch can result on unexpected issues, and at the same time contains steps that are not requied.
In the end you want to test overflow and in a full list how it is working, why not to create a full list and put a new one dending in what you want to test?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstorecache_test.go
willallves: `%!s(int=1)}" could not be applied builtin len()`
[link for failed CI](https://github.com/willallves/spire/actions/runs/3221520111/jobs/5269567154#step:5:76)
Reversed commit

pkg/agent/plugin/workloadattestor/k8s/k8s_posix.go
```diff
@@ -175,3 +211,25 @@ func canonicalizePodUID(uid string) types.UID {
...
+	if c.SkippedImages != nil {
+		client.AddSkippedImage(c.SkippedImages)
+	}
+	client.EnableAllowSubjectList(c.AllowedSubjectListEnabled)
+	client.SetLogger(log)
+	client.ClearAllowedSubjects()
+	for _, subject := range c.AllowedSubjects {
+		client.AddAllowedSubject(subject)
+	}
+	rekorURL := "http://rekor.sigstore.dev/" // default rekor url
```
amartinezfayo, line 227: This should be https://rekor.sigstore.dev/

doc/plugin_agent_workloadattestor_k8s.md
```diff
@@ -40,24 +40,53 @@ server name validation against the kubelet certificate.
...
+| `allowed_subjects_list` | The list of allowed subjects enabled by `enable_allowed_subjects_list` each entry represents subject e-mail. Defaults to empty list. |
+| `rekor_url` | The URL for the rekor STL Server to use with cosign. Defaults to 'http://rekor.sigstore.dev/', Rekor's public instance. |
+
+> **Note** Cosign discourages the use of image tags for referencing docker images, and this plugin does not support attestation of sigstore selectors for workloads running on containers using tag-referenced images, which will then fail attestation for both sigstore and k8s selectors. In cases where this is necessary, add the digest string for the image in the `skip_signature_verification_image_list` setting (eg. `"sha256:abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789"`). Note that sigstore signature attestation will still not be performed, but this will allow k8s selectors to be returned, along with the `"k8s:sigstore-validation:passed"` selector.  
+
+> **Note** Since the Spire Agent can also go through workload attestation, it will also need to be included in the skip list if either its image is not signed or has a digest reference string.
```
amartinezfayo, line 71:
```suggestion
> **Note** Since the SPIRE Agent can also go through workload attestation, it will also need to be included in the skip list if either its image is not signed or has a digest reference string.  
```

doc/plugin_agent_workloadattestor_k8s.md
```diff
@@ -40,24 +40,53 @@ server name validation against the kubelet certificate.
...
+| `certificate_path` | The path on disk to client certificate used for kubelet authentication |
+| `private_key_path` | The path on disk to client key used for kubelet authentication |
+| `use_anonymous_authentication` | If true, use anonymous authentication for kubelet communication |
+| `node_name_env` | The environment variable used to obtain the node name. Defaults to `MY_NODE_NAME`. |
+| `node_name` | The name of the node. Overrides the value obtained by the environment variable specified by `node_name_env`. |
+| `experimental` |  The experimental options that are subject to change or removal. |
+
+| Experimental options | Description |
+| -------------------- | ----------- |
+| `sigstore` | Sigstore options. Options described below. |
```
amartinezfayo, line 60: I would also reference the "Sigstore workload attestor for SPIRE" section here.

doc/plugin_agent_workloadattestor_k8s.md
```diff
@@ -40,24 +40,53 @@ server name validation against the kubelet certificate.
...
+
+> **Note** Since the Spire Agent can also go through workload attestation, it will also need to be included in the skip list if either its image is not signed or has a digest reference string.  
+
+> **Note** The sigstore project contains a transparency log called Rekor that provides an immutable, tamper-resistant ledger to record signed metadata to an immutable record. While it is possible to run your own instance, a public instance of rekor is available at rekor.sigstore.dev, and cosign defaults to using the public instance.
+
+### Sigstore workload attestor for SPIRE
```
amartinezfayo, line 75: A "Platform support" sub-section should be added to note that this is only supported on Unix systems.
Something like:
```
#### Platform support

This capability is only supported on Unix systems.
```

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
```diff
@@ -0,0 +1,453 @@
...
+
+func (s *sigstoreImpl) SetRekorURL(rekorURL string) error {
+	if rekorURL == "" {
+		return errors.New("rekor URL is empty")
+	}
+	rekorURI, err := url.Parse(rekorURL)
+	if err != nil {
+		return fmt.Errorf("failed parsing rekor URI: %w", err)
+	}
+	if rekorURI.Scheme != "" && rekorURI.Scheme != "https" && rekorURI.Scheme != "http" {
```
amartinezfayo, line 339: We should only allow secure transports, http should not be allowed.

doc/plugin_agent_workloadattestor_k8s.md
```diff
@@ -40,24 +40,53 @@ server name validation against the kubelet certificate.
...
+| Experimental options | Description |
+| -------------------- | ----------- |
+| `sigstore` | Sigstore options. Options described below. |
+
+| Sigstore options | Description |
+| ---------------- | ----------- |
+| `skip_signature_verification_image_list` | The list of images, described as digest hashes, that should be skipped in signature verification. Defaults to empty list. |
+| `enable_allowed_subjects_list` | Enables a list of allowed subjects that are trusted and are allowed to sign container images artificats. Defaults to 'false'. If true and `allowed_subjects_list` is empty, no workload will pass signature validation. |
+| `allowed_subjects_list` | The list of allowed subjects enabled by `enable_allowed_subjects_list` each entry represents subject e-mail. Defaults to empty list. |
+| `rekor_url` | The URL for the rekor STL Server to use with cosign. Defaults to 'http://rekor.sigstore.dev/', Rekor's public instance. |
```
haydentherapper, line 67: What is STL?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
```diff
@@ -0,0 +1,453 @@
...
+	}
+	if rekorURL.Path == "" {
+		return nil, errors.New("rekor URL path is empty")
+	}
+
+	rootCerts, err := fulcio.GetRoots()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get fulcio root certificates: %w", err)
+	}
+
```
haydentherapper, line 109: I would add `co.IntermediateCerts, err = fulcio.GetIntermediates()`, since the TUF root stores both the roots and intermediates.

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
```diff
@@ -0,0 +1,453 @@
...
+	if selectors.LogID != "" {
+		selectorsString = append(selectorsString, fmt.Sprintf("%s:image-signature-logid:%s", containerID, selectors.LogID))
+	}
+	if selectors.IntegratedTime != "" {
+		selectorsString = append(selectorsString, fmt.Sprintf("%s:image-signature-integrated-time:%s", containerID, selectors.IntegratedTime))
+	}
+	return selectorsString
+}
+
+func certSubject(c *x509.Certificate) string {
```
haydentherapper, line 422: FYI, we added another subject recently that could be stored in a different x.509 extension, see https://github.com/sigstore/cosign/blob/983c36449ad0b97130bf0392ffda8ce0158de4e3/pkg/signature/keys.go#L238

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
```diff
@@ -0,0 +1,453 @@
...
+	Subject        string
+	Content        string
+	LogID          string
+	IntegratedTime string
+}
+
+func New(cache Cache, logger hclog.Logger) Sigstore {
+	return &sigstoreImpl{
+		functionHooks: sigstoreFunctionHooks{
+			verifyFunction:             cosign.VerifyImageSignatures,
```
haydentherapper, line 79: Happy to see this being used rather than reimplementing verification logic! A small nit, it was hard to tell that this was being used, it may be worth adding a comment around here noting the importance of `verifyFunction`

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
```diff
@@ -0,0 +1,453 @@
...
+
+	rootCerts, err := fulcio.GetRoots()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get fulcio root certificates: %w", err)
+	}
+
+	co := &cosign.CheckOpts{
+		// Set the rekor client
+		RekorClient: rekor.NewHTTPClientWithConfig(nil, rekor.DefaultTransportConfig().WithBasePath(rekorURL.Path).WithHost(rekorURL.Host)),
+		RootCerts:   rootCerts,
```
haydentherapper, line 113: I would also add `EnforceSCT: true` - This adds a requirement that the Fulcio certificate was written to the public CT (certificate transparency, separate from Rekor) log. You will likely need to make this configurable though, for private deployments that don't use CT.

haydentherapper, line 113: You can add `CertIdentity` and `CertOidcIssuer` as check options to delegate the check for identity to Cosign's validation function.

haydentherapper (commented):
Hi! I'm a maintainer from Sigstore, someone pointed me to this PR. Left a few comments. Very happy to see `cosign.VerifyImageSignatures` being used here!

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
```diff
@@ -0,0 +1,453 @@
...
+	s.allowListEnabled = flag
+}
+
+func (s *sigstoreImpl) AttestContainerSignatures(ctx context.Context, status *corev1.ContainerStatus) ([]string, error) {
+	skip, err := s.ShouldSkipImage(status.ImageID)
+	if err != nil {
+		return nil, fmt.Errorf("failed attesting container signature: %w", err)
+	}
+	if skip {
+		return []string{signatureVerifiedSelector}, nil
```
lumjjb, line 294: should another annotation also be added in the instance that an image is skipped.. Surfacing this information would be helpful from an audit perspective

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
```diff
@@ -0,0 +1,453 @@
...
+		if sigSelectors != nil {
+			selectors = append(selectors, *sigSelectors)
+		}
+	}
+	return selectors
+}
+
+// SelectorValuesFromSignature extracts selectors from a signature.
+// returns a list of selectors.
+func (s *sigstoreImpl) SelectorValuesFromSignature(signature oci.Signature, containerID string) (*SelectorsFromSignatures, error) {
```
lumjjb, line 183: nit : not used?

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
joaoguazzelli: I don't think that's necessary, because the only cert we use are the root certs. I'll message you in slack to better discuss this issue

pkg/agent/plugin/workloadattestor/k8s/k8s_posix.go
```diff
@@ -175,3 +211,25 @@ func canonicalizePodUID(uid string) types.UID {
...
+	client.EnableAllowSubjectList(c.AllowedSubjectListEnabled)
+	client.SetLogger(log)
+	client.ClearAllowedSubjects()
+	for _, subject := range c.AllowedSubjects {
+		client.AddAllowedSubject(subject)
+	}
+	rekorURL := "https://rekor.sigstore.dev/" // default rekor url
+	if c.RekorURL != nil {
+		rekorURL = (*c.RekorURL)
+	}
```
dennisgove, line 230: Is it necessary to include this default URL in the compiled binary? While one is able to to override that value, if it's not overridden (on purpose or by mistake) then in-use images will become public knowledge. The act of pulling image signatures from https://rekor.sigstore.dev has the add-on effect of telling https://rekor.sigstore.dev  that a particular `image#tag` is in use. That information can be used to inform multiple attack vectors.

I would prefer if this URL is a required configuration value with no default. This minor change will cause every organization using Sigstore during workload attestation to make an informed decision about this setting, and will make it apparent when there are calls being made to an external URL. Some organizations will choose to not use the public endpoint and will instead stand up their own instance. But by making that a necessary choice it forces the conversation and decision.

As coded now, those outgoing calls may come as an unintended surprise.

conf/agent/agent_full.conf
```diff
@@ -359,6 +359,30 @@ plugins {
             # node_name: The name of the node. Overrides the value obtained by
             # the environment variable specified by node_name_env.
             # node_name = ""
+
+            # experimental: Experimental features.
+            experimental {
+                # sigstore: sigstore options. Enables signature checking.
+                # sigstore {
+                    # rekor_url: The URL for the rekor STL Server to use with cosign.
+                    # rekor_url = "https://rekor.sigstore.dev"
```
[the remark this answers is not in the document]
rodrigolc: We are removing the Rekor public STL URL as a default, so the bundle will not be included. Good catch though!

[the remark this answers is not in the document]
rodrigolc: as per https://github.com/spiffe/spire/pull/3159#discussion_r1023305037

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
rodrigolc: That's a good point, we won't be including this change on initial release, but we will be following it up on an issue at this PR's origin repo https://github.com/willallves/spire/issues/173

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore_test.go
```diff
@@ -0,0 +1,2045 @@
...
+					fetchBinding:     createFetchFunction(&remote.Descriptor{Manifest: []byte("sometext")}, nil),
+					checkOptsBinding: createCheckOptsFunction(defaultCheckOpts, nil),
+				},
+				rekorURL: rekorDefaultURL(),
+			},
+			args: args{
+				imageName: "docker-registry.com/some/image@sha256:5fb2054478353fd8d514056d1745b3a9eef066deadda4b90967af7ca65ce6505",
+			},
+			wantedFetchArguments: fetchFunctionArguments{
+				called:  true,
```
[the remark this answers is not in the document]
rodrigolc: > what is the use of `called` attributes? are you validating that the function was called? do you want to test if that function is wanted to be called and assert it in the function body?

Yes, the created functions will set called as true on the args struct, so we can test the functions are being called along with testing the args received by the functions are the ones we want

MarcosDY (commented):
Great work!! some minor comments

pkg/agent/plugin/workloadattestor/k8s/sigstore/sigstore.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+	return certOIDCProvider(cert)
+}
+
+func certOIDCProvider(cert *x509.Certificate) (string, error) {
+	if cert == nil {
+		return "", errors.New("certificate is nil")
+	}
+	// OIDC token issuer Object Identifier
+	objectIdentifier := asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 57264, 1, 1}
```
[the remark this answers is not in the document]
rodrigolc: This is the OID for the OIDC token issuer, it's a registration number for certificate extensions (in this case, sigstore owns OID "1, 3, 6, 1, 4, 1, 57264" and its child OIDs, in this case the OIDC Token Issuer), which we need to retrieve the issuer for the allowlist. In any case, this will be moved to a constant.

MarcosDY (approved):
Thanks!!! this looks good!

MarcosDY (approved):

amartinezfayo (approved):
Thanks to all that contributed to this effort!

Issue #2940, opened 2022-04-12
[RFC] SPIRE Agent Cache Redesign Proposal


prasadborole1:
###  Background

The [RFC](https://github.com/spiffe/spire/issues/2593) provides the background and discusses the scaling problem caused due to current SPIRE Agent cache design which stores X.509-SVIDs. This proposal builds on top of approaches discussed in [RFC](https://github.com/spiffe/spire/issues/2593) and provides a more detailed redesign proposal.

###  Existing Behavior
All SPIRE Agent RPCs and SPIRE Agent-Server sync implementations are based on the fact that all authorized entries and corresponding SVIDs are cached locally. The existing cache implementation is as following:
```
type Cache struct {
	…
	// records holds the records for registration entries, keyed by registration entry ID
	records map[string]*cacheRecord

	// selectors holds the selector indices, keyed by a selector key
	selectors map[selector]*selectorIndex
}

type selectorIndex struct {
	// subs holds the subscriptions related to this selector
	subs map[*subscriber]struct{}

	// records holds the cache records related to this selector
	records map[*cacheRecord]struct{}
}

type cacheRecord struct {
	entry *common.RegistrationEntry
	svid  *X509SVID

	// this is unused today
	subs  map[*subscriber]struct{}
}
```
#### SPIRE Agent-Server sync
During the periodic sync, SPIRE Agent fetches all authorized entries and bundles from the SPIRE Server. It calculates the missing/deleted entries and expiring SVIDs and updates them in the cache. The subscribers of these affected SVIDs are notified accordingly.
####  SPIRE Agent RPCs
SPIRE Agent supports 2 types of RPCs:
1. Streaming RPCs: The RPCs like FetchX509SVID subscribe to cache changes based on the selector set and wait for updates to the stream until it is closed.
2. FetchJWTSVID: The RPCs like FetchJWTSVID(unary) retrieve SPIFFE IDs from cache based on the selector set so that they can prepare relevant JWT-SVIDs

![SPIRE Server-Agent Sync](https://user-images.githubusercontent.com/55859111/163025944-60196a04-606b-48fa-9210-00f538c5eab3.jpeg)

In a nutshell, SVIDs get added/updated/removed to/from cache during SPIRE Agent-Server sync. The RPCs which need SVIDs, are listening on grpc streams for SVID updates. Some RPCs(like FetchJWTSVID) need details about SPIFFE IDs/ authorized registration entries which the Agent is responsible for in order to validate request parameters.

###  Proposal
[...]

PR #3181, opened 2022-06-22
Implement LRU cache for storing SVIDs in SPIRE Agent

prasadborole1:
Updating SPIRE agent SVID cache to be LRU cache. This cache has experimental config fields like `MaxSvidCacheSize` and `SVIDCacheExpiryPeriod`.
 1. Size limit of SVID cache is a soft limit which means if SVID has a subscriber present then that SVID is never removed from cache.
 2. Least recently used SVIDs are removed from cache only after the cache expiry period has passed. This is done to reduce the overall cache churn.
 3. Last access timestamp for SVID cache entry is updated when a new subscriber is created.
 4. When a new subscriber is created and if there is a cache miss then subscriber needs to wait for next SVID sync event to receive WorkloadUpdate with newly minted SVID

More context: #2940

### Testing:
In addition to new integration test, also tested locally with 9k registrations per agent and validated:
1. SVIDs are properly getting cached as per subscriber selector set
2. SVIDs are getting removed after expiry
3. Tested with workload associated with 8k entries, cli command fails with expected error codes `DeadlineExceeded` or `ResourceExhausted` due to huge response size. But validated that SVIDs cache is properly getting constructed.


azdagron (dismissed):
Thanks much!

azdagron:
Sorry, I accidentally accepted this PR by mistake.

cmd/spire-agent/cli/api/common.go
```diff
@@ -73,7 +73,7 @@ func adaptCommand(env *cli.Env, clientsMaker workloadClientMaker, cmd command) *
 		clientsMaker: clientsMaker,
 		cmd:          cmd,
 		env:          env,
-		timeout:      cli.DurationFlag(time.Second),
+		timeout:      cli.DurationFlag(2 * time.Second),
```
amoore877, line 76: nit: would be nice to extract out timing to a constant at top of file

cmd/spire-agent/cli/run/run.go
```diff
@@ -400,6 +402,17 @@ func NewAgentConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool)
 		}
 	}
 
+	if c.Agent.Experimental.MaxSvidCacheSize != 0 {
```
amoore877, line 405:
```suggestion
	if c.Agent.Experimental.MaxSvidCacheSize > 0 {
```

amoore877: and then unit test negative value rejection probably?

maybe negative values should be logged and/or error out since it means operator has misconfigured

pkg/agent/api/delegatedidentity/v1/service.go
```diff
@@ -82,24 +82,24 @@ func (s *Service) isCallerAuthorized(ctx context.Context, log logrus.FieldLogger
 		}
 	}
 
-	identities := s.manager.MatchingIdentities(callerSelectors)
-	numRegisteredIDs := len(identities)
+	entries := s.manager.MatchingRegistrationEntries(callerSelectors)
+	numRegisteredIDs := len(entries)
```
amoore877, line 86:
```suggestion
	numRegisteredEntries := len(entries)
```
the thing being measured has slightly changed

pkg/agent/manager/manager.go
```diff
@@ -94,7 +96,8 @@ type manager struct {
 
 	// backoff calculator for fetch interval, backing off if error is returned on
 	// fetch attempt
-	backoff backoff.BackOff
+	backoff         backoff.BackOff
```
amoore877, line 99: It may be confusing having a general `backoff` and then a specific `svidSyncBackoff`
Can we name the former something more specific?

test/integration/setup/debugagent/main.go
```diff
@@ -52,26 +54,40 @@ func run() error {
...
 
 func agentEndpoints(ctx context.Context) error {
+	s,err := retrieveDebugPage(ctx)
+	if err == nil {
+		log.Printf("Debug info: %s", string(s))
+	}
+	return nil
+}
+
+func printDebugPage(ctx context.Context) error {
```
amoore877, line 64: maybe add a comment to `printDebugPage` that this is here in support of testing

pkg/agent/manager/cache/cache.go
```diff
@@ -75,6 +81,22 @@ type X509SVID struct {
 // selector it encounters. Each selector index tracks the subscribers (i.e
 // workloads) and registration entries that have that selector.
 //
+// The LRU-like SVID cache has configurable size limit and expiry period.
+// 1. Size limit of SVID cache is a soft limit which means if SVID has a subscriber present then
```
amoore877, line 85:
```suggestion
// 1. Size limit of SVID cache is a soft limit. If SVID has a subscriber present then
```

pkg/agent/manager/cache/cache.go
```diff
@@ -75,6 +81,22 @@ type X509SVID struct {
...
+//    that SVID is never removed from cache.
+// 2. Least recently used SVIDs are removed from cache only after the cache expiry period has passed.
+//    This is done to reduce the overall cache churn.
+// 3. Last access timestamp for SVID cache entry is updated when a new subscriber is created
+// 4. When a new subscriber is created and if there is a cache miss
```
amoore877, line 90:
```suggestion
// 4. When a new subscriber is created and there is a cache miss
```

pkg/agent/manager/cache/cache.go
```diff
@@ -75,6 +81,22 @@ type X509SVID struct {
...
+//    then subscriber needs to wait for next SVID sync event to receive WorkloadUpdate with newly minted SVID
+//
+// The advantage of above approach is that if agent has entry count less than cache size
+// then all SVIDs are cached at all times. If agent has entry count greater than cache size then
+// subscribers will continue to get SVID updates (potential delay for first WorkloadUpdate if cache miss)
+// and least used SVIDs will be removed from cache which will save memory usage.
+// It will allow agent to support large number of registrations.
```
amoore877, line 97:
```suggestion
// This allows agent to support environments where the active simultaneous workload count is a small percentage of the large number of registrations assigned to the agent.
```

pkg/agent/manager/cache/cache.go
```diff
@@ -113,22 +136,40 @@ type Cache struct {
 	// selectors holds the selector indices, keyed by a selector key
 	selectors map[selector]*selectorIndex
 
-	// staleEntries holds stale registration entries
+	// staleEntries holds stale or new registration entries which require new SVID to be stored in cache
```
amoore877, line 139: maybe we should rename the var given its new intention

pkg/agent/manager/cache/cache.go
```diff
@@ -113,22 +136,40 @@ type Cache struct {
...
+
+	// maxSVIDCacheSize is a soft limit of max number of SVIDs that would be stored in cache
+	maxSvidCacheSize int
+
+	// svidCacheExpiryPeriod is a period after which svids that don't have subscribers will be removed from cache
+	svidCacheExpiryPeriod time.Duration
 }
 
-// StaleEntry holds stale entries with SVIDs expiration time
+// StaleEntry holds stale or outdated entries which require new SVID with old SVIDs expiration time (if present)
```
amoore877, line 155: same comment as for the `staleEntries` map. We may want to rename this struct for clarity

pkg/agent/manager/cache/cache.go
```diff
@@ -113,22 +136,40 @@ type Cache struct {
...
 	// Entry stale registration entry
 	Entry *common.RegistrationEntry
 	// SVIDs expiration time
 	ExpiresAt time.Time
 }
 
-func New(log logrus.FieldLogger, trustDomain spiffeid.TrustDomain, bundle *Bundle, metrics telemetry.Metrics) *Cache {
+func New(log logrus.FieldLogger, trustDomain spiffeid.TrustDomain, bundle *Bundle, metrics telemetry.Metrics,
+	maxSvidCacheSize int, svidCacheExpiryPeriod time.Duration, clk clock.Clock) *Cache {
+	if maxSvidCacheSize == 0 {
```
amoore877, line 165:
```suggestion
	if maxSvidCacheSize > 0 {
```
and maybe logging / error on negative value?

same for `svidCacheExpiryPeriod`

pkg/agent/manager/cache/cache.go
```diff
@@ -445,6 +536,104 @@ func (c *Cache) GetStaleEntries() []*StaleEntry {
...
+func (c *Cache) syncSVIDs() (map[string]struct{}, []record) {
+	activeSubs := make(map[string]struct{})
+	lastAccessTimestamps := make([]record, len(c.records))
+
+	i := 0
+	for id, record := range c.records {
+		for _, sel := range record.entry.Selectors {
+			if index, ok := c.selectors[makeSelector(sel)]; ok && index != nil {
+				if len(index.subs) > 0 {
+					if _, ok := c.svids[record.entry.EntryId]; !ok {
```
amoore877, line 609: Since there's 5 nested operations, I think a comment block about the logic is warranted.

pkg/agent/manager/cache/cache_test.go
```diff
@@ -644,6 +620,183 @@ func TestSubscriberNotNotifiedOnOverlappingSVIDChanges(t *testing.T) {
 	assertNoWorkloadUpdate(t, sub)
 }
 
+func TestSVIDCacheExpiry(t *testing.T) {
+	clk := clock.NewMock()
+	cache := newTestCacheWithConfig(10, 1*time.Minute, clk)
+
+	clk.Add(1 * time.Second)
+	foo := makeRegistrationEntry("FOO", "A")
```
amoore877, line 628: for readability, I think it would be helpful to keep foo/bar naming consistent; so changing `subA` and selectors for instance to `subFoo` and so on

pkg/agent/manager/cache/cache_test.go
```diff
@@ -644,6 +620,183 @@ func TestSubscriberNotNotifiedOnOverlappingSVIDChanges(t *testing.T) {
...
+	}
+
+	// Move clk by 58 sec so that a minute has passed since last foo was accessed
+	// svid for foo should be deleted
+	clk.Add(58 * time.Second)
+	cache.UpdateEntries(updateEntries, nil)
+
+	subA = cache.SubscribeToWorkloadUpdates(makeSelectors("A"))
+	defer subA.Finish()
+	//cache.NotifyBySelectorSet(makeSelectors("A"))
```
amoore877, line 694:
```suggestion
```

pkg/agent/manager/cache/cache_test.go
```diff
@@ -644,6 +620,183 @@ func TestSubscriberNotNotifiedOnOverlappingSVIDChanges(t *testing.T) {
...
+	updateEntries.RegistrationEntries[foo.EntryId] = foo
+
+	cache.UpdateEntries(updateEntries, nil)
+
+	// Create a subscriber for foo
+	subA := cache.SubscribeToWorkloadUpdates(foo.Selectors)
+	defer subA.Finish()
+	require.Len(t, cache.GetStaleEntries(), 0)
+
+	// After SyncSVIDsWithSubscribers foo should be marked as stale
```
amoore877, line 776:
```suggestion
	// After SyncSVIDsWithSubscribers foo should be marked as stale, requiring signing
```

pkg/agent/manager/cache/cache_test.go
```diff
@@ -689,6 +842,36 @@ func BenchmarkCacheGlobalNotification(b *testing.B) {
...
+}
+
+func createUpdateEntries(numEntries int, bundles map[spiffeid.TrustDomain]*bundleutil.Bundle) *UpdateEntries {
+	updateEntries := &UpdateEntries{
+		Bundles:             bundles,
+		RegistrationEntries: make(map[string]*common.RegistrationEntry, numEntries),
+	}
+
+	for i := 0; i < numEntries; i++ {
+		entryID := fmt.Sprintf("00000000-0000-0000-0000-%012d", i)
```
amoore877, line 864: maybe specify a limit in comments (or fail test) if `numEntries` is too large and thus will lead to dupe reg IDs?

cmd/spire-agent/cli/run/run.go
prasadborole1: added check and unit tests for negative values.

pkg/agent/manager/cache/cache.go
prasadborole1: Added unit test here but ignoring bad values as we added validations when parsing config.

pkg/agent/manager/cache/cache_test.go
prasadborole1: Added comment but we would not create these many entries in unit tests :)

pkg/agent/manager/cache/cache.go
prasadborole1: This variable name change would also impact other cache implementations hence I'd be curious to know what @rturner3 and @azdagron think about new variable name.

amoore877: any name changes that would add such complexity could be done as a follow-on PR maybe?

cmd/spire-agent/cli/run/run.go
```diff
@@ -97,7 +97,9 @@ type experimentalConfig struct {
 	SyncInterval  string `hcl:"sync_interval"`
 	TCPSocketPort int    `hcl:"tcp_socket_port"`
 
-	UnusedKeys []string `hcl:",unusedKeys"`
+	UnusedKeys            []string `hcl:",unusedKeys"`
+	MaxSvidCacheSize      int      `hcl:"max_svid_cache_size"`
+	SVIDCacheExpiryPeriod string   `hcl:"svid_cache_expiry_interval"`
```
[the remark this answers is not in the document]
rturner3: HCL name and internal field should match

rturner3 (commented):
Hey @prasadborole1 , I'm still working my way through these changes, but leaving some early comments of low-hanging fruit that can be addressed async while I continue to get through the diff.

cmd/spire-agent/cli/run/run.go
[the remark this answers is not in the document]
rturner3: After reviewing how expiry is used, I'm not convinced it's really needed. I have a couple concerns with attaching expiries to SVIDs in the cache:

1. In the end, it's adding another indirect way to influence the maximum cache size. I feel that adding expiry to cache entries creates some unneeded complexity without an obvious value-add from just tweaking the max cache size configurable.
2. It's not clear how a user would determine what to set this to. The cache expiry also has an indirect connection with SVID TTLs, which are configurable at both the Server and entry level, which further complicates what this value should be set to. This feels like a potential footgun that may cause unintended consequences such as cache churn or cache bloat when combined with the max SVID cache configuration and SVID TTLs.

pkg/agent/manager/cache/cache.go
```diff
@@ -445,6 +536,109 @@ func (c *Cache) GetStaleEntries() []*StaleEntry {
...
+	// iterate over all selectors from cached entries and obtain:
+	// 1. entries that have active subscribers
+	//   1.1 if those entries don't have corresponding SVID cached then put them in staleEntries
+	//       so that SVID will be cached in next sync
+	// 2. get lastAccessTimestamp of each entry
+	for id, record := range c.records {
+		for _, sel := range record.entry.Selectors {
+			if index, ok := c.selectors[makeSelector(sel)]; ok && index != nil {
+				if len(index.subs) > 0 {
+					if _, ok := c.svids[record.entry.EntryId]; !ok {
```
rturner3, line 614:
```suggestion
					if _, svidCached := c.svids[record.entry.EntryId]; !svidCached {
```
This improves the readability as well as disambiguates the value of `ok` across the levels of nested `if`s, since this statement as written overrides the previous value of `ok`.

pkg/agent/manager/cache/cache.go
```diff
@@ -744,10 +955,31 @@ func sortIdentities(identities []Identity) {
...
 	return Identity{
 		Entry:      record.entry,
-		SVID:       record.svid.Chain,
-		PrivateKey: record.svid.PrivateKey,
+		SVID:       svid.Chain,
+		PrivateKey: svid.PrivateKey,
 	}
 }
+
+type record struct {
```
rturner3, line 978: This name is kind of confusing because it's not a cache record per se, and there's also a type called `cacheRecord` in this file/package. What about something like `recordAccessEvent`?

rturner3 (commented):
Hey @prasadborole1 , just letting you know I'm still working through the changes, but leaving my latest feedback.

cmd/spire-agent/cli/run/run.go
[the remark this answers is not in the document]
amoore877: I think the initial intention was "what if SVID _foo_ loses all subscribers, but a new subscriber comes up soon after", and trying to avoid an extra signing operation. However since _foo_ was recently subscribed to this would only be a concern if all other SVIDs up to (or past) max cache were also being subscribed to.
In that heavy subscriber load scenario, one extra SVID signing probably won't be a concern.

test/integration/suites/fetch-svids/07-fetch-svids
```diff
@@ -0,0 +1,28 @@
...
+
+# Call agent debug endpoints and check if svid count is equal to cache size limit
+check-svid-count "spire-agent" $CACHESIZE
+
+# introduce some delay between two fetch calls so that we can validate cache cleanup of svids from first fetch.
+sleep 5
+
+docker-compose exec -u 1001 -T spire-agent \
+  /opt/spire/bin/spire-agent api fetch x509 \
+  -socketPath /opt/spire/sockets/workload_api.sock || fail-now "x509-SVID check failed"
```
rturner3, line 17:
```suggestion
  -socketPath /opt/spire/sockets/workload_api.sock || fail-now "X.509-SVID check failed"
```

test/integration/suites/fetch-svids/07-fetch-svids
```diff
@@ -0,0 +1,28 @@
...
+
+# Call agent debug endpoints and check if svid count is equal to 17 (registration entry count for uids 1001 and 1002)
```
rturner3, line 19:
```suggestion
# Call agent debug endpoints and check if X.509-SVID count is equal to 17 (registration entry count for uids 1001 and 1002)
```

test/integration/suites/fetch-svids/07-fetch-svids
```diff
@@ -0,0 +1,28 @@
...
+check-svid-count "spire-agent" 17
+
+# Call agent debug endpoints and check if svid count is equal to 12
```
rturner3, line 22:
```suggestion
# Call agent debug endpoints and check if X.509-SVID count is equal to 12
```

test/integration/suites/fetch-svids/07-fetch-svids
```diff
@@ -0,0 +1,28 @@
...
+# 17(svid-count) - 5(svids from entries with uuid 1002 will be removed first after svid_cache_expiry_interval) = 12
+check-svid-count "spire-agent" 12
+
+# Call agent debug endpoints and check if svid count is equal to 8
```
rturner3, line 26:
```suggestion
# Call agent debug endpoints and check if X.509-SVID count is equal to 8
```

test/integration/common
```diff
@@ -83,6 +83,25 @@ check-synced-entry() {
   fail-now "timed out waiting for agent to sync down entry"
 }
 
+check-svid-count() {
```
rturner3, line 86:
```suggestion
check-x509-svid-count() {
```

test/integration/common
```diff
@@ -83,6 +83,25 @@ check-synced-entry() {
...
+  MAXCHECKS=50
+  CHECKINTERVAL=1
+
+  for ((i=0;i<=MAXCHECKS;i++)); do
+    if (( $i>=$MAXCHECKS )); then
+      fail-now "svid count validation failed"
```
rturner3, line 92:
```suggestion
      fail-now "X.509-SVID count validation failed"
```
1. This `if` condition can be eliminated and the call to `fail-now` can be moved to the end of the function after the `for` loop
2. It would be helpful to print the expected and actual count of SVIDs in the cache for debugging

test/integration/common
```diff
@@ -83,6 +83,25 @@ check-synced-entry() {
...
+    fi
+    log-info "check svid count on agent debug endpoint ($(($i+1)) of $MAXCHECKS max)..."
+    COUNT=`docker-compose exec -T $1 /opt/spire/conf/agent/debugclient -testCase "printDebugPage" | jq '."svidsCount"'`
```
rturner3, line 95:
```suggestion
    COUNT=`docker-compose exec -T $1 /opt/spire/conf/agent/debugclient -testCase "printDebugPage" | jq '.svidsCount'`
```
The double quotes are not necessary in the `jq` command

rturner3: FYI, I filed https://github.com/spiffe/spire/issues/3256, as it is currently ambiguous what the `svidsCount` field should represent. Since today this value is the count of the number of X.509-SVIDs in the primary cache, I would propose we make this distinction clear in the test by using the term `X.509-SVID` rather than the more generic term `SVID` in all the respective variable names and log messages in this test.

test/integration/common
```diff
@@ -83,6 +83,25 @@ check-synced-entry() {
...
+    log-info "svidsCount: ${COUNT}"
+    if [ "$COUNT" -eq "$2" ]; then
+      log-info "SVID count of $COUNT from cache matches the expected count of $2"
```
rturner3, line 98:
```suggestion
      log-info "X.509-SVID count of $COUNT from cache matches the expected count of $2"
```

test/integration/suites/fetch-svids/07-fetch-svids
```diff
@@ -0,0 +1,28 @@
...
+# 12 - 8(cache size) = 4 extra svids from entries with uuid 1001 will be removed after svid_cache_expiry_interval
```
rturner3, line 27:
```suggestion
# 12 - 8(cache size) = 4 extra X.509-SVIDs from entries with uuid 1001 will be removed after svid_cache_expiry_interval
```

pkg/agent/config.go
```diff
@@ -52,6 +52,12 @@ type Config struct {
 	// SyncInterval controls how often the agent sync synchronizer waits
 	SyncInterval time.Duration
 
+	// MaxSvidCacheSize is a soft limit of max number of SVIDs that would be stored in cache
+	MaxSvidCacheSize int
```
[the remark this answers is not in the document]
rturner3: Thinking on it a bit more, I think we should clarify that the configuration is meant only for the X.509-SVID cache. I edited the configurable naming suggestions accordingly.

rturner3 (commented):
Ok, I have finally had a chance to review all the changes. Thanks for your patience on the review, @prasadborole1.

In terms of the overall approach, my only significant feedback is about whether we can simplify the cache design a bit to not attach an expiry to the cache records.

The rest of my comments are more minor in nature, such as implementation details that might be simplified, style consistency, etc.

Once we settle on the configuration surface area, we should also document the configuration properties in `doc/spire_agent.md`.

Lastly, thank you for all your hard work on this PR and for adding a lot of test coverage. It is really encouraging to see these changes being thoroughly verified.

pkg/agent/manager/cache/cache_test.go
prasadborole1: I'll do it in follow-up PR. This one is getting too big for variable renaming.

cmd/spire-agent/cli/run/run.go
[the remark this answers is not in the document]
prasadborole1: Removed `SVIDCacheExpiryPeriod` from config and implementation.

pkg/agent/manager/config.go
```diff
@@ -55,7 +57,8 @@ func newManager(c *Config) *manager {
 		c.Clk = clock.New()
 	}
 
-	cache := cache.New(c.Log.WithField(telemetry.SubsystemName, telemetry.CacheManager), c.TrustDomain, c.Bundle, c.Metrics)
+	cache := cache.New(c.Log.WithField(telemetry.SubsystemName, telemetry.CacheManager), c.TrustDomain, c.Bundle,
```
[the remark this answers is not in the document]
prasadborole1: I'll address it in follow-up PR along with some test variable renames.

pkg/agent/manager/cache/cache.go
```diff
@@ -445,6 +536,109 @@ func (c *Cache) GetStaleEntries() []*StaleEntry {
...
+	defer c.mu.Unlock()
+
+	c.syncSVIDs()
+}
+
+// Notify subscribers of selector set only if all SVIDs for corresponding selector set are cached
+// It returns whether all SVIDs are cached or not.
+// This method should be retried with backoff to avoid lock contention.
+func (c *Cache) Notify(selectors []*common.Selector) bool {
+	c.mu.Lock()
```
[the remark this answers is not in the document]
prasadborole1: Good catch!

pkg/agent/manager/manager.go
```diff
@@ -145,7 +150,17 @@ func (m *manager) Run(ctx context.Context) error {
 }
 
 func (m *manager) SubscribeToCacheChanges(selectors cache.Selectors) cache.Subscriber {
-	return m.cache.SubscribeToWorkloadUpdates(selectors)
+	subscriber := m.cache.SubscribeToWorkloadUpdates(selectors)
+	backoff := backoff.NewBackoff(m.clk, svidSyncInterval)
```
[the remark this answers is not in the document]
prasadborole1: Yes testability and potential cache lock contention were main reasons. Also I feel retrying for SVIDs is better outside of Cache implementation.

test/integration/common
prasadborole1: sounds good!

pkg/agent/manager/manager_test.go
```diff
@@ -984,6 +1245,44 @@ func TestStorableSVIDsSync(t *testing.T) {
...
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case <-waitCh:
+				return
+			case <-m.clk.After(svidSyncInterval):
+				require.NoError(t, m.syncSVIDs(context.Background()))
```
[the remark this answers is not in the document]
prasadborole1: I simplified these tests - getting rid of svid_cache_interval helped cleaning all unit tests :) The blocking `SubscribeToCacheChanges` retries for svids as `clk` moves along hence it needs to be moved in separate routine. I also added comments in tests.

pkg/agent/manager/manager_test.go
```diff
@@ -721,6 +729,256 @@ func TestSubscribersGetUpToDateBundle(t *testing.T) {
...
+	m := newManager(c)
+	if err := m.Initialize(context.Background()); err != nil {
+		t.Fatal(err)
+	}
+
+	// After Initialize, just 1 SVID should be cached
+	require.Equal(t, 1, m.CountSVIDs())
+	waitCh := make(chan struct{})
+
+	closer := runSVIDSync(t, waitCh, m)
```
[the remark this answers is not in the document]
prasadborole1: I simplified these tests - getting rid of svid_cache_interval helped cleaning all unit tests :) The blocking `SubscribeToCacheChanges` retries for svids as `clk` moves along hence it needs to be moved in separate routine. I also added comments in tests.

pkg/agent/manager/manager_test.go
```diff
@@ -984,6 +1245,44 @@ func TestStorableSVIDsSync(t *testing.T) {
...
+func moveClkAfterInterval(clk *clock.Mock, interval, period time.Duration, waitCh chan struct{}) (closer func()) {
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for {
+			select {
+			case <-waitCh:
+				return
+			case <-time.After(interval):
```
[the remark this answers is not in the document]
prasadborole1: Explained in previous comment.

test/integration/setup/debugagent/main.go
```diff
@@ -52,26 +54,41 @@ func run() error {
...
+	}
+	return nil
+}
+
+// printDebugPage allows integration tests to easily parse debug page with jq
+func printDebugPage(ctx context.Context) error {
+	s, err := retrieveDebugPage(ctx)
+	if err == nil {
+		fmt.Println(s)
+	}
```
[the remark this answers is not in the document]
prasadborole1: updated to return `err`.

pkg/agent/manager/manager.go
```diff
@@ -145,7 +150,17 @@ func (m *manager) Run(ctx context.Context) error {
...
+	// block until all svids are cached and subscriber is notified
+	for {
+		if m.cache.Notify(selectors) {
+			return subscriber
+		}
+		m.c.Log.WithField(telemetry.Selectors, selectors).Info("Waiting for SVID to get cached")
+
+		<-m.clk.After(backoff.NextBackOff())
```
rturner3, line 157: We need to accept a `context.Context` in this method now so that we can be informed of context cancellations and return from this method. e.g.
```
select {
case <-ctx.Done():
    return nil, ctx.Err()
case <-m.clk.After(backoff.NextBackoff()):
}
```

pkg/agent/manager/manager.go
```diff
@@ -139,8 +150,30 @@ func (m *manager) Run(ctx context.Context) error {
...
+	return m.subscribeToCacheChanges(ctx, selectors, nil)
+}
+
+func (m *manager) subscribeToCacheChanges(ctx context.Context, selectors cache.Selectors, notifyCallbackFn func()) (cache.Subscriber, error) {
+	subscriber := m.cache.SubscribeToWorkloadUpdates(selectors)
+	bo := m.subscribeBackoffFn()
+	// block until all svids are cached and subscriber is notified
+	for {
+		svidsInCache := m.cache.Notify(selectors)
+		if notifyCallbackFn != nil {
```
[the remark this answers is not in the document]
prasadborole1: Added comment, its used for unit testing.

pkg/agent/api/delegatedidentity/v1/service_test.go
```diff
@@ -692,6 +689,14 @@ func (m *FakeManager) FetchJWTSVID(ctx context.Context, spiffeID spiffeid.ID, au
 	}, nil
 }
 
+func (m *FakeManager) MatchingRegistrationEntries(selectors []*common.Selector) []*common.RegistrationEntry {
+	out := make([]*common.RegistrationEntry, 0, len(m.identities))
+	for _, identity := range m.identities {
```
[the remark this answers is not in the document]
prasadborole1: I was planning to refactor tests using `identities` in separate PR as this one has already become large and there are bunch of tests which reference them.

pkg/agent/config.go
```diff
@@ -59,6 +59,9 @@ type Config struct {
 	// SyncInterval controls how often the agent sync synchronizer waits
 	SyncInterval time.Duration
 
+	// X509SVIDCacheMaxSize is a soft limit of max number of SVIDs that would be stored in cache
+	X509SVIDCacheMaxSize int
```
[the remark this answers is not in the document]
prasadborole1: It's a soft limit as we need to provide identity to all active subscribers even if goes beyond `X509SVIDCacheMaxSize`. I've set 1000 as default value.

pkg/agent/endpoints/workload/handler_test.go
```diff
@@ -1300,8 +1300,12 @@ type FakeManager struct {
 	err         error
 }
 
-func (m *FakeManager) MatchingIdentities(selectors []*common.Selector) []cache.Identity {
-	return m.identities
+func (m *FakeManager) MatchingRegistrationEntries(selectors []*common.Selector) []*common.RegistrationEntry {
+	out := make([]*common.RegistrationEntry, 0, len(m.identities))
+	for _, identity := range m.identities {
```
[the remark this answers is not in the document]
prasadborole1: same as previous comment.

pkg/agent/manager/cache/cache.go
```diff
@@ -708,9 +903,9 @@ func (c *Cache) getSelectorIndexForRead(s selector) *selectorIndex {
 }
 
 type cacheRecord struct {
-	entry *common.RegistrationEntry
-	svid  *X509SVID
-	subs  map[*subscriber]struct{}
+	entry               *common.RegistrationEntry
+	subs                map[*subscriber]struct{}
+	lastAccessTimestamp int64
```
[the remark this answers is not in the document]
prasadborole1: If users start getting lot of cache misses then recommendation would be to increase the `X509SVIDCacheMaxSize` instead of introducing complicated logic. The rational being any solution we provide will be at best probabilistic. That's the reason I removed configurable `X509SVIDCacheExpiryPeriod` field from previous revision of this PR.

pkg/agent/manager/cache/cache.go
```diff
@@ -616,12 +794,33 @@ func (c *Cache) matchingIdentities(set selectorSet) []Identity {
 	// TODO: figure out how to determine the "default" identity
 	out := make([]Identity, 0, len(records))
 	for record := range records {
-		out = append(out, makeIdentity(record))
+		if svid, ok := c.svids[record.entry.EntryId]; ok {
```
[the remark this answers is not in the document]
prasadborole1: Yes but once new SVID is cached we would send update to all existing subscribers with new SVID.

pkg/agent/manager/cache/cache.go
```diff
@@ -445,6 +524,105 @@ func (c *Cache) GetStaleEntries() []*StaleEntry {
...
+	// 1. entries that have active subscribers
+	//   1.1 if those entries don't have corresponding SVID cached then put them in staleEntries
+	//       so that SVID will be cached in next sync
+	// 2. get lastAccessTimestamp of each entry
+	for id, record := range c.records {
+		for _, sel := range record.entry.Selectors {
+			if index, ok := c.selectors[makeSelector(sel)]; ok && index != nil {
+				if len(index.subs) > 0 {
+					if _, ok := c.svids[record.entry.EntryId]; !ok {
+						c.staleEntries[id] = true
```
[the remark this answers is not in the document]
prasadborole1: I don't think so as its updated under single lock.

pkg/agent/manager/cache/cache.go
```diff
@@ -445,6 +524,105 @@ func (c *Cache) GetStaleEntries() []*StaleEntry {
...
+	lastAccessTimestamps := make([]recordAccessEvent, 0, len(c.records))
+
+	// iterate over all selectors from cached entries and obtain:
+	// 1. entries that have active subscribers
+	//   1.1 if those entries don't have corresponding SVID cached then put them in staleEntries
+	//       so that SVID will be cached in next sync
+	// 2. get lastAccessTimestamp of each entry
+	for id, record := range c.records {
+		for _, sel := range record.entry.Selectors {
+			if index, ok := c.selectors[makeSelector(sel)]; ok && index != nil {
```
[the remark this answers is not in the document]
prasadborole1: I believe not

pkg/agent/manager/cache/cache.go
```diff
@@ -375,6 +420,40 @@ func (c *Cache) UpdateEntries(update *UpdateEntries, checkSVID func(*common.Regi
...
 	}
 
+	// entries with active subscribers which are not cached will be put in staleEntries map
+	activeSubsByEntryID, recordsWithLastAccessTime := c.syncSVIDs()
+	extraSize := len(c.svids) - c.svidCacheMaxSize
+
+	// delete svids without subscribers and which have not been accessed since svidCacheExpiryTime
+	if extraSize > 0 {
+		// sort recordsWithLastAccessTime
+		sortTimestamps(recordsWithLastAccessTime)
```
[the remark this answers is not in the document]
prasadborole1: Log is added in `subscribeToCacheChanges` method in `manager.go`. I believe that's better place for notifying cache miss cause such logs in sync methods would result in high cardinality.

pkg/agent/manager/cache/cache.go
```diff
@@ -375,6 +420,40 @@ func (c *Cache) UpdateEntries(update *UpdateEntries, checkSVID func(*common.Regi
 		}
 	}
 
+	// entries with active subscribers which are not cached will be put in staleEntries map
+	activeSubsByEntryID, recordsWithLastAccessTime := c.syncSVIDs()
+	extraSize := len(c.svids) - c.svidCacheMaxSize
+
+	// delete svids without subscribers and which have not been accessed since svidCacheExpiryTime
+	if extraSize > 0 {
```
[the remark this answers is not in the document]
prasadborole1: The entire `UpdateEntries` method is ripe for refactoring:) 
I've added more comments let me know if you still think it would still be better to kick deletion in separate method.

pkg/agent/manager/cache/cache.go
```diff
@@ -375,6 +420,40 @@ func (c *Cache) UpdateEntries(update *UpdateEntries, checkSVID func(*common.Regi
...
+					delete(c.svids, record.id)
+					extraSize--
+				}
+			}
+		}
+	}
+
+	// Update all stale svids or svids whose registration entry is outdated
+	for id, svid := range c.svids {
+		if _, ok := outdatedEntries[id]; ok || (checkSVID != nil && checkSVID(nil, c.records[id].entry, svid)) {
```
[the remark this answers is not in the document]
prasadborole1: We would need to iterate over all new entries and identify entries which are updated. `outdatedEntries` store those entries. 
`checkSVID` tells us which SVIDs are expiring and need to be rotated.

pkg/agent/manager/cache/cache.go
```diff
@@ -375,6 +420,40 @@ func (c *Cache) UpdateEntries(update *UpdateEntries, checkSVID func(*common.Regi
 		}
 	}
 
+	// entries with active subscribers which are not cached will be put in staleEntries map
+	activeSubsByEntryID, recordsWithLastAccessTime := c.syncSVIDs()
```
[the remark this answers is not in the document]
MarcosDY: yeap, but in this possible scenario I was thinking, 
the issue was with completing the list before we know we are not going to add more entries to that list.

Basically we have limit of `10`, and our agent has `20` entries. our cache has only `7` now,  so we complete the list to 3 more from this call to `syncSVIDs()`.
But after that call we found that we need to get SVIDs for another 2 items, that are not currently in our cache,
so our code will try to get `5` new SVIDs instead of the 2 required + 1 to complete the list of cached SVIDs.
All this in the same sync call.
Makes it sense?

pkg/agent/manager/cache/cache.go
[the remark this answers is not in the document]
MarcosDY: that is the log I was asking for.
https://github.com/prasadborole1/spire/blob/lru-agent-cache/pkg/agent/manager/sync.go#L104-L107

with this checkSVID to be always nil, we'll lose it

pkg/agent/manager/cache/cache.go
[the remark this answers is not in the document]
prasadborole1: > But after that call we found that we need to get SVIDs for another 2 items, that are not currently in our cache
`NewSubscriber` method would be waiting for lock until `UpdateEntries` finishes. Therefore, we won't need to get 2 new SVIDs in middle of `UpdateEntries` execution. And those 2 new subscribers will be serviced in next iteration of sync.

pkg/agent/manager/cache/cache.go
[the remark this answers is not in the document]
prasadborole1: got it. Added that log. Thanks!

pkg/agent/endpoints/sdsv2/handler.go
```diff
@@ -64,7 +64,11 @@ func (h *Handler) StreamSecrets(stream discovery_v2.SecretDiscoveryService_Strea
 		return err
 	}
 
-	sub := h.c.Manager.SubscribeToCacheChanges(selectors)
+	sub, err := h.c.Manager.SubscribeToCacheChanges(stream.Context(), selectors)
+	if err != nil {
+		log.WithError(err).Error("Subscribe to cache changes failed")
```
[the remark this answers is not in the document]
prasadborole1: unit tests here are not setup for error checking from manager or attestor. Is it okay if we address this separately essentially making it like sdsv3?

rturner3 (dismissed):
Awesome work here, @prasadborole1!!

pkg/agent/manager/cache/cache.go
```diff
@@ -445,6 +527,99 @@ func (c *Cache) GetStaleEntries() []*StaleEntry {
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.syncSVIDsWithSubscribers()
+}
+
+// Notify subscribers of selector set only if all SVIDs for corresponding selector set are cached
+// It returns whether all SVIDs are cached or not.
+// This method should be retried with backoff to avoid lock contention.
+func (c *Cache) Notify(selectors []*common.Selector) bool {
```
azdagron, line 543: I _think_ the current cache behavior is such that we'll return a partial list.... should this function return true if at least one SVID is available?

azdagron (commented):
Thanks for your patience here. Other than the comments I mentioned, I think the last bit here is to only enable this when the experimental flag is set. Like we discussed in the meeting, i think we can probably get away with forking the cache package. The code added to the manager could be moved into the new cache or it could be skipped if the feature isn't enabled (the former seems cleaner).

pkg/agent/manager/cache/cache.go
prasadborole1: Yep that's the intentional change..we want to return complete list of identities to users mainly for users would just assume first result has all identities it needs(e.g cli users).

rturner3 (commented):
Just a few minor comments, otherwise LGTM

pkg/agent/manager/manager_test.go
```diff
@@ -817,6 +820,251 @@ func TestSubscribersGetUpToDateBundle(t *testing.T) {
...
+		subErrCh <- nil
+	}(ctx)
+
+	syncErrCh := make(chan error, 1)
+	// run svid sync
+	go func(ctx context.Context) {
+		if err := m.runSyncSVIDs(ctx); err != nil {
+			syncErrCh <- err
+		}
+		syncErrCh <- nil
```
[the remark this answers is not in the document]
prasadborole1: Done

pkg/agent/manager/cache/cache_test.go
```diff
@@ -689,6 +689,47 @@ func BenchmarkCacheGlobalNotification(b *testing.B) {
...
 
+func TestMatchingRegistrationEntries(t *testing.T) {
+	cache := newTestCache()
+
+	// populate the cache with FOO and BAR without SVIDS
+	foo := makeRegistrationEntry("FOO", "A")
+	bar := makeRegistrationEntry("BAR", "B")
+
+	// check empty result
+	assert.Equal(t, []*common.RegistrationEntry{},
```
rturner3, line 700: `assert.Empty()` is also nice for this sort of thing. No need to change here though. We've iterated a lot on this PR already 😀

rturner3 (approved):

pkg/agent/manager/sync.go
```diff
@@ -36,6 +36,18 @@ type Cache interface {
 	GetStaleEntries() []*cache.StaleEntry
 }
 
+func (m *manager) syncSVIDs(ctx context.Context) (err error) {
+	// perform syncSVIDs only if using LRU cache
+	if m.c.SVIDCacheMaxSize > 0 {
+		m.cache.SyncSVIDsWithSubscribers()
+		staleEntries := m.cache.GetStaleEntries()
+		if len(staleEntries) > 0 {
+			return m.updateSVIDs(ctx, staleEntries, m.cache)
```
azdagron, line 45: Will this contend with the normal sync goroutine? could this cause two goroutines to pick up the same set of stale entries and effectively request the server sign the same SVID twice?

prasadborole1: Good point. There's a chance that this may happen. Updated code to put fetching of SVIDs under lock.

test/integration/suites/fetch-x509-svids/conf/server/server.conf
```diff
@@ -0,0 +1,26 @@
+server {
+	bind_address = "0.0.0.0"
+	bind_port = "8081"
+	trust_domain = "domain.test"
+	data_dir = "/opt/spire/data/server"
+	log_level = "DEBUG"
+	ca_ttl = "1h"
+	default_svid_ttl = "10m"
```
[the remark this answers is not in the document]
prasadborole1: The experimental flag is added to agent config and not server config.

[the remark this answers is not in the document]
azdagron: Whoops. Wires crossed :) Thanks.

azdagron (approved):

azdagron:
Thanks, @prasadborole1 ! I think we're ready to get this in so we can get some operational experience with it to iron out any remaining issues.

PR #3192, opened 2022-06-27
Windows support: prevent the creation of directories for UDS endpoints

### LINKED ISSUE #3188: Windows support: directories for UDS endpoints are created
On posix platforms, we create directories for Unix Domain Sockets endpoints.
In case of Windows, named pipes are used instead of Unix Domain Sockets, but we still create the directories.
Prevent the creation of those directories on Windows.

amartinezfayo:
Prevent the creation of directories for UDS endpoints.

Fixes #3188.


MarcosDY (commented):
code looks great my only comment is that if we may spend some time to create an unit tests to validate than folders are not created when using windows

amartinezfayo:
Unfortunately, our current testing for the `spire-server run` and `spire-agent run` commands does not currently cover the actual running of the server and the agent (the `Run` function is not tested). I think that ideally, this should be tested as part of the testing of the `Run` function, which seems to be a separate effort. I've created #3199 for that.
@MarcosDY  Would you be OK if that is addressed separately?

PR #3219, opened 2022-07-05
Add integration test to validate CRD mode

MarcosDY:
Add integration test to validate CRD mode and validate that SVID contains DNS


test/integration/suites/k8s-crd-mode/README.md
```diff
@@ -0,0 +1,11 @@
+# Kubernetes Suite
```
amartinezfayo, line 1: Looks like this needs to be updated with the tests specifically done in this suite?

amartinezfayo (commented):
Thank you @MarcosDY for this!

PR #3273, opened 2022-07-25
Add option to use anonymous authentication in k8s WorkloadAttestor

rturner3:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
`k8s` WorkloadAttestor plugin integration with kubelet

**Description of change**
The `k8s` WorkloadAttestor plugin does not provide a way to obtain selectors from the kubelet over the secure port when the kubelet is configured to use anonymous authentication. Add a configuration field `use_anonymous_authentication` that allows the plugin to send unauthenticated requests to the kubelet secure port.

**Which issue this PR fixes**
#3193

Tested this with the `k8s-reconcile` test with these changes applied in my local branch: https://github.com/spiffe/spire/commit/0d59a37267d564a8e042d07c3c94cd2859109b3b

### DIFF
--- a/doc/plugin_agent_workloadattestor_k8s.md
+++ b/doc/plugin_agent_workloadattestor_k8s.md
@@ -31,6 +31,12 @@ server name validation against the kubelet certificate.
 > mitigate this issue. A large cache ttl value is not recommended however, as
 > that can impact permission revocation.
 
+> **Note** Anonymous authentication with the kubelet requires that the
+> kubelet be started with the `--anonymous-auth` flag. It is discouraged to use anonymous
+> auth mode in production as it requires authorizing anonymous users to the `nodes/proxy`
+> resource that maps to some privileged operations, such as executing commands in
+> containers and reading pod logs.
+
 **Note** To run on Windows containers, Kubernetes v1.24+ and containerd v1.6+ are required,
 since [hostprocess](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/) container is required on the agent container.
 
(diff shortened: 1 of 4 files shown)


doc/plugin_agent_workloadattestor_k8s.md
```diff
@@ -43,6 +49,7 @@ since [hostprocess](https://kubernetes.io/docs/tasks/configure-pod-container/cre
 | `token_path` | The path on disk to the bearer token used for kubelet authentication. Defaults to the service account token `/run/secrets/kubernetes.io/serviceaccount/token` |
 | `certificate_path` | The path on disk to client certificate used for kubelet authentication |
 | `private_key_path` | The path on disk to client key used for kubelet authentication |
+| `use_anonymous_authentication` | If true, use anonymous authentication for kubelet communication |
```
MarcosDY, line 52: maybe it is an overkill... but, what do you think about say again that kubelet must be started with `--anonymous-auth` in order to use this feature? I'm worried someone reading this fields think than just setting `use_anonymous_authentication` is enough. (maybe adding to agen_full.conf is enough)

MarcosDY (commented):
Great!!! Code looks good, but can you add the new configuration with some notes on https://github.com/spiffe/spire/blob/main/conf/agent/agent_full.conf#L319?

doc/plugin_agent_workloadattestor_k8s.md
rturner3: I thought about it, but we don't go into that level of detail for the other authentication-related configuration properties in this table, so I followed suit here and left the more detailed note above. I guess I would prefer to keep that consistent, and I feel like it's probably good enough to have the note above.

MarcosDY (approved):
LGTM!!!

PR #3274, opened 2022-07-26
Fix deletegatedidentity integration test

prasadborole1:
`SubscribeToX509SVIDs` call should not be made with empty selectors.

Context: Fails integration tests in https://github.com/spiffe/spire/pull/3181

I'll create follow-up PR to validate selectors are not empty in client methods.


amartinezfayo (approved):
Thanks for the fix, @prasadborole1!

Issue #1848, opened 2020-09-18
Extensible claims in JWT SVIDs


lumjjb:
Following up on an action item from 09/17/2020 SIG-Spec meeting on the discussions on Extensible claims in SVIDs.

During the call, we discussed a couple topics around extensible claims in SVIDs, mainly around usecases, mechanism of extensible claims and several considerations in use.

### Labeling workloads to indicate an attested property

This usecase involves the ability to make authorization decisions on certain attested properties of a workload. This can be a logical aspect of the computation of the underlying workload, like the datacenter it is running in, or go into hardware backed claims such as ownership of a certain asset tag bootstrapped into a Trusted Platform Module. This way, authorization decisions can be made on geolocation properties and trust properties of the underlying compute a workload is running on. 

This is helpful when authorization frameworks such as [Vault Sentinel](https://www.vaultproject.io/docs/enterprise/sentinel), or  ABAC based auth/iam solutions allow creation of policies based on these claims. Having claims be part of the SVID makes extension of claims seamless and not requiring an additional structural schema definition (i.e. regexp on common name)  that needs to be defined between versions of identity solutions and authorization of various services and apps.

### Application specific use for logic, i.e. data/medical use case 

Applications may choose to base logic on properties/claims of a workload identity. For example, in the data scenario, based on the type of workload running, as well as where the workload is running. For example, if a workload is in Europe, it would be given Europe based contextual information. A medical scenario was also mentioned around giving organization specific flows/data based on certain group and organization attributes of a workload.

## Mechanism of extensible claims

We discussed the way the claims would be included in the SVIDs, the initial scope discussed was for the JWT SVID, as claim fields in the JWT. x509 was discussed as well, i.e. x509v3 ext. attributes, but was critical to the use cases presented in the near term.

### Claims collision and meaning

There was a worry about claim name collisions and what it they would mean. It was discussed that for JWT SVIDs. These would refer to the JWT [RFC7519 Section 4.3](https://tools.ietf.org/html/rfc7519#section-4.3)) on Private claims. Use of the claims should follow the guidance from the RFC.

evan2645:
thank you very much for opening this @lumjjb! we've been discussing it a little bit, and I think we've got some brainstorming to do on the best way to support it. We'll probably have some more questions in the coming weeks

edwarnicke:
A couple of things to think about when looking at multiple claims:

[...]

PR #3279, opened 2022-07-28
Add config option for verbose split claims in JWT-SVIDs

Celant:
Signed-off-by: Josh Harris <me@celant.co.uk>

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
JWKS-SVID signer

**Description of change**
Adds a new `jwt_verbose_claims_enabled` server config option. In addition to the subject claim, if enabled, it will also add a `trust` and `workload` claim. These contain the Trust Domain and the Workload Identifier split out from the SPIFFE ID.

**Which issue this PR fixes**


amartinezfayo (commented):
Thank you for this contribution, @Celant!

JWT-SVIDs issued by SPIRE already have the `sub` claim set to the SPIFFE ID of the workload to which it is issued.
Having the trust domain of the SPIFFE ID and the path of the SPIFFE ID inside the trust domain in separate claims doesn't seem to be needed in general, since the consumer of the JWT-SVID should be able to obtain that from the `sub` claim that we already have.

Could you tell us a little more about your use case and if it would be possible for your applications to parse the `sub` claim to obtain the same?

Celant:
@amartinezfayo Unfortunately parsing the `sub` claim out of the JWT isn't working well for us. Our use-case here is for Vault, and using the JWT-SVID in Vault policy templates. We can parse out any claim in the JWT for use in our policy templates, but the default `sub` claim is causing us problems. The `spiffe://` prefix of the SPIFFE ID is causing us headaches, as a few different systems are detecting the URI scheme as an empty folder due to the double-slash and stripping it out.

This 'verbose claims' option simply takes the existing SPIFFE ID, and adds both the workload identifier and trust domain as explicit extra claims. This means we're no longer needing to parse a full URI, and it will allow us to use the workload identifier in our Vault policy templates without issues.

I'd also like to adapt this change in a follow-up PR at some point, to rework claims in a way that also covers #1848. In my mind, a similar system to Vault's policy templating could really suit this. I.e. adding a list of extra claims to the SPIRE config. Each claim can contain a Golang template, so the users have full control over the contents of the claim, and the template can be populated with attested properties of the workload. That, however, is a discussion for another time as this PR is the urgent first step for us 🙂

amartinezfayo:
Thank you for the details, @Celant. Are the JWT-SVIDs being parsed by Vault's built-in JWT/OIDC Auth Method? Or is this consumed by a custom plugin in Vault? We are trying to understand if there is a lack of flexibility in Vault or what is causing the problem with the content of the `sub` claim.

amartinezfayo:
Discussed this offline with @Celant. #3253 would provide a solution for this.

PR #3288, opened 2022-08-01
Remove unofficial SVID Store protos

MarcosDY:
Removes unofficial SVID Store proto and only support spire-plugin-sdk protos


azdagron (approved):
Huzzah!

PR #3308, opened 2022-08-03
Shutdown agent when it is banned

### LINKED ISSUE #3278: Banned agent keep alive
When a SPIRE Agent is banned, it stops getting SVIDs and entries updates, but that agent keeps alive.

That means that it will be able to attest workloads and provided SVIDs it has on cache,
and a workload may get compromised SVIDs or any reason an admin decides to ban that agent.
In case a banned agent is trying to reattest, it is  not allowed and the agent fails to start.

In case of evict (delete) an agent, this agent has the capability to recognize that it was deleted and it shut down itself, 
with this approach the agent (in case of K8S) is going to restart and reattest. 

But is it correct to keep agents alive? 
I can think in 2 possible options here:
- Shutdown as it is happening with evicted agents: the bad side of this is that the agent will keep trying to reattest forever and it will keep failing until the agent is evicted to have a successfully attestation.
- Clean cache: with this agent will keep alive but will not be able to provide new SVIDs in successful attestations, in case an agent is evicted after being banned the agent will get that eviction state and shutdown.

MarcosDY:
Make Agent shutdown when it is banned

**Which issue this PR fixes**
fixes #3278


pkg/agent/svid/rotator_test.go
```diff
@@ -152,14 +157,93 @@ func TestRotator(t *testing.T) {
 	}
 }
 
+func TestRotationFails(t *testing.T) {
+	caCert, caKey := testca.CreateCACertificate(t, nil, nil)
+
+	expiredStatus := status.New(codes.PermissionDenied, "agent is expired")
```
amartinezfayo, line 163: nit:
Even if it's a test, it may be confusing
```suggestion
	expiredStatus := status.New(codes.PermissionDenied, "agent is not active")
```

test/integration/suites/evict-agent/08-delete-agent
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+log-debug "deleting agent..."
+
+docker-compose exec -T spire-server \
+    /opt/spire/bin/spire-server agent evict \
+    -spiffeID "spiffe://domain.test/spire/agent/x509pop/$(fingerprint conf/agent/agent.crt.pem)"
+
+# Check at most 30 times (with one second in between) than agent is going shutdown
```
amartinezfayo, line 9:
```suggestion
# Check at most 30 times (with one second in between) that the agent is going to shut down
```

test/integration/suites/evict-agent/08-delete-agent
```diff
@@ -0,0 +1,21 @@
...
+for ((i=1;i<=MAXCHECKS;i++)); do
+    log-info "checking for agent to get notification and try to reattest ($i of $MAXCHECKS max)..."
+    docker-compose logs spire-agent
+    if docker-compose logs spire-agent | grep "Agent needs to re-attest; removing SVID and shutting down"; then
+	exit 0
+    fi
+    sleep "${CHECKINTERVAL}"
+done
+
+fail-now "timed out waiting for agent to sync down entry"
```
amartinezfayo, line 21:
```suggestion
fail-now "timed out waiting for agent to shut down"
```

test/integration/suites/evict-agent/05-agent-failed-to-start
```diff
@@ -0,0 +1,19 @@
...
+
+log-debug "starting agent again..."
+docker-up spire-agent
+
+# Check at most 30 times (with one second in between) that the agent is not able to get new 
+# workload entries.
+MAXCHECKS=30
+CHECKINTERVAL=1
+for ((i=1;i<=MAXCHECKS;i++)); do
+    log-info "checking agent is not able to start ($i of $MAXCHECKS max)..."
```
amartinezfayo, line 11:
```suggestion
    log-info "checking that the agent is not able to start ($i of $MAXCHECKS max)..."
```

amartinezfayo (commented):
Thank @MarcosDY, looks great!

amartinezfayo (approved):
🎉

PR #3317, opened 2022-08-08
Upgrade golangci-lint to v1.48.0

MarcosDY:
Upgrade golangci-lint to v1.48.0 that is compatible with golang v1.19 and solve revive issues


azdagron (approved):
\o/

PR #3347, opened 2022-08-12
Replace go.uber.org/atomic with sync/atomic

rturner3:
In Go 1.19, the `sync/atomic` package gained new `Int32`, `Int64`, `Uint32`, and `Uint64` types. These provide very similar functionality to the types in `go.uber.org/atomic`. We only use this package in one place for tests, so remove the package dependency in favor of the standard library.

Replace usage of `go.uber.org/atomic` `Uint32` type with `sync/atomic` `Int32` since `Uint32` in `sync/atomic` does not provide a method to subtract values, and it is not possible to declare a negative `Uint32` in the call to `Add()`. This is a safer type to use since subtracting from `0` with a `Uint32` will cause overflow.


amartinezfayo (approved):
LGTM!

PR #3350, opened 2022-08-12
Remove old server entrycache

rturner3:
This old cache code was left around when the full in-memory cache was added in case any issues arised from the in-memory cache. The in-memory cache has been running stably since v0.12.0. Remove this old code that is no longer referenced.

This also removes a dependency on github.com/hashicorp/golang-lru.


azdagron (approved):
Awesome! Thanks, @rturner3!

PR #3403, opened 2022-09-01
[dependabot] Ignore patch versions of AWS SDK

rturner3:
The AWS SDK v2 is decomposed into Go modules per service, and each of these modules has new patch releases almost weekly, if not more frequently. There is not currently a lot of active development of AWS-related features in SPIRE. Handling these patch updates has proved to be very time-consuming with little to no value.

We used to ignore these updates for the AWS SDK v1 before we fully upgraded to using the v2 SDK. Update the dependabot settings to ignore all the AWS SDK v2 module patch updates.


azdagron (approved):
\o/ thank you so much.... so very much.

PR #3437, opened 2022-09-20
Fixes #3436: Properly asserts expected error is returned

### LINKED ISSUE #3436: Test Execution Bug: sqlstore_test.go checks the wrong err object
The test code within [`sqlstore_test::TestUpdateRegistrationEntryWithMask`](https://github.com/spiffe/spire/blob/971443afc532349dd6f54105c9142a17a2eb130d/pkg/server/datastore/sqlstore/sqlstore_test.go#L2464-L2467) is checking the wrong value for an expected returned error.

```go
updatedRegistrationEntry, err := s.ds.UpdateRegistrationEntry(ctx, updateEntry, tt.mask)

if tt.err != nil {
  s.Require().Error(tt.err)
  return
}
```

The purpose of this block reads as "if the test expects an error to have been returned, then assert that the error was in fact returned". However, as currently coded with `Error(tt.err)` it isn't actually asserting on the returned error value. The result is that error cases expecting an error to be returned aren't actually being tested as such.

The block should be updated to the following, which will verify (1) an error is returned and (2) the error message contains the expected string.

```go
if tt.err != nil {
  s.Require().ErrorContains(err, tt.err.Error())
  return
}
```

After making this change we see two previously passing tests now fail.

[`Update Selectors, Bad Data, Mask True`](https://github.com/spiffe/spire/blob/971443afc532349dd6f54105c9142a17a2eb130d/pkg/server/datastore/sqlstore/sqlstore_test.go#L2370-L2373) is now failing due to a misconfigured test setup. The mask value is `false` when it should be `true`. As such, no error is returned from the method call.

[`Update StoreSvid, Invalid selectors, Mask True`](https://github.com/spiffe/spire/blob/971443afc532349dd6f54105c9142a17a2eb130d/pkg/server/datastore/sqlstore/sqlstore_test.go#L2406-L2416) is now failing due to an expected error not being returned. The test appears to be configured correctly so this may be an actual SPIRE code bug, but more research is necessary to find where.

I will submit a PR fixing this issue.

dennisgove:
This fix includes two test setup fixes as well. Both tests expected an error to be returned but the actual test setup code did not result in an error. The bug in `s.Require().Error(tt.err)` missed that failed check. After changing to `s.Require().ErrorContains(err, tt.err.Error())` we saw those two tests now fail. Fortunately, the bug was in the test setup instead of application code. Application code was/is properly checking for valid entry content on an update action.

Fixes #3436 

Signed-off-by: Dennis Gove <dgove1@bloomberg.net>

**Pull Request check list**

- [ ] Documentation updated?

### DIFF
--- a/pkg/server/datastore/sqlstore/sqlstore_test.go
+++ b/pkg/server/datastore/sqlstore/sqlstore_test.go
@@ -2291,7 +2291,7 @@ func (s *PluginSuite) TestUpdateRegistrationEntryWithMask() {
 		EntryExpiry:   1000,
 		DnsNames:      []string{"dns2"},
 		Downstream:    false,
-		StoreSvid:     false,
+		StoreSvid:     true,
 	}
 	badEntry := &common.RegistrationEntry{
 		ParentId:      "not a good parent id",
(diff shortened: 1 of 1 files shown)


azdagron (approved):
Awesome. Thanks for catching this, @dennisgove !

PR #3443, opened 2022-09-21
Add known issues to k8s workload attestor

evan2645:
The issue described in #3092 is unlikely to be solved any time. Adding it to the docs so its more easily discovered.

Signed-off-by: Evan Gilman <evan2645@gmail.com>


azdagron (approved):
\o/, thanks @evan2645

PR #3461, opened 2022-09-23
Pin k8s image version in integration tests

evan2645:
Previously, the integration tests would rely on the default k8s image version, which differs based on the version of kind we're using. This commit explicitly pins the image version such that 1) it's clear what version of k8s is in use, and 2) it's easily changed when needed.

Signed-off-by: Evan Gilman <evan2645@gmail.com>


amartinezfayo (commented):
Thanks @evan2645.
Should we update https://github.com/spiffe/spire/blob/main/doc/supported_integrations.md#kubernetes?

evan2645:
> Should we update https://github.com/spiffe/spire/blob/main/doc/supported_integrations.md#kubernetes?

Yikes ... I suspect this is pretty inaccurate at this point. I'm happy to update it, but I actually don't know what it should be. I can lower it to 1.16 based on my previous tests, and can also try bumping to a later version. That would at least bring the doc up to date, but alas it's likely to fall out of date again quickly.

One idea could be to have a k8s subsystem maintainer, who is responsible for maintaining all k8s related integrations, and can stay on top of changes and make sure we continue to support the latest and oldest etc

azdagron (approved):

PR #3468, opened 2022-09-29
.Pod.pod_name should be .Pod.Name in example

keeganwitt:
This is just a documentation fix. The example doesn't use the correct variable to reference to pod name.


azdagron (approved):
Thanks!

PR #3473, opened 2022-09-30
Fix markdown escapes

keeganwitt:
The first commit adds a missing escape of a pipe in a Markdown table.

The second commit replaces the backslash escapes with HTML entities because some parsers don't like the backslash (like IntelliJ's Markdown renderer, for example).


keeganwitt:
I did not adjust the spacing around the pipes in Markdown tables because I couldn't tell what formatting rules the project was using.

azdagron (approved):
Thanks!

I wish we were consistent on our markdown styling but you'll find that it varies per md doc, sometimes varying within the same document. There's definitely room for improvement here.

PR #3505, opened 2022-10-17
Improve run command test coverage

### LINKED ISSUE #3199: Improve the test coverage of the `spire-server run` and `spire-agent run` commands
Current tests don't cover the `Run` function (in the server and in the agent) that runs the server and the agent.
There are things like endpoints preparation (create / not create directories depending on the transport) that are not currently tested and should be tested.

guilhermocc:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
No functionality affected

**Description of change**
Add spire agent and server `run` command test cases to increase coverage

**Which issue this PR fixes**
fixes #3199


cmd/spire-agent/cli/run/run_posix_test.go
```diff
@@ -5,15 +5,121 @@ package run
...
+				allowUnknownConfig: false,
+			},
+			want: want{
+				code:               1,
+				agentUdsDirCreated: false,
+				dataDirCreated:     false,
+			},
+		},
+		{
+			name:     "creates spire-agent uds and data dirs",
```
guilhermocc, line 67: wasn't able to explore more scenarios loading config successfully due to fflag singleton (after it's first instantiation, no flags can be loaded anymore, ref pkg/common/fflag/fflag.go:67) Any suggestions?

cmd/spire-agent/cli/run/run_posix_test.go
```diff
@@ -5,15 +5,121 @@ package run
...
+		{
+			name:     "don't create any dir when error loading config",
+			osTarget: "linux",
+			args: args{
+				args: []string{},
+			},
+			fields: fields{
+				logOptions: []log.Option{},
+				env: &commoncli.Env{
+					Stderr: io.Discard,
```
amartinezfayo, line 56: I think that it would be good to test the content of Stderr, so we can assert the reason of the failure. We may have a `*bytes.Buffer` var here that you can then convert to a string for the comparison.
There is an example of how this can be done here: https://github.com/spiffe/spire/blob/main/cmd/spire-agent/cli/healthcheck/healthcheck_test.go#L67

cmd/spire-agent/cli/run/run_posix_test.go
```diff
@@ -5,15 +5,121 @@ package run
...
+				allowUnknownConfig: testCase.fields.allowUnknownConfig,
+			}
+
+			result := cmd.Run(testCase.args.args)
+			assert.Equal(t, testCase.want.code, result)
+
+			_, agentSocketDirErr := os.Stat(testAgentSocketDir)
+
+			if testCase.want.agentUdsDirCreated {
+				assert.Nilf(t, agentSocketDirErr, "spire-agent uds dir should be created")
```
amartinezfayo, line 108: When asserting errors, the specific functions should be used to have a better assertion and diagnostics. Error and NoError functions are provided for this purpose.

```suggestion
				assert.NoError(t, agentSocketDirErr, "spire-agent uds dir should be created")
```

cmd/spire-agent/cli/run/run_posix_test.go
amartinezfayo: I think that running `os.Exit(0)` at the end of each test run should do the trick. By default, golang will panic unless `test.paniconexit0` is set to false. I'm curious to know if @azdagron has a suggestion here.

amartinezfayo (commented):
Thank you for this, @guilhermocc!

cmd/spire-agent/cli/run/run_posix_test.go
amartinezfayo: We discussed this in the contributors sync, and the consensus was that enhancing the fflag package to be able to externally set `singleton.loaded` to false is probably the best solution for this.

guilhermocc: I will proceed by extending fflag package with this feature 👍

pkg/common/fflag/fflag.go
```diff
@@ -91,6 +91,23 @@ func Load(rc RawConfig) error {
 	return nil
 }
 
+// Unload resets the feature flags states to its default values.
+func Unload() error {
+	singleton.mtx.Lock()
+	defer singleton.mtx.Unlock()
+
+	if !singleton.loaded {
+		return errors.New("feature flags has not been loaded")
```
amartinezfayo, line 100:
```suggestion
		return errors.New("feature flags have not been loaded")
```

pkg/common/fflag/fflag.go
```diff
@@ -91,6 +91,23 @@ func Load(rc RawConfig) error {
 	return nil
 }
 
+// Unload resets the feature flags states to its default values.
```
amartinezfayo, line 94: I think that we need to update the comment for the Load function, to reflect the fact that Unload can be called to reset the state. It would be probably a good idea to point out that the Unload function is only designed for testing, and it's not expected to be called by the normal execution of SPIRE.

cmd/spire-server/cli/run/run_posix_test.go
```diff
@@ -4,18 +4,168 @@
...
+			defer os.RemoveAll(testTempDir)
+
+			cmd := &Command{
+				logOptions:         testCase.fields.logOptions,
+				env:                testCase.fields.env,
+				allowUnknownConfig: testCase.fields.allowUnknownConfig,
+			}
+
+			go func() {
+				time.Sleep(1 * time.Second)
```
amartinezfayo, line 143: Generally speaking, it's not a good idea to sleep the process to wait for an event to occur. Is this to wait that the server loaded completely? Could we inspect the output to see the progress of the loading and have a timeout?

cmd/spire-server/cli/run/run_posix_test.go
```diff
@@ -4,18 +4,168 @@
...
+	"github.com/spiffe/spire/pkg/common/log"
 	"github.com/spiffe/spire/pkg/server"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 const (
-	configFile = "../../../../test/fixture/config/server_good_posix.conf"
+	configFile  = "../../../../test/fixture/config/server_good_posix.conf"
+	testTempDir = "/tmp/spire-server-test"
```
amartinezfayo, line 24: This should be handled by calling t.TempDir().

cmd/spire-server/cli/run/run_posix_test.go
```diff
@@ -4,18 +4,168 @@
...
+				code:           0,
+				dataDirCreated: fmt.Sprintf("%s/data", testTempDir),
+			},
+		},
+	}
+	for _, testCase := range tests {
+		t.Run(testCase.name, func(t *testing.T) {
+			_ = fflag.Unload()
+			os.RemoveAll(testTempDir)
+			defer os.RemoveAll(testTempDir)
```
amartinezfayo, line 134: If t.TempDir() is used, this is not needed.

cmd/spire-server/cli/run/run_posix_test.go
guilhermocc: Yeah, apparently i can use spiretest lib to inspect the output logs for waiting right?

cmd/spire-server/cli/run/run_posix_test.go
guilhermocc: In this specific test i wasn't able to to use  t.TempDir() because I needed to interpolate this dir path into the loaded conf file ( server_run_posix.conf), but I see that in another test case we used env vars for making this interpolation (server_good_templated.conf), I'm going to take the same approach here 👍

amartinezfayo (approved):
Thank you @guilhermocc@

PR #3561, opened 2022-11-02
Update versions in main

rturner3:


azdagron (approved):
Thanks, @rturner3 !

test/integration/suites/upgrade/versions.txt
```diff
@@ -3,3 +3,4 @@
 1.4.2
 1.4.3
 1.4.4
+1.4.5
```
amartinezfayo, line 6: 1.5.0 is missing here.

rturner3: Ah, you're right. I always get confused which versions belong in here.

MarcosDY (approved):
LGTM!

PR #3563, opened 2022-11-03
Update release documentation to reflect more currently followed practices

rturner3:
Much of the release documentation has fallen out of date. There are some key differences to how we manage the project that are not accounted for in the currently documented process:

- Correlation of PRs to releases is tracked with GitHub milestones
- Every release has its own dedicated release branch
- We are not triggering releases directly off of the HEAD of main branch
- We are post-1.0, where the digits of the version strings have different meanings than in pre-1.0 releases. In many cases we were using the term "major" where we should be using "minor", and "minor" when we should be using "patch".
- The maintainers follow an on-call style rotation, where duties are distributed across the maintainers depending on when they are on-call.

Out-of-scope in this PR is defining a process for major version releases, e.g. 2.x, because we have not defined a conclusive process for this yet post-1.0.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -68,13 +68,18 @@ The above list is advisory, and is meant only to get the mind going.
 The SPIRE project maintains active support for both the current and the previous major versions. All active development occurs in the `main` branch. Version branches are used for minor releases of the previous major version when necessary.
 
 ### Version Branches
-When a bug is discovered in the latest release that also affects releases of the prior major version, it is necessary to backport the fix.
+Each release must have its own release branch following the naming convention `release/vX.Y.Z` where `X` is the major version, `Y` is the minor version, and `Z` is patch version.
 
-If it is the first time that the prior major version is receiving a backported patch, then a version branch is created to track it. The version branch is named `vX.Y` where X and Y are the two most significant digits in the semantic version number. Its base is the last tag present in main for the release in question. For example, if SPIRE is on version 0.9.3, and the last 0.8 release was 0.8.4, then a `v0.8` branch is created with its base being the main commit tagged with `v0.8.4`.
+The base commit of the release branch is based on the type of release being generated:
+- Patch release for older minor release series. In this case, the new release branch is based off of the previous patch release branch for the same minor release series. Example: the latest release is v1.5.z, and the release being prepared is v1.4.5. The base commit should be the `release/v1.4.4` branch.
+- Security release for current minor release series. In this case, the new release branch should be based off of the previous release branch for the same minor release series. Example: the latest release is v1.5.0, and the release being prepared is v1.5.1. The base commit should be the `release/v1.5.0` branch.
+- Scheduled patch release for current minor release series OR scheduled minor release. In this case, the new release branch should be based off of a commit on the `main` branch. Example: the latest release is v1.5.0, and the release being prepared is v1.5.1. The base commit should be the candidate commit selected from the `main` branch.
+
+When a bug is discovered in the latest release that also affects releases of the prior minor version, it is necessary to backport the fix.
 
 Once the version branch is created, the patch is either cherry picked or backported into a PR against the version branch. The version branch is maintained via the same process as the main branch, including PR approval process etc.
 
-Releases for the previous major version are made directly from its version branch. Ensure that the CHANGELOG is updated in both the main and the version branch to reflect the new release.
+Ensure that the CHANGELOG is updated in both `main` and the version branch to reflect the new release.
 
 ### Releasing
 The SPIRE release machinery is tag-driven. When the maintainers are ready to release, a tag is pushed referencing the release commit. While the CI/CD pipeline takes care of the rest, it is important to keep an eye on its progress. If an error is encountered during this process, the release is aborted.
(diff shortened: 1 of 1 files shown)


MAINTAINERS.md
```diff
@@ -88,41 +93,43 @@ A simple majority vote is required to authorize a SPIRE release at a specific co
...
-  * Cherry-pick into the version branch the commits for all the changes that must be included in the release.
+  * Cherry-pick into the version branch the commits for all the changes that must be included in the release. Ensure the PRs for these commits all target the release milestone in GitHub.
+
+The following steps must be completed by the primary on-call maintainer at least one day before the release (preferably the Monday of the release week):
+* Create a draft pull request against the main branch with the updates to the CHANGELOG following [these guidelines](doc/changelog_guidelines.md). This allows those tracking the project to have early visibility into what will be included in the upcoming release and an opportunity to provide feedback. The release date can be set as "TBD" while it is a draft.
 
-**If this is a major release**, the following steps must be completed before releasing:
+**If this is a major or minor release**, the following steps must be completed by the secondary on-call maintainer before releasing:
```
rturner3, line 108: The delegation of this responsibility to the secondary on-call maintainer is a new proposal in order to share some of the release burden load across maintainers for major/minor releases.

MAINTAINERS.md
```diff
@@ -88,41 +93,43 @@ A simple majority vote is required to authorize a SPIRE release at a specific co
...
+* Create Git tags (not annotated) with the name `vX.Y.Z` in the [spire-api-sdk](https://github.com/spiffe/spire-api-sdk) and [spire-plugin-sdk](https://github.com/spiffe/spire-plugin-sdk) repositories for the HEAD commit of the main branch.
+* Close the GitHub issue created to track the release process.
+* Broadcast news of release to the community via available means: SPIFFE Slack, Twitter, etc.
 * Open and merge a PR to bump the SPIRE version to the next projected version and [update the upgrade integration test](test/integration/suites/upgrade/README.md#maintenance).
   * For example, after releasing 0.10.0, update the version to 0.10.1, since it is more likely to be released before 0.11.0.
   * Ideally, this is the first commit merged following the release.
+* Create a new GitHub milestone for the next release, if not already created.
 
-**If this is a major release**, the following steps must be completed no later than one week after the release:
+**If this is a major or minor release**, the following steps must be completed by the secondary on-call maintainer no later than one week after the release:
```
rturner3, line 130: The delegation of this responsibility to the secondary on-call maintainer is a new proposal in order to share some of the release burden load across maintainers for major/minor releases.

MAINTAINERS.md
evan2645: I think this is a great responsibility to delegate to the secondary, however I'm not sure that we want to do it for _every_ release. It's a fairly heavy lift to walk through everything. Breaking changes are most important, which is why this was originally limited to majors only.

That said, I think we've been pretty lax on this bit in practice, even for majors only

rturner3: Oh, I was using the semver terminology here - by major I meant x.0.0 and minor I meant x.y.0. Were you thinking of major as x.y.0 and minor as x.y.z?

MAINTAINERS.md
```diff
@@ -88,41 +93,43 @@ A simple majority vote is required to authorize a SPIRE release at a specific co
...
 * Push the annotated tag to SPIRE, and watch the build to completion.
   * If the build fails, or anything unusual is encountered, abort the release.
     * Ensure that the GitHub release, container images, and release artifacts are deleted/rolled back if necessary.
-* Visit the releases page on GitHub, copy the release notes, click edit and paste them back in. This works around a GitHub rendering bug that you will notice before completing this task.
-* Close the GitHub project created to track the release process.
+* Visit the releases page on GitHub, copy the release notes, click edit and paste them back in. This works around a GitHub Markdown rendering bug that you will notice before completing this task.
+* Create Git tags (not annotated) with the name `vX.Y.Z` in the [spire-api-sdk](https://github.com/spiffe/spire-api-sdk) and [spire-plugin-sdk](https://github.com/spiffe/spire-plugin-sdk) repositories for the HEAD commit of the main branch.
```
MarcosDY, line 122: this ones may have more work, we have releases we need to cherry pick commits from next, and and then we can create the tag, and it is possible we forgot to apply those changes...
So what do you think about creating a DRAFT PR 1w before release (or the same day we start with the release process) so we can be sure we get into all we need? 
and at the same time people that is watching for changes can get noticed...
Another possible approach is that any time we have a PR merged to SPIRE that required changes on SDK we merge them to main...

MarcosDY (commented):
Thanks for this Ryan!!! looks great

MAINTAINERS.md
rturner3: spire-tutorials is using v1.5.0, and there is an open PR to use v1.5.0 in spire-examples: https://github.com/spiffe/spire-examples/pull/52

I think once we've updated the examples to use the latest minor version, usually there shouldn't be too many changes we need to make each minor release other than removing/replacing deprecated configs. spire-tutorials has a CI job that runs all the tutorials for you, so testing is not too bad in that repo. For spire-examples, the examples need to be run manually today, but there are also only a handful of examples, so it usually isn't too much effort.

Any other things you can think of that might be time consuming for updating the examples each minor release? For upgrading from v1.0.z/v1.2.z to v1.5.0 in the PRs I referenced, there wasn't all that much to do, so initially I wasn't all that concerned about this being a lot of unexpected work for the secondary.

MAINTAINERS.md
```diff
@@ -88,41 +93,43 @@ A simple majority vote is required to authorize a SPIRE release at a specific co
...
-  * If there is not a version branch for this release, create a branch following the guidelines described in [Version branches](#version-branches).
-  * Create a GitHub project named `Release vX.X.X` to identify the PRs that will be cherry-picked. The project should have two statuses to track the progress: one to identify the PRs to be cherry-picked and one for those that have been merged in the version branch.
   * Make sure that the [version in the branch](pkg/common/version/version.go) has been bumped to the version that is being released and that the [upgrade integration test is updated](test/integration/suites/upgrade/README.md#maintenance).
-  * Cherry-pick into the version branch the commits for all the changes that must be included in the release.
+  * Cherry-pick into the version branch the commits for all the changes that must be included in the release. Ensure the PRs for these commits all target the release milestone in GitHub.
+
+The following steps must be completed by the primary on-call maintainer at least one day before the release (preferably the Monday of the release week):
```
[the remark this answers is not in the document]
rturner3: I proposed this change mostly because we now track all the PRs for our releases through GitHub milestones, which are also publicly visible. But I think giving people time to review the PR in advance is also good so that other maintainers that are not on-call don't need to be available during release day.

I will change this back to how it was.

MAINTAINERS.md
```diff
@@ -88,41 +93,43 @@ A simple majority vote is required to authorize a SPIRE release at a specific co
...
+* Create a draft pull request against the main branch with the updates to the CHANGELOG following [these guidelines](doc/changelog_guidelines.md). This allows those tracking the project to have early visibility into what will be included in the upcoming release and an opportunity to provide feedback. The release date can be set as "TBD" while it is a draft.
```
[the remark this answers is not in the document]
rturner3: Good catch, this needs to be updated. Back when we weren't using release branches for each release, we used to target the `main` branch. But now that we have release branches for each release, we should target the release branch, then port forward into the `main` branch after the release.

MAINTAINERS.md
rturner3: Yeah this is a good point. I will make a note of making sure the spire-api-sdk and spire-plugin-sdk repositories have  all intended changes merged from `next` to `main` before the changelog PR is prepared one week prior to the release.

azdagron (approved):
LGTM!

MarcosDY (approved):
Looks great!

evan2645 (approved):

PR #3564, opened 2022-11-03
Update SECURITY.md

rturner3:
This document has fallen out of date and does not reflect the currently supported versions. In order to keep this document current without needing to update it for every minor release series, document the policy rather than the exact versions that are supported.


azdagron (approved):
Awesome. Thanks for doing this, @rturner3 !

PR #3578, opened 2022-11-04
Add bundle_file_arn to plugin options for awssecret

### LINKED ISSUE #3554: Add support for additional upstream root certificates to UpstreamAuthority awssecret
When using the `awssecret` UpstreamAuthority plugin, it assumes that the user is specifying a self-signed CA. Unlike the `disk` or `aws_pca` plugins, there is no way to specify additional certificates that chain up to a self-signed root to be included in the trust bundle. 

This means that if the certificate and key pair specified in `cert_file_arn` and `key_file_arn` are themselves an intermediate, the trust bundle served by SPIRE does not contain the full chain required to validate a SPIFFE ID. 

Request: add an equivalent to the `supplemental_bundle_path` or `bundle_file_path` setting allowing users to specify additional certificates that should be included in the trust bundle. 

* **Version**: Latest
* **Platform**: `Linux 7e5256c9610e 5.10.104-linuxkit #1 SMP PREEMPT Thu Mar 17 17:05:54 UTC 2022 x86_64 GNU/Linux`
* **Subsystem**: UpstreamAuthority awssecret

jcrumb-u21:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
`UpstreamAuthority awssecret`

**Description of change**
This will allow us to specify addtional certificates that should be added to the trust bundle when using non-self-signed certificates in the cert_file_arn configuration field. Without this, it is only possible to use the awssecret UpstreamAuthority with a self-signed cert. Otherwise, you will end up with a partial chain when presented with the trust bundle + client certificate chain.

This parameter is optional, and when omitted does not break existing flows.

**Which issue this PR fixes**
Fixes #3554


evan2645:
Thanks so much for opening this @jcrumb-u21!

I've been spending some time reading over your issue and looking at this PR. I think there's a bit of conflation going on between chain management and bundle management. Assuming I understand correctly, it looks like you have an intermediate CA that you're storing in aws secrets, and you're wanting to load it for use as a signing cert in SPIRE ... however, the plugin (as you've observed) doesn't seem to support this. It seems pretty clear to me that this is an oversight in the plugin functionality. This case should definitely be supported. Thank you for sending a patch!!

I think we need to fix the chain management logic in addition to adding `bundle_file_arn`, and update the docs accordingly. Of course, we'll want the change to be backwards compatible. Generally speaking, chains should include all certs except the root. Bundle should include roots only. We want to be sure to preserve this structure in our plugin response, because workloads also need to be able to distinguish between chain and roots (e.g. for TLS), and we preserve this arrangement across the workload API. Additionally, strange things can happen in various client libraries if these get mixed (i.e. if intermediates are present in the bundle).

Sorry for the wall of text. I'm happy to hop on a zoom call or something to discuss further if that helps too. Here's a quick summary of the behavior that I think makes sense for this plugin
* When providing a root for signing, user should be able to set only `cert_file_arn` and `key_file_arn`
* When providing a root for signing, but also wanting to add other CAs to the bundle, the user should be able to set `cert_file_arn` (with the root), `key_file_arn`, and finally `bundle_file_arn` to provide supplementary roots
* When providing an intermediate for signing, user should be able to set `cert_file_arn` with the intermediate plus its chain (if any), `key_file_arn`, and `bundle_file_arn` with one or more roots

I think this necessitates detecting whether or not `cert_file_arn` is a root. There is also some more info on how the response should be arranged here: https://github.com/spiffe/spire-plugin-sdk/blob/main/proto/spire/plugin/server/upstreamauthority/v1/upstreamauthority.proto#L44-L53

jcrumb-u21:
I got pulled down a bit of a rabbit hole and this ended up being a much bigger refactor than I initially thought! I made a few related changes:
- I broke the certificate parsing/validation logic out of the `fetchFromSecretsManager` and into its own function
- I added much stricter verification on startup to ensure that the certificate(s) and bundle provided will properly validate
- The static certs that were used for testing had expired. Rather than just generating new ones that will expire at some point in the future, I adjusted the tests to generate certs + keys on each test run that have expiry dates constrained by the mock clock.

doc/plugin_server_upstreamauthority_awssecret.md
```diff
@@ -12,6 +12,7 @@ The plugin accepts the following configuration options:
 | region            | AWS Region that the AWS Secrets Manager is running in |
 | cert_file_arn     | ARN of the "upstream" CA certificate                  |
 | key_file_arn      | ARN of the "upstream" CA key file                     |
+| bundle_file_arn   | ARN of additional roots to include in the trust bundle. If `cert_file_arn` contains a self-signed certificate this field can be left unset. Otherwise, `bundle_file_arn` must include one or more certificates necessary to establish a valid chain up to a self-signed certificate. |
```
evan2645, line 15:
```suggestion
| bundle_file_arn   | ARN of roots to include in the trust bundle. If `cert_file_arn` contains a self-signed root CA certificate this field can be left unset. Otherwise, `bundle_file_arn` must include one or more root CA certificates |
```

doc/plugin_server_upstreamauthority_awssecret.md
```diff
@@ -12,6 +12,7 @@ The plugin accepts the following configuration options:
 | region            | AWS Region that the AWS Secrets Manager is running in |
 | cert_file_arn     | ARN of the "upstream" CA certificate                  |
```
evan2645, line 13:
```suggestion
| cert_file_arn     | ARN of the "upstream" CA certificate that will be used for signing. If more than one certificate is present, they will be added to the chain in order of appearance, where the first certificate will be the one used for signing.                 |
```

evan2645 (commented):
Thanks so much for this contrib @jcrumb-u21 , and sorry for the delay in review.

I spent an hour or so going over this .. there were a handful of bits that at first appeared confusing then resolved with further review. I think this plugin could benefit from major refactor, along with its test suite .. normally I'd ask for more tests and clearer logic, but I know this change has already creeped on you and I don't expect you to undertake a refactor.

I'll see if anyone else wants to have a look at this before merging ... but it looks alright to me aside from my doc comments. Thanks again for hanging in there

jcrumb-u21:
@evan2645 confirmed the following testcases locally/in our AWS playground with this build:

✅ Running without a `bundle_file_arn` set and a self-signed certificate in `cert_file_arn` still works

✅ Running with a single intermediate in `cert_file_arn` and the root that issued it in `bundle_file_arn` results in an SVID being minted with three certificates: the SVID cert, the generated spire CA cert (signed by the intermediate), and the intermediate cert in that order. Then, the trust bundle contains only the root specified in `bundle_file_arn`.

✅ Confirmed that the above setup plays nicely with Envoy via SDS. SVIDs that are generated from this build with an intermediate in `cert_file_arn` are (finally!) verifiable with Envoy for mTLS purposes. 

❌ If you specify only a single intermediate in `cert_file_arn` but leave `bundle_file_arn` unset, your SVID/bundle combo will not be able to verify properly due to the missing root in the trust bundle. There's no warning or error thrown by spire in this condition; that was already the case though and is still the case in the disk plugin as well. I'd like to leave this as a TODO for a later refactor of these two plugins since they work nearly identically and can probably take much more advantage of code sharing.

evan2645 (dismissed):
> ❌ If you specify only a single intermediate in cert_file_arn but leave bundle_file_arn unset, your SVID/bundle combo will not be able to verify properly due to the missing root in the trust bundle. There's no warning or error thrown by spire in this condition; that was already the case though and is still the case in the disk plugin as well. I'd like to leave this as a TODO for a later refactor of these two plugins since they work nearly identically and can probably take much more advantage of code sharing.

Given that this is the current behavior, I think opening an issue about it is reasonable. It can be solved at another time.

@rturner3 mentioned that he might give this a look in the next week or two. It looks good to me otherwise, should go out with 1.5.4 barring any surprises.

evan2645 (approved):
Since we squash when we merge, it's ok to push merge commits onto this PR/branch. It's actually a bit easier for us, because GH will assist in diff review between the last approval and now, but not if there was a force push. I think it will also not dismiss a review if all you've done is merge main.

I'll keep an eye on these tests and merge when they're green. Thanks @jcrumb-u21 !

PR #3606, opened 2022-11-11
Use `default_x509_svid_ttl` instead of the deprecated `default_svid_ttl` config

amartinezfayo:
Update our release template config files to use `default_x509_svid_ttl` instead of the deprecated `default_svid_ttl` config setting.
Update the tests also (except for one that purposely is kept to exercise the deprecated setting).


azdagron (dismissed):
\o/

azdagron (approved):

PR #3616, opened 2022-11-16
Fix race in TestDisposeActiveCryptoKeys

amartinezfayo:
Fix race in TestDisposeActiveCryptoKeys accessing the stored fake CryptoKeys through the use of locks.


MarcosDY (approved):
LGTM!!!

PR #3631, opened 2022-11-23
Update help in Makefile to include lint commands

amartinezfayo:
Update help in Makefile to include lint commands.


azdagron (approved):
\o/

PR #3636, opened 2022-11-23
Replace usage of alpine images with scratch images

rturner3:
- Remove usage of wait-for-it image from GCR that doesn't exist in GHCR and shouldn't be required.
- Add `/tmp` directory to scratch images so that scratch images can be run as non-root user in upgrade integration test, since only the `root` user has permission to create directories at the root level `/` of the filesystem
- Temporarily move the server socket to the data directory in the upgrade integration test since 1.4.x scratch images don't have `/tmp`
- Add some resilience to the upgrade integration test so that it doesn't depend on near-instantaneous startup of containers
- In some cases where tools are expected in a server or agent container, or where files need to be read/written inside the container as a different user than the uid of spire-server/spire-agent, use an alpine-based image for convenience.


amartinezfayo (commented):
Looks like the private socket is not found.

```
Error: connection error: desc = "transport: error while dialing: dial unix /tmp/spire-server/private/api.sock: connect: no such file or directory"
[2022-12-02T13:23:55Z] step 01-run-upgrade-tests failed
```

rturner3:
> Looks like the private socket is not found.
> 
> ```
> Error: connection error: desc = "transport: error while dialing: dial unix /tmp/spire-server/private/api.sock: connect: no such file or directory"
> [2022-12-02T13:23:55Z] step 01-run-upgrade-tests failed
> ```

Oof, I forgot some backslashes in some multi-line commands I edited in the upgrade test bash script. I'm not sure how it's resulting in that error about the wrong socket path though since the missing backslashes were after the `-socketPath` argument...

rturner3:
> > Looks like the private socket is not found.
> > ```
> > Error: connection error: desc = "transport: error while dialing: dial unix /tmp/spire-server/private/api.sock: connect: no such file or directory"
> > [2022-12-02T13:23:55Z] step 01-run-upgrade-tests failed
> > ```
> 
> Oof, I forgot some backslashes in some multi-line commands I edited in the upgrade test bash script. I'm not sure how it's resulting in that error about the wrong socket path though since the missing backslashes were after the `-socketPath` argument...

I realized I added some comments mid-command in the multiline commands, which broke them. The test should be fixed now.

developer-guy:
That's super cool! But what about creating a secure version of the wait-for-it image by using Chainguard tooling apko + melange? I've created several PRs to build a secure version of the wait-for-it image. 

/cc @anvega @dlorenc

> https://github.com/chainguard-images/images/pull/115

rturner3:
> That's super cool! But what about creating a secure version of the wait-for-it image by using Chainguard tooling apko + melange? I've created several PRs to build a secure version of the wait-for-it image.
> 
> /cc @anvega @dlorenc
> 
> > [chainguard-images/images#115](https://github.com/chainguard-images/images/pull/115)

In our case, the wait-for-it init container was only used as a way to prevent some error logs when starting all the SPIRE components at once with tests and example K8s configs. In practice, users shouldn't need to rely on a wait-for-it init container when deploying SPIRE. It's recommended to deploy the Server first and then deploy the Agent to minimize Agent downtime during upgrades.

support/k8s/k8s-workload-registrar/mode-crd/config/spire-agent.yaml
```diff
@@ -108,16 +108,9 @@ spec:
       hostNetwork: true
       dnsPolicy: ClusterFirstWithHostNet
       serviceAccountName: spire-agent
-      initContainers:
```
marcofranssen, line 111: Without this the spire-agents will crash couple of times until server is up on a fresh install.

image

dlorenc: We'll happily package wait-for-it and maintain images for this if it helps, like @developer-guy suggested above.

marcofranssen:
> minimize

With a helm-chart you don't really control that. Also when deploying with argocd or flux you can't control that really.

support/k8s/k8s-workload-registrar/mode-crd/config/spire-agent.yaml
evan2645: ❤️ thank you @developer-guy and @dlorenc .. we will take you up on that!

We discussed this change today and concluded that it will be best to avoid the use of `wait-for-it` in our test suites, however we still want to use it in all of our tutorials. I left a comment on an existing issue about it here https://github.com/spiffe/spire-tutorials/issues/90#issuecomment-1343280662

Thanks again!

rturner3:
I realized today that I missed the instances in integration tests where we were referring to the `latest-local` images. I will push more changes to migrate those tests to using the locally built scratch images.

test/integration/suites/k8s-crd-mode/Dockerfile
```diff
@@ -1,3 +1,8 @@
-FROM spire-agent:latest-local AS example-crd-agent
-CMD []
+FROM spire-agent-scratch:latest-local AS spire-agent
+
+FROM alpine:3.17 AS example-crd-agent
+RUN apk add --no-cache --update dumb-init
 RUN apk add --no-cache --update openssl
+
+COPY --from=spire-agent /opt/spire/bin/spire-agent /opt/spire/bin/spire-agent
+CMD []
```
azdagron, line 8: maybe move above line 4 for better docker layer caching?

test/integration/suites/nested-rotation/Dockerfile
```diff
@@ -1,3 +1,5 @@
-FROM spire-agent:latest-local AS nested-agent
-CMD []
+FROM spire-agent-scratch:latest-local AS nested-agent
+FROM alpine:3.17 as nested-agent-alpine
 RUN apk add --no-cache --update openssl
+COPY --from=nested-agent /opt/spire/bin/spire-agent /opt/spire/bin/spire-agent
```
azdagron, line 4: I think --from can take `spire-agent-scratch:latest-local` directory. If so, we can get rid of that build stage and keep the name nested-agent instead of having both nested-agent and nested-agent-alpine.

test/integration/suites/spire-server-cli/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM spire-server-scratch:latest-local AS test-spire-server
+FROM alpine:3.17 AS spire-server-alpine
+RUN apk add --no-cache --update openssl
+COPY --from=test-spire-server /opt/spire/bin/spire-server /opt/spire/bin/spire-server
```
azdagron, line 4:
```suggestion
FROM alpine:3.17 AS spire-server-alpine
RUN apk add --no-cache --update openssl
COPY --from=spire-server-scratch:latest-local /opt/spire/bin/spire-server /opt/spire/bin/spire-server
```

test/integration/suites/upstream-authority-cert-manager/03-verify-ca
```diff
@@ -13,10 +13,18 @@ expLeafURI="URI:spiffe://example.org/ns/foo/sa/bar"
 
 log-debug "verifying CA..."
 
-./bin/kubectl exec -n spire $(./bin/kubectl get pod -n spire -o name) -- /opt/spire/bin/spire-server x509 mint -spiffeID spiffe://example.org/ns/foo/sa/bar -write .
-leafURIResult=$(./bin/kubectl exec -n spire $(./bin/kubectl get pod -n spire -o name) -- cat svid.pem | openssl x509 -noout -text | grep URI | sed 's/^ *//g')
-leafIssuerResult=$(./bin/kubectl exec -n spire $(./bin/kubectl get pod -n spire -o name) -- cat svid.pem | openssl x509 -noout -issuer)
-caSubjectResult=$(./bin/kubectl exec -n spire $(./bin/kubectl get pod -n spire -o name) -- cat bundle.pem | openssl x509 -noout -subject)
+mintx509svid_out=mintx509svid-out.txt
```
azdagron, line 16: I'm not sure i understand the reason for this change...

azdagron (commented):
Looking pretty good, @rturner3. Just a few comments...

test/integration/suites/upstream-authority-cert-manager/03-verify-ca
rturner3: The `spire-server x509 mint` command executed in this script was writing to a directory in the container. We don't have the ability to read files from scratch containers because there are no utilities like `cat`. Rather than writing to files in the container, I opted to just parse the command output on `stdout`.

Alternatively, we could try to introduce a bind mount into the container and have the command write its output there.

azdagron: Parsing unstructured text is a little gross. How bad is the bind mount path? I'm might be ok with this for now. Presumably at some point we get --format=json for this CLI command which should improve things significantly.

rturner3: Taking another look at this test, it is deploying the containers with kind + Kubernetes. I'm not too familiar with kind, so it would be nontrivial for me at least, but there's probably a way to make it work.

azdagron: Ugg. Yeah, let's just keep this for now. We can improve it when structured output support lands for this command.

azdagron (approved):
Thanks, @rturner3 !

Dockerfile.scratch
```diff
@@ -7,11 +7,13 @@ ADD go.* ./
 RUN go mod download
 ADD . .
 RUN make build-static
+RUN install -d -o root -g root -m 1777 /newtmp
```
marcofranssen, line 10: does it have to be root or can we also create a new user?

That would also allow me to configure the helm-chart to run a a certain user and resolve some security issues.

test/integration/suites/spire-server-cli/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM alpine:3.17 AS spire-server-alpine
+RUN apk add --no-cache --update openssl
+COPY --from=spire-server-scratch:latest-local /opt/spire/bin/spire-server /opt/spire/bin/spire-server
+ENTRYPOINT ["/opt/spire/bin/spire-server", "run"]
```
marcofranssen, line 4: Do the entrypoint first, this will allow better reuse of layer caching.

Dockerfile.scratch
rturner3: Hi @marcofranssen, I responded in the issue you linked to. We want to do this, but we are planning to handle it in a separate PR.

test/integration/suites/spire-server-cli/Dockerfile
rturner3: Thanks for the suggestion. In this case, you would only notice a caching benefit if you ran this integration test multiple times successively with different scratch server images, which doesn't happen on CI. I agree it's an optimization, but it has pretty minor impact overall, so I think we will merge this as is for now, but we'd be happy to accept another PR with that change.

PR #3638, opened 2022-11-24
Do not try to close `gcp_kms` client if not initialized

amartinezfayo:
If there is a configuration error in the "gcp_kms" plugin, the client will not be initialized and the Close() operation should not be attempted, because there is no client.


guilhermocc (approved):

MarcosDY (approved):
LGTM!

PR #3643, opened 2022-11-29
Auto remove markdown lint container

guilhermocc:
Auto remove markdown lint container after executing it


azdagron (approved):
Good catch!

PR #3653, opened 2022-12-01
Use the UpdateMask field when calling UpdateCryptoKey

amartinezfayo:
If the algorithm for a CryptoKey needs to be changed, a call to UpdateCryptoKey must be performed.
This call requires to specify the UpdateMask field.


MarcosDY (approved):
LGTM!!

PR #3655, opened 2022-12-01
Retry GetPublicKey if needed in gcp_kms plugin

amartinezfayo:
If a CryptoKeyVersion was just generated with a large key size, when attempting to get the public key the CryptoKeyVersion may be still in the `CryptoKeyVersion_PENDING_GENERATION` status.
Implement a retry logic in that case.


MarcosDY (approved):
LGTM!

PR #3656, opened 2022-12-02
Refactor push-images script to reduce duplication

marcofranssen:
Can be used for both scratch and non-scratch
images.

Signed-off-by: Marco Franssen <marco.franssen@gmail.com>

**Pull Request check list**

- [ ] Commit conforms to CONTRIBUTING.md?
- [ ] Proper tests/regressions included?
- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**

PR #3676, opened 2022-12-07
Revert "Refactor push-images script to reduce duplication"

rturner3:
Reverts spiffe/spire#3656

This change causes the k8s-workload-registrar scratch image to be published to GHCR, which is not something we have done in the past and do not want to do for future releases. Backing this change out on main for now to unblock the 1.5.3 release.


rturner3:
Clicking the "Revert" button from the GitHub UI doesn't create a commit that passes DCO. Re-creating this PR manually.

PR #3677, opened 2022-12-07
Revert "Refactor push-images script to reduce duplication (#3656)"

rturner3:
This reverts commit fde87a667d2fdab7086db70c3c3a5041c6bc0099.

This change causes the k8s-workload-registrar scratch image to be published to GHCR, which is not something we have done in the past and do not want to do for future releases. Backing this change out on main for now to unblock the 1.5.3 release.


marcofranssen:
Please hold this and simply let me add the if clause.

marcofranssen:
See https://github.com/spiffe/spire/pull/3678 for the fix.

rturner3:
Closed in favor of #3678.

PR #3696, opened 2022-12-12
Expand gcp kms plugin test scenarios

### LINKED ISSUE #3657: gcp_kms plugin: create a test that exercises the logic that retries getting a public key
Add a test that exercises the logic that retries getting a public key when the CryptoKeyVersion status is still `CryptoKeyVersion_PENDING_GENERATION`.

guilhermocc:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
No functionality is affected

**Description of change**
Create new test scenario for gcp kms plugin, for testing the retry mechanism on key `generation_pending` state

**Which issue this PR fixes**
fixes #3657


faisal-memon:
i can help with review

pkg/server/plugin/keymanager/gcpkms/client_fake.go
```diff
@@ -324,11 +330,35 @@ func (k *fakeKMSClient) setIsKeyDisabled(ok bool) {
 	k.keyIsDisabled = ok
 }
 
-func (k *fakeKMSClient) setGetPublicKeyErr(fakeError error) {
+func (k *fakeKMSClient) setGetPublicKeySequentialErrs(fakeErrors ...error) {
 	k.mu.Lock()
 	defer k.mu.Unlock()
+	errsRing := ring.New(len(fakeErrors))
```
faisal-memon, line 336: Are you using the ring buffer so you don't have to track the index?

pkg/server/plugin/keymanager/gcpkms/client_fake.go
```diff
@@ -709,6 +738,7 @@ func (k *fakeKMSClient) createFakeCryptoKeyVersion(cryptoKey *kmspb.CryptoKey, v
 		},
 		CryptoKeyVersion: &kmspb.CryptoKeyVersion{
 			Name:      path.Join(cryptoKey.Name, "cryptoKeyVersions", version),
+			State:     k.initialCryptoKeyVersionState,
```
faisal-memon, line 741: what is this `initialCryptoKeyVersionState` change accomplishing?

pkg/server/plugin/keymanager/gcpkms/client_fake.go
```diff
@@ -324,11 +330,35 @@ func (k *fakeKMSClient) setIsKeyDisabled(ok bool) {
...
+			k.getPublicKeyErrsRing = k.getPublicKeyErrsRing.Next()
+		}
+		k.mu.Unlock()
+	}()
+
+	if k.getPublicKeyErrsRing == nil || k.getPublicKeyErrsRing.Value == nil {
+		return nil
+	}
+
+	return k.getPublicKeyErrsRing.Value.(error)
```
faisal-memon, line 361: I think the function logic would be simpler if you just make the write lock at the beginning at the cost of a bit of efficiency with the fine grained read/write locking.

guilhermocc: Definitely! since this bit of efficiency is not that critical here, I can give it a try 😄

pkg/server/plugin/keymanager/gcpkms/client_fake.go
guilhermocc: Exactly, i wanted a circular linked list behavior for the errors we can configure to be returned so that the errors will be returned sequentially, and we don't actually need to configure the returned error for every call.

pkg/server/plugin/keymanager/gcpkms/client_fake.go
guilhermocc: So, for testing the scenario in which we keep retrying the retrieval of the public key when the crypto key state is still in `CryptoKeyVersion_PENDING_GENERATION` state, we need to be able to mock this initial state, and since we weren't doing it, every crypto key created by our test was starting with the default value `CryptoKeyVersion_CRYPTO_KEY_VERSION_STATE_UNSPECIFIED`, making it impossible to reach the desired scenario.

guilhermocc: The retry that I mentioned can be seem here: pkg/server/plugin/keymanager/gcpkms/gcpkms.go:1056

pkg/server/plugin/keymanager/gcpkms/client_fake.go
guilhermocc: 7d12460

pkg/server/plugin/keymanager/gcpkms/client_fake.go
faisal-memon: Since you're not actually using the ring functionality and wrapping around, i think a better approach is the one used  in fakedatastore.go that gives you the same benefit: https://github.com/spiffe/spire/blob/84866e1b677c8f8da997fd49d2f7ae43f0a3119e/test/fakes/fakedatastore/fakedatastore.go#L322-L329

guilhermocc: yeah, this approach will give me the sequential order for the errors, but not the circular behavior. I will have to update other tests (maybe just one another that reaches the max attempts for retrieving the public key ) but I think it will be worth having a simple code here 😄

guilhermocc: 632d64e

pkg/server/plugin/keymanager/gcpkms/gcpkms_test.go
```diff
@@ -116,6 +116,11 @@ func setupTest(t *testing.T) *pluginTest {
 }
 
 func TestConfigure(t *testing.T) {
+	getPublicKeyMaxAttempts := 11
```
faisal-memon, line 119: Is 11 as the max number of attempts specified somewhere?

pkg/server/plugin/keymanager/gcpkms/client_fake.go
```diff
@@ -324,11 +329,21 @@ func (k *fakeKMSClient) setIsKeyDisabled(ok bool) {
 	k.keyIsDisabled = ok
 }
 
-func (k *fakeKMSClient) setGetPublicKeyErr(fakeError error) {
+func (k *fakeKMSClient) setGetPublicKeySequentialErrs(fakeErrors ...error) {
```
faisal-memon, line 332: What do you think of having this function take in a length and a single error string and then generating the errors array here?

guilhermocc: So, we don't want the same error to be returned here. To reach the desired scenario that this PR proposes (retry mechanism), we need to return some errors and then return nil.

pkg/server/plugin/keymanager/gcpkms/gcpkms_test.go
guilhermocc: Yes, it is defined as a scoped const on pkg/server/plugin/keymanager/gcpkms/gcpkms.go:1032. The actual value of max attempts is 10, but since we have one attempt prior to that, the actual maximum attempts that it does is 11.

faisal-memon: maybe you can move that definition to the const section so you don't have to redefine it: https://github.com/spiffe/spire/blob/main/pkg/server/plugin/keymanager/gcpkms/gcpkms.go#L36

pkg/server/plugin/keymanager/gcpkms/client_fake.go
faisal-memon: I dont think you need nil in the array. Just return nil when you get to the end

guilhermocc: 5641933

pkg/server/plugin/keymanager/gcpkms/gcpkms_test.go
guilhermocc: 5641933

Issue #3688, opened 2022-12-08
SVID issuance audit log


evan2645:
When SPIRE Server mints an SVID for a remote caller, we should get to know details about both the entry that the SVID was minted against, as well as the SVID itself.

Currently, when SPIRE Server mints an SVID, it logs a message with the SPIFFE ID, Caller ID, Entry ID, and Expiration date, amongst other things. Notably missing is the SVID serial number, and the revision number of the Entry. We should definitely log these as well, in a pair of new fields.

```DEBU[0061] Signed X509 SVID                              authorized_as=agent authorized_via=datastore caller_addr="127.0.0.1:61745" caller_id="spiffe://example.org/spire/agent/join_token/99a9a879-4e18-46bb-8389-2dde59050faa" entry_id=d1004dd5-364f-43a3-bdcb-57cce31256d9 expiration="2022-12-08T20:07:26Z" method=BatchNewX509SVID request_id=4a7b1f7e-96d2-42a5-8c98-a8a8f6e28379 service=svid.v1.SVID spiffe_id="spiffe://example.org/foo" subsystem_name=api```

_Originally requested by @amoore877 in https://github.com/spiffe/spire/issues/3253#issuecomment-1341296289_

alexviktorov:
Adding serial and revision number to Batch/New method and only the serial to Mint method

rturner3:
I chatted with @alexviktorov offline and he volunteered to take up this issue, assigning him here. Thanks, @alexviktorov!

PR #3699, opened 2022-12-13
Add serial number and revision number to svid minting log entries

alexviktorov:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Log entries will include extra fields; datetime of Svid minting audit logs changed format from Unix to RFC3339 

**Description of change**
Added serial number and revision number to svid minting log entries: both for Batch/NewX509Svid and only serial number for MintX509Svid

**Which issue this PR fixes**
#3688


pkg/server/api/svid/v1/service.go
```diff
@@ -115,11 +115,12 @@ func (s *Service) MintX509SVID(ctx context.Context, req *svidv1.MintX509SVIDRequ
 	}
 
 	rpccontext.AddRPCAuditFields(ctx, logrus.Fields{
-		telemetry.ExpiresAt: x509SVID[0].NotAfter.Unix(),
+		telemetry.ExpiresAt: x509SVID[0].NotAfter.Format(time.RFC3339),
```
evan2645, line 118: Was this change intentional? I don't have a strong opinion on the formatting here, the concern from my end is more around backward compatibility for folks using this field today

pkg/server/api/svid/v1/service.go
```diff
@@ -115,11 +115,12 @@ func (s *Service) MintX509SVID(ctx context.Context, req *svidv1.MintX509SVIDRequ
...
 	})
 
 	rpccontext.AuditRPCWithFields(ctx, commonX509SVIDLogFields)
 	log.WithField(telemetry.Expiration, x509SVID[0].NotAfter.Format(time.RFC3339)).
+		WithField(telemetry.SerialNumber, x509SVID[0].SerialNumber).
```
evan2645, line 123: I think this is an x509.Certificate type, and the SerialNumber field there is actually *big.Int which is a struct. I think we need to call the String() method to get something printable

```suggestion
		WithField(telemetry.SerialNumber, x509SVID[0].SerialNumber.String()).
```

evan2645 (commented):
Thanks so much for this contribution @alexviktorov! Couple minor comments, should be good to go after that.

pkg/server/api/svid/v1/service.go
alexviktorov: Yes, just a couple of lines below on line 122  the Expiration is logged in time.RFC3339 format so changed it for consistency's sake after discussing with Ryan. I don't have a strong opinion though, I can easily revert it.

evan2645: Ah gotcha ... alrighty then.

evan2645 (approved):
Thank you @alexviktorov!

PR #3707, opened 2022-12-20
Sign images using Sigstore

marcofranssen:
- Limit workflow job permissions to bare minimum
- Add container signing using Sigstore keyless

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

This allows consumers of the image to verify the signature. E.g. via a policy engine in k8s or via `cosign verify` in workflows etc.

See https://registry-ui.chainguard.app/ to explore the signatures of an image.
and https://rekor.tlog.dev/ to explore the transparancy log (shows which workflow was used to sign the image) This information can be used in policies and attestations. 

**Which issue this PR fixes**


.github/workflows/scripts/push-images.sh
```diff
@@ -63,4 +63,8 @@ for img in "${OCI_IMAGES[@]}"; do
   image_to_push="${registry}/${img}:${version}"
   docker tag "${image_variant}:latest-local" "${image_to_push}"
   docker push "${image_to_push}"
+
+  image_digest="$(docker inspect "${image_to_push}" --format '{{ index .RepoDigests 0 }}' | awk -F '@' '{ print $2 }')"
```
marcofranssen, line 67: Retrieving the digest will have to change if the multi-platform arch is merged first

.github/workflows/nightly_build.yaml
```diff
@@ -14,11 +14,19 @@ jobs:
...
+      id-token: write
       packages: write
 
+    env:
+      COSIGN_EXPERIMENTAL: 1
+
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+      - name: Install cosign
```
MarcosDY, line 26: so, we'll be signing only release and nightly builds, how about regular builds images?  (the ones generated when running PRs) 
I think signing regular PR images can be useful to verify sign still working, and we may benefits with that if we create an IT that verify our images are signed

marcofranssen:
Signatures are stored in the registry. As well in the transparency log. IMHO doesn't make sense to do that.

.github/workflows/nightly_build.yaml
```diff
@@ -4,19 +4,29 @@ on:
...
   build-and-publish-images:
     runs-on: ubuntu-20.04
+
+    permissions:
+      contents: read
+      id-token: write
+      packages: write
+
+    env:
+      COSIGN_EXPERIMENTAL: 1
```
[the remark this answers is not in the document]
marcofranssen: yes

.github/workflows/release_build.yaml
```diff
@@ -570,11 +570,19 @@ jobs:
 
     permissions:
       contents: read
+      id-token: write
       packages: write
 
+    env:
+      COSIGN_EXPERIMENTAL: 1
```
puerco, line 577: Same here :)

.github/workflows/nightly_build.yaml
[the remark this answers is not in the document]
cpanato: this is still required until we cut a cosign release v2

.github/workflows/nightly_build.yaml
```diff
@@ -14,11 +14,19 @@ jobs:
...
       contents: read
+      id-token: write
       packages: write
 
+    env:
+      COSIGN_EXPERIMENTAL: 1
+
     steps:
       - name: Checkout
         uses: actions/checkout@v3
```
cpanato, line 25: a good practice is to use the git hash instead of the tag and have a comment to which release is pointing to. Dependabot when update the release can also update the comment :)

```suggestion
        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
```

same applies to the other actions

.github/workflows/nightly_build.yaml
cpanato: agree, you can sign the image that you generate in the main branch build

cpanato (commented):
some comments, this is super nice to see 🎉

.github/workflows/nightly_build.yaml
marcofranssen: For that there is already a nightly in place. On the main branch no images are released, so also nothing to sign there.

.github/workflows/nightly_build.yaml
marcofranssen: There is another PR adressing this practice already.

amartinezfayo (approved):
Thank you @marcofranssen!

PR #3715, opened 2022-12-21
Remove null values from cli json priting

guilhermocc:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
JSON output results from CLI commands

**Description of change**
As discussed in [this thread](https://github.com/spiffe/spire/pull/3660#discussion_r1042523302) from a previous [PR](https://github.com/spiffe/spire/pull/3660), we find it valuable to remove the noise from JSON output messages by removing fields that have a null value. 

**Which issue this PR fixes**
Ongoing work for https://github.com/spiffe/spire/issues/1354


faisal-memon:
i can review this one.

faisal-memon:
Is it possible to set `EmitUnpopulated` to `false` and/or make use of the `omitempty` fields on the go structs? Instead of manually stripping out nulls?

guilhermocc:
@faisal-memon so, the situation here is a little bit more complicated than this:
- We can't disable `EmitUnpopulated` because fields with a zeroed value that should be included in the printing will not be included; some examples are: `Count: 0` or `SomBoolField: false`.
- This specific part of the code deals with converting proto messages to JSON; currently, all of the proto message structs already include the `omitempty` field tag. Still, the lib we use for marshaling proto messages (google.golang.org/protobuf/encoding/protojson) doesn't consider it in its logic. Even if we were using a basic JSON marshal, using the `omitempty` field tag will lead us to the aforementioned problem.

Those are the reasons for having this custom logic of removing null values.

pkg/common/cliprinter/internal/protojson/protojson.go
```diff
@@ -34,13 +34,66 @@ func Print(msgs []proto.Message, stdout, stderr io.Writer) error {
...
 		jms = append(jms, jb)
 	}
-
 	var err error
-	if len(jms) == 1 {
-		err = json.NewEncoder(stdout).Encode(jms[0])
+
+	parsedJms, err := parseJSONMessages(jms)
+	if err != nil {
+		_ = errorjson.Print(err, stdout, stderr)
```
faisal-memon, line 41: Do we want to `return err` here?

pkg/common/cliprinter/internal/protojson/protojson.go
```diff
@@ -34,13 +34,66 @@ func Print(msgs []proto.Message, stdout, stderr io.Writer) error {
...
+
+	removeNulls(jmMap)
+
+	parsedJm, err := json.Marshal(jmMap)
+	if err != nil {
+		return nil, err
+	}
+
+	return parsedJm, nil
+}
```
faisal-memon, line 80:
```suggestion
	return json.Marshal(removeNulls(jmMap))
}
```

pkg/common/cliprinter/internal/protojson/protojson.go
```diff
@@ -34,13 +34,66 @@ func Print(msgs []proto.Message, stdout, stderr io.Writer) error {
...
+
+func removeNulls(jsonMap map[string]interface{}) {
+	for key, val := range jsonMap {
+		if val == nil {
+			delete(jsonMap, key)
+			continue
+		}
+		switch v := val.(type) {
```
faisal-memon, line 88: Can this be any type other than the two listed?

pkg/common/cliprinter/internal/protojson/protojson.go
guilhermocc: Yes!

pkg/common/cliprinter/internal/protojson/protojson.go
guilhermocc: According to the `json.Unmarshal` documentation, when unmarshalling data into a interface{} value, Unmarshal stores one of these in the interface value:
```go
//
//	bool, for JSON booleans
//	float64, for JSON numbers
//	string, for JSON strings
//	[]interface{}, for JSON arrays
//	map[string]interface{}, for JSON objects
//	nil for JSON null
```
Since this method only looks for removing null values, we only have to check for nil values, or values that can contain nil values, like a `map[string]interface{}` or `[]interface{}`, so this switch clause, with the nil check above it, should be sufficient for this use case.

pkg/common/cliprinter/internal/protojson/protojson.go
guilhermocc: e9029de

pkg/common/cliprinter/internal/protojson/protojson.go
guilhermocc: e9029de

pkg/common/cliprinter/internal/protojson/protojson.go
```diff
@@ -34,13 +34,62 @@ func Print(msgs []proto.Message, stdout, stderr io.Writer) error {
...
+		case map[string]interface{}:
+			removeNulls(v)
+		case []interface{}:
+			for _, elem := range v {
+				if e, ok := elem.(map[string]interface{}); ok {
```
faisal-memon, line 89: Is it possble that the element in the interface array is a type other than the map? ie can it be a bool?

guilhermocc: Yes, but the most important here is that it can have nil values or []interface{} values. I'm gonna update it to include the logic for this.

pkg/common/cliprinter/internal/protojson/protojson.go
faisal-memon: Got it, the other values are ok to leave as is.

pkg/common/cliprinter/internal/protojson/protojson.go
faisal-memon: got it, the other types are ok to leave as is.

guilhermocc: [f05c578](https://github.com/spiffe/spire/pull/3715/commits/f05c578175592723d32170714bfc86822ad5b113)

pkg/common/cliprinter/internal/protojson/protojson.go
```diff
@@ -34,13 +34,62 @@ func Print(msgs []proto.Message, stdout, stderr io.Writer) error {
...
+
+	return json.Marshal(jmMap)
+}
+
+func removeNulls(jsonMap map[string]interface{}) {
+	for key, val := range jsonMap {
+		if val == nil {
+			delete(jsonMap, key)
+			continue
+		}
```
faisal-memon, line 83: I think you can lump this into the switch statement below with a `case nil:`.

pkg/common/cliprinter/internal/protojson/protojson.go
```diff
@@ -34,13 +34,75 @@ func Print(msgs []proto.Message, stdout, stderr io.Writer) error {
...
+	return json.Marshal(jmMap)
+}
+
+func removeNulls(jsonMap map[string]interface{}) {
+	for key, val := range jsonMap {
+		switch v := val.(type) {
+		case nil:
+			delete(jsonMap, key)
+		case map[string]interface{}:
+			removeNulls(v)
```
faisal-memon, line 84: This branch is not being covered in unit tests. Can you update to cover it?

guilhermocc: Sure!

pkg/common/cliprinter/internal/protojson/protojson.go
```diff
@@ -34,13 +34,75 @@ func Print(msgs []proto.Message, stdout, stderr io.Writer) error {
...
+		case nil:
+			continue
+		case map[string]interface{}:
+			removeNulls(v)
+			newSlice = append(newSlice, v)
+		case []interface{}:
+			newSlice = append(newSlice, removeNullsFromSlice(v))
+		default:
+			newSlice = append(newSlice, v)
+		}
```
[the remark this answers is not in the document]
guilhermocc: I wasn't able to cover those scenarios in the Print test using our existing proto messages, so I had to create a specific test for removeNulls functions. [45febb1](https://github.com/spiffe/spire/pull/3715/commits/45febb17934c17507a59733bb62ad6264c622329)

faisal-memon:
Thanks @guilhermocc for adding the additional unit tests. The coverage and change set looks good to me.

amartinezfayo (approved):
Thank you @guilhermocc for the contribution and @faisal-memon for the review.
LGTM!

Issue #3490, opened 2022-10-07
Support Service Hooks on Windows


nweedon-u:
* **Version**: Any
* **Platform**: Windows
* **Subsystem**: Agent

As at stands, the agent cannot be run as a Windows service as it does not integrate [service state transitions](https://learn.microsoft.com/en-us/windows/win32/services/service-status-transitions). As such, we are having to manage the process manually which is a bit more cumbersome than just starting and stopping a Windows service. The agent should support at least `STOPPED`, `STOP_PENDING`, `START_PENDING` and `RUNNING`.

amartinezfayo:
Thank you @nweedon-u for opening this issue. This is something that we considered in the past.
We've discussed this in today's sync and providing native support in SPIRE to run it as a Windows service is something that we want to have.
We think that SPIRE Server should also have the same support, and not only the agent, so they both have consistent support for this.
Something that we discussed is that we want this support to be scoped to these changes, and SPIRE will not be in charge of installing/uninstalling the service. This support will be transparent to the way that SPIRE is deployed in Windows systems.
Thanks again!

PR #3733, opened 2023-01-05
Integration tests for windows service support

guilhermocc:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Running spire as a windows service inside windows containers.

**Description of change**
- Adds integration tests for windows service support.
- Fix windows service support not working when running spire in windows containers.

**Which issue this PR fixes**
Complement for #3490


pkg/common/entrypoint/entrypoint_windows.go
```diff
@@ -20,7 +25,10 @@ type systemCall struct {
 }
 
 func (s *systemCall) IsWindowsService() (bool, error) {
-	return svc.IsWindowsService()
+	// We are using a custom function because the svc.IsWindowsService() one still has an open issue in which it states
+	// that it is not working properly in Windows containers: https://github.com/golang/go/issues/56335. Soon as we have
```
faisal-memon, line 29: Is it worth submitting a PR.? That issue has a help wanted tag and it seems like you have a fix below.

guilhermocc: A PR is already opened for this, in fact, the custom function with the workaround is based on the opened solution proposal: https://github.com/golang/sys/pull/141

test/integration/suites-windows/windows-service/00-setup.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+pwd
```
MarcosDY, line 3: why pwd is required?

test/integration/suites-windows/windows-service/01-start-server-service
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+source ./common
+
+docker-up spire-base
+
+create-service spire-server C:\\spire\\bin\\spire-server.exe
+start-service spire-server run -config C:/spire/conf/server/server.conf
```
MarcosDY, line 7: is it possible to be consistent and use `\\` or `/` in all places?

test/integration/suites-windows/windows-service/05-test-fetch-svid
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+log-debug "test fetch x509 SVID..."
+docker-compose exec -T -u ContainerUser spire-base  \
+    c:/spire/bin/spire-agent api fetch x509 || fail-now "failed to fetch x509"
+
+log-debug "test fetch JWT SVID..."
+docker-compose exec -T -u ContainerUser spire-base \
+    c:/spire/bin/spire-agent api fetch jwt -audience mydb fail-now "failed to fetch x509"
```
MarcosDY, line 9:
```suggestion
docker-compose exec -T -u ContainerUser spire-base \
    c:/spire/bin/spire-agent api fetch jwt -audience mydb || fail-now "failed to fetch JWT"
```

test/integration/suites-windows/windows-service/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM spire-agent-windows:latest-local as spire-agent-windows
+
+FROM spire-server-windows:latest-local as spire-base
+
+COPY --from=spire-agent-windows C:/spire/bin/spire-agent.exe C:/spire/bin/spire-agent.exe
```
MarcosDY, line 5: why you choose to use the same container to start both services?

test/integration/suites-windows/windows-service/00-setup.sh
guilhermocc: Not sure if it is actually required, I took it from another test sample (windows workload attestor). I'm going to check if it is really necessary.

test/integration/suites-windows/windows-service/01-start-server-service
guilhermocc: Sure! nice suggestion

test/integration/suites-windows/windows-service/Dockerfile
guilhermocc: So, in this initial version, I tried to reproduce a test that I ran manually on my local machine (spire server and agent on the same host). Still, I see that a more realistic scenario would be running them in different hosts, right? I can update it :)

MarcosDY: yeah that is 'generally' the case

amartinezfayo (approved):
Thank you @guilhermocc, LGTM!

PR #3759, opened 2023-01-12
Optimize spire server tls connection bundle conversion amount

### LINKED ISSUE #3744: Optimize bundle conversion amount in spire server TLS connections
* **Version**: Any version
* **Platform**: Any Platform
* **Subsystem**: Server

After https://github.com/spiffe/spire/pull/3642 is merged, spire server will perform a  bundle conversion ( *common.bundle to *x509bundle.Bundle) in all tls connections. In the current implementation, we are using a cache for the bundles fetched from the datastore. We could somehow include the parsed version of the bundle in this cache to optimize the conversion amount.

My initial thoughts on this are about not changing the DataStore interface to maintain the scope of the changes to a minimum. I think we can achieve that by:
- Adding the parsed bundle version to the cache entry in DatastoreCache
``` go
type bundleEntry struct {
   mu     sync.Mutex
   ts     time.Time
   bundle *common.Bundle
   x509Bundle   *x509bundle.Bundle
}
```
- Create a new method in DatastoreCache for retrieving this parsed data
``` go
func (ds *DatastoreCache) FetchBundleX509(ctx context.Context, trustDomain string) (*x509bundle.Bundle, error) {
```
- Change the Endpoints struct to use the concrete `DatastoreCache` type instead of the `Datastore` interface; we are already using the DatastoreCache here by default, so this shouldn't break any existing integration test.

I am willing to open a PR on this in case of approval, and I would like to get some opinions here. What do you think?

guilhermocc:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
- Spire server tls connections

**Description of change**
- Add converted x509 bundles to a memory cache for optimizing the amount of conversions during spire server TLS connections.

**Which issue this PR fixes**
fixes #3744


azdagron (approved):
LGTM! Thanks @guilhermocc !

guilhermocc:
Thank you for the review, @azdagron!

PR #3786, opened 2023-01-19
Avoid logging errors that are part of the normal health check operation

### LINKED ISSUE #3784: "No identity issued" error is seen in agent log as part of the health checks execution
The following error is seen in the agent log, as part of the normal execution of the health checks:
```
ERRO[0001] No identity issued                            method=FetchX509Bundles pid=95311 registered=false service=WorkloadAPI subsystem_name=endpoints
```

This should not be logged because is the expected behavior of a healthy agent.

amartinezfayo:
Do not log errors that are part of the normal health check operation

Fixes #3784.


azdagron (approved):
\o/ thanks @amartinezfayo

Issue #2588, opened 2021-10-07
Implement SVID labels in SPIRE


dfeldman:
This is a tracking issue to implement SVID labels in SPIRE. The feature has already been discussed and approved in [SPIFFE #179](https://github.com/spiffe/spiffe/pull/179), but it needs to be implemented in SPIRE.

dfeldman:
Hint is now in the datastore schema. 
I have the needed change in go-spiffe here: https://github.com/spiffe/go-spiffe/pull/179. 
I'm working on step 2, adding it to the API. This might require a feature flag -- not sure.

guilhermocc:
Hey there, I would like to contribute by continuing to resolve this issue, but first, I need to know what is left to be done, could someone help me with that?

evan2645:
Hey @guilhermocc - thank you for this! @dfeldman was working on Step 2 above, but I'm not sure how far he got. I'll go ahead and assign this issue to you

rturner3:
@guilhermocc We will also want to update the [java-spiffe](https://github.com/spiffe/java-spiffe) Workload API client for feature parity in both Go and Java client libraries.

guilhermocc:
@rturner3 ok, I'm going to create a sub-task for it 👍

PR #3787, opened 2023-01-20
Update datastore to persist, update and query registration entries with Hint field

guilhermocc:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Registration entries manipulation in datastore SQL plugin. The following methods are affected:
- `FetchRegistrationEntry`
- `ListRegistrationEntries`
- `CreateRegistrationEntry`
- `UpdateRegistrationEntry`

**Description of change**
Include `Hint` field to `commom.RegistrationEntry` generated struct, enabling spire to persist, update and query registration entries with `Hint` field.

**Which issue this PR fixes**
Ongoing work for #2588


guilhermocc:
I'm aware of broken integration tests, already working on fixing them!

pkg/server/datastore/sqlstore/sqlstore_test.go
```diff
@@ -2145,21 +2159,24 @@ func (s *PluginSuite) testListRegistrationEntries(dataConsistency datastore.Data
...
 				if withPagination {
 					assert.Equal(t, tt.expectPagedTokensIn, tokensIn, "unexpected request tokens")
 				} else {
 					assert.Empty(t, tokensIn, "unexpected request tokens")
 				}
-				assert.Equal(t, expectIDsOut, actualIDsOut, "unexpected response entries")
+
+				assert.Len(t, actualEntriesOut, len(expectedEntriesOut), "unexpected number of entries returned")
+				for id, expectedEntry := range expectedEntriesOut {
+					// Some databases are not returning federated IDs in the same order (e.g. mysql)
```
guilhermocc, line 2176: I've identified this behavior after adding a full message assertion. Should we consider it a bug?

pkg/server/datastore/sqlstore/models.go
```diff
@@ -96,7 +96,7 @@ type RegisteredEntry struct {
 
 	// Hint is a "hint string" passed to the workload to distinguish between
 	// multiple SVIDs
-	Hint string
+	Hint string `gorm:"index"`
```
azdagron, line 99: Yikes. We'll need to add a migration for this. Since it won't impact correctness, we're probably still ok shipping this feature for 1.6.0, but it doesn't mean searching by hint will be slow until they apply migrations.

pkg/server/datastore/sqlstore/sqlstore_test.go
azdagron: Hmm, certainly consistency would be good here, though shouldn't matter for correctness. It might through off some home-grown "compare these two entries for quality" code. We can probably just file an issue.

pkg/server/datastore/datastore.go
```diff
@@ -168,6 +168,7 @@ type ListRegistrationEntriesRequest struct {
 	BySpiffeID      string
 	Pagination      *Pagination
 	ByFederatesWith *ByFederatesWith
+	ByHint          string
```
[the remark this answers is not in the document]
azdagron: I don't think MatchBehavior  really applies to this since it is a single field and not a set. We don't do set matching on characters inside of strings for other fields.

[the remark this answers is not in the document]
guilhermocc: You're right. In fact, I was thinking of something like `match any` here, in which a user could provide a list o hints, and spire would retrieve a list of entries that contains one of those fields. Do you think it could be useful?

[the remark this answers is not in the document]
azdagron: It might be useful, but we don't really have that for other fields and have yet to have someone make that request. I think my preference would be to hold off on adding more features until the use cases and benefits become clear. Implementing this would complicate the already gross entry query code :)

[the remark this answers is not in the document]
guilhermocc: It makes sense, so I will maintain the KISS principle here 😄

pkg/server/datastore/sqlstore/sqlstore_test.go
guilhermocc: I can open an issue for this 😄

proto/spire/common/common.proto
```diff
@@ -106,6 +109,7 @@ message RegistrationEntryMask {
     bool dns_names = 10;
     bool store_svid = 11;
     bool jwt_svid_ttl = 12;
+    bool hint = 14;
```
MarcosDY, line 112: why 14? may we use 13?

guilhermocc: Thanks for pointing this up, @MarcosDY 😄

azdagron (approved):
\o/

PR #3811, opened 2023-01-31
Support dedicated non-root user in release images

rturner3:
Natively support a way for users of the release images to run the images as a non-root user.

SPIRE depends on certain directories to be readable from / writeable to / creatable. With scratch-based images, the root of the filesystem is empty by default and only the root user has permission to create directories at the root level. This means that any directories SPIRE depends on should be pre-created at build time for some dedicated non-root user.

Create all such directories in the SPIRE Server and Agent image builds. The directories chosen were ones defined in the
[Linux Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html).

Additionally, provide sample configurations that use the paths pre-created in the images, and add documentation on how to run the images as a non-root user.


doc/docker_images.md
```diff
@@ -0,0 +1,75 @@
...
+All of these directories referenced in those configuration files are provided in the release images with correct permissions for a user with uid `1000` and gid `1000`.
+To run the SPIRE containers based off these configuration files, run the container as uid `1000` and gid `1000`.
+Note that you will also need to ensure the SPIRE Server configuration file mounted into the container has the correct permissions for uid `1000`.
+Example `docker run` command for SPIRE Server with non-root user configuration:
+
+```bash
+$ docker run \
+    --user 1000:1000 \
+    -p 8081:8081 \
+    -v /path/to/server/config:/etc/spire/server
```
amartinezfayo, line 32:
```suggestion
    -v /path/to/server/config:/etc/spire/server \
```

amartinezfayo (commented):
Thank you @rturner3 for this!
I have some suggestions.

amartinezfayo (approved):
🎉

PR #3828, opened 2023-02-06
Add new migration to create hint field index

guilhermocc:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
New migration version

**Description of change**
Add a new migration version that creates an index for `hint` column. With the new[ SVID label feature](https://github.com/spiffe/spire/issues/2588), we will need to enable querying registration entries by hint, and using an indexed field for making this query will optimize the query cost.

**Which issue this PR fixes**
Ongoing work for https://github.com/spiffe/spire/issues/2588


pkg/server/datastore/sqlstore/migration.go
```diff
@@ -403,6 +406,13 @@ func migrateToV20(tx *gorm.DB) error {
 	return nil
 }
 
+func migrateToV21(tx *gorm.DB) error {
+	if err := tx.AutoMigrate(&RegisteredEntry{}, &Selector{}).Error; err != nil {
```
MarcosDY, line 410: there are no changes on Selector struct
```suggestion
	if err := tx.AutoMigrate(&RegisteredEntry{}).Error; err != nil {
```

guilhermocc:
Thank you for the review, @MarcosDY

PR #3831, opened 2023-02-07
Update linter versions

rturner3:
The latest image version of markdownlint-cli supports arm64 architectures, which is useful for running the linter locally on ARM-based Macs.

Bump golangci-lint version to latest while we're at it.


azdagron (approved):
\o/

PR #3873, opened 2023-02-16
Clean node resolver map entries on attested node deletion

### LINKED ISSUE #3872: Attested node deletion doesn't clean node resolver data
* **Version**: `main` branch
* **Platform**: all platforms
* **Subsystem**: server - datastore

Currently, when the datastore's `DeleteAttestedNode` method is evaluated, It cleans any entry data related to the provided entryID in the `attested_node_entries` table, but related node resolver data on `node_resolver_map_entries` remains in the database. Apparently, we have no usage for this data once the attested node entry is deleted, so this data should also be deleted in the process. I wasn't able to find any logical reason in the code to keep this data in the database after node entry deletion, but if there is, I would be happy to know. 

PS: If we definitely don't have any use for the data that's left behind, I would be happy to act on this fix since it relates to another issue that I'm working on right now (#1836). 😄

guilhermocc:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
- Agent eviction action

**Description of change**
Update `DeleteAttestedNode` method to also delete node selectors

**Which issue this PR fixes**
fixes #3872


azdagron (approved):
Thanks, @guilhermocc! I wonder if we should open a separate issue to provide some database cleanup code that deletes selector rows that have a spiffe ID not represented in the attested nodes table, maybe when the database is first opened, that we could keep in for a minor release cycle to make sure stale entries are removed eventually.

guilhermocc:
> Thanks, @guilhermocc! I wonder if we should open a separate issue to provide some database cleanup code that deletes selector rows that have a spiffe ID not represented in the attested nodes table, maybe when the database is first opened, that we could keep in for a minor release cycle to make sure stale entries are removed eventually.

It makes sense since this PR will only prevent the existence of new stale node resolver map entries, but will not clean existing ones. I can create it :)

guilhermocc:
@azdagron update about the created issue: https://github.com/spiffe/spire/issues/3919

PR #3878, opened 2023-02-17
Skip MySQL 5.7 integration tests on non-amd64 architectures

### LINKED ISSUE #3874: datastore-mysql and datastore-mysql-replication tests fail with MySQL 5.7 on arm64
```
Pulling mysql-5-7 (mysql:5.7)... 5.7: Pulling from library/mysql ERROR: no matching manifest for linux/arm64/v8 in the manifest list entries
```

Oracle doesn't publish MySQL 5.7 arm64 images. They only publish arm64 images for MySQL 8.0 today: https://hub.docker.com/r/arm64v8/mysql/

Considering Oracle doesn't release [arm64 binaries for 5.7,](https://dev.mysql.com/downloads/mysql/5.7.html) I think it's unlikely they'll publish arm64 Docker images for 5.7 given its community end of life date is October 2023.

I don't think we want to maintain our own MySQL 5.7 image just for this integration test. We could detect in the test when we are running on arm64 and skip testing 5.7 with a warning message. I think it's worth considering to just drop testing against 5.7 at the end of this year given it's close to EOL.

rturner3:
Oracle only publishes binaries + images for MySQL 5.7 for amd64 architecture. Ignore running on other architectures, since the amd64 image is not compatible with other architectures.

Fixes #3874.


rturner3:
I verified this works locally on my ARM-based Mac. Also verified that in the CI logs for this PR that the tests do get run for x86_64.

guilhermocc:
Both tests are passing in linux arm64, and the unsupported tests are correctly being skipped :+1: 

image
image

test/integration/suites/datastore-mysql-replication/01-test-variants
```diff
@@ -118,5 +118,11 @@ test-mysql-replication() {
     docker-stop "${readwrite_service_name}" "${readonly_service_name}"
 }
 
-test-mysql-replication mysql-5-7 || exit 1
+arch=$(uname -m)
```
azdagron, line 121: If you'd like to dedup this logic, the `test/integration/suites/common` file is sourced into each script before it is executed. You could add a function there.

azdagron (approved):
LGTM, with one comment on how you can consolidate the "is this amd64" logic.

test/integration/suites/datastore-mysql-replication/01-test-variants
rturner3: Makes sense. This would also affect the approach in https://github.com/spiffe/spire/pull/3884. I can do this as a follow-up PR so we can try to get all the integration test fixes in before 1.6.0.

azdagron: sgtm!

PR #3880, opened 2023-02-17
Update agents listing service with "CanReAttest" filter

guilhermocc:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
spire-server agents listing.

**Description of change**
- Add "CanReAttest" boolean filter in list agents method
- Update spire-server `agent` and `entry` commands to add new response fields: registration entry `hint` and agent `canReAttest`.

**Which issue this PR fixes**
Ongoing work for #1836

**Dependencies**
⚠️ This PR depends on https://github.com/spiffe/spire-api-sdk/pull/36. After it's merge, go mod file should be updated to remove temporary local spire-api-sdk reference. ⚠️


faisal-memon:
I can review

cmd/spire-server/cli/agent/agent_test.go
```diff
@@ -266,7 +266,7 @@ func TestList(t *testing.T) {
 			expectedReturnCode:   0,
```
MarcosDY, line 266: can you add a test case where an agent that is re_attestable is returned?

cmd/spire-server/cli/entry/create_test.go
```diff
@@ -289,6 +289,7 @@ StoreSvid        : true
       },
       "entry": {
         "id": "entry-id",
+        "hint": "",
```
MarcosDY, line 292: why hint is required here? looks like it is not related with this PR (the same comment is for all files with the same change)

guilhermocc: The new version of spire-api-sdk comes with the hint field change; that's why I had to include those here. This field will be effectively included in https://github.com/spiffe/spire/pull/3926

cmd/spire-server/cli/agent/agent_test.go
guilhermocc: 0db980d

MarcosDY (approved):

PR #3882, opened 2023-02-17
Fix Envoy integration tests on arm64 architecture

### LINKED ISSUE #3875: envoy-sds-v2, envoy-sds-v3, and envoy-sds-v3-spiffe-auth tests fail on arm64
```
---> [Warning] The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested  ---> Running in 15761d6f4957 standard_init_linux.go:219: exec user process caused: exec format error
```

We can fix these tests by migrating from the `envoyproxy/envoy-alpine` image to the `envoyproxy/envoy` image. The latter image is published for arm64 for the versions we depend on (1.16+): https://hub.docker.com/r/envoyproxy/envoy/tags

rturner3:
Our Envoy tests were using Alpine-based release images that are only published for amd64 architecture. However, Envoy also publishes `envoyproxy/envoy` Ubuntu-based release images for both amd64 and arm64 architectures. These images are around 45 MB in size.

Switch the tests to the multiarch Envoy images so we can run them on arm64 machines.

Fixes #3875

### DIFF
--- a/test/integration/common
+++ b/test/integration/common
@@ -133,10 +133,11 @@ build-mashup-image() {
 
     cat > Dockerfile <<EOF
 FROM spire-agent:latest-local as spire-agent
-FROM envoyproxy/envoy-alpine:${ENVOY_IMAGE_TAG} AS envoy-agent-mashup
+FROM envoyproxy/envoy:${ENVOY_IMAGE_TAG} AS envoy-agent-mashup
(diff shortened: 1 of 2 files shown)


amartinezfayo (approved):
Thanks @rturner3, LGTM!

guilhermocc:
All envoy-related integration tests are now passing in arm64 Linux machines 👍

PR #3883, opened 2023-02-17
Fix upstream-authority-cert-manager integration test on arm64

### LINKED ISSUE #3876: upstream-authority-cert-manager test fails on arm64
```
[2023-02-03T13:46:06Z] downloading kind from https://github.com/kubernetes-sigs/kind/releases/download/v0.11.1/kind-linux-amd64... [2023-02-03T13:46:09Z] downloading kubectl from https://storage.googleapis.com/kubernetes-release/release/v1.21.1/bin/linux/amd64/kubectl... [2023-02-03T13:46:13Z] starting cluster... [2023-02-03T13:46:13Z] unable to create cluster
```

kind and kubectl publish arm64 release binaries. We just need to add some architecture detection logic in [test/integration/common](https://github.com/spiffe/spire/blob/65c8e0a239d92a7e5c550c0c1271ddeba4c1e907/test/integration/common#L199-L227) by parsing uname -m, mapping `{x86_64, amd64}` -> `amd64` and `{aarch64, arm64}` -> `arm64`, and including `amd64`/`arm64` in the correct part of the path of the release binary URLs.

We probably also need to do the same for the [kindest/node image](https://github.com/spiffe/spire/blob/65c8e0a239d92a7e5c550c0c1271ddeba4c1e907/test/integration/common#L229-L238), which is published for arm64 today: https://hub.docker.com/r/kindest/node/tags

rturner3:
The test was always trying to use amd64 kind and kubectl binaries. Added architecture detection logic to pull the arm64 binaries when running on arm64.

Fixes #3876


rturner3:
The test passes on my local ARM-based Mac with this change.

guilhermocc:
`upstream-authority-cert-manager` test is now passing on arm64 linux machines 👍

test/integration/common
```diff
@@ -227,7 +241,7 @@ download-kubectl() {
 }
 
 start-kind-cluster() {
-    K8SIMAGE=kindest/node:v1.21.1@sha256:fae9a58f17f18f06aeac9772ca8b5ac680ebbed985e266f711d936e91d113bad
+    K8SIMAGE=kindest/node:v1.21.1
```
rturner3, line 244: I removed the sha256 pin here because the sha is specific to an architecture, i.e. amd64 and arm64 will have different shas.

amartinezfayo (approved):
🎉

PR #3884, opened 2023-02-17
Skip upgrade test on arm64 until 1.7.0

### LINKED ISSUE #3877: upgrade test fails on arm64
```
spire-server-1.5.0           | standard_init_linux.go:219: exec user process caused: exec format error
```

This is going to fail until the 1.7 release series. We should detect if the test is running on arm64, and if so, skip it. If we ignore this test, we should also add a TODO comment to remind us to re-enable the test in 1.7.0.

I think with these changes, we should be able to run make integration on an arm64-based system without seeing any errors.

rturner3:
We are publishing arm64 images starting with the 1.6.0 release. Since the upgrade test uses images from the previous minor release series, this means the upgrade test cannot be run on arm64 until 1.7.0, at which point the prior minor release series will all have arm64 images.

Fixes #3877

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -303,6 +303,9 @@ endif
 integration:
 ifeq ($(os1), windows)
 	$(error Integration tests are not supported on windows)
+else ifeq (,$(filter $(arch2),arm64,aarch64))
+	# TODO: Remove this special handling of arm64 in 1.7.0
+	$(E)(export IGNORE_SUITES=suites/upgrade && ./test/integration/test.sh $(SUITES))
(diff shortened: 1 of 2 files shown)


rturner3:
The test passes with the expected warning logs with these changes on my ARM-based Mac.

amartinezfayo (approved):
LGTM!

guilhermocc:
`upgrade` test is being correctly skipped with warning logs on arm64 Linux machines 👍 .

PR #3927, opened 2023-02-28
Fix trust domain json logging in federated bundle updater

StupidScience:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

Currently if json log format is in use `trust_domain` isn't logged properly for bundle manager, instead we're getting smth like:
```json
{"at":"2023-02-28T14:27:53Z","level":"debug","msg":"Scheduling next bundle refresh","subsystem_name":"bundle_client","time":"2023-02-28T14:26:38Z","trust_domain":{}}
```
As you can see `"trust_domain":{}` because underlying variable value is object `spiffeid.TrustDomain` with private properties only.

**Description of change**

Use `String()` method of `spiffeid.TrustDomain` to set log field value.


evan2645 (approved):
Thank you!

PR #3934, opened 2023-03-01
Honor the `ca_ttl` server config

amartinezfayo:
The `ca_ttl` server config was not being set in the `credtemplate.Config` struct, hence it was being ignored.
Add the setting of `X509CATTL` with the configured value.


azdagron (approved):
Thanks for hunting this down, @amartinezfayo !

PR #3941, opened 2023-03-02
Clean stale node resolver map entries in database initialization

### LINKED ISSUE #3919: Clean stale node resolver map entries in database startup
## Context 
After the identified [issue](https://github.com/spiffe/spire/issues/3872) regarding the stale node resolver map entries data that was left on the database after a node deletion, we already have a solution for ensuring this data is deleted on node deletion (#3873), but nodes that were deleted prior to this change, will still have stale node resolver map entries persisted. 

## Proposal
This issue proposes adding a logic for cleaning up existing stale node resolver map entries (those that reference an unexistent node spiffeID) on the database startup. With this, nodes that were deleted prior to (#3873) will also have their stale resolver data deleted from the database.

guilhermocc:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
No functionality affected

**Description of change**
Add logic for deleting existing stale node resolver entries in database initialization.

**Which issue this PR fixes**
Fixes #3919


pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -739,11 +739,49 @@ func (ds *Plugin) openDB(cfg *configuration, isReadOnly bool) (*gorm.DB, string,
...
+		// Delete this function in a future release.
+		if err := cleanStaleNodeResolverEntries(db, ds.log); err != nil {
+			ds.log.WithError(err).Error("Failed to clean stale node resolver entries")
+		}
 	}
 
 	return db, version, supportsCTE, dialect, nil
 }
 
+func cleanStaleNodeResolverEntries(tx *gorm.DB, log logrus.FieldLogger) error {
```
azdagron, line 752: Instead of hydrating all of the ids for rows that need to be deleted, what do you think about doing something simple like:

```
DELETE FROM node_resolver_map_entries WHERE spiffe_id NOT IN (SELECT spiffe_id FROM attested_node_entries)
```

guilhermocc: Initially, I thought about using subquery like that, but from what I know, subquery tends to be less performant than joining tables. This query will hold spire initialization only for the first time, so I preferred to use a more performant query here

azdagron: I appreciate that you are being performance minded :) I wrote some code to build up a dataset and did an "EXPLAIN ANALYZE" and indeed your version is faster by a large margin, at least with Postgres.

Carry on!

guilhermocc: I've evaluated an Explain Analyse in both queries using some sample data, resulting in 667 stale entries being deleted, with 652 attested nodes and 1000 node resolver entries.

Using Join:
image

Using Subquery
image

Apparently my optimization attempt was unsuccessful, I'm going to update it :stuck_out_tongue:

azdagron: @guilhermocc and I got different results. We did some more analysis and the subquery version is indeed faster and less costly and has the benefits of not having to stash the row ids in memory on the server side before deleting them.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -739,11 +739,31 @@ func (ds *Plugin) openDB(cfg *configuration, isReadOnly bool) (*gorm.DB, string,
 			db.Close()
 			return nil, "", false, nil, err
 		}
+		// TODO: we should keep this logic for a minor release cycle to make sure stale entries are removed eventually.
+		// Delete this function in a future release.
```
azdagron, line 743: We should put the exact release we want to remove this in. Makes it easy to grep for. We should also open an issue for this.
```suggestion
		// TODO: we should keep this logic for a minor release cycle to make sure stale entries are removed eventually.
		// Remove in SPIRE 1.8.0
```

azdagron: I'll open the issue.

azdagron (approved):
\o/

PR #3955, opened 2023-03-09
Azure KMS (Key Vault) Server KeyManager plugin

moe-omar:
This PR adds a server KMS plugin that uses Azure's Key Vault.  

Pull Request check list

- [x]  Commit conforms to CONTRIBUTING.md?
- [x]  Proper tests/regressions included?
- [x]  Documentation updated?

Affected functionality
No existing functionality affected. New server plugin added

Description of change
A new server keymanager plugin that uses Azure's Key Vault.  

**What/How was this new change tested?**

- Tested using all the supported key types (`rsa-2048`, `rsa-4096`, `ec-p256`, and `ec-p384`) for both JWT keys and CA keys.
- Tested key rotation, refresh and keys cleanup for stale/unused keys belonging to the trust domain the server belongs to. 
- Tested authentication using Azure client/application credentials as well as using MSI in an Azure VM.
- This involved running the server, registering agents and workloads and making sure SVID's are issued (signed by Key Vault keys) and can be validated using the bundle.
- Added unit tests.


pkg/server/plugin/keymanager/azurekms/azurekms_test.go
```diff
@@ -0,0 +1,1048 @@
...
+			_ = waitForSignal(t, refreshKeysSignal)
+			// move the clock forward so the task is run
+			ts.clockHook.Add(6 * time.Hour)
+			// wait for refresh keys to be run
+			err = waitForSignal(t, refreshKeysSignal)
+
+			// assert
+			if tt.updateKeyErr != "" {
+				require.NotNil(t, err)
+				require.Equal(t, tt.err, err.Error())
```
MarcosDY, line 805: you can use require.EqualError() that validate error is not nil

pkg/server/plugin/keymanager/azurekms/azurekms_test.go
```diff
@@ -0,0 +1,1048 @@
...
+
+			ts.plugin.hooks.disposeKeysSignal = disposeKeysSignal
+			ts.plugin.hooks.scheduleDeleteSignal = scheduleDeleteSignal
+
+			// exercise
+			_, err := ts.plugin.Configure(ctx, tt.configureRequest)
+			require.NoError(t, err)
+
+			// Wait for dispose disposeCryptoKeysTask to be initialized.
+			_ = waitForSignal(t, disposeKeysSignal)
```
MarcosDY, line 897: how about validating errors from all this waitForSignal calls?

pkg/server/plugin/keymanager/azurekms/azurekms_test.go
```diff
@@ -0,0 +1,1048 @@
...
+			_ = waitForSignal(t, scheduleDeleteSignal)
+			// Wait for the 3rd key to be deleted
+			_ = waitForSignal(t, scheduleDeleteSignal)
+			// Wait for the 4th key to be deleted
+			_ = waitForSignal(t, scheduleDeleteSignal)
+			// Wait for the 5th key to be deleted
+			_ = waitForSignal(t, scheduleDeleteSignal)
+
+			// assert errors
+			if tt.err != "" {
```
MarcosDY, line 917: err is never set on test cases so this is never called

pkg/server/plugin/keymanager/azurekms/azurekms_test.go
```diff
@@ -0,0 +1,1048 @@
...
+	}
+}
+
+func serializedConfiguration(keyMetadataFile string, keyVaultURI string, tenantID string, subscriptionID string, appID string, appSecret string) string {
+	return fmt.Sprintf(`{
+        "key_metadata_file":"%s",
+        "key_vault_uri":"%s",
+		"tenant_id":"%s",
+		"subscription_id":"%s",
+        "app_id":"%s",
```
MarcosDY, line 947: can you resolve indentation issues in all configurations?

pkg/server/plugin/keymanager/azurekms/client_fake.go
```diff
@@ -0,0 +1,435 @@
...
+		return azkeys.CreateKeyResponse{}, k.createKeyErr
+	}
+
+	var publicKey *azkeys.JSONWebKey
+	var privateKey crypto.Signer
+	keyOperations := getKeyOperations()
+	kmsKeyID := path.Join(k.vaultURI, keyName)
+	switch {
+	case *parameters.Kty == azkeys.JSONWebKeyTypeEC && *parameters.Curve == azkeys.JSONWebKeyCurveNameP256:
+		privateKey = k.testKeys.NewEC256(k.t)
```
MarcosDY, line 168: with this approach, every time we call `testKeys.NewXXXXX` we are asking for a new key to fixture (that is why you had to add more entries to fixture) so every time we want to add a new test case we'll need to add new keys.

It looks like an overkill we dont want to test key generation but that a key is returned, so what do you think about loading keys when fake is created and just use the same keys (depending on key types of course)

MarcosDY (commented):
Looks great!!!

pkg/server/plugin/keymanager/azurekms/azurekms_test.go
```diff
@@ -0,0 +1,1048 @@
...
+		err              string
+		code             codes.Code
+		configureRequest *configv1.ConfigureRequest
+		fakeEntries      []fakeKeyEntry
+		listKeysErr      string
+		getKeyErr        string
+		getPublicKeyErr  string
+	}{
+		{
+			name:             "pass with keys",
```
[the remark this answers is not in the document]
moe-omar: 👍🏾 done for all three

pkg/server/plugin/keymanager/azurekms/azurekms_test.go
moe-omar: These are strings we're comparing here though not errors

pkg/server/plugin/keymanager/azurekms/client_fake.go
moe-omar: Done 👍🏾

amartinezfayo (commented):
Thank you very much @moe-omar for this contribution! We really appreciate the effort put in this.

I have a couple of concerns:
1) With the current logic, it seems that key versions will be kept in the Key Vault forever, unless the server starts with a different server ID (e.g. the metadata file is lost).
This may not be super critical from the costing perspective, since the pricing model of the service is based on the number of transactions performed rather than the number of objects maintained. Still, it's concerning to me to keep keys that have been rotated and shouldn't be usable anymore.

2) It doesn't seem that is possible to customize the access control of the keys created by SPIRE. This has been proved to be problematic in the past with other providers (see #2424 for details). Could we support influencing the access control of keys managed by the plugin?

moe-omar:
> Thank you very much @moe-omar for this contribution! We really appreciate the effort put in this.
> 
> I have a couple of concerns:
> 
> 1. With the current logic, it seems that key versions will be kept in the Key Vault forever, unless the server starts with a different server ID (e.g. the metadata file is lost).
>    This may not be super critical from the costing perspective, since the pricing model of the service is based on the number of transactions performed rather than the number of objects maintained. Still, it's concerning to me to keep keys that have been rotated and shouldn't be usable anymore.
> 2. It doesn't seem that is possible to customize the access control of the keys created by SPIRE. This has been proved to be problematic in the past with other providers (see [Make the KMS Key manager to create keys using a role-based policy when possible #2424](https://github.com/spiffe/spire/pull/2424) for details). Could we support influencing the access control of keys managed by the plugin?

Thanks @amartinezfayo !

For 1, Key Vault doesn't support deleting specific key versions unfortunately, but I think it's possible to "disable" a key version (it can be reenabled though). You think that would be good enough?

Regarding 2, I looked at the [original issue](https://github.com/spiffe/spire/issues/2352) that motivated adding policies to AWS KMS keys, I think it's a little different with Key Vault. Policies are set at the Key Vault level not individual keys like in AWS KMS and I also don't think that you get a default policy when you create the Key Vault itself (I had to create a policy to give SPIRE access to keys in the Key Vault). So the way I think this would work is that a Key Vault + access policies to give SPIRE the appropriate access are provisioned separately (via Terraform for example) and the plugin is then configured to use the Key Vault.

 Let me know if you think I'm missing something.

doc/plugin_server_keymanager_azure_kms.md
```diff
@@ -0,0 +1,87 @@
...
+### Use of key versions
+
+In Key Vault, the cryptographic key material that is used to sign data is stored
+in a key version. A key can have zero or more key versions.
+
+For each SPIRE Key ID that the server manages, this plugin maintains a Key.
+When a key is rotated, a new version is added to the Key.
+
+Note that Azure does not support deleting individual key versions, instead, the key itself is deleted by the plugin
+when it's no longer being used by a server in the trust domain the server belongs to.
```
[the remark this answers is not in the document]
moe-omar: Yes that's correct, when a key is rotated, the older versions are still active but they won't be used by the SPIRE server (the plugin will only use the current version). So even though the older versions are active, they are not usable in the SPIRE context.

amartinezfayo:
>Key Vault doesn't support deleting specific key versions unfortunately, but I think it's possible to "disable" a key version (it can be reenabled though). You think that would be good enough?

I think that we should try to avoid having the plugin creating key versions indefinitely. Key Vault does not restrict the number of versions on a key, but storing a large number of versions may have impact in the performance of operations, and may be confusing to still have them showing in the console.
I would rather prefer that we delete the rotated keys. They will still be recoverable for a retention period but will not be showed by default in the console (you will only see them if you choose to manage the deleted keys).
We would create a new key when we need to generate a new one, and once it's created, we can delete the old one. What do you think?

>Policies are set at the Key Vault level not individual keys like in AWS KMS and I also don't think that you get a default policy when you create the Key Vault itself (I had to create a policy to give SPIRE access to keys in the Key Vault). So the way I think this would work is that a Key Vault + access policies to give SPIRE the appropriate access are provisioned separately (via Terraform for example) and the plugin is then configured to use the Key Vault.

While I believe that the access control in Key Vault can also be defined at the key level (that's the resource scope in the case of Key Vault - https://learn.microsoft.com/en-us/azure/role-based-access-control/role-assignments-portal#step-1-identify-the-needed-scope) I agree that since there is no default policy applied, the situation is different than the other KMS plugins so it seems that we should be good here.

amartinezfayo:
Closing in favor of #4455. Thank you @moe-omar for this contribution!

Issue #3968, opened 2023-03-13
Upstream Authority "vault" Plugin doesn't make X509 Certificate with SAN


inajob:
* **Version**: v1.6.1
* **Platform**: `Linux HOSTANAME 3.10.0-1160.80.1.el7.x86_64 #1 SMP Tue Nov 8 15:48:59 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`
* **Subsystem**: server

We use SPIRE with Upstream Authority "vault" Plugin.
When we try to upgrade Spire 1.5.5 -> 1.6.1, SPIRE Server can'r start with below error.
> level=error msg="Fatal run error" error="rpc error: code = InvalidArgument desc = X509 CA minted by upstream authority is invalid: invalid upstream-signed X509 CA: invalid X509 CA: missing URI SAN"

It seems to be no URI SAN field in upsrtream minted X509 Certificate.

I found the [interface of Vault](https://developer.hashicorp.com/vault/api-docs/secret/pki#uri_sans-2), it requires `uri_sans` field if I want to get X509 Certificate with SAN.
but plugin's code doesn't specify this filed ( https://github.com/spiffe/spire/blob/v1.6.1/pkg/server/plugin/upstreamauthority/vault/vault_client.go#L353 )

How to fix this issue?

I don't know which is the best solution.

1. fix vault plugin to speficy this field.
2. fix validation logic to ignore SAN filed.

Or are there other solution about this probrem?

hiyosi:
If a value of URI SAN(SPIFFE ID) is required in certificates for X509SVID CA, I will open a PR ( #3971 ).

amartinezfayo:
Fixed by https://github.com/spiffe/spire/issues/3968.

PR #3971, opened 2023-03-13
Make sure to set uri_sans parameter

hiyosi:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

Fixed a issue in which intermediate certificates issued by the HashiCorp Vault PKI engine do not contain a URI SAN. 

**Which issue this PR fixes**

https://github.com/spiffe/spire/issues/3968


pkg/server/plugin/upstreamauthority/vault/vault_test.go
```diff
@@ -259,7 +259,7 @@ func TestMintX509CA(t *testing.T) {
 				},
 			},
 			authMethod:              TOKEN,
-			expectX509CA:            []string{"spiffe://intermediate-vault", "spiffe://intermediate"},
+			expectX509CA:            []string{"spiffe://intermediate-spire", "spiffe://intermediate-vault"},
```
hiyosi, line 262: Before this PR, values of URI SAN are little confusing.

amartinezfayo (commented):
Thank you @hiyosi for this contribution!
This is in draft mode, are you expecting to add more commits?

amartinezfayo:
>I'm not sure a value of URI SAN(SPIFFE ID) must be set or not in X509SVID CA certificates.

While it would be desirable that a signing certificate itself should an SVID, it's not mandatory.
A more strict validation was introduced in SPIRE 1.6 that made this a hard requirement. After discussing this during our last maintainer's sync, we decided that it's better to relax that requirement so the process doesn't fail. There are chances that an upstream authority plugin can't control this, so we will change this to not have an unnecessary limitation there.

We still welcome this change though :)

pkg/server/plugin/upstreamauthority/vault/vault_client.go
```diff
@@ -350,10 +350,19 @@ func (c *Client) LookupSelf(token string) (*vapi.Secret, error) {
 func (c *Client) SignIntermediate(ttl string, csr *x509.CertificateRequest) (*SignCSRResponse, error) {
 	csrPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE REQUEST", Bytes: csr.Raw})
 
+	var uris []string
+	for _, uri := range csr.URIs {
+		uris = append(uris, uri.String())
+	}
+	if len(uris) == 0 {
+		return nil, status.Errorf(codes.Internal, "CSR must have least one URIs")
```
amartinezfayo, line 358: Since the error is due to an invalid input argument and not an internal failure I think that we should use codes.InvalidArgument.

```suggestion
		return nil, status.Errorf(codes.InvalidArgument, "CSR must have at least one URI")
```

hiyosi: fixed via [f9b6103](https://github.com/spiffe/spire/pull/3971/commits/f9b610336ff31073fc7ba84b1b68d120fc8ff1ed)

pkg/server/plugin/upstreamauthority/vault/vault_client.go
```diff
@@ -350,10 +350,19 @@ func (c *Client) LookupSelf(token string) (*vapi.Secret, error) {
 func (c *Client) SignIntermediate(ttl string, csr *x509.CertificateRequest) (*SignCSRResponse, error) {
 	csrPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE REQUEST", Bytes: csr.Raw})
 
+	var uris []string
+	for _, uri := range csr.URIs {
+		uris = append(uris, uri.String())
+	}
+	if len(uris) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "CSR must have least one URIs")
```
amartinezfayo, line 358:
```suggestion
		return nil, status.Errorf(codes.InvalidArgument, "CSR must have at least one URI")
```

amartinezfayo (commented):
Last final thing and we are ready to go. Sorry that I didn't suggest this in the first review.

pkg/server/plugin/upstreamauthority/vault/vault_client.go
hiyosi: Fixed via [f34b40b](https://github.com/spiffe/spire/pull/3971/commits/f34b40b511ba53b0b1211355adb1cdbae40f6706)

amartinezfayo (approved):
Thank you @hiyosi!

PR #3979, opened 2023-03-14
Force Rotation: Implement RevokeX509CA in datastore layer

### LINKED ISSUE #3890: Implement RevokeX509CA in datastore
Implement X.509 CA revocation in datastore

Depends on #3887

MarcosDY:
**Which issue this PR fixes**
fixes #3890

### DIFF
--- a/pkg/server/datastore/sqlstore/sqlstore.go
+++ b/pkg/server/datastore/sqlstore/sqlstore.go
@@ -1136,14 +1136,9 @@ func taintX509CA(tx *gorm.DB, trustDomainID string, taintedPublicKey crypto.Publ
 }
 
 func revokeX509CA(tx *gorm.DB, trustDomainID string, publicKey crypto.PublicKey) error {
-	model := &Bundle{}
-	if err := tx.Find(model, "trust_domain = ?", trustDomainID).Error; err != nil {
-		return sqlError.Wrap(err)
-	}
-
-	bundle, err := modelToBundle(model)
+	_, bundle, err := getBundle(tx, trustDomainID)
 	if err != nil {
-		return status.Errorf(codes.Internal, "failed to unmarshal bundle: %v", err)
+		return err
 	}
 
 	found := false
--- a/pkg/server/datastore/sqlstore/sqlstore.go
+++ b/pkg/server/datastore/sqlstore/sqlstore.go
@@ -1161,7 +1161,7 @@ func revokeX509CA(tx *gorm.DB, trustDomainID string, publicKey crypto.PublicKey)
 
 		if ok {
 			if !ca.TaintedKey {
-				return status.Error(codes.Internal, "it is not possible to revoke an untainted root CA")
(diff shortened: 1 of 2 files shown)


pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1052,6 +1059,47 @@ func pruneBundle(tx *gorm.DB, trustDomainID string, expiry time.Time, log logrus
...
 
+func revokeX509CA(tx *gorm.DB, trustDomainID string, publicKey crypto.PublicKey) error {
+	model := &Bundle{}
+	if err := tx.Find(model, "trust_domain = ?", trustDomainID).Error; err != nil {
+		return sqlError.Wrap(err)
+	}
+
+	bundle, err := modelToBundle(model)
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed to unmarshal bundle: %v", err)
```
guilhermocc, line 1070: @MarcosDY Why we are returning GRPC errors here?

guilhermocc (approved):

pkg/server/datastore/sqlstore/sqlstore.go
MarcosDY: I found some related code was returning GRPCs 
```
	if err := tx.Find(model, "trust_domain = ?", trustDomainID).Error; err != nil {
		return sqlError.Wrap(err)
	}
```
So for consistency I moved all to statuses, but I was not sure if we really wanted that...

guilhermocc: We are indeed returning GRPC status error in other sqlstore methods as well, I wasn't aware of it.

amartinezfayo: Yes, we should return the status here. Since the `sqlstore` package is better positioned than anyone to know the exact reason of the error, I think that it makes sense to bubble up the status from here rather than having to parse the error and assign a status in an upper layer.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1052,6 +1059,52 @@ func pruneBundle(tx *gorm.DB, trustDomainID string, expiry time.Time, log logrus
...
+		}
+
+		ok, err := cryptoutil.PublicKeyEqual(cert.PublicKey, publicKey)
+		if err != nil {
+			return status.Errorf(codes.Internal, "failed to compare public key: %v", err)
+		}
+
+		if ok {
+			if !ca.TaintedKey {
+				return status.Error(codes.Internal, "it is not possible to revoke an untainted root CA")
```
amartinezfayo, line 1088: The error is due to an invalid input argument, not an internal failure.

```suggestion
				return status.Error(codes. InvalidArgument, "it is not possible to revoke an untainted root CA")
```

amartinezfayo (approved):
🎉

PR #3981, opened 2023-03-14
Force Rotation: Implement TaintJWTKey in datastore

### LINKED ISSUE #3889: Implement  TaintJWTKey in datastore
Implement TaintJWKKey

Depends on #3887

MarcosDY:
**Which issue this PR fixes**
fixes #3889


pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1052,6 +1059,55 @@ func pruneBundle(tx *gorm.DB, trustDomainID string, expiry time.Time, log logrus
...
+	}
+
+	var taintedKey *common.PublicKey
+	for _, jwtKey := range bundle.JwtSigningKeys {
+		if jwtKey.Kid != taintedKeyID {
+			continue
+		}
+
+		if jwtKey.TaintedKey {
+			return nil, status.Error(codes.Internal, "key is already tainted")
```
amartinezfayo, line 1080: The error is due to an invalid input argument, not an internal failure.

```suggestion
			return nil, status.Error(codes.InvalidArgument, "key is already tainted")
```

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1052,6 +1059,55 @@ func pruneBundle(tx *gorm.DB, trustDomainID string, expiry time.Time, log logrus
 	return changed, nil
 }
 
+func taintJWTKey(tx *gorm.DB, trustDomainID string, taintedKeyID string) (*common.PublicKey, error) {
```
amartinezfayo, line 1062: Since the function _will_ taint the key, the taintedKeyID argument is confusing, because with that name I would expect to provide a key that is already tainted.

Maybe keyIDToTaint or just keyID are better options?

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1052,6 +1059,55 @@ func pruneBundle(tx *gorm.DB, trustDomainID string, expiry time.Time, log logrus
...
+	model := &Bundle{}
+	if err := tx.Find(model, "trust_domain = ?", trustDomainID).Error; err != nil {
+		return nil, sqlError.Wrap(err)
+	}
+
+	bundle, err := modelToBundle(model)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to unmarshal bundle: %v", err)
+	}
```
amartinezfayo, line 1071: This code is repeated across the taint/revoke functions. We may have a function that could be called from them.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1052,6 +1059,55 @@ func pruneBundle(tx *gorm.DB, trustDomainID string, expiry time.Time, log logrus
...
+		}
+
+		if jwtKey.TaintedKey {
+			return nil, status.Error(codes.Internal, "key is already tainted")
+		}
+
+		// Breaking here to be sure that we have only one
+		// key with the giver KeyID, it must never happens
+		// since we does not allow to have repeated key IDs
+		if taintedKey != nil {
```
amartinezfayo, line 1086: Where does `taintedKey` can be assigned a value after the var declaration?

pkg/server/datastore/sqlstore/sqlstore.go
MarcosDY: I'll update to keyID

pkg/server/datastore/sqlstore/sqlstore.go
amartinezfayo: Nevermind, I just realized that this can be assigned after this line, and it's inside a for loop.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1052,6 +1059,55 @@ func pruneBundle(tx *gorm.DB, trustDomainID string, expiry time.Time, log logrus
...
+			continue
+		}
+
+		if jwtKey.TaintedKey {
+			return nil, status.Error(codes.Internal, "key is already tainted")
+		}
+
+		// Breaking here to be sure that we have only one
+		// key with the giver KeyID, it must never happens
+		// since we does not allow to have repeated key IDs
```
amartinezfayo, line 1085: Maybe this could be rephrased:

```suggestion
		// Check if a JWT Key with the provided keyID was already
		// tainted in this loop. This is purely defensive since we do not
		// allow to have repeated key IDs.
```

amartinezfayo (approved):
🎉

PR #3997, opened 2023-03-17
Do not require that X509 CAs minted by an upstream authority have a URI SAN

### LINKED ISSUE #3996: Do not require that X509 CAs minted by an upstream authority have a URI SAN
It should not be required that X509 CAs minted by an upstream authority have a URI SAN.
Upstream authority plugins may not have control over that, which could be problematic. Since this is not a hard requirement in the specification, relax the current validation to not require this.

amartinezfayo:
Changed the validation in the `credvalidator` package to not require that X509 CAs minted by an upstream authority have a URI SAN.
When a CA certificate does have a URI SAN, other validations still apply: https://github.com/spiffe/spiffe/blob/main/standards/X509-SVID.md#32-signing-certificates

Fixes #3996.


azdagron (approved):
\o/ thanks @amartinezfayo !

PR #4003, opened 2023-03-21
Update GitHub runners to use ubuntu-22.04

rturner3:
Ubuntu 22.04 is the latest LTS version supported by GitHub actions as a builtin runner type.

Also update the dev Dockerfile to be based on Ubuntu 22.04 Jammy, since it is currently using a really old version of Ubuntu (16.04) that is no longer supported.

### DIFF
--- a/.github/workflows/nightly_build.yaml
+++ b/.github/workflows/nightly_build.yaml
@@ -10,7 +10,7 @@ env:
 
 jobs:
   build-and-publish-images:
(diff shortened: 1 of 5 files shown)


rturner3:
@azdagron reminded me that this could cause compatibility issues because our release binaries currently need to link to glibc, which may be a different version across Ubuntu releases. Closing this until we address #4008.

PR #4020, opened 2023-03-22
Force Rotation:  local authority scaffolding and implements X.509 authority endpoints

### LINKED ISSUE #3895: Create LocalAuthority Scaffolding
Create LocalAuthority Scaffolding based on API SDK

Depends on #3885

### LINKED ISSUE #3897: Implement X.509 Local Authority RPCs
Implement:
- Implement GetX509AuthorityState
- implement PrepareX509Authority
- implement ActivateX509Authority
- Implement taintX509Authiority
- implement RevokeX509Authority

Depends on #3895 #3885

MarcosDY:
Force Rotation:  local authority scaffolding and implements X.509 authority endpoints

**Which issue this PR fixes**
fixes #3895 and fixes #3897


pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,278 @@
...
+
+	rpccontext.AuditRPC(ctx)
+
+	return &localauthorityv1.ActivateX509AuthorityResponse{
+		ActivatedAuthority: state,
+	}, nil
+}
+
+func (s *Service) TaintX509Authority(ctx context.Context, req *localauthorityv1.TaintX509AuthorityRequest) (*localauthorityv1.TaintX509AuthorityResponse, error) {
+	parseRequest := func() logrus.Fields {
```
guilhermocc, line 168: @MarcosDY What do you think about deanonymizing this function? it is duplicated in the method below

pkg/server/api/localauthority/v1/service_test.go
```diff
@@ -0,0 +1,1024 @@
...
+					Message: "API accessed",
+					Data: logrus.Fields{
+						telemetry.Status: "success",
+						telemetry.Type:   "audit",
+					},
+				},
+			},
+		},
+		{
+			name:        "slot has no public key",
```
guilhermocc, line 125:
```suggestion
			name:        "next slot has no public key",
```

pkg/server/api/localauthority/v1/service_test.go
```diff
@@ -0,0 +1,1024 @@
...
+	"testing"
+
+	"github.com/andres-erbsen/clock"
+	"github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/test"
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	localauthorityv1 "github.com/spiffe/spire-api-sdk/proto/spire/api/server/localauthority/v1"
+	"github.com/spiffe/spire/pkg/common/telemetry"
+	"github.com/spiffe/spire/pkg/server/api"
+	localauthority "github.com/spiffe/spire/pkg/server/api/localauthority/v1"
```
guilhermocc, line 18: Redundant alias

pkg/server/api/localauthority/v1/service_test.go
```diff
@@ -0,0 +1,1024 @@
...
+		keyToRevoke []byte
+
+		expectKeyToTaint crypto.PublicKey
+		expectLogs       []spiretest.LogEntry
+		expectCode       codes.Code
+		expectMsg        string
+		expectResp       *localauthorityv1.RevokeX509AuthorityResponse
+	}{
+		{
+			name:        "taint old authority",
```
guilhermocc, line 717:
```suggestion
			name:        "revoke old authority",
```

pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		states = append(states, state)
+	}
+
+	next := s.ca.GetNextX509CASlot()
+	// when next has a key indicates that it was initialized
+	if next.GetPublicKey() != nil {
+		status := localauthorityv1.AuthorityState_PREPARED
+		// CA is removed from slot on rotation
+		if next.IsEmpty() {
```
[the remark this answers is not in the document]
MarcosDY: no, when we reset a slot, the x509 field is set to be `nil` but we keep the public key, so with that we know than that slot contains an OLD authority (you can take a look to [RotateX509 ](https://github.com/spiffe/spire/blob/main/pkg/server/ca/manager/manager.go#L246-L252)

pkg/server/api/localauthority/v1/service_test.go
MarcosDY: it is not, if we does not set alias it will try to use `v1`

guilhermocc: I've tried to run the tests without this alias, and they are all passing; the package name being imported from github.com/spiffe/spire/pkg/server/api/localauthority/v1 is already "localauthority".

pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,279 @@
...
+
+	// TODO: implement a way to activate an OLD authority
+	if req.AuthorityId != "" {
+		log = log.WithField(telemetry.LocalAuthorityID, req.AuthorityId)
+		return nil, api.MakeErr(log, codes.InvalidArgument, "activating an old authority is not supported yet", nil)
+	}
+
+	if s.ca.GetNextX509CASlot().IsEmpty() {
+		return nil, api.MakeErr(log, codes.Internal, "no prepared authority found", nil)
+	}
```
evan2645, line 154: Good check here ... I was surprised to see that CA manager doesn't do this check .. maybe it should?

pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,279 @@
...
+
+	rpccontext.AuditRPC(ctx)
+	log.Info("X.509 authority revoked successfully")
+
+	return &localauthorityv1.RevokeX509AuthorityResponse{
+		RevokedAuthority: state,
+	}, nil
+}
+
+// getX509PublicKey validates provided authority ID, and return OLD associated publick key, in case of authority ID is not provided, use the current OLD authority
```
evan2645, line 231:
```suggestion
// getX509PublicKey validates provided authority ID, and return OLD associated public key, in case of authority ID is not provided, use the current OLD authority
```

Relying on empty authority id to mean the old authority feels kind of dangerous (related to above comment about missing authority id in req)

pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,279 @@
...
+		cert, err := x509.ParseCertificate(ca.DerBytes)
+		if err != nil {
+			return "", nil, err
+		}
+
+		subjectKeyID := x509util.SubjectKeyIDToString(cert.SubjectKeyId)
+		if authorityID == subjectKeyID {
+			return subjectKeyID, cert.PublicKey, nil
+		}
+	}
```
evan2645, line 268: Should we be searching the bundle for these keys? I think one side effect of this is I can taint a key in the bundle that isn't managed locally ... Since it's the Local Authority API, kind of feels like you should be interacting with the server instance which owns the key? Or are there use cases that require this kind of thing?

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1148,56 +1143,76 @@ func taintX509CA(tx *gorm.DB, trustDomainID string, taintedPublicKey crypto.Publ
...
+		if ok {
+			taintedKeyFound = true
+			continue
+		}
+
+		taintedKeys = append(taintedKeys, eachTaintedKey)
+	}
+
+	if !taintedKeyFound {
+		return status.Error(codes.InvalidArgument, "it is not possible to revoke an untainted root CA")
```
evan2645, line 1174: Good check! I think it may be better to handle in local authority api though?

proto/spire/common/common.proto
```diff
@@ -125,9 +125,6 @@ message RegistrationEntries {
 /** Certificate represents a ASN.1/DER encoded X509 certificate */
 message Certificate {
     bytes der_bytes = 1;
-
-    /** Whether or not the key used to sign the certificate is tainted */
-    bool tainted_key = 2;
```
evan2645, line 130: Why move this to the top level? Feels natural to have on a per-cert and per-key basis ... it remains set on PublicKey message

evan2645 (commented):
Very nice work @MarcosDY thank you so much! Code quality is high and it was easy to read through ❤️ 🙏

pkg/common/x509util/keyid.go
```diff
@@ -27,3 +28,21 @@ func GetSubjectKeyID(pubKey interface{}) ([]byte, error) {
 	keyID := sha1.Sum(subjectKeyInfo.SubjectPublicKey.Bytes) //nolint: gosec // usage of SHA1 is according to specification
 	return keyID[:], nil
 }
+
+// SubjectKeyIDToString parse Subject Key ID into string
+func SubjectKeyIDToString(ski []byte) string {
```
evan2645, line 33: In https://github.com/spiffe/spire/pull/4216 @rturner3 landed on lowercase hex, no colon

pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,279 @@
...
+func (s *Service) RevokeJWTAuthority(context.Context, *localauthorityv1.RevokeJWTAuthorityRequest) (*localauthorityv1.RevokeJWTAuthorityResponse, error) {
+	return nil, status.Error(codes.Unimplemented, "unimplemented")
+}
+
+func (s *Service) GetX509AuthorityState(ctx context.Context, _ *localauthorityv1.GetX509AuthorityStateRequest) (*localauthorityv1.GetX509AuthorityStateResponse, error) {
+	log := rpccontext.Logger(ctx)
+
+	var states []*localauthorityv1.AuthorityState
+	current := s.ca.GetCurrentX509CASlot()
+	if !current.IsEmpty() {
```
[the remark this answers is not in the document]
MarcosDY: technically this must never happens... we can just not validate it.

pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,279 @@
...
+		// CA is removed from slot on rotation
+		if next.IsEmpty() {
+			status = localauthorityv1.AuthorityState_OLD
+		}
+
+		states = append(states, &localauthorityv1.AuthorityState{
+			AuthorityId: next.AuthorityID(),
+			Status:      status,
+		})
+	}
```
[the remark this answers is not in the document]
MarcosDY: Actually we have an State field on journal now, and we can just parse that status into a  `localauthorityv1.AuthorityState` I was thinking on do that on this PR but I choose to keep it out for now, 
About the third `slot` approach, are you ok with creating another issue for that? 
I can start working on that and we can adapt this code if required, I `think` since it is an ID we'll not affect key managers but I will need to review that,

pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if err := s.ca.PrepareX509CA(ctx); err != nil {
+		return nil, api.MakeErr(log, codes.Internal, "failed to prepare X.509 authority", err)
+	}
+
+	slot := s.ca.GetNextX509CASlot()
+	// Prepare is going to use current slot when it is empty
+	if slot.IsEmpty() {
+		slot = s.ca.GetCurrentX509CASlot()
+	}
```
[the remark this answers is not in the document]
MarcosDY: agree it must never happen, I'm happy to return an error instead, but we must never get into this case, even when initializing, we must load authorities before open APIs, so I dont expect it to happen

pkg/server/api/localauthority/v1/service.go
MarcosDY: not sure if this should validate this on Activate, since we depends on dates, on normal rotations this must never happens, the issue here is that basically we are forcing this, and preparation may not be done yet,
I prefer to return error here since we allow users to force a preparation and be sure that the prepared bundle is propagated before they activate it.

pkg/server/api/localauthority/v1/service.go
MarcosDY: so do you suggest here to make authorityID required, but it must be the old on?

pkg/server/api/localauthority/v1/service.go
MarcosDY: searching be publickey was my initial idea too, as you said we may be able to taint a no local authority, I dont have a use case about that "yet",
So are you ok with getting authorityID and validate if it is used in the Old authority and reject if that is not the case? (we can have the Old one once we add the third Slot)

pkg/server/datastore/sqlstore/sqlstore.go
MarcosDY: I was not sure about that, I'm ok in moving it to service layer, but revoking on datastore without this validation worry me

proto/spire/common/common.proto
MarcosDY: the issue here is that we does not have intermediate certificates on bundle when UpstreamAuthority is used, 
So we keep "only" bundle here in that case, but intermediates lives on Journal,
So in order to taint a key in that case we must store the public key in some place

pkg/server/api/localauthority/v1/service.go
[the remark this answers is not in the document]
evan2645: > are you ok with creating another issue for that?

Yes for sure! Didn't intend for all that to be in scope for this PR, just that it seems like a good future direction

pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,279 @@
...
+		states = append(states, &localauthorityv1.AuthorityState{
+			AuthorityId: next.AuthorityID(),
+			Status:      status,
+		})
+	}
+
+	rpccontext.AuditRPC(ctx)
+
+	return &localauthorityv1.GetX509AuthorityStateResponse{
+		States: states,
```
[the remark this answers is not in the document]
evan2645: This feels much better 😁 but I also know that we already reviewed the API etc and I missed this, so I don't think we should block on it here

pkg/server/api/localauthority/v1/service.go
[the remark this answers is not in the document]
evan2645: If it's unexpected, feels like a good idea to return an error

pkg/server/datastore/sqlstore/sqlstore.go
evan2645: Ok .. it felt weird to me for datastore to be knowing about this kind of thing, but I think it's fine to leave as a defensive measure 👍

pkg/server/api/localauthority/v1/service.go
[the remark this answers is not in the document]
MarcosDY: It is always expected to have changes on implementation, based on how things looks, 
I like to have improvements as soon as possible, 
I created a [PR](https://github.com/spiffe/spire-api-sdk/pull/43) on spire-api-sdk,
as soon it is merged I'll apply changes here

pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	nextSlot := s.ca.GetNextX509CASlot()
+
+	switch {
+	// Authority ID is required
+	case req.AuthorityId == "":
+		return nil, api.MakeErr(log, codes.InvalidArgument, "no authority ID provided", nil)
+
+		// It is not possible to taint Active authority
```
evan2645, line 191: non-blocking nit: indentation error

pkg/server/api/localauthority/v1/service.go
```diff
@@ -0,0 +1,302 @@
...
+	case req.AuthorityId == s.ca.GetCurrentX509CASlot().AuthorityID():
+		return nil, api.MakeErr(log, codes.InvalidArgument, "unable to taint current local authority", nil)
+
+	// Only next local authority can be tainted
+	case req.AuthorityId != nextSlot.AuthorityID():
+		return nil, api.MakeErr(log, codes.InvalidArgument, "unexpected authority ID", nil)
```
evan2645, line 197: We only want to taint "old" authorities, right?

pkg/server/api/localauthority/v1/service.go
evan2645: Yea that sounds good .. probably better to start with local only, and extend it if/when we have a good use case for that

evan2645 (commented):
Left a couple more comments, I think this is good to go otherwise. Thanks again @MarcosDY 🙏

pkg/server/api/localauthority/v1/service.go
MarcosDY: yeap, the next case validates that the next is not in prepared status (so only old ones can be tainted)

evan2645 (approved):
Amazing work @MarcosDY thank you!! 🎸

PR #4045, opened 2023-04-05
Use go spiffe types server unit

guilhermocc:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
No functionality affected

**Description of change**
Use go-spiffe types in spire-server units

**Which issue this PR fixes**
fixes https://github.com/spiffe/spire/issues/1828


pkg/common/bundleutil/unmarshal_test.go
```diff
@@ -4,23 +4,26 @@ import (
...
+	"github.com/spiffe/go-spiffe/v2/bundle/spiffebundle"
 	"github.com/spiffe/go-spiffe/v2/spiffeid"
 	"github.com/stretchr/testify/require"
 )
 
 func TestUnmarshal(t *testing.T) {
 	rootCA := createCACertificate(t)
 	trustDomain := spiffeid.RequireTrustDomainFromString("domain.test")
+	emptyBundle := spiffebundle.New(trustDomain)
+	emptyBundle.SetRefreshHint(0)
```
MarcosDY, line 16: why do you need to set hint? it will be nil as default, is that causing issues?

pkg/server/plugin/upstreamauthority/spire/spire_test.go
```diff
@@ -326,9 +326,10 @@ func TestPublishJWTKey(t *testing.T) {
 	require.NotNil(t, upstreamJwtKeys)
 
 	require.Len(t, upstreamJwtKeys, 3)
-	assert.Equal(t, upstreamJwtKeys[0].Kid, "C6vs25welZOx6WksNYfbMfiw9l96pMnD")
-	assert.Equal(t, upstreamJwtKeys[1].Kid, "gHTCunJbefYtnZnTctd84xeRWyMrEsWD")
-	assert.Equal(t, upstreamJwtKeys[2].Kid, "kid-2")
+	keyIDs := []string{upstreamJwtKeys[0].Kid, upstreamJwtKeys[1].Kid, upstreamJwtKeys[2].Kid}
+	assert.Contains(t, keyIDs, "C6vs25welZOx6WksNYfbMfiw9l96pMnD")
+	assert.Contains(t, keyIDs, "gHTCunJbefYtnZnTctd84xeRWyMrEsWD")
+	assert.Contains(t, keyIDs, "kid-2")
```
MarcosDY, line 332: so order is no longer important?

guilhermocc: @MarcosDY actually, this is happening due to changes on fake_handlers_test; this sorting break is not happening in the real implementation, I'm going to update it!

pkg/common/bundleutil/unmarshal_test.go
guilhermocc: @MarcosDY emptyBundle holds the expected value for the empty bundle ("{}"); this is just an adjustment for ensuring the refreshHint will be set to zero by default and not nil (to maintain the current behavior).

PR #4062, opened 2023-04-10
Fixes #2247 - checksum algorithm in release package artifact filename

### LINKED ISSUE #2247: Indicate checksum algorithm used for release package artifacts in checksum filenames
The SPIRE and SPIRE extras release package files have corresponding checksum files that can be used to verify the integrity of the release bits fetched from GitHub. These checksum filenames follow the naming convention:
```
spire-<version>-linux-x86_64-glibc_checksums.txt
spire-extras-<version>-linux-x86_64-glibc_checksums.txt
```

The contents of these files appear to be the output of `shasum -a 256 <package_file>` on Linux. Ideally the checksum filenames should indicate that SHA-256 is the algorithm used to compute the checksum over the release package files so that users know which checksum to compute locally to verify their downloaded release packages were not tampered in transit.

edwbuck:
Additionally, now supports multiple checksum algorithms on release, which demonstrates the file name and algorithms match.

Additionally, now releases all checksums in the same file, supporting 'sha256sum --check file.txt' verifying all downloadables in one go.

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Release functionality only.  Same checksum data in different files.  Additional checksum files for different algorithms on Linux only.  Common tools of md5sum and shasum must be installed.

**Description of change**
Centralize the checksum / hashes of released items into a per-release file for each hashing algorithm.  Add in hashing algorithms on Linux platform.  Push the algorithm name into the file name.

All algorithms can now be verified with 'sha256sum --check <checksum_file>_sha256sum.txt'

**Which issue this PR fixes**
Fixes #2247


edwbuck:
Odds are that we should trim which of the checksums we will support, and perhaps add the relevant Windows-side ones.

The major change here is that they get all lumped into a single per-algorithm txt file.  This works well with the often supported '--check <filename>' option of various hashing tools, but it does mean that the "extra artifacts" checksum file gets rolled into the "per algorithm" file.

rturner3:
Hi @edwbuck, thanks for opening this! I noticed this PR is trying to do a few different things, but I'm not sure whether all these changes are needed.

Regarding adding more checksum algorithms to the release artifacts, did you have any particular use case in mind where verifying a SHA-256 checksum would not be suitable? If there is a need to support additional checksum algorithms, I think it would be good to open an issue for that where we can discuss the requirements and decide on which, if any, new algorithms we want to support in our release artifacts. Since these get included in each release artifact, it would be helpful to know if they are going to serve any class of users.

My suggestion would be that we scope this PR to just address the issue described in #2247, because I think that part is a bit more well understood and accepted.

edwbuck:
This commit does three things

1. It combines the two checksum files into one.  These never should have been individual files, as all checksum verification tools has a `--check` option to verify all files in the checksum file with one command line call.  By shipping an individual checksum file with each artifact, support for algorithms makes the checksum file count grow at 0(n^2) (algorithms * distributables) instead of 0(n) (algorithms)
2. It adds support for different checksum algorithms.  While it is not a critical matter, reflecting the algorithm in the checksum file name is the point of 2247, If we don't parameterize this early, then the script will suffer from cut-and-paste code (which it already does, leading to two checksum files).  As far as "need" for different checksums, it depends on a customer who whishes to verify against two algorithms to ensure that a (one in quadrillion chance) hash collision hasn't occurred.
3. It ensures that the name of the file contains the algorithm.

If you want the name of the file to only contain the algorithm, the old duplicated generation of two file names can be done to satisfy 2247, but then it will be a file rename only, with a fixed string of sha256sum being added in the output name, with two files both stating they are sha256sum, and should we ever need to change the sha256sum, to a sha512sum, you'll have to rewrite the release tooling then

Or, I could submit two additional issues, one that states we shouldn't have two checksum files for one release of two executables, and another that asks for support of multiple algorithms, and then you could review this against three issues.  It's your choice.

edwbuck:
Sorry, first time I read your comment, I saw "not sure it's needed".  Now I realize you mean "not sure is (wanted)".

I'll simply do a file rename, not combine the SPIRE checksums, and not make the system support multiple checksum algorithms.  I guess SPIRE can then rewrite it later should sha256sum fall out of favor and sha512sum become the new player.

rturner3:
> Sorry, first time I read your comment, I saw "not sure it's needed". Now I realize you mean "not sure is (wanted)".

Yeah that's more what I meant, sorry if that didn't come across clearly.

> I'll simply do a file rename, not combine the SPIRE checksums, and not make the system support multiple checksum algorithms. I guess SPIRE can then rewrite it later should sha256sum fall out of favor and sha512sum become the new player.

Great, I think this fits the bill for now.

> It combines the two checksum files into one. These never should have been individual files, as all checksum verification tools has a --check option to verify all files in the checksum file with one command line call. By shipping an individual checksum file with each artifact, support for algorithms makes the checksum file count grow at 0(n^2) (algorithms * distributables) instead of 0(n) (algorithms)

I'm open to this idea. I've personally seen projects do this different ways, where some have individual checksum files and some have one `SHASUMS` file like you suggested. I don't think any one of these approaches is necessarily incorrect since tools like `shasum` typically support passing the checksum file as an argument, so `shasum --check` could be invoked multiple times with different arguments for different artifacts. I do however see that having one file for all this information would be nice if we start publishing binaries in releases for other operating systems/platforms (e.g. `arm64` architecture, `darwin` OS). One checksum file would keep the number of release artifacts trimmed down, as you mentioned.

Feel free to open an issue for this. Note that this would also require some changes on the [spiffe.io](https://spiffe.io/downloads/) website, which currently links out to SPIRE release assets on its Downloads page.

rturner3 (approved):

PR #4087, opened 2023-04-14
Fix Linux release artifact checksum files

rturner3:
The checksum files published for Linux release artifacts are published in a format that is not consumable by tools like
`shasum --check`/`sha256sum --check` because these tools expect two spaces between the checksum and the filename. Add the missing space to fix this.

Tested on macOS with `shasum --check <file>` and on Ubuntu with `sha256sum --check <file>`.


amartinezfayo (approved):
Thank you @rturner3!

rturner3:
Closing in favor of #4096, which is more robust solution to this problem.

PR #4096, opened 2023-04-19
Directly write the output of the hashing command into the file

### LINKED ISSUE #4095: Don't parse and then reformat shasum hash lines
When generating sha256sum hashes of the release executables, we cut the output to obtain the hash and then join that back to the file name, in a way that ensures the output file name lacks any parent directory paths.

The problem is that this destroys the specific formatting of the hashing algorithm, such that on some platforms the line isn't recognized as valid input to the hashing algorithm's --check command line option.

Instead of cutting the output line, we need to change to the directory of the file to be hashed and then request the hashing of the filename only, lacking any parent paths.  The output of that command should be written to the hash file without parsing.

All of this should be done in a sub-shell, avoiding the chance that a command leaves the current working directory in the wrong state should any of the internal commands to generate the hash or write it to a file fail.

edwbuck:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
checksum file generation approach, output should be identical.

**Description of change**
Directly write the output of the hashing command into the checksum files.

This avoids the cut and pasting that we do to take the hash from the hashing algorithm and format it with the file from the trimmed full-file path.  It does so by opening a sub-shell, changing directory to the parent of the file to be hashed, generating the hash on the file name only (relative to the current working directory), and then writing the output to the checksum file.

This avoids the need to cut the hashing output to get the hash so we can then reconstruct the hash line with a file name lacking parent directories.  This will greatly improve the ability of the --check command to work, as now we don't alter the format of the sha256sum line.

**Which issue this PR fixes**
Fixes #4095

### DIFF
--- a/script/build-artifact.sh
+++ b/script/build-artifact.sh
@@ -19,6 +19,7 @@ fi
 # handle the case that we're building for alpine
 if [ "$OS" = "linux" ]; then
     case $(ldd --version 2>&1) in
+        *GNU\ libc*) LIBC="-glibc" ;;
         *GLIB*) LIBC="-glibc" ;;
         *muslr*) LIBC="-musl" ;;
         *) LIBC="-unknown" ;;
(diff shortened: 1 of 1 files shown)


script/build-artifact.sh
```diff
@@ -77,13 +77,13 @@ if [ $OS == "windows" ]; then
     (cd "${TMPDIR}/spire"; zip -rv "${ARTIFACT}" -- *)
     (cd "${TMPDIR}/spire-extras"; zip -rv "${EXTRAS_ARTIFACT}" -- *)
 
-    echo "$(CertUtil -hashfile "${ARTIFACT}" SHA256 | sed -n '2p') $(basename "${ARTIFACT}")" > "${CHECKSUM}"
-    echo "$(CertUtil -hashfile "${EXTRAS_ARTIFACT}" SHA256 | sed -n '2p') $(basename "${EXTRAS_ARTIFACT}")" > "${EXTRAS_CHECKSUM}"
+    (cd "$(dirname "${ARTIFACT}")"; CertUtil -hashfile "$(basename "${ARTIFACT}")" SHA256 > "${CHECKSUM}")
```
rturner3, line 80: To keep potential error output tidy and accurate, in each of these new subshell invocations, we will want to use `&&` rather than `;` to separate the commands, such that the command to generate the hash file is not invoked if we cannot even change to the directory where the artifact is located (maybe because of a bug where `ARTIFACT` is not set properly).
```suggestion
    (cd "$(dirname "${ARTIFACT}")" && CertUtil -hashfile "$(basename "${ARTIFACT}")" SHA256 > "${CHECKSUM}")
```

rturner3: I take it back, I see that we are setting `set -e` in this script, so this is good as written.

rturner3 (approved):

PR #4111, opened 2023-04-26
datastore: use UNION ALL for postgres list entries query

sorindumitru:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Listing registration entries.

**Description of change**

This cuts the cost in time, as determined by EXPLAIN ANALYZE, in half. For a database with ~20000 entries and ~50000 selectors the time goes from ~200ms to ~100ms.

The difference between UNION and UNION ALL is that the former also does a DISTINCT step on the entries, removing any duplicates. It's possible I'm missing something simple, but I don't think that actually does anything for us here since most of the sub-queries also return an unique primary key so they are already unique.

Similar change can probably be done for the MySQL CTE query, but I haven't tested that yet. Can do that if this is considered an acceptable change.


rturner3 (approved):
This looks good, nice catch @sorindumitru!

Issue #4083, opened 2023-04-13
Publish release Linux arm64 binary archives


rturner3:
The project now publishes releases Linux arm64 Docker images, but it only publishes binary archives for x86_64 architecture on Linux. Ideally, the project should consistently publish the same set of release artifacts for each OS + architecture combination.

Since the project currently only supports arm64 on Linux, the following artifacts should be published for Linux arm64:

- SPIRE binaries + configuration archive (and checksum file)
- SPIRE extras archive (and checksum file)

PR #4143, opened 2023-05-05
Publish arm64 artifacts in release and PR builds for Linux

rturner3:
The project started publishing release Docker images for Linux arm64 starting in SPIRE v1.6.0, but the Linux release binary artifacts were still only generated for amd64 architecture.

Publish binary artifacts in releases for Linux arm64 to provide consistent architecture support for both binaries and Docker images. Also build and publish the Linux arm64 artifacts on PR builds to prevent breaking changes for this platform from being merged into the project.

**Which issue this PR fixes**
#4083


amartinezfayo (approved):
Thank you @rturner3!

PR #4188, opened 2023-05-30
Force Rotation: Add status to journal

### LINKED ISSUE #4187: Update journal to persist entry Status
Originally Journal depends on dates to know when an entry is marked as OLD keep only for historical purposed.
But it is not longer possible to depends on that, now that we can force rotations and prepare new authorities.
So we can have bundles that are not expired (OLD) but those authorities was forced to rotate, but we still need to keep them 
on journal file to keep historic.
With this issue it is possible to load and "OLD" authority into ACTIVE or even Prepared status when the intention was to make that authority as no longer used.

So this proposal is to add next statuses:
- Active: authority is currently used to sign SVIDs
- Old: Previously active authority, but no longer used
- Prepared: authority prepared in advance to be propagated, but it is not used for signing yet
- Unknown: this is the default status but will not be used except when upgrading for older spire server

MarcosDY:
Add current status to stored entries on journal file

**Which issue this PR fixes**
fixes #4187


pkg/server/ca/manager/slot.go
```diff
@@ -86,74 +84,178 @@ func (s *SlotLoader) Load(ctx context.Context) (*Journal, map[SlotPosition]Slot,
...
+// - If all status are unknown, choose the two newest on the list
+// - Active entry is returned on current if set
+// - Newest Prepared or Old entry is returned on next
+func (s *SlotLoader) getX509CASlots(ctx context.Context, entries []*X509CAEntry) (*X509CASlot, *X509CASlot, error) {
+	var current *X509CASlot
+	var next *X509CASlot
+
+	// Search from oldest
+	for i := len(entries) - 1; i >= 0; i-- {
+		slot, err := s.tryLoadX509CASlotFromEntry(ctx, entries[i])
```
amartinezfayo, line 127: Since this is now done in every entry, a lot of possible unusable slots will be found and logged. We should probably update `tryLoadX509CASlotFromEntry` to log more information about the entry, like the status and issuance time. Also, I'm wondering if entries with OLD status should really log a warning. Maybe debug level is more appropriate,.

pkg/server/ca/manager/slot.go
```diff
@@ -86,74 +84,178 @@ func (s *SlotLoader) Load(ctx context.Context) (*Journal, map[SlotPosition]Slot,
...
+		current = newX509CASlot("A")
+		next = newX509CASlot("B")
 	}
 
-	if len(entries.JwtKeys) > 0 {
-		nextJWTKey, err = s.tryLoadJWTKeySlotFromEntry(ctx, entries.JwtKeys[len(entries.JwtKeys)-1])
+	return current, next, nil
+}
+
+// getJWTKeysSlots returns slots based on status
```
amartinezfayo, line 189:
```suggestion
// getJWTKeysSlots returns JWTKey slots based on the status of the slots.
```

pkg/server/ca/manager/slot.go
```diff
@@ -86,74 +84,178 @@ func (s *SlotLoader) Load(ctx context.Context) (*Journal, map[SlotPosition]Slot,
...
+// - If all status are unknown, choose the two newest on the list
```
amartinezfayo, line 190:
```suggestion
// - If all the statuses are unknown, the two most recent slots are returned.
```

pkg/server/ca/manager/slot.go
```diff
@@ -86,74 +84,178 @@ func (s *SlotLoader) Load(ctx context.Context) (*Journal, map[SlotPosition]Slot,
...
+// - Active entry is returned on current if set
+// - Newest Prepared or Old entry is returned on next
```
amartinezfayo, line 192: I think that these could benefit from some rewording, to make them more clear.

pkg/server/ca/manager/slot.go
```diff
@@ -86,74 +84,178 @@ func (s *SlotLoader) Load(ctx context.Context) (*Journal, map[SlotPosition]Slot,
...
+func (s *SlotLoader) getJWTKeysSlots(ctx context.Context, entries []*journal.JWTKeyEntry) (*JwtKeySlot, *JwtKeySlot, error) {
+	var current *JwtKeySlot
+	var next *JwtKeySlot
+
+	// Search from oldest
+	for i := len(entries) - 1; i >= 0; i-- {
+		slot, err := s.tryLoadJWTKeySlotFromEntry(ctx, entries[i])
```
amartinezfayo, line 199: The same comment done in `getX509CASlots` applies here.

pkg/server/ca/manager/slot.go
MarcosDY: it is not really loading all, just until you found the active and the first prepared or old entry, generally you will be getting only the first 2 elements as before, except there was some force rotations envolved where user prepared more than one CA in advance (no regular case)

pkg/server/ca/manager/slot.go
```diff
@@ -86,74 +84,178 @@ func (s *SlotLoader) Load(ctx context.Context) (*Journal, map[SlotPosition]Slot,
...
+		}
+
+		switch slot.Status() {
+		// ACTIVE entry must go into current slot
+		case journal.Status_ACTIVE:
+			current = slot
+
+		// Status can be UNKNOWN only after an upgrade when this happens,
+		// we must first set next position, and then current is the next one
+		// DEPRECATED: this validation must be removed from v1.9
```
amartinezfayo, line 149:
```suggestion
		// TODO: status in journal has been introduced in v1.7.
		// Keep this validation in v1.7.x and v1.8.x. Remove this in v1.9.
```

pkg/server/ca/manager/slot.go
```diff
@@ -86,74 +84,178 @@ func (s *SlotLoader) Load(ctx context.Context) (*Journal, map[SlotPosition]Slot,
...
```
amartinezfayo, line 221:
```suggestion
		// TODO: status in journal has been introduced in v1.7.
		// Keep this validation in v1.7.x and v1.8.x. Remove this in v1.9.
```

pkg/server/ca/manager/slot.go
```diff
@@ -86,74 +84,178 @@ func (s *SlotLoader) Load(ctx context.Context) (*Journal, map[SlotPosition]Slot,
...
-			currentX509CA, err = s.tryLoadX509CASlotFromEntry(ctx, entries.X509CAs[len(entries.X509CAs)-2])
-			if err != nil {
-				return nil, nil, err
+
+		// Unable to load slot
+		// TODO: before this changes, we analize only latest two entries,
+		// and if this happens to both of them, we created new slots, but now
+		// we will iterate in all the file, to try to get a useful one,
+		// may we prevent this to happens? maybe just verifying if bundle
+		// is not expired?
```
amartinezfayo, line 137:
```suggestion
		// TODO: the previous implementation analized only the last two entries,
		// and if those slots were empty, we created new slots.
		// Now we iterate through all the file, to try to get a useful slot.
		// Maybe there is room for improvement here, by just verifying if the
		// bundle is not expired?
```

pkg/server/ca/manager/slot.go
```diff
@@ -86,74 +84,178 @@ func (s *SlotLoader) Load(ctx context.Context) (*Journal, map[SlotPosition]Slot,
...
-			currentJWTKey, err = s.tryLoadJWTKeySlotFromEntry(ctx, entries.JwtKeys[len(entries.JwtKeys)-2])
-			if err != nil {
-				return nil, nil, err
+
+		// Unable to load slot
+		// TODO: before this changes, we analize only latest two entries,
+		// and if this happens to both of them, we created new slots, but now
+		// we will iterate in all the file, to try to get a useful one,
+		// may we prevent this to happens? maybe just verifying if bundle
+		// is not expired?
```
amartinezfayo, line 209:
```suggestion
		// TODO: the previous implementation analized only the last two entries,
		// and if those slots were empty, we created new slots.
		// Now we iterate through all the file, to try to get a useful slot.
		// Maybe there is room for improvement here, by just verifying if the
		// bundle is not expired?
```

pkg/server/ca/manager/slot.go
```diff
@@ -53,11 +54,8 @@ func (s *SlotLoader) Load(ctx context.Context) (*Journal, map[SlotPosition]Slot,
 	log := s.Log
 	journalPath := s.journalPath()
 
-	var currentX509CA *X509CASlot
-	var nextX509CA *X509CASlot
-	var currentJWTKey *JwtKeySlot
-	var nextJWTKey *JwtKeySlot
-
+	// TODO: looks like it is old and no longer required
+	// DEPRECATED: remove on 1.7?
```
amartinezfayo, line 58: It's probably better to create an issue to look into this rather than having this comment here?

amartinezfayo (dismissed):
Thank you @MarcosDY!

Issue #3825, opened 2023-02-03
[RFE] allow bring your own certs for the oidc-disicovery-provider components


raffaelespazzoli:
currently the oidc discovery provider component assumes it can use let's encrypt to create it;s own certs. In general this is a strong assumption that will impede adoption as not all scenarios allow for that. This RFE is about enbaling the use case of bring your own pre-existing certs.

evan2645:
Hello @raffaelespazzoli, thanks for opening this and sorry for the delay. The project has previously avoided allowing these kinds of configurations because it felt antithetical to the goals of SPIFFE/SPIRE, which includes treating rotation as a first class citizen for all keys in the system. With this in mind, an expiring cert/key in a SPIRE deployment is likely to catch folks by surprise, and is not the experience that we want our users to have.

That said, we also recognize the pain that this stance has caused. We've revisited the topic in light of this issue, and came to the conclusion that this should be supported, however there should be adequate warnings in place. I had a look at the current configurables, and am proposing the following:

* New configuration section `static_serving_cert`
  * To be added as a one-of option alongside `acme` and `listen_socket_path`
* `cert_file_path` and `key_file_path` configurables inside `static_serving_cert` section, both mandatory
* When `static_serving_cert` is set, a warning is emitted informing the user that SPIRE is unable to rotate the configured static key, and the day/time that the certificate will expire

The configuration shape of this component has grown organically over the years and is not very ergonomic at all. Unfortunately, my proposed changes above only serve to make it worse. I can't think of any way to improve it that doesn't involve configurable deprecation, which I feel is out-of-scope for this issue.

raffaelespazzoli:
I don't like the name of the parameter: `static_serving_cert` , it assumes that external systems cannot rotate certs.
Instead the code should be monitoring those paths for possible changes. 
It's ironic, in my setup I ended up using let's encrypt, but I need to be in control of the process. So if the oidc-discovery-provider was able to read externally supplied certs, it would see those certs being rotated.

evan2645:
I see - if we want to add logic to detect updates to these files and apply them to the TLS stack without restarting, that sounds like a good idea and in that case the static prefix makes less sense. How about `serving_cert_file` as the section name? I think we still need to log a warning though .. you may know that there exists rotation mechanisms for this file, but SPIRE doesn't and certainly not all users will do that

raffaelespazzoli:
I like it!

evan2645:
[...]

PR #4190, opened 2023-05-30
Serving cert file OIDC provider

guilhermocc:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
oidc discovery provider HTTPS configuration.

**Description of change**
Add new configuration option for oidc discovery provider, so users can now use their own certificates to be used on HTTPS connections. The new configuration structure is described below:

```
serving_cert_file {
	cert_file_path = "path_to_certificate_file"
	key_file_path = "path_to_private_key_file"
}
``` 

Users can now indicate the file path of the certificate and private key that are going to be loaded from the disk and used for establishing TLS connections with clients. Upon start, oidc discovery provider will start watching for file changes in the provided paths and reload the certificate used for connections.

**Which issue this PR fixes**
Fixes https://github.com/spiffe/spire/issues/3825


support/oidc-discovery-provider/cert_manager_test.go
```diff
@@ -0,0 +1,423 @@
...
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/test"
+	"github.com/spiffe/spire/pkg/common/x509util"
+	"github.com/spiffe/spire/test/testkey"
+	"github.com/stretchr/testify/require"
+)
+
+var logger, logHook = test.NewNullLogger()
```
azdagron, line 22: This should probably live in TestTLSConfig?

support/oidc-discovery-provider/cert_manager_test.go
```diff
@@ -0,0 +1,423 @@
...
+
+func TestTLSConfig(t *testing.T) {
+	oidcServerKey := testkey.MustEC256()
```
azdagron, line 25: nit: it's ok here, but usually i try not to use `testkey` package level functions except at the package scope. If somebody runs `go test -count=1000`, it can end up auto-generating a ton of keys :)

My suggestion would be to move this key var to the package scope

support/oidc-discovery-provider/cert_manager_test.go
```diff
@@ -0,0 +1,423 @@
...
+
+	certTmpl.Subject.Country = []string{"US"}
+
+	oidcServerCertUpdated3, err := x509util.CreateCertificate(certTmpl, certTmpl, oidcServerKey.Public(), oidcServerKeyNew)
+	require.NoError(t, err)
+
+	tmpDir := t.TempDir()
+
+	err = writeFile(tmpDir+keyFilePath, oidcServerKeyPem)
+	require.NoError(t, err)
```
azdagron, line 70: nit: these test helpers could take the *testing.T and do the require.NoError internally instead of returning an error that has to be checked.

support/oidc-discovery-provider/cert_manager_test.go
```diff
@@ -0,0 +1,423 @@
...
+			if err != nil {
+				return false
+			}
+			require.Len(t, cert.Certificate, 1)
+			x509Cert, err := x509.ParseCertificate(cert.Certificate[0])
+			if err != nil {
+				return false
+			}
+			return reflect.DeepEqual(oidcServerCertUpdated1, x509Cert)
+		}, 1*time.Second, 101*time.Millisecond, "Failed to assert updated certificate")
```
azdagron, line 212: Some of these Eventually's might take longer on slow, resource constrained CI/CD environments. I'd bump the timeouts up on these to at least 10 seconds.

support/oidc-discovery-provider/cert_manager.go
```diff
@@ -0,0 +1,177 @@
...
+	keyFilePath      string
+	certLastModified time.Time
+	keyLastModified  time.Time
+	fileSyncInterval time.Duration
+	certMtx          sync.RWMutex
+	cert             *tls.Certificate
+	log              logrus.FieldLogger
+}
+
+func NewDiskCertManager(config *Config, log logrus.FieldLogger) (*DiskCertManager, error) {
```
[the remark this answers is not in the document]
guilhermocc: Nope, we can use ServingCertFile

support/oidc-discovery-provider/cert_manager_test_windows.go
```diff
@@ -0,0 +1,77 @@
...
+
+var (
+	certFilePath                = "\\oidcServerCert.pem"
+	keyFilePath                 = "\\oidcServerKey.pem"
+	fileDontExistMessage        = "The system cannot find the file specified."
+	filePermissionDeniedMessage = "Access is denied."
+)
+
+func writeFile(name string, data []byte) error {
+	handle, err := windows.Open(name, os.O_RDWR|os.O_CREATE, 0666)
```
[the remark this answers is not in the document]
guilhermocc: We can use os.WriteFile

support/oidc-discovery-provider/cert_manager.go
```diff
@@ -0,0 +1,177 @@
...
+		keyFilePath:      config.ServingCertFile.KeyFilePath,
+		fileSyncInterval: config.ServingCertFile.FileSyncInterval,
+		log:              log,
+	}
+
+	if err := dm.loadCert(); err != nil {
+		return nil, fmt.Errorf("failed to load certificate: %w", err)
+	}
+
+	go dm.watchFileChanges()
```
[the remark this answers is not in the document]
guilhermocc: I prefer the first option! But I will need to create the context hierarchy that we currently don't have for oidc-discovery-provider process

support/oidc-discovery-provider/main.go
```diff
@@ -112,9 +97,51 @@ func run(configPath string) error {
...
+		log.WithFields(logrus.Fields{
+			telemetry.Network: listener.Addr().Network(),
+			telemetry.Address: listener.Addr().String(),
+		}).Info("Serving HTTP")
+	case config.ServingCertFile != nil:
+		listener, err = newListenerWithServingCert(ctx, log, config)
+		if err != nil {
+			return nil, err
+		}
+		log.Info("Serving HTTPS using certificate loaded from disk")
```
MarcosDY, line 134: may we add fields for used files?

support/oidc-discovery-provider/main.go
```diff
@@ -140,6 +167,25 @@ func newSource(log logrus.FieldLogger, config *Config) (JWKSSource, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	go func() {
+		certManager.WatchFileChanges(ctx)
+	}()
+
+	tlsConfig := certManager.TLSConfig()
+
+	tcpListener, err := net.ListenTCP("tcp", &net.TCPAddr{Port: 8080})
```
MarcosDY, line 181: is this hardcoded port ok?

guilhermocc: It should be 443

support/oidc-discovery-provider/cert_manager_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+	chInfo := &tls.ClientHelloInfo{
+		ServerName: "oidc-provider-discovery.example.com",
+	}
+
+	ctx, cancelFn := context.WithCancel(context.Background())
+	certManager, err := NewDiskCertManager(&ServingCertFileConfig{
+		CertFilePath:     tmpDir + certFilePath,
+		KeyFilePath:      tmpDir + keyFilePath,
+		FileSyncInterval: 10 * time.Millisecond,
```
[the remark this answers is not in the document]
guilhermocc: This could make the test run faster, but we would still need to use require.Eventually to make some assertions since the time difference between loading a cert and moving the clock + making an assertion can cause errors.

support/oidc-discovery-provider/cert_manager_test.go
```diff
@@ -0,0 +1,334 @@
...
+	err := os.WriteFile(name, data, 0600)
+	require.NoError(t, err)
+}
+
+func removeFile(t *testing.T, name string) {
+	err := os.Remove(name)
+	require.NoError(t, err)
+}
+
+func assertFileDontExist(t *testing.T, filePath string, err error) {
```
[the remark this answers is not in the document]
guilhermocc: I did that initially, but I didn't like to have two extra files for differentiating only two string constants. Do you think it is worth it?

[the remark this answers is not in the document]
guilhermocc: I like the idea!

support/oidc-discovery-provider/main.go
```diff
@@ -91,10 +97,19 @@ func run(configPath string) error {
 		Handler:           handler,
 		ReadHeaderTimeout: 10 * time.Second,
 	}
+
+	go func() {
+		<-ctx.Done()
+		err = server.Shutdown(context.Background())
+		if err != nil {
```
azdagron, line 104: In general, it isn't safe to modify the `err` variable from the goroutine. In practice in this code it is fine because it isn't used after this point. It would be safer to use a new `err` variable, ideally scoped to the shutdown statement:
```
if err := server.Shutdown(context.Background()); err != nil {
    log.Error(err)
}
```

support/oidc-discovery-provider/main.go
```diff
@@ -152,15 +167,18 @@ func newSource(log logrus.FieldLogger, config *Config) (JWKSSource, error) {
...
 }
 
-func newListenerWithServingCert(log logrus.FieldLogger, config *Config) (net.Listener, error) {
+func newListenerWithServingCert(ctx context.Context, log logrus.FieldLogger, config *Config) (net.Listener, error) {
 	certManager, err := NewDiskCertManager(config.ServingCertFile, log)
 	if err != nil {
 		return nil, err
 	}
+	go func() {
+		certManager.WatchFileChanges(ctx)
```
azdagron, line 180: nit: I'll accept this code as-is since it's happening in `main`, but ideally the lifetime of this goroutine is controlled by the net.Listener it returns, i.e. when the listener is closed, this goroutine shuts down and the listener waits for it to close.

support/oidc-discovery-provider/main.go
azdagron: The acme code for sure used 443. Now that we're not bound to ACME, this could probably be configurable so you don't need `CAP_NET_BIND_SERVICE` capabilities when testing it out...

guilhermocc: You're right, we can provide more flexibility now, I will introduce a new addr configuration in this section.

guilhermocc:
@azdagron thank you very much for the review so far, I've addressed the last comments you made, could you review it again? :smile:

support/oidc-discovery-provider/README.md
```diff
@@ -91,11 +91,13 @@ will terminate if another domain is requested.
...
-| `key_file_path`      | string   | required  | The private key file path, the file must contain PEM encoded data. |          |
-| `file_sync_interval` | duration | optional  | Controls how frequently the service polls the files for changes.   | 1 minute |
+| Key                  | Type     | Required? | Description                                                        | Default    |
+|----------------------|----------|-----------|--------------------------------------------------------------------|------------|
+| `cert_file_path`     | string   | required  | The certificate file path, the file must contain PEM encoded data. |            |
+| `key_file_path`      | string   | required  | The private key file path, the file must contain PEM encoded data. |            |
+| `file_sync_interval` | duration | optional  | Controls how frequently the service polls the files for changes.   | 1 minute   |
+| `file_sync_interval` | duration | optional  | Controls how frequently the service polls the files for changes.   | 1 minute   |
+| `addr`               | string   | optional  | Exposes the service on the given address.                          | :433 |
```
azdagron, line 100: 443 should be the default (multiple places to fix)

azdagron (approved):
Thanks, @guilhermocc !

PR #4216, opened 2023-06-02
Produce serial number selector in x509pop NodeAttestor

### LINKED ISSUE #4215: Produce serial number node selector in x509pop NodeAttestor plugin
[RFC 5280 Section 4.1.2.2](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.2) states that CAs MUST assign and set a unique, positive integer value in the Serial Number of X.509 certificates they sign. It may be useful to manage `x509pop` node alias registrations by the serial number of the attestation X.509 certificate.

The `x509pop` NodeAttestor plugin already has all the information about the leaf attestation certificate and should be able to produce a new selector with the attestation certificate serial number. It also already allows templating of Agent SPIFFE IDs using the attestation certificate serial number, so it may as well be a node selector too.

rturner3:
The serial number of an X.509 certificate is supposed to be unique and may be a useful way to organize authorized workload entries.

**Which issue this PR fixes**
Fixes #4215


amartinezfayo (approved):
Looks great @rturner3, thank you!

PR #4230, opened 2023-06-06
Remove Journal logic to migrate old json format into proto format

### LINKED ISSUE #4229: Remove migrate journal code
Historically we used to have a journal file that persist only used authorities, but a [refactor](https://github.com/spiffe/spire/commit/0ce38f3d1c3ad22713549b5045d08578966ca437#diff-55147e5334575b7e2e2f9d0a397218f88a45d993340ad33f2a7abd4817ad599aR160) was introduced to persist multiple authorities into a proto struct.
A [migration code](https://github.com/spiffe/spire/blob/main/pkg/server/ca/manager/journal.go#L160) was introduced to parse json to the new format, 
after some releases that code is no longer necessary, may we remove it?

MarcosDY:
Remove journal migration from json to proto file

**Which issue this PR fixes**
fixes #4229


azdagron (approved):
\o/ this is a long time coming :)

PR #4154, opened 2023-05-13
spire-server fix run_posix_test case

zmt:
- [x] Commit conforms to CONTRIBUTING.md?

**Affected functionality**
Change expected response code for a single test case in spire-server `run_posix_test.go` (`TestCommand_Run/create_data_dir_when_config_is_loaded_and_server_stops`).

**Description of change**
I noticed a single consistently failing test case on main in my linux dev environment. This change passes ci today. I bisected my local failure to the commit associated with #3505.

I'm not sure why this test case would not fail elsewhere or otherwise escaped detection by ci. It's possible that [killServerOnStart](https://github.com/spiffe/spire/blob/main/cmd/spire-server/cli/run/run_posix_test.go#L188) could be flaky depending on system capabilities, but that's just a guess.

PR #4250, opened 2023-06-12
Fix spire server run command tests

guilhermocc:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
No functionality affected

**Description of change**
Use available ephemeral port for spire server run command test; this will prevent errors in scenarios in which the spire server port the test tries to use is already being used by another process.

**Which issue this PR fixes**
fix for https://github.com/spiffe/spire/pull/4154


cmd/spire-server/cli/run/run_posix_test.go
```diff
@@ -267,3 +274,12 @@ func newServerConfigCasesOS() []newServerConfigCase {
...
 	assert.Equal(t, c.Server.SocketPath, "/tmp/spire-server/private/api-test.sock")
 }
+
+func getAvailablePort() (string, error) {
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		return "", err
+	}
+	defer l.Close()
+	return strconv.Itoa(l.Addr().(*net.TCPAddr).Port), nil
```
rturner3, line 284:
```suggestion
	addrPort, err := netip.ParseAddrPort(l.Addr().String())
	if err != nil {
		return "", err
	}
	
	return strconv.Itoa(int(addrPort.Port())), nil
```
This is a safer way to parse the port number from the address, because `net.Listener.Addr()` returns an implementation of `net.Addr` that is not guaranteed to be `*net.TCPAddr`.

rturner3 (approved):
LGTM, thanks @guilhermocc

cmd/spire-server/cli/run/run_posix_test.go
```diff
@@ -267,3 +275,18 @@ func newServerConfigCasesOS() []newServerConfigCase {
 func testParseConfigGoodOS(t *testing.T, c *Config) {
 	assert.Equal(t, c.Server.SocketPath, "/tmp/spire-server/private/api-test.sock")
 }
+
+func getAvailablePort() (string, error) {
```
zmt, line 279: Ugh - this was classic port collision and I didn't immediately see it.

zmt, line 279: Ugh - a classic port collision. As written, this `getAvailablePort()` could potentially still race against other processes. The kernel could reallocate the port between the time the deferred `l.Close()` occurs and the time that the test code attempts to bind to the port returned from this helper.

Also nit: should mark this as a helper with `t.Helper()`.

PR #4262, opened 2023-06-15
Implement structured logging for gRPC errors on SPIRE agent

### LINKED ISSUE #2613: Structured logging of SPIRE Agent errors from gRPC requests to SPIRE Server
When SPIRE Agent encounters an error in an RPC to SPIRE Server, it logs the full gRPC error in the `error` log field, e.g. https://github.com/spiffe/spire/blob/main/pkg/agent/client/client.go#L317

gRPC errors have a specific formatted structure with a status code and message, e.g.
```
rpc error: code = PermissionDenied desc = agent "spiffe://example.org/spire/agent/foo" expected to have serial number "1234567890"; has "0987654321"
```

It would be convenient to be able to filter these gRPC error logs by status code. This should be doable if SPIRE Agent parses the error using [status.FromError](https://pkg.go.dev/google.golang.org/grpc/status#FromError) and logs the [Code](https://pkg.go.dev/google.golang.org/grpc@v1.42.0/internal/status#Status.Code) and [Message](https://pkg.go.dev/google.golang.org/grpc@v1.42.0/internal/status#Status.Message) fields as separate fields.

Neniel:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

SPIRE agent's client

**Description of change**

This pull request is intended to improve logging of gRPC errors from the SPIRE agent by separating code and message.

**Which issue this PR fixes**

fixes #2613


pkg/agent/client/client.go
rturner3, line 1: It would also be nice to get the gRPC status and message in the error logs after the call to `GetFederatedBundle()` on line 358.

rturner3 (commented):
Thanks for sending this, @Neniel! I had a couple small suggestions, but overall this is headed in the right direction.

pkg/agent/client/client.go
```diff
@@ -171,8 +171,16 @@ func (c *client) RenewSVID(ctx context.Context, csr []byte) (*X509SVID, error) {
...
 		c.release(connection)
+		errToReturn := fmt.Errorf("failed to renew agent: %w", err)
+		if s, ok := status.FromError(err); ok {
+			c.c.Log.WithFields(logrus.Fields{
+				"code":    s.Code(),
+				"message": s.Message(),
+			}).WithError(err).Error("Failed to renew agent")
+			return nil, errToReturn
+		}
 		c.c.Log.WithError(err).Error("Failed to renew agent")
```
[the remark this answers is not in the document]
Neniel: Thanks @rturner3! Fixed on [1bba9e2](https://github.com/spiffe/spire/pull/4262/commits/1bba9e28a704cfabfb86fe8134974bff79bbc10b)
I was considering creating constants for the strings `code` and `message` somewhere on the `names.go` file, which is in the `telemetry` package. Do you think it would be useful? I do, because we could reference them when handling the errors of others gRPC calls.

Also, I noticed that the following block is being repeated. Is there a package where I could create a function that set those fields to the logger? (if you think it would be useful).
```go
if s, ok := status.FromError(err); ok {
	log = log.WithFields(logrus.Fields{
		"code":    s.Code(),
		"message": s.Message(),
	})
}
```

pkg/agent/client/client.go
```diff
@@ -171,7 +171,14 @@ func (c *client) RenewSVID(ctx context.Context, csr []byte) (*X509SVID, error) {
 	})
 	if err != nil {
 		c.release(connection)
-		c.c.Log.WithError(err).Error("Failed to renew agent")
+		log := c.c.Log
+		if s, ok := status.FromError(err); ok {
```
MarcosDY, line 175: can you add a helper function to reduce code? (and use constants), something like:
```
func addErrorFields(log logrus.FieldLogger, err error) logrus.FieldLogger {
   if s, ok := status.FromError(err); ok {
        log = log.WithFields(logrus.Fields{                
		telemetry.StatusCode:    s.Code(),
		telemetry.StatusMessage: s.Message(),
	})
   }
   return log
}
```
and you can call
```
   log := addErrorFields(c.c.Log, err)
```

Neniel: Yes, that's what I asked Taylor in a previous comment but I'm not sure where to create that function. Is there a helper package or something similar that is suitable to hold this kind of stuff?

MarcosDY: add as private on the same file, since, or do you think you need to use in on different packages?

Neniel: For now it will only be used on the same package.

Neniel: Fixed on 6c8d983, let me know in case there's more suggestions :)

pkg/agent/client/client.go
Neniel: Done as part of 6c8d983

pkg/agent/client/client.go
```diff
@@ -465,3 +465,15 @@ func (c *client) newAgentClient(ctx context.Context) (agentv1.AgentClient, *node
...
+func (c *client) addErrorFields(err error) logrus.FieldLogger {
+	if s, ok := status.FromError(err); ok {
+		c.c.Log = c.c.Log.WithFields(logrus.Fields{
+			telemetry.StatusCode:    s.Code(),
+			telemetry.StatusMessage: s.Message(),
+		})
+	}
+
+	return c.c.Log.WithError(err)
+}
```
rturner3, line 479:
```suggestion
func (c *client) withErrorFields(err error) logrus.FieldLogger {
	if err == nil {
		return c.c.Log
	}
	
	logger := c.c.Log.WithError(err)
	if s, ok := status.FromError(err); ok {
		logger = c.c.Log.WithFields(logrus.Fields{
			telemetry.StatusCode:    s.Code(),
			telemetry.StatusMessage: s.Message(),
		})
	}

	return logger
}
```
We shouldn't modify the logger in the `client` struct since the added fields will continue to be emitted in subsequent log messages. We only want these fields to be emitted in the next log message, so we need to return a new `logrus.FieldLogger` reference.

A couple other suggestions:
- If we check if `err` is `nil`, then we can just call this method after every SPIRE RPC rather than having to call it for each possible error log statement, since in some cases we log different messages based on the error.
- `withErrorFields()` is a more consistent name with Logrus's API design

pkg/agent/client/client_test.go
```diff
@@ -909,3 +911,61 @@ type testClient struct {
...
+					telemetry.StatusMessage: "NotFound",
+				}).WithError(status.Error(codes.NotFound, "NotFound")),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &client{
+				c: tt.fields.c,
+			}
+			assert.Equalf(t, tt.want, c.addErrorFields(tt.args.err), "addErrorFields(%v)", tt.args.err)
```
rturner3, line 968: As `addErrorFields()` is a private method of `client`, I think it would be better to test that each public method in `client` produces the expected error logs when the RPCs those public methods call return errors. However, I acknowledge that would significantly increase the scope of this PR because there is no testing of log messages in this struct today. We do similar log assertions in other parts of the codebase, but it looks like it was missed in this package for some reason.

If you feel up for adding that test support, it would be much appreciated since there are currently uncovered branches, and it's a bit difficult to tell that the usages of this new method are working as expected. If not, I think it would be good to create a follow-up issue and link it as a comment somewhere in this file (maybe above each test that is covering error cases).

pkg/agent/client/client.go
```diff
@@ -362,9 +362,9 @@ func (c *client) fetchBundles(ctx context.Context, federatedBundles []string) ([
 		case codes.OK:
 			bundles = append(bundles, bundle)
 		case codes.NotFound:
-			c.c.Log.WithError(err).WithField(telemetry.FederatedBundle, b).Warn("Federated bundle not found")
+			c.addErrorFields(err).WithField(telemetry.FederatedBundle, b).Warn("Federated bundle not found")
```
rturner3, line 365: Related to my suggestion below, if we check for `err == nil` in this new method, we can call it once above this `switch` block, rather than for each error log in the switch cases.

pkg/agent/client/client_test.go
Neniel: I've just started working on that. Please check [29961ed](https://github.com/spiffe/spire/pull/4262/commits/29961ed199fc31639c958c19182dcdcbff42d33d) for the first tests that covers some cases. I'll be adding the remaining ones.

Neniel: Please dismiss my previous comment, I need to re-write the tests due to the changes of the ex `addErrorFields` method.

pkg/agent/client/client.go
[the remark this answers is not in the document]
rturner3: Sorry, I missed this comment before, but I agree that this is a good suggestion and think your approach for this in the PR looks good.

pkg/agent/client/client_test.go
Neniel: I've just updated the new tests at [7d46273](https://github.com/spiffe/spire/pull/4262/commits/7d46273bdf1e92fda70019c64c7f82397c043dfb)

pkg/agent/client/client.go
Neniel: It would be called once for each one of the `federatedBundles` that are being processed and use the logger that that method returns in each one of the cases, right?

rturner3: Right, if you declare the local logger with the error fields before the switch block within the for loop, it will only use that logger for the federated bundle currently being examined.

pkg/agent/client/client_test.go
```diff
@@ -241,6 +243,12 @@ func TestRenewSVID(t *testing.T) {
 			agentErr: errors.New("renew fails"),
 			err:      "failed to renew agent: renew fails",
 		},
+		{
+			name:     "call to RenewAgent fails",
```
rturner3, line 247: Can we also test that there is an error level log with the expected error fields?

pkg/agent/client/client_test.go
```diff
@@ -344,6 +352,90 @@ func TestNewX509SVIDs(t *testing.T) {
 	assertConnectionIsNotNil(t, client)
 }
 
+func TestNewX509SVIDsFails(t *testing.T) {
```
rturner3, line 355: Most of the code in this test is the same as in `TestNewX509SVIDs`. Can we re-use the existing setup logic in `TestNewX509SVIDs` and refactor it to be a table-driven test with this new test case?

pkg/agent/client/client_test.go
```diff
@@ -344,6 +352,90 @@ func TestNewX509SVIDs(t *testing.T) {
...
+	}
+
+	tc.svidClient.x509SVIDs = map[string]*types.X509SVID{
+		"entry-id": {
+			Id:        &types.SPIFFEID{TrustDomain: "example.org", Path: "/path"},
+			CertChain: [][]byte{{11, 22, 33}},
+		},
+	}
+
+	tc.svidClient.batchSVIDErr = status.Error(codes.NotFound, "not found when executing BatchNewX509SVID")
```
rturner3, line 410: Can we also test that there is an error level log with the expected error fields?

pkg/agent/client/client_test.go
```diff
@@ -696,6 +824,19 @@ func TestFetchJWTSVID(t *testing.T) {
 			},
 			err: "JWTSVID issued after it has expired",
 		},
+		{
+			name: "grpc call to NewJWTSVID fails",
+			setupTest: func(err error) {
+				tc.svidClient.jwtSVID = &types.JWTSVID{
+					Token:     "token",
+					ExpiresAt: issuedAt,
+					IssuedAt:  expiresAt,
```
rturner3, line 833: These fields look mixed up, is it intentional?

pkg/agent/client/client_test.go
```diff
@@ -696,6 +824,19 @@ func TestFetchJWTSVID(t *testing.T) {
 			},
 			err: "JWTSVID issued after it has expired",
 		},
+		{
+			name: "grpc call to NewJWTSVID fails",
```
rturner3, line 828: Can we also test that there is an error level log with the expected error fields?

pkg/agent/client/client.go
Neniel: Fixed at 55c88fa

pkg/agent/client/client.go
```diff
@@ -465,3 +466,20 @@ func (c *client) newAgentClient(ctx context.Context) (agentv1.AgentClient, *node
...
+
+// withErrorFields add fields of gRPC call status in logger
+func (c *client) withErrorFields(err error) logrus.FieldLogger {
+	if err == nil {
+		return c.c.Log
+	}
+
+	logger := c.c.Log.WithError(err)
+	if s, ok := status.FromError(err); ok {
+		logger = c.c.Log.WithFields(logrus.Fields{
```
MarcosDY, line 478: logger variable is not used to add fields, and setted with a new one, so `WithError` is missed, 
you must use logger.WithFields instead

pkg/agent/client/client_test.go
Neniel: Done at [5818be9](https://github.com/spiffe/spire/pull/4262/commits/5818be93f6e17adccd2a16d2a8a3faa4430479cd)

pkg/agent/client/client_test.go
Neniel: Thanks. I forgot to change it after copying the previous test. Fixed at 7488427c

pkg/agent/client/client.go
Neniel: Fixed at [7b9fc68](https://github.com/spiffe/spire/pull/4262/commits/7b9fc681670652ed020d29f652e02a8c8fbf530f)

rturner3:
Hey @Neniel, I see there's some commits from `main` that somehow got pulled into the branch supporting this PR. It looks like they were interleaved between the changes you're making, maybe there was a rebase that got messed up at some point?

In any case, the branch history for this PR can be cleaned up to just the changes you're proposing by running `git rebase -i` locally in your branch, and marking each commit unrelated to this PR as `drop`. Once the branch history includes only your commits on top of the `main` branch, doing a `git push -f <remote> <branch>` should clean things up here.

Also just as a suggestion for the future, running `git merge main` in the development branch is usually the best way to keep your development branch in sync with the latest. Additionally, doing a "force push" of development branches using `git push -f` makes it a bit harder for us as reviewers to see what has changed since the last time you pushed changes to the PR. Generally `git push` without `-f` is recommended (which is possible if you use `git merge` rather than `git rebase` to keep your development branch in sync).

pkg/agent/client/client_test.go
```diff
@@ -569,6 +615,41 @@ func TestFetchUpdatesReleaseConnectionIfItFails(t *testing.T) {
...
+	client, tc := createClient()
+
+	tc.entryClient.err = status.Error(codes.Internal, "call to grpc method fetchEntries has failed")
+
+	update, err := client.FetchUpdates(context.Background())
+	assert.Nil(t, update)
+	assert.Error(t, err)
+	assertConnectionIsNil(t, client)
+
+	expectedLog := log.WithFields(
```
rturner3, line 628: Ideally we would check that a log message was emitted that contained these fields rather than just checking if the logger had the fields added to it. We have methods like `AssertLogs()` in `test/spiretest` for this purpose. Since this kind of test coverage was not present in this package already, I'm ok if we defer this to a follow-up issue.

I've created #4474 to track the follow-up work.

pkg/agent/client/client_test.go
rturner3: Can be done separately to address #4474.

rturner3 (approved):

PR #4269, opened 2023-06-16
admin-endpoints integration test fixes

rturner3:
- Directly pipe output of `spire-server bundle show` to `spire-server bundle set` to avoid synchronization delays between host paths and mounts in the containers, which is known to be notoriously slow on macOS. The current approach of using an intermediate file for the bundle saved in a host path mounted by the spire-server containers causes this test to consistently fail on my Mac.
- Fix federation config, which had an incorrect endpoint address and trust domain name for one of the servers.
- Remove unnecessary command to start SPIRE Server in step 2, because step 1 already did the same thing.


azdagron (approved):
Thanks, @rturner3!

PR #4279, opened 2023-06-21
[SPIRE Agent] Add size based backoff strategy for fetchSVIDs

### LINKED ISSUE #4264: SPIRE Agent svid signing requests should support backoff retries
SPIRE Agent syncs with SPIRE Server every 5 sec and can request for upto 500 SVID signings requests. But if the previous requests was failed due to some server issue then next request will again happen in next 5 sec with upto 500 csrs.

If SPIRE Server is running at capacity then this behavior will cause snowball effect.

Backoff strategy should include both time based as well as len of csrs

prasadborole1:
- Implemented size based backoff that reduces the num of csrs by half if encountered failure and doubles on success.
- Currently synchronization backoff task has maxInterval of 2 mins (for 5 sec of sync interval). Adding Option to set maxInterval that will override default.

Size based backoff will recover in 8 steps (40 secs for 5 sec of sync interval) for current limit of 500.

Failure | num of CSRs 
-- | --
1 | 250
2 | 125
3 | 62
4 | 31
5 | 15
6 | 7
7 | 3
8 | 1

<br class="Apple-interchange-newline">

fixes #4264


pkg/agent/common/backoff/size_backoff_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func TestRequestSizeBackOff(t *testing.T) {
+	maxRequestSize := 1000
+	b := NewSizeLimitedBackOff(maxRequestSize)
+
+	// Initial backoff value should be equal to the maxRequestSize
+	assert.Equal(t, maxRequestSize, b.NextBackOff())
+
+	// After multiple successes, the backoff value should cap at maxRequestSize
+	b.Success()
```
MarcosDY, line 17: success is not tested, since backoff was max from the start

pkg/agent/manager/manager.go
```diff
@@ -22,9 +22,16 @@ import (
...
 	"github.com/spiffe/spire/pkg/common/telemetry"
 	"github.com/spiffe/spire/pkg/common/util"
+	"github.com/spiffe/spire/pkg/server/api/limits"
 	"github.com/spiffe/spire/proto/spire/common"
 )
 
+const (
+	maxSVIDSyncInterval = 4 * time.Minute
+	// for sync interval of 5 sec this will result in max of 4 mins of backoff
+	synchronizeMaxIntervalMultiple = 48
```
MarcosDY, line 32: if a user set the interval to be 1m this means we may wait 48m in the worst scenario, may we add a max value?

pkg/agent/manager/sync.go
```diff
@@ -136,8 +136,10 @@ func (m *manager) updateSVIDs(ctx context.Context, log logrus.FieldLogger, c SVI
 
 		update, err := m.fetchSVIDs(ctx, csrs)
 		if err != nil {
+			m.csrSizeLimitedBackoff.Failure()
```
MarcosDY, line 139: may we add a defer, and validate if no error returned, and based on that play with backoff?

MarcosDY (commented):
some minor comments, 
I think you will need to update documentation, at least the one related with sync interval, 
because if some user set it to a big number this backoff can result confusing

pkg/agent/common/backoff/size_backoff.go
```diff
@@ -0,0 +1,59 @@
...
+		r.currentSize = r.maxSize
+	}
+}
+
+func (r *sizeLimitedBackOff) Failure() {
+	r.currentSize /= 2
+	if r.currentSize < 1 {
+		r.currentSize = 1
+	}
+}
```
amoore877, line 45: we're applying calculations to `r.currentSize` and then making a correction based on max / min values. we *probably* don't need to worry about race conditions, but having a single assignment to the struct in the function would still at least put me more at ease

```suggestion
	newSize := r.currentSize / 2
	if newSize < 1 {
		newSize = 1
	}
	r.currentSize = newSize
}
```

pkg/agent/manager/manager.go
amoore877: should such a max be configurable, or should that be an opinionated sane hard-coded value?

pkg/agent/common/backoff/size_backoff.go
prasadborole1: done

pkg/agent/manager/sync.go
prasadborole1: moved it closer to server call

pkg/agent/manager/manager.go
prasadborole1: Hard coded the max value to 8 mins

pkg/agent/common/backoff/size_backoff_test.go
prasadborole1: done

prasadborole1:
> some minor comments, I think you will need to update documentation, at least the one related with sync interval, because if some user set it to a big number this backoff can result confusing

@MarcosDY Can you please point to docs which needs update or where to add new backoff details?

amoore877 (dismissed):
My comments are addressed. Looks like just some a couple open things from @MarcosDY needing changes still.

Thanks for this!

MarcosDY:
looks like agent sync interval was never documented https://github.com/spiffe/spire/blob/main/doc/spire_agent.md#agent-configuration-file
what do you think about adding that inside experimental table? with the default value? 
I'm not sure how much information add about this new backoff strategy but a single explanation about it will works with a backoff can be very useful for people that is debugging it

MarcosDY (dismissed):

pkg/agent/common/backoff/backoff.go
```diff
@@ -30,7 +35,23 @@ func NewBackoff(clk clock.Clock, interval time.Duration) BackOff {
...
+	return backoffOption{maxInterval: maxInterval}
+}
+
+type backoffOption struct {
+	maxInterval time.Duration
+}
+
+func (b backoffOption) applyOption(backoff *backoff.ExponentialBackOff) {
+	backoff.MaxInterval = b.maxInterval
+}
```
rturner3, line 57:
```suggestion
func (b backoffOption) applyOption(bo *backoff.ExponentialBackOff) {
	bo.MaxInterval = b.maxInterval
}
```
Using the variable name `backoff` obscures the import of the `github.com/cenkalti/backoff/v3` package, please use a different variable name to disambiguate

pkg/agent/common/backoff/size_backoff_test.go
```diff
@@ -0,0 +1,54 @@
+package backoff
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRequestSizeBackOff(t *testing.T) {
```
rturner3, line 9: nit: I see at least three test cases in this test method:
1. Verify the maxRequestSize upper bound
2. Verify the general behavior of the request size halving on failure and doubling on success
3. Verify the minRequestSize lower bound

Could we organize this test into sub-tests to cover these verifications separately? That will make the test a little more readable. e.g.
```go
func TestRequestSizeBackOff(t *testing.T) {
    t.Run("max request size does not go above configured maximum", func(t *testing.T) {
        ...
    })
    t.Run("min request size does not go below 1", func(t *testing.T) {
        ...
    })
    // etc.
}
```

rturner3 (commented):
This looks good to me, just a couple small comments

pkg/agent/common/backoff/backoff.go
prasadborole1: Updated

pkg/agent/common/backoff/size_backoff_test.go
prasadborole1: Updated

rturner3 (approved):

PR #4296, opened 2023-06-28
Clamp bundle refresh interval to a maximum

### LINKED ISSUE #4297: Add the ability to configure the refresh hint of the local bundle
spire-server refreshes the trust bundle for domains it federates with based on the `spiffe_refresh_hint` in the trust bundle, or in the absence of it based on the lifetime of the X509 CAs from it. 

This is what the specifications says that the refresh hint should be used for, but it
can miss updates if it follows that advice. For example:
* You increase the number of replicas for spire-servers in your trust domain. Those new instances have no key to use so they create one, add it to the bundle and start using it for signing SVIDs. Trust domains federating with us won't learn about these new keys until the next refresh interval happens which can be hours or even days away.
* A key gets revoked and removed from the bundle. Trust domains federating with us will still accept SVIDs signed by this key until the next refresh interval.

spire-server should refersh as often as possible to learn about these unexpected changes to the trust bundle of a trust domain. Question is how often. Having this be  once every couple of minutes is usually enough to minimise issues, but it also depends a bit on what the endpoint hosting the bundle can handle so maybe it makes sense to have this configurable.

Maybe this should go even further to deprecating the refresh hint in the specification since I think most implementation run into these kinds of issues where unscheduled updates of the trust bundle happen.

* **Version**: any
* **Platform**: any
* **Subsystem**: server, federation

sorindumitru:
The refresh hint for a trust domain is useful to know when we expect to see updates to the bundle on a standard schedule, but there are situations where we get updates more often, for example when a spire-server doesn't have any available key (e.g. the first time it's started) and needs to create a new one and start using it immediately.

Because of this it's useful to clamp the refresh hint to a maxium so we can learn about those keys sooner rather than later.

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**
fixes #4297

### DIFF
--- a/pkg/common/bundleutil/refreshhint.go
+++ b/pkg/common/bundleutil/refreshhint.go
@@ -13,6 +13,10 @@ const (
 	// MinimumRefreshHint is the smallest refresh hint the client allows.
 	// Anything smaller than the minimum will be reset to the minimum.
 	MinimumRefreshHint = time.Minute
+
+	// MaximumRefreshHint is the maximum refresh hint the client allows.
+	// Anything larger than the maximum will be reset to the maximum.
+	MaximumRefreshHint = 10 * time.Minute
 )
 
 // CalculateRefreshHint is used to calculate the refresh hint for a given
@@ -50,6 +54,9 @@ func CalculateRefreshHint(bundle *spiffebundle.Bundle) time.Duration {
 func safeRefreshHint(refreshHint time.Duration) time.Duration {
 	if refreshHint < MinimumRefreshHint {
 		return MinimumRefreshHint
+	}
+	if refreshHint > MaximumRefreshHint {
+		return MaximumRefreshHint
 	}
 	return refreshHint
 }
--- a/pkg/common/bundleutil/refreshhint_test.go
+++ b/pkg/common/bundleutil/refreshhint_test.go
@@ -13,6 +13,9 @@ import (
 func TestCalculateRefreshHint(t *testing.T) {
 	trustDomain := spiffeid.RequireTrustDomainFromString("domain.test")
 	emptyBundle := spiffebundle.New(trustDomain)
+	emptyBundleWithShortRefreshHint := spiffebundle.New(trustDomain)
+	emptyBundleWithShortRefreshHint.SetRefreshHint(time.Minute)
+
 	emptyBundleWithRefreshHint := spiffebundle.New(trustDomain)
 	emptyBundleWithRefreshHint.SetRefreshHint(time.Hour * 1)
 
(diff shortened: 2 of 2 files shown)


sorindumitru:
Does this sound like the right approach here? I can fix up the tests if that's the case.

amartinezfayo (commented):
Thank you @sorindumitru for this contribution. The problem described in #4296 is certainly an issue that we want to address.

The `safeRefreshHint` function is called by `bundleutil.CalculateRefreshHint`, which is called when calculating the time for the next federated bundle update (client side) but also to calculate the refresh hint value of a SPIFFE bundle that is being served by SPIRE Server when configured to serve the local bundle exposing a SPIFFE Bundle endpoint server (server side).
The implications of this change are that 1) refresh hints defined by the servers will be effectively ignored if the value is longer than 10 minutes (and 10 minutes will be used), and 2) SPIRE servers serving the local bundle will set a refresh hint to a maximum of 10 minutes.
While this can improve the situation described in #4297, it has the side effect of not honoring the refresh hints defined by SPIFFE Bundle servers (being served by a SPIRE Server or other programs).
A server may set a specific refresh hint due to a variety of reasons, including the intention to limit the load of requests based on its resources. If clients ignore that, there is a risk of overloading servers with more requests than they are able to handle.

I think that a possible solution for this problem could be the introduction of a configurable setting in SPIRE Server to be able to set the refresh hint of the local bundle.
When configured to expose a SPIFFE Bundle endpoint server, SPIRE Server currently sets the refresh hint of the exposed local bundle to one tenth of the smallest lifetime of the X.509 authorities (in the absence of a refresh hint). This behavior could be changed with the introduction of this new setting, and use the configured value instead of that calculation when it's present. In the absence of the setting, we can eventually set the default refresh hint value to something like 5 minutes (on the server side) and have SPIRE refreshing foreign bundles every 5 minutes in the absence of a refresh hint in the foreign bundle. But I don't think that we can introduce the change of setting a 5 minutes default refresh hint when serving bundles immediately, because it would be an unexpected change in the behavior for users updating existing deployments of SPIRE.

A possible solution path could be:
1) Introduce a new setting under the `bundle_endpoint` setting in the `federation` section of the SPIRE Server configuration to be able to configure the refresh hint of the local bundle. Probably log a warning if this is set to a high value.
2) When SPIRE updates foreign bundles, look at the refresh hint of the foreign bundle. If it's set, refresh the bundle at the specified interval (keeping the current logic of making sure that we refresh at a safe interval not shorter than 1 minute). If it's not set, default to 5 minutes.
3) The ultimate goal is that SPIRE migrates to a 5 minutes default interval for foreign bundle refreshing across the board. To that end, we could start logging a warning when the new configurable is absent in the `bundle_endpoint` configuration, saying that it will take that default value in a future version of SPIRE. I think that we will need to wait until 1.9 to make the switch to that default.

What do you think?

sorindumitru:
That sounds reasonable and I think it should fix the issues.  I can try implementing that when I have some time, if nobody else is already planning on doing that.

amartinezfayo:
>I can try implementing that when I have some time, if nobody else is already planning on doing that.

Thank you @sorindumitru, that would be awesome.

I'm closing this PR, as we will not be merging the changes contained here. I've added a comment in #4297 with the possible solution that we can implement.

PR #4298, opened 2023-06-29
Bump generated test certificate TTLs to 1 hour

rturner3:
Tests relying on these certs have been seen to fail due to the certs expiring during the test run when run in GitHub infrastructure. Increase the cert TTL to be more accommodating to longer-running tests and test environments where there may be larger clock skew out of our control.


azdagron (approved):
\o/, thanks so much @rturner3 !

PR #4379, opened 2023-07-31
Add entry and node event tables

faisal-memon:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Affects the database model

**Description of change**
Adds the entry and node event tables from https://github.com/spiffe/spire/pull/4227

**Which issue this PR fixes**
First part to address #2182


pkg/server/datastore/sqlstore/models.go
```diff
@@ -162,6 +162,30 @@ func (FederatedTrustDomain) TableName() string {
...
+// EntryEvent holds the entry id of an entry that had an event
+type EntryEvent struct {
+	Model
+
+	EntryID string `gorm:"unique_index"`
+}
+
+// TableName gets table name for Entry Events
+func (EntryEvent) TableName() string {
+	return "entry_events"
```
MarcosDY, line 174: NIT: since entry table is called `registered_entries` may we set table name `registered_entries_events`? (similar comment to `attested_node_entries`)

edwbuck: > NIT: since entry table is called `registered_entries` may we set table name `registered_entries_events`? (similar comment to `attested_node_entries`)

It was done many commits back, which was a bit long in various places.  That led to a shortening suggestion that turned it into "events" which was then too short to be descriptive.  "entries_events" should be fine in my opinion, even if the other item is called "registered_entries" because we don't have any other kind of entries.

Personally, I'd love to see "registered" dropped from entries.  It's like saying "stored_entries" when there's no place to store "not stored entries" :)

faisal-memon: Though it is a bit long, I prefer the `registered_entries_events` as it makes it clear what its attached to.

edwbuck: Not a point that I'm against, go for it!

faisal-memon: names updated.

pkg/server/datastore/sqlstore/migration.go
```diff
@@ -188,11 +188,13 @@ import (
 // | v1.7.0  |        |                                                                           |
 // |---------|        |                                                                           |
 // | v1.7.1  |        |                                                                           |
+// |---------|--------|---------------------------------------------------------------------------|
+// | v1.7.2  | 22     | Add registered_entries_events and attested_node_entries_events tables         |
```
MarcosDY, line 192: can you solve this indentation?

faisal-memon: fixed.

MarcosDY (dismissed):
LGTM!

MarcosDY (approved):

PR #4389, opened 2023-07-31
spire-server: marshall sequence_number for trust bundles

sorindumitru:
The CLI worked ok, it was using the Marshaller from go-spiffe, but this one didn't set the sequence number to anything. We now also include the sequence number in the bundle endpoint.

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**


pkg/common/bundleutil/marshal.go
```diff
@@ -72,8 +72,15 @@ func Marshal(bundle *spiffebundle.Bundle, opts ...MarshalOption) ([]byte, error)
...
+
+	sequenceNumber, ok := bundle.SequenceNumber()
+	if !ok {
+		sequenceNumber = 0
+	}
+
 	c := &marshalConfig{
-		refreshHint: refreshHint,
+		refreshHint:    refreshHint,
+		sequenceNumber: sequenceNumber,
```
MarcosDY, line 83: can you update unit tests to validate changes?

MarcosDY (commented):
Thansk @sorindumitru!!! changes looks great (good catch)

pkg/common/bundleutil/marshal.go
sorindumitru: Good idea :), I've updated them now.

Issue #4205, opened 2023-06-01
Remove SPIRE Agent Envoy SDS v2 API in v1.8.0


rturner3:
In v1.8.0, remove the SPIRE Agent Envoy SDS v2 API and all related code/tests.

PR #4444, opened 2023-08-17
Remove SDS v2 API

rturner3:
The SDS v2 API has been removed from Envoy for several years. We cannot reasonably support it any longer, and we expect most users should no longer be using it anymore.

**Which issue this PR fixes**
#4205


MarcosDY (approved):
Thanks Ryan!!!

PR #4455, opened 2023-08-23
Introduce the `azure_key_vault` KeyManager plugin

amartinezfayo:
This PR introduces the `azure_key_vault` KeyManager plugin that was contributed by @moe-omar.
Thank you so much @moe-omar for the contribution!

On top of the original contribution (#3955), I've added commits that:
- Change the implementation to create a new key when we need to generate a new one instead of adding a new version to an existing key. Once the key is created, the old one is deleted. Since key versions can't be deleted in Key Vault, this allows us to delete rotated keys.
- Avoid disposing keys that belong to the current server.
- Rename the plugin to `azure_key_vault`.
- Fix tests.


moe-omar:
Thank you for making the improvement, @amartinezfayo !

amartinezfayo:
I've noticed that there are problems with DCO. I've closed this PR to have that fixed in a new PR.

PR #4458, opened 2023-08-25
azure key vault plugin

moe-omar:
This PR adds a server KMS plugin that uses Azure's Key Vault.  

Pull Request check list

- [x]  Commit conforms to CONTRIBUTING.md?
- [x]  Proper tests/regressions included?
- [x]  Documentation updated?

Affected functionality
No existing functionality affected. New server plugin added

Description of change
A new server keymanager plugin that uses Azure's Key Vault.  

**What/How was this new change tested?**

- Tested using all the supported key types (`rsa-2048`, `rsa-4096`, `ec-p256`, and `ec-p384`) for both JWT keys and CA keys.
- Tested key rotation, refresh and keys cleanup for stale/unused keys belonging to the trust domain the server belongs to. 
- Tested authentication using Azure client/application credentials as well as using MSI in an Azure VM.
- This involved running the server, registering agents and workloads and making sure SVID's are issued (signed by Key Vault keys) and can be validated using the bundle.
- Added unit tests.


amartinezfayo:
Thank you @moe-omar for this contribution!

I've added a commit on top of the original contribution to:
- Change the implementation to create a new key when we need to generate a new one instead of adding a new version to an existing key. Once the key is created, the old one is deleted. Since key versions can't be deleted in Key Vault, this allows us to delete rotated keys.
- Avoid disposing keys that belong to the current server.
- Rename the plugin to azure_key_vault.
- Fix tests and linter.

MarcosDY (approved):

Issue #4462, opened 2023-08-29
On-disk data dependency: add table to store CA journal data


amartinezfayo:
Add a table to hold information about prepared, active, and old X509 and JWT authorities of servers sharing the same database. This will effectively replace the information that is stored today in the journal file.

PR #4465, opened 2023-08-29
On-disk data dependency: add table to store CA journal data

amartinezfayo:
Add a table to hold information about prepared, active, and old X509 and JWT authorities of servers sharing the same database.
The information stored in this table will effectively replace the information that is stored today in the journal file.

Fixes: #4462.


azdagron (approved):
LGTM!

PR #4476, opened 2023-09-04
Add profile configuration to  bundle endpoint 

### LINKED ISSUE #2670: SPIRE bundle endpoint server should be configurable in terms of profile
When configuring federation in SPIRE Server, remote bundle endpoint servers are configured in terms of what profile they are using (i.e. `https_web` or `https_spiffe`). This is in contrast to SPIRE Server's bundle endpoint server, the configuration of which takes only address/port and an optional ACME config. If ACME is configured, the bundle endpoint server uses the `https_web` profile, and if not then it uses the `https_spiffe` profile.

This distinction isn't documented anywhere, and it is also confusing to users who now have to choose a "profile" when they didn't have to originally. Instead, SPIRE Server's bundle endpoint server should be configurable in terms of profile the same way that the client is, and the ACME configurables should simply be a part of the `https_web` profile configuration.

Propose then implement a new configuration shape for the bundle endpoint server that is more closely aligned with the client configuration and the specification terminology

MarcosDY:
Add profile configuration to  bundle endpoint 

**Which issue this PR fixes**
fixes #2670


cmd/spire-server/cli/run/run.go
```diff
@@ -654,22 +667,65 @@ func NewServerConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool
...
+			acmeConfig = configToACMEConfig(acme, dataDir)
+		}
+		return acmeConfig, nil
+
+	// For now ignore SPIFFE configuration
+	case profileConfig.HTTPSSPIFFE != nil:
+		return nil, nil
+
+	default:
+		return nil, errors.New(`no bundle endpoint profile defined; current supported profiles are "https_spiffe" and 'https_web"`)
```
evan2645, line 709: I think this would be a bit clearer if we talk about configuration explicitly. If we're hitting this, then a profile is set but it's unknown? It would be nice if we could also say what the value of the unknown profile was so the user can more easily find it. I don't think we should block on that though

```suggestion
		return nil, errors.New(`unknown bundle endpoint profile configured; current supported profiles are "https_spiffe" and 'https_web"`)
```

cmd/spire-server/cli/run/run.go
```diff
@@ -695,6 +751,23 @@ func parseBundleEndpointProfile(config federatesWithConfig) (trustDomainConfig *
 	}, nil
 }
 
+func parseSingleAstNode(node ast.Node) (string, error) {
```
evan2645, line 754: We may consider renaming this function since it has logic specifically for bundle endpoint profile handling e.g. `parseBundleEndpointProfileASTNode` or maybe something shorter 😂

doc/spire_server.md
```diff
@@ -186,14 +186,15 @@ The `federation.federates_with` section is also optional and is used to configur
...
+| address                                       | IP address where this server will listen for HTTP requests                                                                                                                                                                                                                                                  |
+| port                                          | TCP port number where this server will listen for HTTP requests                                                                                                                                                                                                                                             |
+| acme                                          | Automated Certificate Management Environment configuration section (see below)                                                                                                                                                                                                                              |
```
evan2645, line 193: Should we remove this from the docs since we're trying to deprecate it?

evan2645 (commented):
Thanks for this PR Marcos 🙌 one less point of confusion 🚤

doc/spire_server.md
```diff
@@ -159,11 +159,13 @@ server {
         bundle_endpoint {
             address = "0.0.0.0"
             port = 8443
-            acme {
-                domain_name = "example.org"
-                email = "mail@example.org"
-            }
             refresh_hint = "10m"
+            profile "https_web" {
```
evan2645, line 163: non-blocking: Sorry I missed this before, but I think we need to document this in `### Configuration options for federation.bundle_endpoint` section?

evan2645 (approved):

Issue #4459, opened 2023-08-28
Using k8s_psat attestation in sidecar model


netiperher:
We would like to start using Spire in a highly restricted multi-tenant Kubernetes cluster. In this environment it is not desirable to run the Spire agent as a DaemonSet due to it requiring elevated permissions (run as root, access to kubelet and so on).

Instead we would like to run the spire agent as a sidecar container together with the normal workload. In this model the attestation between server and agent uses the `k8s_psat` plugin, where a Spire node entry is created for each type of workload/service account. Attestation between agent and workload can use either the `k8s` or `unix` (requires `shareProcessNamespace` in pod) plugins.

A problem with the `k8s_psat` plugin in this configuration is that server creates an "agent id" with the node uuid in the path. https://github.com/spiffe/spire/blob/main/pkg/server/plugin/nodeattestor/k8spsat/psat.go#L209. So sidecar model only works when a single agent runs on the node. Would it be possible to make the "agent id" configurable, e.g. with a template approach that the `x509pop` plugin uses? For sidecar model using something like the pod uuid or pod name would work better than the node uuid.

* **Version**: All versions
* **Platform**: All platforms
* **Subsystem**: server, agent

edwbuck:
The design of SPIRE has a SPIRE Agent that manages the Workload Attestation and Certificate issuances for all Workloads within a Node.  While it might seem like the Agent could be deployed multiple times within the same node, there are issues that arise when attempting to do so.  Some of those issues are due to the architectural choices made by SPIRE in the past, and could be altered; however, some of the issues intertwine with the distribution of SPIRE Entries which control if a SPIRE Agent should respond to a request for a Workload.  

Currently, SPIRE believes an Agent is the sole distribution point of a Node.  This means that the Agent will report aspects of the Node's identity in the Node/Agent Attestation to the pool of servers.  When two Agent instances of the same Agent Attester report to the same set of SPIRE Servers, the competition of the two instances can invalidate the other Agent's identity.  

Additionally, the Agent UNIX domain socket is intended to have a single location per node, which is important for the workload to communicate with the Agent as the Workload must be configured to use the correct UNIX domain socket.  While a Kubernetes Pod provides a private filesystem space such that two Agents would not conflict, SPIRE must also support non-Kubernetes deployments where the single operating system namespace would prevent all but the first Agent from launching.

[...]

PR #4479, opened 2023-09-06
Agent manager retries sync during init

### LINKED ISSUE #2287: Cache miss in manager initialization for the 1st time agent
* **Subsystem**: server

Co-authored by @poorkenny

When an agent finishes the node attestation, it will create a manager object by calling newManager, where a synchronization will be triggered in [initialization](https://github.com/spiffe/spire/blob/main/pkg/agent/manager/manager.go#L118) and authorized entries will be fetched from server's full in-memory cache. Then the agent will sync entries from server every 5 seconds as default sync interval. 
However, if it is the 1st time node attestation for this agent, the synchronization in manager initialization will always return 0 entry, due to the cache miss in server. The agent has to wait for a while to get entries in some synchronization thereafter. It will be great if the entries can be fetched in manager initialization for the 1st time agent.

nweisenauer-sap:
Agent manager retries every 5 seconds for a total of 1 minute to fetch SVIDs from the server during initalization.

**Pull Request check list**

- [ ] Documentation updated? _not sure if this needs to be added somewhere_

**Affected functionality**
When an agent finishes the node attestation, it will create a manager object by calling newManager, where a synchronization will be triggered in the initialization and authorized entries will be fetched from the server's full in-memory cache. Then the agent will sync entries from the server every 5 seconds as default sync interval (this can be configured with the experimental config section).
However, if it is the first time this agents performs the node attestation, the synchronization in the manager's initialization will return 0 entries, due to the cache miss in the server. The agent has to wait for a while to get entries in some synchronization thereafter. 
If the sync_interval is set to a large duration, like several minutes, the agent has to wait a long time until it receives its entries and is basically useless until then.

**Description of change**
The spire-agent retries the initial synchronisation with the server every 5 seconds for up to 1 minute until it gets more than 0 SVIDs as a response.

**Which issue this PR fixes**
fixes #2287


rturner3:
Hi @nweisenauer-sap, just in case you didn't know, there is some current work in flight to reduce overall propagation delay of entries and the network traffic required to keep the agent and server in sync (#3496, #2182). Since you mentioned that the problem you're seeing is with an agent configured with a higher sync interval, I'm curious if the solutions to the issues I referenced might address any concerns you have with using the default agent sync interval of 5 seconds. Long term, we hope to remove the `sync_interval` configuration once these scale-related challenges are resolved (#4484).

It would also be helpful if you could share how much of an issue this is for you with your deployment and why you decided to change the agent's sync interval.

rturner3:
Thanks for explaining your use case a bit more and sorry for the delayed response.

One thing I would say first is that running the agent as a sidecar process is not really endorsed as a supported deployment pattern at the moment. There has been some recent preliminary discussion about possible solutions for cases when there might be limiting factors that prevent running a host agent: https://github.com/spiffe/spire/issues/4459#issuecomment-1720017161. I would encourage you to bring some of your requirements and environment limitations into that discussion to see how we can find a more well-supported path for this in the project (not sure if the issue I referenced is related to your deployment).

Going back to this PR specifically though, I have one concern with the changes proposed here. As currently implemented in this PR, the agent's API endpoints will not be available while the agent is retrying its sync due to not receiving SVIDs on init. This can lead to other issues upstream, such as health checks failing (which may result in restart or rollback of the agent) or workloads reporting extended errors while the agent is unavailable. I think it will be safer to do this retry after the agent has initialized its endpoints to avoid these kinds of problems.

In general, it should be valid for an agent to be up and healthy and not have any authorized identities cached yet. This can be the case in several scenarios, such as during bootstrap of a node/SPIRE deployment, if there are no workloads requiring identity running on a host, etc. If we move the retries to after the endpoints are started, we can still guarantee that an agent can report itself as healthy while waiting to receive a non-empty update from the server for authorized entries/SVIDs.

nweisenauer-sap:
Hi @rturner3,

 

I added a commit to address the problem of delaying the initialization of the agent. With the newest changes the agent's initialization remains unchanged, but the synchronization loop is changed to run every 5 seconds (or every "sync_interval"-seconds, whichever is lower) if the number of currently cached SVIDs in the agent is zero. This makes configuring very large sync_intervals viable and should not change the behavior of the agent when default or low sync_intervals are configured.

 

Regarding your comment against running the agent as a sidecar, while we are running in Kubernetes environments as well, the scenario that I described above (which benefits from high sync_intervals) is not a Kubernetes deployment. There is actually only one agent per "node", even though my description may have been a bit confusing. Nevertheless, for our K8s environment, we are actually interested in exploring a sidecar approach as mentioned in #4459 and will join the discussion on that one.

Cheers

pkg/agent/manager/manager.go
```diff
@@ -159,22 +161,31 @@ func (m *manager) Initialize(ctx context.Context) error {
...
-	if nodeutil.ShouldAgentReattest(err) {
+	err := m.handleSyncError(m.synchronize(ctx))
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (m *manager) handleSyncError(err error) error {
```
rturner3, line 179: I don't really see a benefit of making this a dedicated method since it's only used to handle an error from one specific function call. Would you mind moving the error handling back to where it was before?

pkg/agent/manager/manager_test.go
```diff
@@ -963,6 +963,98 @@ func TestSynchronizationWithLRUCache(t *testing.T) {
...
+
+	// initialize will generate the first attempt at fetching entries
+	if err := m.Initialize(ctx); err != nil {
+		t.Fatal(err)
+	}
+	initTime := clk.Now()
+	require.Equal(t, initTime, m.GetLastSync())
+
+	// keep clk moving so that synchronizer retries sync
+	go func(ctx context.Context) {
```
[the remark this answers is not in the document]
nweisenauer-sap: I agree with changing the test to not have races anymore and submitted a new variant in my most recent commit, however, I did not adopt the proposed solution with the sync Channel. The issue I saw with it is that it would block the production code without something to drain the channel, as the test would do. Additionally, it would not really help in getting the "runSynchronizer" function past the "select" statement.
Instead I decided to reduce complexity and just use a real time clock paired with very low defaultSyncInterval settings for the test.

pkg/agent/manager/manager.go
```diff
@@ -294,6 +297,13 @@ func (m *manager) runSynchronizer(ctx context.Context) error {
 			m.c.Log.WithError(err).Error("Synchronize failed")
 		default:
 			m.synchronizeBackoff.Reset()
+			syncInterval = m.synchronizeBackoff.NextBackOff()
```
MarcosDY, line 300: is not it going to result in never changing backoff?
syncInternval will no change when `err != nil` and  we are getting nextBackOff after reset, so syncInterval will never change

nweisenauer-sap: Good catch @MarcosDY , I missed to add a call to NextBackOff() in the event of an error.

pkg/agent/manager/manager_test.go
```diff
@@ -963,6 +964,86 @@ func TestSynchronizationWithLRUCache(t *testing.T) {
 	require.Equal(t, clk.Now(), m.GetLastSync())
 }
 
+func TestSyncRetriesWithDefaultIntervalOnZeroSVIDSReturned(t *testing.T) {
+	dir := spiretest.TempDir(t)
+	km := fakeagentkeymanager.New(t, dir)
+
+	clk := realClock.New()
```
rturner3, line 971: So as I mentioned in my previous review, we need to come up with a solution here that doesn't involve a real clock. We have run into a lot of flakiness with tests that rely on timing that is problematic for several reasons (unclear if behavior works or not, can require multiple re-runs of CI which is time-consuming and incurs unnecessary cost).

I proposed a solution for this in the past review, did you give that a try or run into any problems with it?

nweisenauer-sap: Hi @rturner3, 
yes, unfortunately the proposed solution did not work. The channel would block the production code without something to drain the channel, as the test would do. Additionally, it would not really help in getting the "runSynchronizer" function past the "select" statement.
Nevertheless, I can see the issue with using a real clock in this test and we found another way of testing our scenario by injecting a hook into the mock clock. Please take a look at the proposed test from the most recent commit.
Cheers

rturner3 (approved):
Thanks for the patience @nweisenauer-sap , this is looking good to me.

Issue #4530, opened 2023-09-27
Spire agent should error if -insecureBootstrap and -trustBundleUrl are given


mnp:
* **Version**: 
 1.7.2
* **Platform**: 
 Linux da497fc50492 5.10.124-linuxkit #1 SMP Thu Jun 30 08:19:10 UTC 2022 x86_64 GNU/Linux
* **Subsystem**: 
agent

[Docs say](https://spiffe.io/docs/latest/deploying/spire_agent/#initial-trust-bundle-configuration) only one of the three options can be set at once, but the agent will not error out with that message if you try.

Eg `spire-agent run -config /hacked.conf -joinToken ... -insecureBootstrap -trustBundleUrl https://abc.xy:8081` will carry on trying to call the URL instead of bailing early.

This noob user was confused for a while before more RTFM and realized it's an invalid usage.

mnp:
Yeah, if I get a minute.  Anyone else, feel free though.

SilvaMatteus:
Hi guys,

I was looking at the issues to contribute again after some time not being able to do so.

There is a test in https://github.com/spiffe/spire/blob/main/cmd/spire-agent/cli/run/run_test.go#L744. However this test does not catch the right error. I forced the test case to fail by setting `expectError: false`:

```
--- FAIL: TestNewAgentConfig (0.00s)
    --- FAIL: TestNewAgentConfig/insecure_bootstrap_and_trust_bundle_url_cannot_both_be_set (0.00s)
        run_test.go:935:
            	Error Trace:	/home/matteus/spire/cmd/spire-agent/cli/run/run_test.go:935
            	Error:      	Received unexpected error:
            	            	only one of trust_bundle_url or trust_bundle_path can be specified, not both
            	Test:       	TestNewAgentConfig/insecure_bootstrap_and_trust_bundle_url_cannot_both_be_set
FAIL
```

It is actually failing because of "only one of trust_bundle_url or trust_bundle_path can be specified, not both". 

I would use a function `expectErrorContains` to catch the right error, if you guys do not mind me tackling this issue.

SilvaMatteus:
@evan2645, should we close this now that #4532 is merged?

PR #4532, opened 2023-10-01
Fix: only one of trust_bundle_path, trust_bundle_url, or insecure_bootstrap can be set

SilvaMatteus:
@mnp reported in issue #4530  that it was possible to set trust_bundle_url and insecure_bootstrap in the Agent configuration. There was a test for this case. However, the test was just checking if there was an error. There was an error but not the expected one. This commit also adds expectErrorContains to the test case struct so tests can check the expected error message. Also, more tests added.

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

SPIRE Agent config validation.

**Description of change**
Now, the validation code checks when the insecure bootstrap option is used with trust bundle config options and returns proper error messages. Tests were updated.

**Which issue this PR fixes**

#4530

### DIFF
--- a/cmd/spire-agent/cli/run/run_test.go
+++ b/cmd/spire-agent/cli/run/run_test.go
@@ -30,11 +30,12 @@ type mergeInputCase struct {
 }
 
 type newAgentConfigCase struct {
-	msg         string
-	expectError bool
-	input       func(*Config)
-	logOptions  func(t *testing.T) []log.Option
-	test        func(*testing.T, *agent.Config)
+	msg                string
+	expectError        bool
+	requireErrorPrefix string
+	input              func(*Config)
+	logOptions         func(t *testing.T) []log.Option
+	test               func(*testing.T, *agent.Config)
 }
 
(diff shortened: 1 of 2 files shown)


cmd/spire-agent/cli/run/run_test.go
```diff
@@ -931,6 +992,9 @@ func TestNewAgentConfig(t *testing.T) {
 			ac, err := NewAgentConfig(input, logOpts, false)
 			if testCase.expectError {
 				require.Error(t, err)
+				if testCase.expectErrorContains != "" {
+					require.ErrorContains(t, err, testCase.expectErrorContains)
```
MarcosDY, line 996: may we use startwith?

MarcosDY (commented):
Thanks!! it looks great just some minor nits

SilvaMatteus:
Hey @MarcosDY, thank you for the review. I addressed your comments. Could you take a second look at it?

MarcosDY (approved):
Thanks!!!

PR #4534, opened 2023-10-03
[spire-agent] Added a jitter in spire agent svid renewal

stevend-uber:
Introduces a jitter to the renewal of SVIDs in spire-agent which will allow a more uniform distribution of requests to spire-server rather than thundering herds at the same time.

https://github.com/spiffe/spire/issues/4268

### DIFF
--- a/pkg/common/rotationutil/rotationutil.go
+++ b/pkg/common/rotationutil/rotationutil.go
@@ -2,6 +2,7 @@ package rotationutil
 
 import (
 	"crypto/x509"
+	"math/rand"
 	"time"
 
 	"github.com/spiffe/spire/pkg/agent/client"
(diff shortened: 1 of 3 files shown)


pkg/common/rotationutil/rotationutil_test.go
```diff
@@ -86,3 +95,13 @@ func TestJWTSVIDExpiresSoon(t *testing.T) {
 
 	assert.True(t, JWTSVIDExpiresSoon(expiredJWT, mockClk.Now()))
 }
+
+// halfLife(beginTime, expiryTime time.Time)
+func TestHalfLife(t *testing.T) {
```
amartinezfayo, line 100: I'm wondering if we really want to have this test function. Ideally, we should not need to test non-exported functions (that usually leads to fragile tests), but rather cover them through test cases of the exported function. Are there more improvements that we could do in the test ShouldRotateX509 to make sure that a proper jitter is added, within the ranges that we expect?

stevend-uber:
Thanks @amartinezfayo all comments have been addressed :)

pkg/common/rotationutil/rotationutil.go
```diff
@@ -38,6 +39,31 @@ func JWTSVIDExpired(svid *client.JWTSVID, now time.Time) bool {
...
+
+	return ttl <= jitteredHalfLife
+}
+
+// jitterDelta is a calculated delta centered to the half-life of the SVID.
+// It's to spread out the renewal of SVID rotations to avoid spiky renewal requests.
+// The jitter is calculated as ± 10% of the half-life of the SVID.
+func jitterDelta(halfLife time.Duration) time.Duration {
+	// ± 10% of the half-life
+	return time.Duration(halfLife / 10)
```
amartinezfayo, line 63:
```suggestion
	return halfLife / 10
```

amartinezfayo (commented):
Linter complains about the unnecessary conversion.

amartinezfayo (approved):
Thank you @stevend-uber!

PR #4561, opened 2023-10-14
Fix documentation for agent KeyManager "disk"

edwbuck:
Text of the documentation was correct, but the example was for the server KeyManager "disk".


rturner3 (approved):
Good catch, thanks @edwbuck!

PR #4568, opened 2023-10-17
server/plugins/azure: use the default credentials API

shashankram:
Uses the NewDefaultAzureCredential API to fetch
client credentials. This API wraps different
mechanisms to obtain credentials using a chained
token credential mechanism. By doing so, the Azure plugins are able to obtain a token using any of the supported mechanisms: env vars, MSI, workload identity, without needing separate config input for each.

This is a part of #4485 to enable obtaining API tokens using Azure workload identity.

**Pull Request check list**

- [x] Documentation updated?

### DIFF
--- a/pkg/server/plugin/keymanager/azurekeyvault/azure_key_vault.go
+++ b/pkg/server/plugin/keymanager/azurekeyvault/azure_key_vault.go
@@ -68,7 +68,7 @@ type keyEntry struct {
 type pluginHooks struct {
 	newKeyVaultClient func(creds azcore.TokenCredential, keyVaultUri string) (cloudKeyManagementService, error)
 	clk               clock.Clock
-	msiCredential     func() (azcore.TokenCredential, error)
+	fetchCredential   func() (azcore.TokenCredential, error)
 	// Used for testing only.
 	scheduleDeleteSignal chan error
 	refreshKeysSignal    chan error
@@ -117,8 +117,8 @@ func newPlugin(
 		hooks: pluginHooks{
 			newKeyVaultClient: newKeyVaultClient,
 			clk:               clock.New(),
-			msiCredential: func() (azcore.TokenCredential, error) {
-				return azidentity.NewManagedIdentityCredential(nil)
+			fetchCredential: func() (azcore.TokenCredential, error) {
+				return azidentity.NewDefaultAzureCredential(nil)
 			},
 		},
 		scheduleDelete: make(chan string, 120),
@@ -171,17 +171,20 @@ func (p *Plugin) Configure(ctx context.Context, req *configv1.ConfigureRequest) 
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "failed to create Key Vault client with client credentials: %v", err)
 		}
+
 	case config.UseMSI:
-		cred, err := p.hooks.msiCredential()
+		p.log.Warn("use_msi is deprecated and will be removed in a future release")
+		fallthrough // use default credential which attempts to fetch credentials using MSI
+
+	default:
+		cred, err := p.hooks.fetchCredential()
 		if err != nil {
-			return nil, status.Errorf(codes.Internal, "unable to get MSI credential: %v", err)
+			return nil, status.Errorf(codes.Internal, "unable to fetch client credential: %v", err)
 		}
 		client, err = p.hooks.newKeyVaultClient(cred, config.KeyVaultURI)
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "failed to create Key Vault client with MSI credential: %v", err)
 		}
-	default:
-		return nil, status.Errorf(codes.InvalidArgument, "either MSI (use_msi) must be set to true or app authentication (tenant_id, subscription_id, app_id and app_secret) must be set")
 	}
 
 	fetcher := &keyFetcher{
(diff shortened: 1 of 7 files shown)


shashankram:
@MarcosDY could you review this soon? This is blocking us from using SPIRE in Azure.

pkg/server/plugin/keymanager/azurekeyvault/azure_key_vault.go
```diff
@@ -171,17 +171,20 @@ func (p *Plugin) Configure(ctx context.Context, req *configv1.ConfigureRequest)
...
 			return nil, status.Errorf(codes.Internal, "failed to create Key Vault client with client credentials: %v", err)
 		}
+
 	case config.UseMSI:
-		cred, err := p.hooks.msiCredential()
+		p.log.Warn("use_msi is deprecated and will be removed in a future release")
+		fallthrough // use default credential which attempts to fetch credentials using MSI
+
+	default:
+		cred, err := azidentity.NewDefaultAzureCredential(nil)
```
MarcosDY, line 180: can you add a test case for this? 
and may you remove the hook? or replace it with a new one but to replace direct use of azidentityNewDefaultAzureCredential

pkg/server/plugin/nodeattestor/azuremsi/msi.go
```diff
@@ -203,11 +200,9 @@ func (p *MSIAttestorPlugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServ
 	}
 
 	var selectorValues []string
-	if tenant.client != nil {
```
MarcosDY, line 206: is it not longer possible to have a nil client?

pkg/server/plugin/nodeattestor/azuremsi/msi.go
```diff
@@ -277,26 +272,34 @@ func (p *MSIAttestorPlugin) Configure(_ context.Context, req *configv1.Configure
...
+			p.log.Warn("use_msi is deprecated and will be removed in a future release")
+			fallthrough // use default credential which attempts to fetch credentials using MSI
+
+		default:
 			instanceMetadata, err := p.hooks.fetchInstanceMetadata(http.DefaultClient)
 			if err != nil {
 				return nil, err
 			}
-			cred, err := p.hooks.msiCredential()
+			cred, err := azidentity.NewDefaultAzureCredential(
```
MarcosDY, line 285: can you add a test case for this failing?

pkg/server/plugin/nodeattestor/azuremsi/msi.go
```diff
@@ -277,26 +272,34 @@ func (p *MSIAttestorPlugin) Configure(_ context.Context, req *configv1.Configure
...
 			}
 			client, err = p.hooks.newClient(instanceMetadata.Compute.SubscriptionID, cred)
 			if err != nil {
-				return nil, status.Errorf(codes.Internal, "unable to create client with MSI credential: %v", err)
+				return nil, status.Errorf(codes.Internal, "unable to create client with default credential: %v", err)
 			}
 		}
 
 		// If credentials are not configured then selectors won't be gathered.
 		// TODO: make this an error condition in a future release
```
MarcosDY, line 300: can you remove todo?

pkg/server/plugin/nodeattestor/azuremsi/msi.go
shashankram: If the plugin's `Configure()` fails, is it even possible to call `Attest()` on it? With this change, a nil client will result in `Configure()` producing an error.

pkg/server/plugin/nodeattestor/azuremsi/msi.go
shashankram: With credential chaining happening in the SDK, errors are ignored. We would need to wrap this call behind a custom hook to test this scenario

pkg/server/plugin/nodeattestor/azuremsi/msi.go
shashankram: done

pkg/server/plugin/nodeattestor/azuremsi/msi.go
MarcosDY: you are right client is required now, so removing the if looks fine

MarcosDY:
@shashankram can you resolve DCO?

shashankram:
> @shashankram can you resolve DCO?

yes will address your comments and fix DCO

shashankram:
@MarcosDY Had to force-push to fix DCO, I addressed the comments

pkg/server/plugin/keymanager/azurekeyvault/azure_key_vault.go
shashankram: I have modified the existing test. Otherwise, it's just an API wrapper from before.

MarcosDY (dismissed):

MarcosDY:
@shashankram code looks good, 
are you planning in update plugin documentation in a separated PR? (we need at least notify that use_msi is deprecated)

shashankram:
> @shashankram code looks good, are you planning in update plugin documentation in a separated PR? (we need at least notify that use_msi is deprecated)

I'll push an update to this PR with the doc changes, thanks for the reminder!

shashankram:
@MarcosDY I updated the docs, let me know if it makes sense to you

MarcosDY (approved):

shashankram:
There seems to be a few unrelated test flakes. How do I rerun these?

PR #4580, opened 2023-10-19
Group k8s.io/* dependency updates

rturner3:
The Kubernetes client libraries prefixed by `k8s.io/*` follow the same versioning scheme. Have them be upgraded in unison to minimize overhead of merging each individual dep bump.

Also alphabetize the dependabot dependency groups.


amartinezfayo (approved):
🚀

PR #4591, opened 2023-10-21
Bump golangci-lint to v1.55.0

rturner3:
Fix code locations now affected by updated linter rules.


rturner3:
@amartinezfayo Any ideas on how to address the Windows lint failures? Not sure if https://github.com/spiffe/spire/issues/4596 is related.

amartinezfayo:
@rturner3 Yes, this is because of the use of `reflect.SliceHeader` (it's not in the `syscall` package) which is deprecated.
I filed https://github.com/spiffe/spire/issues/4603 for this. I can take care of that.

Thank you for bringing this up!

rturner3:
> @rturner3 Yes, this is because of the use of `reflect.SliceHeader` (it's not in the `syscall` package) which is deprecated. I filed #4603 for this. I can take care of that.
> 
> Thank you for bringing this up!

Great! I'm going to move this PR to draft state for now since it depends on the fix for #4603.

amartinezfayo (approved):
🚀

Issue #4341, opened 2023-07-14
Allow spire-agent to retry connection to the server indefinitely instead of crashing during initial node attestation


shashankram:
When spire-agent starts up and it can't connect to the server to perform node attestation, it crashes. To improve resiliency, it would be nice if the agent could delay node attestation until it connects, where it retries the connection to the server indefinitely vs crashing. A config option for this would make sense to not break backward compatibility.

shashankram:
Hi @evan2645, the error pertains to the agent crashing when it tries to perform the attestation upon startup and can't connect to the server.

Logs:
```
time="2023-07-18T17:46:10Z" level=warning msg="Current umask 0002 is too permissive; setting umask 0027"
time="2023-07-18T17:46:10Z" level=info msg="Starting agent with data directory: \"/var/lib/spire/data/agent\""
time="2023-07-18T17:46:10Z" level=info msg="Plugin loaded" external=false plugin_name=disk plugin_type=KeyManager subsystem_name=catalog
time="2023-07-18T17:46:10Z" level=info msg="Plugin loaded" external=false plugin_name=join_token plugin_type=NodeAttestor subsystem_name=catalog
time="2023-07-18T17:46:10Z" level=info msg="Plugin loaded" external=false plugin_name=unix plugin_type=WorkloadAttestor subsystem_name=catalog
time="2023-07-18T17:46:10Z" level=info msg="Bundle loaded" subsystem_name=attestor trust_domain_id="spiffe://example.org"
time="2023-07-18T17:46:10Z" level=debug msg="No pre-existing agent SVID found. Will perform node attestation" subsystem_name=attestor
time="2023-07-18T17:46:10Z" level=info msg="SVID is not found. Starting node attestation" subsystem_name=attestor trust_domain_id="spiffe://example.org"
time="2023-07-18T17:46:10Z" level=debug msg="Closing catalog" subsystem_name=catalog
time="2023-07-18T17:46:10Z" level=debug msg="Unloading plugin" external=false plugin_name=unix plugin_type=WorkloadAttestor subsystem_name=catalog
time="2023-07-18T17:46:10Z" level=debug msg="Plugin deinitialized" external=false plugin_name=unix plugin_type=WorkloadAttestor subsystem_name=catalog
time="2023-07-18T17:46:10Z" level=info msg="Plugin unloaded" external=false plugin_name=unix plugin_type=WorkloadAttestor subsystem_name=catalog
[...]

Issue #4359, opened 2023-07-25
SPIRE Agent: allow to configure a bootstrap retry timeout


amartinezfayo:
When SPIRE Agent starts and needs to go through the attestation process, connection to the SPIPRE Server may fail.
This is currently an unrecoverable error and the agent exits.

Provide a configurable `bootstrap_retry_timeout` that instructs the agent to retry (connecting to the server) with backoff up until the configured timeout.

SilvaMatteus:
Hi @amartinezfayo,

I would be happy to implement this, if there is no one working on it yet.

amartinezfayo:
Great, thank you @SilvaMatteus!

amartinezfayo:
Fixed in #4597, thank you @SilvaMatteus for this contribution!

PR #4597, opened 2023-10-21
Add `retry_bootstrap`: if true, the agent retries bootstrap with backoff

SilvaMatteus:
This commit introduces the `retry_bootstrap` configurable. If true, the agent retries bootstrap with backoff.

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Connection to the server for attestation and manager initialization.

**Description of change**
This commit introduces the `retry_bootstrap` configurable. If true, the agent retries bootstrap with backoff.

**Which issue this PR fixes**

This PR adresses #4359.


amartinezfayo (commented):
Thank you @SilvaMatteus for this contribution.

I don't think that these changes will make any difference during the bootstrapping of agents.
The proposed changes in this PR set a configured timeout to the context when dialing to the server for many different operations. There are sync operations that can fail (syncing of entries and agent SVID rotation), which happen at scheduled intervals, so the retry logic is already in place to handle the failures. For that reason, we set the `grpc.FailOnNonTempDialError(true)` option so we fail the connection during the initial dial on non-temporary dial errors and we just wait for the next sync. Due to that, the changes in this PR don't make any difference during the initial dialing, which is what we want to change in order to fix #4341. If we wouldn't have the `FailOnNonTempDialError` option set, we would be piling up retries in each sync. Also, note that these changes would also affect when connecting to the server when dialing the server for calling all the RPCs for Agent, Bundle, Entry and SVID APIs.

The agent can crash at startup when failing to connect to the server if 1) [needs to attest and can't attest](https://github.com/spiffe/spire/blob/main/pkg/agent/agent.go#L91-L94), or 2) [fails to synchronize entries during the initialization of the agent cache manager](https://github.com/spiffe/spire/blob/main/pkg/agent/agent.go#L98-L101).
In order to fix #4341, I think that we should scope the changes only to retry on those operations when the agent is starting.

In the case of 2), the retry logic should probably retry the [synchronize operation during initialization](https://github.com/spiffe/spire/blob/main/pkg/agent/manager/manager.go#L171).

What we are trying to improve here is the resiliency when connecting to the server on agent startup, so we should check for the error code and not retry if the error is due to a permission denied situation (which means that it could connect to the server).

Also, it's probably better to have a boolean setting that enables this retry with backoff behavior during bootstrapping, instead of a configurable timeout that could be difficult to tune. Maybe we could come up with a timeout that could work in most cases? Maybe something between one and five minutes?

What do you think?

SilvaMatteus:
> Thank you @SilvaMatteus for this contribution.
> 
> I don't think that these changes will make any difference during the bootstrapping of agents. The proposed changes in this PR set a configured timeout to the context when dialing to the server for many different operations. There are sync operations that can fail (syncing of entries and agent SVID rotation), which happen at scheduled intervals, so the retry logic is already in place to handle the failures. For that reason, we set the `grpc.FailOnNonTempDialError(true)` option so we fail the connection during the initial dial on non-temporary dial errors and we just wait for the next sync. Due to that, the changes in this PR don't make any difference during the initial dialing, which is what we want to change in order to fix #4341. If we wouldn't have the `FailOnNonTempDialError` option set, we would be piling up retries in each sync. Also, note that these changes would also affect when connecting to the server when dialing the server for calling all the RPCs for Agent, Bundle, Entry and SVID APIs.
> 
> The agent can crash at startup when failing to connect to the server if 1) [needs to attest and can't attest](https://github.com/spiffe/spire/blob/main/pkg/agent/agent.go#L91-L94), or 2) [fails to synchronize entries during the initialization of the agent cache manager](https://github.com/spiffe/spire/blob/main/pkg/agent/agent.go#L98-L101). In order to fix #4341, I think that we should scope the changes only to retry on those operations when the agent is starting.
> 
> In the case of 2), the retry logic should probably retry the [synchronize operation during initialization](https://github.com/spiffe/spire/blob/main/pkg/agent/manager/manager.go#L171).
> 
> What we are trying to improve here is the resiliency when connecting to the server on agent startup, so we should check for the error code and not retry if the error is due to a permission denied situation (which means that it could connect to the server).
> 
> Also, it's probably better to have a boolean setting that enables this retry with backoff behavior during bootstrapping, instead of a configurable timeout that could be difficult to tune. Maybe we could come up with a timeout that could work in most cases? Maybe something between one and five minutes?
> 
> What do you think?

Hi @amartinezfayo, 

I was in wrong direction. Thanks for the detailed clarification. I will update the PR soon.

go.mod
```diff
@@ -277,6 +277,7 @@ require (
 	github.com/secure-systems-lab/go-securesystemslib v0.8.0 // indirect
 	github.com/segmentio/asm v1.2.0 // indirect
 	github.com/segmentio/ksuid v1.0.4 // indirect
+	github.com/sethvargo/go-retry v0.2.4 // indirect
```
amartinezfayo, line 280: Instead of introducing a new dependency, we should use the "github.com/spiffe/spire/pkg/agent/common/backoff" package that we already have

pkg/agent/agent.go
```diff
@@ -208,7 +218,47 @@ func (a *Agent) attest(ctx context.Context, sto storage.Storage, cat catalog.Cat
 		ServerAddress:     a.c.ServerAddress,
 		NodeAttestor:      na,
 	}
-	return node_attestor.New(&config).Attest(ctx)
+
```
amartinezfayo, line 221: I would prefer to have the retry logic in the caller, and not inside this function. We want to retry this when called from a specific place, rather than always.

amartinezfayo (commented):
Hey @SilvaMatteus, thank you for updating the PR with the new approach. I left some comments.

go.mod
SilvaMatteus: I see. I will prepare a "final" version. Thank you for the comments.

pkg/agent/agent.go
```diff
@@ -89,9 +91,47 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			as, err = a.attest(ctx, sto, cat, metrics, nodeAttestor)
+			if err == nil {
+				break
+			}
+
+			switch status.Convert(err).Code() {
+			//TODO: what additional errors we do need here?
+			// https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+			case codes.InvalidArgument:
```
SilvaMatteus, line 114: @amartinezfayo , do you have any suggestions of more error codes we should check here?

pkg/agent/agent.go
```diff
@@ -234,11 +274,42 @@ func (a *Agent) newManager(ctx context.Context, sto storage.Storage, cat catalog
...
+				return nil, ctx.Err()
+			default:
+			}
+
+			err := mgr.Initialize(ctx)
+			if err == nil {
+				return mgr, nil
+			}
+
+			//TODO: what errors should make us return immediately?
```
SilvaMatteus, line 292: @amartinezfayo , in this case, should we verify for specific errors to return immediately?

pkg/agent/agent.go
```diff
@@ -89,9 +91,47 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			if err == nil {
+				break
+			}
+
+			switch status.Convert(err).Code() {
+			//TODO: what additional errors we do need here?
+			// https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+			case codes.InvalidArgument:
+				return err
+			}
```
amartinezfayo, line 116:
```suggestion
			if status.Code(err) == codes.PermissionDenied {
				return err
			}
```

pkg/agent/agent.go
```diff
@@ -89,9 +91,47 @@ func (a *Agent) Run(ctx context.Context) error {
...
+
+			nextDuration := attBackoff.NextBackOff()
+			if nextDuration == backoff.Stop {
+				return err
+			}
+
```
amartinezfayo, line 122: We need to inform that there was an error. Something like this may work:
```
			a.c.Log.WithFields(logrus.Fields{
				telemetry.Error:         err,
				telemetry.RetryInterval: nextDuration,
			}).Warn("Attestation failed")

```

pkg/agent/agent.go
amartinezfayo: I think that we should check if we know that the agent should reattest or shutdown, in which case we should not retry. We have functions to know that based on the error. Something like:

```

			if nodeutil.ShouldAgentReattest(err) || nodeutil.ShouldAgentShutdown(err) {
				return nil, err
			}
```

pkg/agent/agent.go
```diff
@@ -234,11 +274,42 @@ func (a *Agent) newManager(ctx context.Context, sto storage.Storage, cat catalog
...
+			//TODO: what errors should make us return immediately?
 
-	return mgr, nil
+			nextDuration := initBackoff.NextBackOff()
+			if nextDuration == backoff.Stop {
+				return nil, err
+			}
+
```
amartinezfayo, line 298: We should also log the error here.

amartinezfayo (commented):
Hi @SilvaMatteus, thank you for updating the PR. I left some suggestions.

pkg/agent/agent.go
```diff
@@ -234,11 +288,54 @@ func (a *Agent) newManager(ctx context.Context, sto storage.Storage, cat catalog
...
+			select {
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			case <-initBackoffClock.After(nextDuration):
+				continue
+			}
+		}
+	} else {
+		err := mgr.Initialize(ctx)
+		if err != nil {
```
amartinezfayo, line 334: nit:
```suggestion
		if err := mgr.Initialize(ctx); err != nil {
```

pkg/agent/agent.go
```diff
@@ -89,9 +98,54 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			bootstrapBackoffInterval,
+			backoff.WithMaxElapsedTime(bootstrapBackoffMaxElapsedTime),
+		)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			default:
+			}
```
amartinezfayo, line 116: Is this really needed?

pkg/agent/agent.go
```diff
@@ -234,11 +288,54 @@ func (a *Agent) newManager(ctx context.Context, sto storage.Storage, cat catalog
...
+			bootstrapBackoffInterval,
+			backoff.WithMaxElapsedTime(bootstrapBackoffMaxElapsedTime),
+		)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			default:
+			}
```
amartinezfayo, line 304: Is this really needed?

amartinezfayo (commented):
Thank you @SilvaMatteus for addressing the comments. I think that we are pretty close here. I've left a few more comments.

pkg/agent/agent.go
SilvaMatteus: Revisiting this code, I think it is not needed. I was overthinking. Thanks!

amartinezfayo (approved):
Thank you @SilvaMatteus!

PR #4617, opened 2023-10-27
Make agent fail when getting unknown authority

MarcosDY:
Make agent fails when getting unknown authority trying to get authorized entries from spire-server.
This can happen when not using an upstream authority and the server crashes. 
If the server crashes, and it is no able to load previous bundles, Server will sign new authorities. In this case, the agent will still have an agent SVID using the old bundle, and communication will be broken for ever.


pkg/agent/manager/manager.go
```diff
@@ -285,6 +285,9 @@ func (m *manager) runSynchronizer(ctx context.Context) error {
 
 		err := m.synchronize(ctx)
 		switch {
+		case nodeutil.IsUnknownAuthorityError(err):
+			m.c.Log.WithError(err).Info("Synchronize failed, non-recoverable error")
+			return fmt.Errorf("agent SVID is signed by unknow authority: %w", err)
```
evan2645, line 290: Agent SVID or Server SVID?

I think the error message we print with `%w` is going to be roughly the same as this one? Does it include any add'l information? Maybe the string should be something like `failed to sync with SPIRE Server: %w`?

evan2645 (approved):
Thank you @MarcosDY! One small nit about error message, rest looks good.

evan2645 (approved):

PR #4654, opened 2023-11-09
Improve log message while renewing SVID

### LINKED ISSUE #2438: confusing logging for "Renewing X509-SVID"
Ref https://github.com/spiffe/spire/blob/04bee9741ac3b8c4b204e6367d0bd6146f78a8c7/pkg/agent/manager/sync.go#L129

Today when X509 CSRs are being sent for signing, we log the message `Renewing X509-SVID`.

There are similar "renewal" statements at:
- https://github.com/spiffe/spire/blob/04bee9741ac3b8c4b204e6367d0bd6146f78a8c7/pkg/agent/manager/sync.go#L82
- https://github.com/spiffe/spire/blob/04bee9741ac3b8c4b204e6367d0bd6146f78a8c7/pkg/agent/manager/sync.go#L125

However, this specific logic does not handle *only* SVID renewals. It also handles initial signings for registrations the Agent has just learned about. So it is confusing on Agent startup and on creation of new registrations that the Agent would log it is "renewing" the SVID.

It would be nice if either:
- there was differentiating in logging messages for if we are "renewing" and SVID or getting the initial signing
- the language was changed to encompass both renewal and initial signing

We might do a mix of these things depending on the log message location in the logic. I think it is handy for operators to have as granular of information as possible (the differentiation option).

FedeNQ:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Renewing SVID log messages

**Description of change**
Refined log messages for SVID renewal. When the ExpiresAt time is zero, the log will now indicate "Creating x509-SVID." In other cases, it will show "Renewing x509-SVID" as before.

**Which issue this PR fixes**
fixes #2438


amartinezfayo (approved):
LGTM, thank you @FedeNQ!

PR #4670, opened 2023-11-15
Use Go from build directory in Windows integration tests instead of requiring a Go installation

amartinezfayo:
If the PATH is not properly set to use Go from the SPIRE .build directory, it either fails because Go is not found in the system, or uses the Go from the system (which is not ideal since the targeted Go version for the SPIRE release should be used).
This PR fixes that for Windows integration tests (in the Makefile). We already did this for other integration tests (#4055).


MarcosDY (approved):
LGTM!

PR #4700, opened 2023-11-28
Remove KMS requiring metadata files (closes #4375)

### LINKED ISSUE #4375: Remove dependency on key metadata files of KMS plugins
As part of the goal of removing server dependency on on-disk data, we need the KMS plugins to avoid having to store a file that  stores an auto-generated server ID, that's used to keep track of the keys managed by the plugin.
This file is referenced as the key metadata file by the KMS plugins.

keeganwitt:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Removes requirement to have KMS metadata kept in files.

**Description of change**
Adds new config options to avoid the need to use metadata files.

**Which issue this PR fixes**
#4375


pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -832,8 +836,8 @@ func parseAndValidateConfig(c string) (*Config, error) {
 		return nil, status.Error(codes.InvalidArgument, "configuration is missing a region")
 	}
 
-	if config.KeyMetadataFile == "" {
-		return nil, status.Error(codes.InvalidArgument, "configuration is missing server id file path")
+	if config.KeyMetadataFile == "" && config.KeyMetadata == "" {
```
amartinezfayo, line 839: We should also check for the case were both settings are configured, and disallow having both of them configured at the same time. It would be confusing if we allow that.

pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -131,11 +132,14 @@ func (p *Plugin) Configure(ctx context.Context, req *configv1.ConfigureRequest)
 		return nil, err
 	}
 
-	serverID, err := loadServerID(config.KeyMetadataFile)
-	if err != nil {
-		return nil, err
+	var serverID = config.KeyMetadata
+	if config.KeyMetadata == "" {
```
amartinezfayo, line 136: This logic provides precedence to the `KeyMetadata` configuration over the `KeyMetadataFile`, allowing to have both configured, which looks confusing to me. I think that we should allow to have only one of the two. Please see my comment related with that.

pkg/server/plugin/keymanager/awskms/awskms.go
amartinezfayo: I think that we should have some validation of the `KeyMetadata` value that was configured. Values containing [characters that are not valid for a KMS Alias name](https://docs.aws.amazon.com/kms/latest/APIReference/API_CreateAlias.html#API_CreateAlias_RequestSyntax) will result in errors. I think that ensuring well-formed aliases _before_ trying to create them is important because the creation of the alias happens after the creation of the key, so users having a `KeyMetada` value with invalid characters will end up having an orphan key (until is disposed through the plugin routine to dispose stale keys) if the alias cannot be created.

pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -96,6 +96,7 @@ type Config struct {
 	SecretAccessKey string `hcl:"secret_access_key" json:"secret_access_key"`
 	Region          string `hcl:"region" json:"region"`
 	KeyMetadataFile string `hcl:"key_metadata_file" json:"key_metadata_file"`
+	KeyMetadata     string `hcl:"key_metadata" json:"key_metadata"`
```
amartinezfayo, line 99: I'm not sure if `key_metadata` is really a config name that properly describes what the user needs to set here. This is actually originated by the fact that `key_metadata_file` may not be the best name either :)
To provide some context: `key_metadata_file` was chosen with the idea of pointing to a file that could have a bunch of metadata describing the key. What we have learned is that the file hasn't grown to have anything else rather than an identifier that's tied to the server.
Since we are adding a config that's related to this, we may take the opportunity to rename `key_metadata_file` and have names that better describe what the configs are for. Otherwise we would keep adding more confusion.
Maybe something like `key_identifier_file` and `key_identifier_value` are good options, or something along those lines. What do you think?
We should rename `key_metadata_file` in a backwards compatible way. This means that we can introduce the `key_identifier_file` setting while keeping `key_metadata_file` in 1.8.x, deprecate `key_metadata_file` in 1.9.0 and remove it in 1.10.

I know that this config name change may be out of scope for this PR. If you prefer to have that addressed separately, I'm happy to help with that.

What do you think?

amartinezfayo (commented):
Thank you @keeganwitt for this contribution!

I have some concerns, related with the config name, allowing to have both the `key_metadata_file` and the `key_metadata` setting at the same time, and the validation of the `key_metadata` value configured.
I've commented on the AWS plugin, but the same comments apply to the others. Please let me know what you think. Thanks again!

pkg/server/plugin/keymanager/awskms/awskms.go
keeganwitt: Do you know what's valid for Azure? GCP is a UUID which is already validated by the current code.

keeganwitt: Added checks for AWS.

pkg/server/plugin/keymanager/awskms/awskms.go
keeganwitt: Makes sense. Fixed.

pkg/server/plugin/keymanager/awskms/awskms.go
keeganwitt: Yea, it wasn't a great name, I was just following the convention we'd started. I added this change.

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
```diff
@@ -1103,8 +1117,14 @@ func parseAndValidateConfig(c string) (*Config, error) {
 		return nil, status.Error(codes.InvalidArgument, "configuration is missing the key ring")
 	}
 
-	if config.KeyMetadataFile == "" {
-		return nil, status.Error(codes.InvalidArgument, "configuration is missing server ID file path")
+	if config.KeyMetadataFile == "" && config.KeyIdentifierFile == "" && config.KeyIdentifierValue == "" {
```
amartinezfayo, line 1120: In the case of the Google Cloud Key Management Service, we leverage the use of labels to identify which keys belongs to a server. We should validate the configured `key_identifier_value` according with the requirements for the label values (63 chars length and other restrictions): https://cloud.google.com/kms/docs/creating-managing-labels#requirements

doc/plugin_server_keymanager_aws_kms.md
```diff
@@ -6,13 +6,15 @@ The `aws_kms` key manager plugin leverages the AWS Key Management Service (KMS)
...
+|----------------------|--------|---------------------------------------|--------------------------------------------------------------------------------------------|---------------------------------------------------------|
+| access_key_id        | string | see [AWS KMS Access](#aws-kms-access) | The Access Key Id used to authenticate to KMS                                              | Value of the AWS_ACCESS_KEY_ID environment variable     |
+| secret_access_key    | string | see [AWS KMS Access](#aws-kms-access) | The Secret Access Key used to authenticate to KMS                                          | Value of the AWS_SECRET_ACCESS_KEY environment variable |
+| region               | string | yes                                   | The region where the keys will be stored                                                   |                                                         |
+| key_metadata_file    | string | yes                                   | A file path location where information about generated keys will be persisted (deprecated) |                                                         |
```
amartinezfayo, line 14:
```suggestion
| key_metadata_file    | string | Required if key_identifier_value is not set | A file path location where information about generated keys will be persisted (deprecated, use key_identifier_file instead) |                                                         |
```

doc/plugin_server_keymanager_aws_kms.md
```diff
@@ -6,13 +6,15 @@ The `aws_kms` key manager plugin leverages the AWS Key Management Service (KMS)
...
+| key_identifier_file  | string | yes                                   | A file path location where information about generated keys will be persisted              |                                                         |
```
amartinezfayo, line 15:
```suggestion
| key_identifier_file  | string | Required if key_identifier_value is not set | A file path location where information about generated keys will be persisted              |                                                         |
```

doc/plugin_server_keymanager_aws_kms.md
```diff
@@ -6,13 +6,15 @@ The `aws_kms` key manager plugin leverages the AWS Key Management Service (KMS)
...
+| key_identifier_value | string | yes                                   | A static identifier for the SPIRE server instance (used instead of `key_metadata_file`)    |                                                         |
```
amartinezfayo, line 16:
```suggestion
| key_identifier_value | string | Required if key_identifier_file is not set | A static identifier for the SPIRE server instance (used instead of `key_metadata_file`)    |                                                         |
```

amartinezfayo (commented):
Thank you @keeganwitt for your patience. I went over the validations that should be applied in all the cases and left a couple of comments related with that. I also have some suggestions for the documentation (I did it only for the AWS KMS case, but applies to the rest also).
We should be ready to go after this is addressed. Thanks again!

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
keeganwitt: The current implementation requires a UUID in the file. Are we saying we'd allow non-UUIDs in that file to match the verification behavior you've linked?

https://github.com/spiffe/spire/blob/f8f6c0e46d67b99ee4c91afe80766699c4bc063b/pkg/server/plugin/keymanager/gcpkms/gcpkms.go#L1024

pkg/server/plugin/keymanager/azurekeyvault/azure_key_vault.go
```diff
@@ -685,8 +698,23 @@ func parseAndValidateConfig(c string) (*Config, error) {
 		return nil, status.Error(codes.InvalidArgument, "configuration is missing the Key Vault URI")
 	}
 
-	if config.KeyMetadataFile == "" {
-		return nil, status.Error(codes.InvalidArgument, "configuration is missing server ID file path")
+	if config.KeyIdentifierValue != "" {
+		re := regexp.MustCompile(".*[^A-z0-9/_-].*")
```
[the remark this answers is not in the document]
keeganwitt: Actually, this implementation requires a UUID in the file as well currently (https://github.com/spiffe/spire/blob/f8f6c0e46d67b99ee4c91afe80766699c4bc063b/pkg/server/plugin/keymanager/azurekeyvault/azure_key_vault.go#L704).

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
keeganwitt: I changed the value validation. But probably should consider whether the file behavior should match.

amartinezfayo: >Are we saying we'd allow non-UUIDs in that file to match the verification behavior you've linked?

No, I think that the content of the file should remain to be opaque to the user.
The user is not meant to manipulate the content of the file configured in key_identifier_file (key_metadata_file), only its path/name. I think that the behavior should remain the same in case that a key_identifier_file is configured. The logic here is "let SPIRE manage the key_identifier_value".

In the case that a key_identifier_value is configured, I personally think that having a more relaxed format there (following the restrictions imposed by the service) would be useful, so a value that's meaningful for the environment where SPIRE Server is running can be used, probably expanding environment variables. This would be particularly useful in the k8s case.

keeganwitt: OK.

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
```diff
@@ -1103,13 +1120,41 @@ func parseAndValidateConfig(c string) (*Config, error) {
...
+	if config.KeyMetadataFile != "" && config.KeyIdentifierFile != "" {
+		return nil, status.Error(codes.InvalidArgument, "configuration must not contain both 'key_identifier_file' and deprecated 'key_metadata_file'")
+	}
+	if config.KeyIdentifierValue != "" {
+		if !validateCharacters(config.KeyIdentifierValue) {
+			return nil, status.Error(codes.InvalidArgument, "Key identifier must contain only alphanumeric characters, underscores (_), and dashes (-)")
+		}
+		if !unicode.IsLetter(rune(config.KeyIdentifierValue[0])) {
+			return nil, status.Error(codes.InvalidArgument, "Key identifier must start with a letter character")
+		}
```
amartinezfayo, line 1138: This restriction is only for the key of a label, not the value. As an example, UUIDs start with a number, and that should continue to be allowed.

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
```diff
@@ -1103,13 +1120,41 @@ func parseAndValidateConfig(c string) (*Config, error) {
...
+	}
+	if (config.KeyMetadataFile != "" || config.KeyIdentifierFile != "") && config.KeyIdentifierValue != "" {
+		return nil, status.Error(codes.InvalidArgument, "configuration must not contain both server id and server id file path")
+	}
+	if config.KeyMetadataFile != "" && config.KeyIdentifierFile != "" {
+		return nil, status.Error(codes.InvalidArgument, "configuration must not contain both 'key_identifier_file' and deprecated 'key_metadata_file'")
+	}
+	if config.KeyIdentifierValue != "" {
+		if !validateCharacters(config.KeyIdentifierValue) {
+			return nil, status.Error(codes.InvalidArgument, "Key identifier must contain only alphanumeric characters, underscores (_), and dashes (-)")
```
amartinezfayo, line 1134: We should validate that only lowercase letters are used (only lowercase letters, numeric characters, underscores, and dashes).

keeganwitt: Are you saying we shouldn't allow international characters here? I thought the documents were saying they could.

amartinezfayo: >Actually I think what you were saying was I missed the requirement to be lowercase before.

Yes, I was referring to that :)

doc/plugin_server_keymanager_aws_kms.md
```diff
@@ -6,13 +6,15 @@ The `aws_kms` key manager plugin leverages the AWS Key Management Service (KMS)
...
+| access_key_id        | string | see [AWS KMS Access](#aws-kms-access) | The Access Key Id used to authenticate to KMS                                              | Value of the AWS_ACCESS_KEY_ID environment variable     |
+| secret_access_key    | string | see [AWS KMS Access](#aws-kms-access) | The Secret Access Key used to authenticate to KMS                                          | Value of the AWS_SECRET_ACCESS_KEY environment variable |
+| region               | string | yes                                   | The region where the keys will be stored                                                   |                                                         |
+| key_metadata_file    | string | Required if key_identifier_value is not set | A file path location where information about generated keys will be persisted (deprecated, use key_identifier_file instead) |                                                         |
+| key_identifier_file  | string | Required if key_identifier_value is not set | A file path location where information about generated keys will be persisted              |                                                         |
```
evan2645, line 15: Are both of these files required when `key_identifier_value` is not set?

evan2645: Apologies, I missed the deprecated notice. You may consider removing more completely e.g.

```suggestion
| key_metadata_file    | string | Deprecated. Use key_identifier_file instead |                                                         |
| key_identifier_file  | string | Required if key_identifier_value is not set | A file path location where information about generated keys will be persisted              |                                                         |
```

Or

```suggestion
| key_identifier_file  | string | Required if key_identifier_value is not set | A file path location where information about generated keys will be persisted (previously `key_metadata_file`)             |                                                         |
```

keeganwitt: `key_metadata_file` is deprecated and replaced with `key_identifier_file`. So maybe the required column for `key_metadata_file` should just say "no"?

amartinezfayo: >So maybe the required column for key_metadata_file should just say "no"?

Sounds reasonable to me.

amartinezfayo (commented):
Just a couple of final comments:
- The `gcp_kms` plugin is missing the deprecation warning log when using `key_metadata_file`.
- Could you update `server_full.conf` with the changes?

Thanks!

doc/plugin_server_keymanager_aws_kms.md
keeganwitt: Done.

keeganwitt:
> * The `gcp_kms` plugin is missing the deprecation warning log when using `key_metadata_file`.

I think I have this already? https://github.com/spiffe/spire/pull/4700/files#diff-3dc348b71fdca86ef32e199c452c68d0bc8c70438010305d256fd94fa112ef80R1135

> * Could you update `server_full.conf` with the changes?

Done.

amartinezfayo:
I was referring that the `log.Warn("'key_metadata_file' is deprecated in favor of 'key_identifier_file' and will be removed in a future version")` warning seems to be missing in the gcp_kms plugin. I can see it in the aws_kms and azure_key_vault but not in gcp_kms?

keeganwitt:
> I was referring that the `log.Warn("'key_metadata_file' is deprecated in favor of 'key_identifier_file' and will be removed in a future version")` warning seems to be missing in the gcp_kms plugin. I can see it in the aws_kms and azure_key_vault but not in gcp_kms?

Ah, gotcha. Added.

amartinezfayo (approved):
🎉 Thank you @keeganwitt!

PR #4717, opened 2023-12-07
Remove saving of bundle and svid in independent files

sorindumitru:
The json file used for storing them together has been available since 1.4 or so, so users should have upgraded by now

**Pull Request check list**

- [ ] Documentation updated?


evan2645:
@sorindumitru can we delete the files as well with this PR? In the off chance someone was looking at or otherwise relying on this file, deleting it sends a stronger signal than allowing it to hang around and go stale

evan2645 (approved):
Thank you @sorindumitru

evan2645:
@sorindumitru we're missing a DCO on your latest commit.

georgi-lozev:
Hello friends,

Just a small feedback from our side which you hopefully consider for future.

It's probably been a while since this is marked as legacy and users should not use it, but it caught us off guard and broke some internal tests.

The unexpected part for us is it's introduced in a patch version increment, while we're more prepared for similar changes to come with minor upgrades.

I know it was likely not the brightest idea from our side to rely on it and it doesn't technically violates any compatibility matrix SPIRE has documented, but just wanted to share with the community.

evan2645:
Thanks for letting us know @georgi-lozev, and sorry for the disruption.

We wondered if this might be the case somewhere, and another option on the table was to discontinue updating the file, but that route was deemed more confusing than simply removing it and allowing the user to see that the file is gone (as opposed to, e.g., assuming rotation has stopped).

While we don't plan for folks relying on "internals" in this way, it is still good to know that it does happen. In the future, sticking to the APIs present in the SDK, as well as the telemetry, is a good way to avoid breakage since we treat these interfaces as public (and thus adhering to upgrade/downgrade guarantees). We will think further about how to avoid breakage if/when people rely on such internal behaviors

Out of curiosity, what were you trying to monitor? Maybe we are missing something in our logs/API/telemetry?

georgi-lozev:
It's a high level acceptance tests that tries to put all the things end-to-end together.
We want to assert certain detailed attributes(key extensions, order of subject atts. and etc.)  of the Agent SVID, so we're sure we don't violate our certificates policy.

When we introduced this test suite I don't think any Agent API existed for this purpose, so we directly took the agent_svid.der and parsed the certificate chain.

Please let me know if there are any Agent APIs which we might have missed that return the full Agent SVID chain.
I don't thin the metrics or logs has this type of detailed information.

georgi-lozev:
Thank you both for getting back and suggestions.

Will discuss internally and see if we want to extend this API and what other implications it will add.

Don't get me wrong, I know we're in the grey area with the way we use Agent internals at the moment.

My main point is to 1) let you know there are some people using it out there and 2) to at least consider making those kind of change in major version increments, e.g. 1.9.X instead of 1.8.7 patch.

Issue #2010, opened 2020-12-09
Hardening the Agent - Rate Limiting UDS


jkevlin:
* **Version**: all
* **Platform**: all
* **Subsystem**: agent

The possibility of a DoS attack on an Agent UDS is a known issue.  Rate limiting the connections to the UDS socket would be a possible solution to preventing a DoS attack through the UDS vector.

jkevlin:
Research:
https://godoc.org/golang.org/x/time/rate
https://github.com/grpc-ecosystem/go-grpc-middleware/tree/master/ratelimit
https://stackoverflow.com/a/62932668
https://github.com/juju/ratelimit
https://github.com/uber-go/ratelimit
https://dev.to/plutov/rate-limiting-http-requests-in-go-based-on-ip-address-542g

evan2645:
Thank you for opening this @jkevlin 

We need to put some thought into how we key the ratelimiting logic. For example, if we ratelimit based on PID, then nothing stops the abuser from continually forking and hitting the API. Alternatively, we could ratelimit by user ID, however I'm not sure if this would have other implications (are there legitimate use cases that involve a large number of workloads running as the same uid on the same host?).

Since we need to apply the ratelimiting _prior_ to exercising all of our attestation logic, I think our options are limited, and may even be platform-specific (a function of what all we can get inside peertracker).

jkevlin:
I kind of assumed that the rateLimiter would be an injected dependency configured in config. Have 2 or 3 options, including no limit and all be configurable.

rturner3:
One issue we've encountered related to this is when consumers of SPIRE Agent write their integrations incorrectly to recreate clients/connections for each use or on retry. This can result in clients opening and retaining lots of file descriptors for the connections to the Agent UDS which affects overall health of SPIRE Agent and the host. It hasn't always been easy for these consumers to trace the symptom of a steady increase in open file descriptors on the host to a faulty integration with SPIRE Agent.

Returning some more descriptive RPC errors around rate limiting or overall number of connections per client might help identify these bad integrations more easily, in addition to the hardening benefits already described in this issue.

amoore877:
Re-opening as this is still a very relevant problem at least for my organization

sorindumitru:
[...]

PR #4750, opened 2023-12-21
[peertracker] Wrap the *net.UnixListener with netutil.LimitListener

zmt:
**Pull Request check list**

- [ ] Proper tests/regressions included
- [ ] Documentation updated

**Affected functionality**
Use `golang.org/x/net/netutil.LimitListener` to set a load shedding limit
calculated from the current rlimit for the SPIRE Agent to prevent file
descriptor exhaustion from too many open connections.

Addresses #4665, #2010

**Description of change**
This WIP depends on a replace directive to local files to export the
`LimitListenerConn` type from `golang.org/x/net/netutil`.

The patch there is literally just this:
```
% git diff -U0
diff --git a/netutil/listen.go b/netutil/listen.go
index f8b779e..5caf94e 100644
--- a/netutil/listen.go
+++ b/netutil/listen.go
@@ -68 +68 @@ func (l *limitListener) Accept() (net.Conn, error) {
-       return &limitListenerConn{Conn: c, release: l.release}, nil
+       return &LimitListenerConn{Conn: c, release: l.release}, nil
@@ -77 +77 @@ func (l *limitListener) Close() error {
-type limitListenerConn struct {
+type LimitListenerConn struct {
@@ -83 +83 @@ type limitListenerConn struct {
-func (l *limitListenerConn) Close() error {
+func (l *LimitListenerConn) Close() error {
```

I filed https://github.com/golang/go/issues/64830 to track the feature request.


pkg/common/peertracker/listener_posix.go
```diff
@@ -34,8 +43,24 @@ func (lf *ListenerFactory) listenUnix(network string, laddr *net.UnixAddr) (*Lis
...
+func (lf *ListenerFactory) limitListenerBelowRlimit(l net.Listener) (net.Listener, error) {
+	var (
+		concurrency   int
+		rlimit        syscall.Rlimit
+		rlimitPercent uint64 = 99
+	)
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rlimit); err != nil {
+		return l, err
+	}
+	concurrency = int(rlimit.Cur * rlimitPercent / 100)
```
zmt, line 61: As written, this is essentially proof of concept that we could potentially ratelimit/loadshed in agent, but not representative of an optimum algorithm to calculate the actual concurrency threshold.

zmt:
Summary of highlights from discussion in contributor sync today:
* better to move any ratelimit/loadshed outside the critical path between Accept and peertracker reading the metadata
    * OK to layer in a custom implementation rather than trying to stick to stdlib netutil concurrency limiter
* really difficult to auto-calculate a sensible concurrency limit
* the configurable/tunable threshold approach would also be very problematic

evan2645:
After discussion, we'll implement a new rate limiter using caller PID as the unique caller ID. Thanks very much @zmt

PR #4846, opened 2024-01-30
Log partial selectors from workload attestation on context cancellation

rturner3:
In the case when a client hangs up a connection to SPIRE Agent during workload attestation (e.g. timeout, process dies, etc.) while the agent is still discovering selectors from workload attestation plugins, it can be helpful to log whatever selectors the agent could find for debugging purposes. Today, only the pid is captured in logs in this case, which is often not enough information to trace back the caller since:
- The caller process may have already died (e.g. crashlooping on startup)
- pid alone is only useful if you can query the state of running processes on the host around the time of workload attestation (often not practical)

Logging the partial set of selectors fetched by the agent before the caller hung up can help diagnose affected workloads in case a workload attestation plugin is misbehaving or a downstream dependency required by the plugin is down/not responding with accurate information.


amartinezfayo (approved):
Thank you @rturner3 for this, LGTM!

PR #4847, opened 2024-01-30
Default to 5 minutes for refresh hint

sorindumitru:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
spire-server: undle refresh hint

**Description of change**
Default the bundle refresh hint to 5 minutes.

**Which issue this PR fixes**
Fixes: https://github.com/spiffe/spire/issues/4297


cmd/spire-server/cli/run/run.go
```diff
@@ -454,9 +454,8 @@ func NewServerConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool
 
 				sc.Federation.BundleEndpoint.RefreshHint = &refreshHint
 			} else {
-				sc.Log.Warn("Bundle endpoint refresh_hint is not set. This configuration " +
-					"will default to 5 minutes in a future release; please check if you " +
-					"need to specify it")
+				refreshHint := 5 * time.Minute
```
amartinezfayo, line 457: Thank you @sorindumitru for this, looks great!

I only have one minor suggestion:
I think that it may be worth logging the value that the refresh hint is taking.
We could probably add one more field when we log "Serving bundle endpoint" here: https://github.com/spiffe/spire/blob/main/pkg/server/endpoints/config.go#L96
Something like: `c.Log.WithField("addr", c.BundleEndpoint.Address).WithField("refresh_hint", c.BundleEndpoint.RefreshHint).Info("Serving bundle endpoint")`

sorindumitru: I've done that and a bunch of cleanup because the RefreshHint is now non-optional, so no reason to store it as a `*time.Duration`.

amartinezfayo (approved):
Thank you @sorindumitru!

PR #4852, opened 2024-01-31
LRU subscribers failed to start when no selector was provided

### LINKED ISSUE #4827: LRU cache enabled agent, healthcheck API does not respond the status if the attestor plugin returns error
* **Version**: nightly
* **Platform**: linux
* **Subsystem**: agent

In case of the agent with LRU cache enabled, healthcheck API does not return response if the workload attestation is failed.

```
> spire-agent healthcheck -verbose
Checking agent health...

// never returned response
```

My quick debug, it seems to be blocked because updates are not passed at the following codes.

https://github.com/spiffe/spire/blob/2d8555c16d42cab81783e6c70e1b4d962db6efec/pkg/agent/endpoints/workload/handler.go#L232-L242

In v1.8.7( or LRU cache disabled), even if the workload attestor returns error, healthcheck endpoint returns healthy.

Should we avoid at least a situation the response is blocked?

MarcosDY:
LRU was failing into no returning notifications when no selector was found on attestation process, 
this caused regular healthchecks to fail.
Used this opportunity to no longer add debug logs when no SVID was updated on LRU

**Which issue this PR fixes**
fixes #4827


pkg/agent/manager/cache/lru_cache_test.go
```diff
@@ -807,6 +807,68 @@ func TestNotify(t *testing.T) {
...
+
+		if _, err := u1.Bundle.GetBundleForTrustDomain(trustDomain1); err != nil {
+			subErrCh <- err
+			return
+		}
+
+		subErrCh <- nil
+	}()
+
+	// Wait until subscribe is created and got a notification
```
amartinezfayo, line 854:
```suggestion
	// Wait until subscriber is created and got a notification
```

pkg/agent/manager/cache/lru_cache_test.go
```diff
@@ -807,6 +807,68 @@ func TestNotify(t *testing.T) {
...
+		defer sub1.Finish()
+
+		u1 := <-sub1.Updates()
+		if len(u1.Identities) > 0 {
+			subErrCh <- fmt.Errorf("no identity expected, got: %d", len(u1.Identities))
+			return
+		}
+
+		if len(u1.Bundle.X509Authorities()) != 1 {
+			subErrCh <- fmt.Errorf("a single bundle is expected")
```
amartinezfayo, line 842:
```suggestion
			subErrCh <- fmt.Errorf("a single bundle is expected but got %d", len(u1.Bundle.X509Authorities()))
```

amartinezfayo (commented):
Thank you @MarcosDY for this fix! I tested it in my environment and fixed the issue.
It's also good that you removed the debug messages "Updating SVIDs..." that were being logged every 5 seconds and now is logged when it's needed only.

I have only very small suggestions.

amartinezfayo (approved):

PR #4860, opened 2024-02-01
Create event when entry is pruned

faisal-memon:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Events based cache

**Description of change**
Create an event when entries are pruned due to being expired.

**Which issue this PR fixes**
N/A


pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -3818,6 +3818,9 @@ func pruneRegistrationEntries(tx *gorm.DB, expiresBefore time.Time, logger logru
 		if err := deleteRegistrationEntrySupport(tx, entry); err != nil {
 			return err
 		}
+		if err := createRegistrationEntryEvent(tx, entry.EntryID); err != nil {
```
zmt, line 3821: Can we add coverage for actually pruning events? It looks like this test case prunes zero and expects zero entries: https://github.com/spiffe/spire/blob/main/pkg/server/datastore/sqlstore/sqlstore_test.go#L4028-L4032

faisal-memon: updated existing unit test to cover this case. lmk if it looks ok to you.

azdagron (approved):
Thanks @faisal-memon

amartinezfayo:
Hi @faisal-memon, sign-off is missing is missing in one commit:
```
Commit sha: [98db1cf](https://github.com/spiffe/spire/pull/4860/commits/98db1cfba69182ba1a928cdc94fa0c7f46d371df), Author: Faisal Memon, Committer: Faisal Memon; The sign-off is missing.
```
Could you please take a look at that?

faisal-memon:
dco fixed

pkg/server/datastore/sqlstore/sqlstore.go
zmt: It's very similar to @stevend-uber's adjustment to this unit test. It covers event creation so, yes, looks OK. Thanks for updating.

zmt (approved):

pkg/server/datastore/sqlstore/sqlstore.go
faisal-memon: Thats good. Thanks for confirming

PR #4871, opened 2024-02-06
Fix use of ARG for users in Dockerfile

jonjohnsonjr:
An ARG is scoped to a specific build stage, so these later stages end up with the user being ":" instead of "1000:1000".

https://docs.docker.com/engine/reference/builder/#scope

```
$ crane config ghcr.io/spiffe/oidc-discovery-provider:nightly | jq .config.User
":"
```


amartinezfayo:
Thank you @jonjohnsonjr for this contribution.
While I was reviewing this, I've noticed that the Dockerfile does not honor the specified ${spireuid}:${spiregid} in the USER instructions, which explains why the build fails with the proposed changes. I've filed #4903 to address that.
We will need to fix that before taking this contribution.
Thanks again!

amartinezfayo:
This change requires some additional changes in the Dockerfile and fixes in the integration tests. I've opened #4967 to address all that, so the changes in this PR are already contained there. I'll go ahead and close this one in favor of #4967.
Thank you again @jonjohnsonjr for bringing this up and make this contribution!

PR #4879, opened 2024-02-12
Upstream CA disk bugfixes: pass RawSubject and preserve Extensions

nikotih:
**Pull Request check list**

- [x] Documentation updated? - not needed

**Affected functionality**

Upstream CA or local file system upstream authority plugins: disk + aws secret. 

**Description of change**

When using upstream authority disk plugin CSR is being passed as DER format which is later on parsed with x509.ParseCertificateRequest. Its not a problem on its own, but rather x509 golang package that is bad. 

**Problem a):**
Subject RDN is wrong if you use csr.Subject. Order is not preserved and multi entry is turned into multi-valued
Background:
- https://github.com/golang/go/issues/24254
- https://github.com/golang/go/issues/40876

**To fix this**: Need to pass RawSubject to template instead of pkix.Name Subject

**Problem b):**
ParseCertificateRequest does not populate csr.ExtraExtensions, but rather puts everything into csr.Extensions
https://pkg.go.dev/crypto/x509#CertificateRequest
```
	// ExtraExtensions contains extensions to be copied, raw, into any CSR
	// marshaled by CreateCertificateRequest. Values override any extensions
	// that would otherwise be produced based on the other fields but are
	// overridden by any extensions specified in Attributes.
	//
	// The ExtraExtensions field is not populated by ParseCertificateRequest,
	// see Extensions instead.
```

**To fix this**: Need to pick csr.Extensions instead


azdagron (approved):
Thanks @nikotih! Looks good and great catch.

PR #4881, opened 2024-02-13
[delegated-identity] correctly parse and log workload spiffe_id

chiragk25:
Currently the API logs SPIFFEID directly from the protobuf string conversion, eg `trust_domain:"example.org" path:"/path"` which makes searching logs based on SPIFFEID string difficult. This change updates the log message to use the SPIFFEID string in the log field.

This change also updates the delegated identity API first update latency metric to be only emitted when the cache update has at least one identity, to improve the correctness of the metric

Tested locally and log shows full spiffe_id string -

```
DEBU[0073] Fetched X.509 SVID for delegated identity     count=1 method=SubscribeToX509SVIDs pid=34774 service=DelegatedIdentity spiffe_id="spiffe://example.org/dummy_entry" ttl=3586.178217
```


pkg/agent/api/delegatedidentity/v1/service.go
```diff
@@ -180,9 +180,14 @@ func sendX509SVIDResponse(update *cache.WorkloadUpdate, stream delegatedidentity
 	// a response has already been sent so nothing is
 	// blocked on this logic
 	for i, svid := range resp.X509Svids {
+		id, err := idutil.IDProtoString(svid.X509Svid.Id)
+		if err != nil {
+			log.WithError(err).Warn("Could not convert SPIFFEID proto to string")
+			continue
```
rturner3, line 186: Although it should never happen, we should still try to log the event in this case without the `spiffe_id` log field

pkg/agent/api/delegatedidentity/v1/service.go
```diff
@@ -180,9 +180,14 @@ func sendX509SVIDResponse(update *cache.WorkloadUpdate, stream delegatedidentity
 	// a response has already been sent so nothing is
 	// blocked on this logic
 	for i, svid := range resp.X509Svids {
+		id, err := idutil.IDProtoString(svid.X509Svid.Id)
+		if err != nil {
+			log.WithError(err).Warn("Could not convert SPIFFEID proto to string")
```
rturner3, line 185: I think we can avoid having an additional log statement in this case and just default to logging an empty string for the value of the `spiffe_id` log field. This case is anyways very unlikely to ever happen, but this message could be "spammy" if there was some reason why we couldn't parse the SPIFFE ID.

chiragk25: updated

pkg/agent/api/delegatedidentity/v1/service.go
chiragk25: updated

rturner3 (approved):

rturner3:
@chiragk25 It looks like DCO is missing on the latest commit, could you fix that up and force push?

chiragk25:
> @chiragk25 It looks like DCO is missing on the latest commit, could you fix that up and force push?

thanks, added!

rturner3 (approved):

Issue #4884, opened 2024-02-15
Fix a confusing typo in readme


Zhengneng:
It's pretty confusing in this [profiling_names](https://spiffe.io/docs/latest/deploying/spire_agent/#profiling-names) session as saying set in the `profiling_freq` . I'd guess it might be a typo, which should be `profiling_names` instead.

[Update] SPIRE server also has the same issue.

For more details, please refer this in [this slack thread](https://spiffe.slack.com/archives/C7XDP01HB/p1707870319905059).

evan2645:
Yes it is a typo, thank you for filing this. PRs welcome ❤️

zhengneng-cb:
Will file a PR later.

evan2645:
Fixed in https://github.com/spiffe/spire/pull/4890 thank you @Zhengneng!

PR #4889, opened 2024-02-16
Fixed confusing typo in readme

MeenuyD:
In the profiling configuration, it seems there was a typo where `profiling_freq` was mentioned instead of `profiling_names`. This might have caused confusion or incorrect behavior in the profiling settings.
 Corrected the typo in the configuration file from `profiling_freq` to `profiling_names`.
#4884


amartinezfayo:
Thank you @MeenuyD. This was already addressed in #4890.

Issue #4788, opened 2024-01-10
DNS/HTTP Node Attestor


kfox1111:
On bare metal nodes without TPM's, it would be very nice if using HTTP/DNS like ACME does for initial attestation could be used for bootstrapping rather then needing to ssh in (and accept an untrusted key) and using a join token. It wouldn't need to be ACME itself, but something that functions similarly.

kfox1111:
I'm currently thinking, I start with the x509pop plugin, copy it to a plugin named 'http', then modify it as follows:

For the server plugin, change its Attest function, removing the x509 cert validation bits. Then change the challenge to generate a 'token' as per https://datatracker.ietf.org/doc/html/rfc8555#section-8.3. It is returned to the agent.

The agent would then start a webserver on port 80 (default) or any configured port. (If port is != 80, something else needs to proxy on the host from 80->the chosen port). The agent would share out just "/.well-known/acme-challenge/$token" as per the acme rfc. The content would be the token

Once the webserver is started, the agent would respond to the server that its ready, along with its proposed dns name.
The Server plugin would first validate the dns name against a regex list its configured with of valid dns names it is willing to test. If it matches, it fetch the document from the agents webserver and validates the token matches. If it all passes, it generates a node identity with a selector matching the dns name attested.

aaomidi:
So the concern I have with this flow is:

Some assumptions:

- Client: Agent
- Server: The aforementioned server plugin
- Finalize: Similar to ACME's Finalize method, but with a public key and dns name
- The server is only TLS protected for this attestation, and not mTLS (yet).

The ideal scenario is:

- Agent reaches out to the server, and says give me a challenge.
- Agent gets the token from the challenge, and puts up a web resource.
- Agent reaches out to the server, acknowledging that it's ready to respond to the challenge.
- Server confirms that the token matches.
- Agent reaches out to the server with a finalize message.

Now imagine this scenario:

- Agent reaches out to the server, and says give me a challenge.
- Agent gets the token from the challenge, and puts up a web resource.
- Agent reaches out to the server, acknowledging that it's ready to respond to the challenge.
- Server confirms that the token matches.
- _Bad actor has entered the communication between the two_
- Bad actor reaches out with something like "finalize" with a public key

At this point the bad actor has successfully hijacked the issued identity.

Note: I may be making some wrong assumptions here, that may make this not really a possible attack.

aaomidi:
[...]

PR #4909, opened 2024-02-23
Add http node attestor

kfox1111:
Adds an http node attestor

Fixes: https://github.com/spiffe/spire/issues/4788

- [x] Documentation updated?


doc/plugin_agent_nodeattestor_httppop.md
```diff
@@ -0,0 +1,47 @@
...
+
+```xml
+spiffe://<trust_domain>/spire/agent/httppop/<hostname>
+```
+
+| Configuration     | Description                                                                                                                                      | Default   |
+|-------------------|--------------------------------------------------------------------------------------------------------------------------------------------------|-----------|
+| `hostname`        | Hostname to use for handshaking. If unset, it will be automatically detected.                                                                    |           |
+| `agentname`       | Name of this agent on the host. Useful if you have multilpe agents bound to different spire servers on the same host.                            | "default" |
+| `port`            | The port to listen on.                                                                                                                           | 80        |
```
evan2645, line 19: I think we should choose a random port here by default. Low port numbers require root which we don't always have. Defaulting to a static number can be error prone since the port might be in use.

doc/plugin_server_nodeattestor_httppop.md
```diff
@@ -0,0 +1,53 @@
...
+
+```xml
+spiffe://<trust_domain>/spire/agent/httppop/<hostname>
+```
+
+| Configuration           | Description                                                                                                                                               | Default                             |
+|-------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------|
+| `dns_patterns`          | A list of regular expressions to apply to the hostname being attested. If none match, attestation will fail. If unset, all hostnames are allowed.         |                                     |
+| `allow_alternate_ports` | Set to true to allow ports other then 80 to be specified by the agent and honored during the handshake. If false, ports other then 80 will be rejected.   | false                               |
+| `allow_non_root_ports`  | Set to true to allow ports >= 1024 to be used by the agents with the advertised_port                                                                      | false                               |
```
evan2645, line 19: Is there danger in enabling these? Should we just always allow it?

pkg/agent/plugin/nodeattestor/httppop/httppop.go
```diff
@@ -0,0 +1,203 @@
...
+
+	m sync.Mutex
+	c *Config
+}
+
+func New() *Plugin {
+	return &Plugin{}
+}
+
+func (p *Plugin) ServeNonce(agentName string, nonce string) (err error) {
```
evan2645, line 59: I think these functions should be unexported?

pkg/common/plugin/httppop/httppop.go
```diff
@@ -0,0 +1,100 @@
...
+
+func GenerateChallenge() (*Challenge, error) {
+	nonce, err := generateNonce()
+	if err != nil {
+		return nil, err
+	}
+	return &Challenge{Nonce: nonce}, nil
+}
+
+func CalculateResponse(challenge *Challenge) (*Response, error) {
```
evan2645, line 54: What do you envision happening here? Since the challenge is fulfilled by out of band call back from the server, I don't think we need to send anything back on the stream?

evan2645 (commented):
Thank you for this @kfox1111 and for your patience. I left a handful of high level comments/questions. I had many smaller comments that I held back, I think we're clear to move this out of draft and add tests etc whenever you have a chance

doc/plugin_agent_nodeattestor_httppop.md
```diff
@@ -0,0 +1,47 @@
+# Agent plugin: NodeAttestor "httppop"
```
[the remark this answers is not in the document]
kfox1111: That may be why they called it acme rather then anything more specific....

acme has dns mode and http mode. Both do pretty different things. This plugin is most akin to the acme http protocol and may be a little confusing to people calling it dns as it doesn't do proof of possession over dns records like acme dns.

It still may be a little bit clearer I think being httppop, as the proof of possession token is hosted out of a http server. almost all http servers use dns, so thats a bit implied?

naming things is hard. :/

doc/plugin_agent_nodeattestor_httppop.md
kfox1111: acme chose to standardize on port 80 as its most likely to make it across the internet unhurt. If we choose a random port, I think we're probably going to get a bunch of support requests asking why the plugin is broken? :/

But maybe I'm assuming something here. Do we think intranet usage will be the most common and spire-agent -> spire-server will be uncommon?

doc/plugin_server_nodeattestor_httppop.md
kfox1111: Some situations I can see (not exhaustive)

For allow_alternate_ports, if you have an internet facing spire-server, agents can make the server spend extra time/resources asking for callbacks to ports that are more likely to get blackholed I think. It could take longer for the server to decide to give up. Forcing it to be just one port controls the issue somewhat. Same with unintentially asking for an arbitrary port and have an intermediate firewall just block non port 80/443 and users wonder why the plugin is broken.

allow_non_root_ports being false adds an extra bit of security to things like NFS does. Say you have a shared unix box where multiple untrusted users can login, but only as their own users (no sudo root). They could http attest with a high port and get their own agent running on the node under their own user, when the system admin wants to use http attestation for the whole node with a root owned agent. These type nodes are common in HPC environments amongst others.

pkg/common/plugin/httppop/httppop.go
kfox1111: Mostly a copy/paste thing from other plugins that use it...

But, there does need to be a message from the agent to the server after the agent starts up the http server and hosts the token to tell the server it can now try and call it. So, there is a Response, even though its blank.

kfox1111:
@evan2645 Thanks for the review and the discussion! All good things to consider. :)

Paul-Luciano-2003 (approved):

doc/plugin_agent_nodeattestor_httppop.md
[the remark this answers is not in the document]
evan2645: > This plugin is most akin to the acme http protocol and may be a little confusing to people calling it dns as it doesn't do proof of possession over dns txt records like acme dns.

Hmm ... that is a good point. I feel it's more about reachability when a certain DNS record is used, proving that you can serve traffic for a record .. I also see what you mean by ACME prior art, we are not proving control over DNS, but proving that we can serve a DNS name. I guess I'll spend some more time thinking about it

doc/plugin_agent_nodeattestor_httppop.md
evan2645: > acme chose to standardize on port 80 as its most likely to make it across the internet unhurt

Well it's also the standard HTTP port for web facing services, and ACME traditionally fills requests for web facing services. This case feels different.

> But maybe I'm assuming something here. Do we think intranet usage will be the most common and spire-agent -> spire-server over the internet will be uncommon?

I do think internet traversal of this traffic is uncommon. We use mTLS there, which frequently has trouble across the internet (e.g. corporate and ISP TLS interception). My bet is that cost/benefit will outweigh use of port 80 - cons: root required, may already be in use ... pros: less likely to be filtered by a firewall. Agent/server traffic currently defaults to port 8081. If I can't or don't want to use port 80, I have to choose some other random static port number, which also feels funny

kfox1111: > Well it's also the standard HTTP port for web facing services, and ACME traditionally fills requests for web facing services. This case feels different.

I have seen a fair amount of use of certbot in http mode for not webservers. But, it is much more common to usefor webservers.

> > But maybe I'm assuming something here. Do we think intranet usage will be the most common and spire-agent -> spire-server over the internet will be uncommon?
> 
> I do think internet traversal of this traffic is uncommon. We use mTLS there, which frequently has trouble across the internet (e.g. corporate and ISP TLS interception). My bet is that cost/benefit will outweigh use of port 80 - cons: root required, may already be in use ... pros: less likely to be filtered by a firewall. Agent/server traffic currently defaults to port 8081. If I can't or don't want to use port 80, I have to choose some other random static port number, which also feels funny

Where I think it may get used on the internet is for things like edge computing, where you have one spire server on the internet, and then you have a spire-agent at multiple different organizations. Likely in that scenerio, spire-server would be setup on port 443 which would make it out of all the orgs that were deploying the agent, and would need port 80 made back into each organization. So, getting N number of firewall teams to let in a particular port 80 to a host might be easier then random port at N orgs.

Again, not sure how common this will be, but somehow supporting the mode of operation for those that may need to do this kind of thing seems useful.

kfox1111: I'm kind of iffy on the con list too:
> cons: root required, may already be in use

I kind of see those in some ways as extra security checks rather then cons. But I agree some may see it that way, and 
thats one of the reasons to have flags to allow_non_root_ports and allow_alternate_ports.
Maybe we should have an agent slide "use_random_port" flag too? Then it could be configured both ways.

kfox1111: How about we change up the args...
On the agent, if no port is specified, pick a random one. This still allows port 80 when desired.

On the server, allow all non root and alternate ports by default but still keep them for those that want to lock down the system further?

Paul-Luciano-2003 (changes requested):
helllo, just saying hi. 
   i have to build a better domain, pitch in for team players.

doc/plugin_agent_nodeattestor_httppop.md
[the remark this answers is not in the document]
evan2645: How about "http_challenge"

doc/plugin_agent_nodeattestor_httppop.md
evan2645: > On the agent, if no port is specified, pick a random one. This still allows port 80 when desired.
> On the server, allow all non root and alternate ports by default but still keep them for those that want to lock down the system further?

❤️ I like this much better

doc/plugin_server_nodeattestor_httppop.md
evan2645: > allow_non_root_ports being false adds an extra bit of security to things like NFS does. Say you have a shared unix box where multiple untrusted users can login, but only as their own users (no sudo root). They could http attest with a high port and get their own agent running on the node under their own user, when the system admin wants to use http attestation for the whole node with a root owned agent. These type nodes are common in HPC environments amongst others.

This is an important observation! It is the same problem as most CSP attestors , and there we work around it using TOFU. Should this attestor also have TOFU behavior for a given DNS name?

doc/plugin_agent_nodeattestor_httppop.md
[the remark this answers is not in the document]
kfox1111: Done

doc/plugin_server_nodeattestor_httppop.md
kfox1111: TOFU sounds like a good option in addition to the allow_non_root_ports feature. It adds a different, and mostly complimentary feature I think.

For bare metal nodes, its more common to need to reuse the same hostname when reinstalling the node then in the cloud I think. Needing to clear out the name out of the spire server so it can be re-provisioned can painful and run into issues with automation. It may be worth it to some users, but not others to use the TOFU option due to this.

Also thinking forward, when spire has support for multiple attestors together so that say, a tpm plugin and an http_challenge plugin are both required, it would be desirable to not TOFU but reattest so that both the tpm and http challenge are valid with regular re-attestation, TOFU wouldn't work in that environment. Unless periodic reattestation could remove TOFU registration when reattesting I guess?

pkg/agent/plugin/nodeattestor/httppop/httppop.go
kfox1111: Updated.

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,272 @@
...
+		notfound = true
+		l := re.FindAllStringSubmatch(attestationData.HostName, -1)
+		if len(l) > 0 {
+			notfound = false
+			break
+		}
+	}
+	if notfound {
+		return status.Errorf(codes.PermissionDenied, "the requested hostname is not allowed to connect")
+	}
```
azdagron, line 114: nit: I think it would be cleaner to extract this to a function. The notFound variable wouldn't be needed then (function could early-return if the hostname matches a pattern).

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,272 @@
...
+		return status.Errorf(codes.InvalidArgument, "port %d is not allowed to be used by this server", attestationData.Port)
+	}
+	if (!config.allowNonRootPorts) && attestationData.Port >= 1024 {
+		return status.Errorf(codes.InvalidArgument, "port %d is not allowed to be >= 1024", attestationData.Port)
+	}
+
+	l := config.agentNamePattern.FindAllStringSubmatch(attestationData.AgentName, -1)
+	if len(l) != 1 || len(l[0]) == 0 || len(l[0]) > 32 {
+		return status.Error(codes.InvalidArgument, "agent name is not valid")
+	}
```
azdagron, line 101: nit: can we extract this to a function, e.g. validateAgentName

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,272 @@
...
+func (p *Plugin) setConfiguration(config *configuration) {
+	p.m.Lock()
+	defer p.m.Unlock()
+	p.config = config
+}
+
+func buildSelectorValues(hostName string) []string {
+	var selectorValues []string
+
+	selectorValues = append(selectorValues, "httpchallenge:hostname:"+hostName)
```
azdagron, line 269: Selector values don't include the name of the plugin.
```suggestion
	selectorValues = append(selectorValues, "hostname:"+hostName)
```

azdagron (commented):
Thanks, @kfox1111. I'm still thinking through the security of the challenge/response but here is some preliminary feedback.

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (p *Plugin) Configure(_ context.Context, req *configv1.ConfigureRequest) (*configv1.ConfigureResponse, error) {
+	// Parse HCL config payload into config struct
+	config := new(Config)
+	if err := hcl.Decode(config, req.HclConfiguration); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "unable to decode configuration: %v", err)
+	}
+
+	// make sure the configuration produces valid data
+	if _, err := loadConfigData(config); err != nil {
```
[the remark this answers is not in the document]
kfox1111: code copied verbatim from:
https://github.com/spiffe/spire/blob/main/pkg/agent/plugin/nodeattestor/x509pop/x509pop.go#L102-L124

With the cert config bits removed.

Should we still make this change or keep it the same for consistency? I'm good either way. Just curious.

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,229 @@
...
+	data, err := p.loadConfigData()
+	if err != nil {
+		return err
+	}
+
+	port := data.port
+
+	if p.l != nil {
+		(*p.l).Close()
+	}
```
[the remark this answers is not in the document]
kfox1111: yeah.... I had it that way before we added random port support. We have to know the port to give to the server early on. The listener is now opened earlier in the process so the random port can be assigned and determined. The way its structured now, the listener is really hard to determine when its finished being used so I had to put some extra safety measures in place to deal with the edge cases.

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,229 @@
...
+		advertisedPort = l.Addr().(*net.TCPAddr).Port
+	}
+
+	attestationPayload, err := json.Marshal(httpchallenge.AttestationData{
+		HostName:  data.hostName,
+		AgentName: data.agentName,
+		Port:      advertisedPort,
+	})
+	if err != nil {
+		l.Close()
```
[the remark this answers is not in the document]
kfox1111: Due to the random port support, defer wasn't an option.

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	challenge := new(httpchallenge.Challenge)
+	if err := json.Unmarshal(resp.Challenge, challenge); err != nil {
+		l.Close()
+		return status.Errorf(codes.Internal, "unable to unmarshal challenge: %v", err)
+	}
+
+	// FIXME open http port and post nonce here. When nonce fetched, auto remove webserver.
+	err = p.serveNonce(l, data.agentName, string(challenge.Nonce))
```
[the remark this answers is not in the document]
kfox1111: oh. so, make serveNonce have access to the data about the spire communcation stream? that could work. But I was under the impression closing the listener would also unwedge things? (related to the random port listener bits)

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,229 @@
...
+func (p *Plugin) serveNonce(l net.Listener, agentName string, nonce string) (err error) {
+	h := http.NewServeMux()
+	s := &http.Server{
+		Handler:      h,
+		ReadTimeout:  10 * time.Second,
+		WriteTimeout: 10 * time.Second,
+	}
+	path := fmt.Sprintf("/.well-known/spiffe/nodeattestor/http_challenge/%s/%s", agentName, nonce)
+	h.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "%s", nonce)
```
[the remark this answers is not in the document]
kfox1111: Is that safe? nonce could potentially have formatting bits in the string? Maybe go automatically ignores that if its a single string? (my go is not great)

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,229 @@
...
+	})
+
+	go func() {
+		err = s.Serve(l)
+		l.Close()
+		if err == http.ErrServerClosed {
+			return
+		}
+		if err != nil {
+			fmt.Printf("Unexpected error while serving. %e", err)
```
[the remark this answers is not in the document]
kfox1111: kk. will try and rework that.

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,229 @@
...
+	h.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "%s", nonce)
+		go func() {
+			_ = s.Shutdown(context.Background())
+		}()
+	})
+
+	go func() {
+		err = s.Serve(l)
+		l.Close()
```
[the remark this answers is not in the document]
kfox1111: See discussion about random port

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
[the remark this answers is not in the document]
azdagron: fmt.(F)Println does not do any formatting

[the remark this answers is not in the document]
kfox1111: Doh. I missed the Println part... gotcha.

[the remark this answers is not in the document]
kfox1111: Updated.

pkg/common/plugin/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,115 @@
...
+	if strings.Contains(string(challenge.Nonce), ".") {
+		return fmt.Errorf("nonce can not contain a dot")
+	}
+	turl := url.URL{
+		Scheme: "http",
+		Host:   net.JoinHostPort(attestationData.HostName, strconv.Itoa(attestationData.Port)),
+		Path:   fmt.Sprintf("/.well-known/spiffe/nodeattestor/http_challenge/%s/%s", attestationData.AgentName, challenge.Nonce),
+	}
+
+	resp, err := http.Get(turl.String())
```
azdagron, line 77: Oh, code should use the context so that it can be cancelled (e.g http.NewRequest+req.WithContext+http.DefaultClient.Do(req))

doc/plugin_agent_nodeattestor_http_challenge.md
```diff
@@ -0,0 +1,48 @@
...
+|-------------------|--------------------------------------------------------------------------------------------------------------------------------------------------|-----------|
+| `hostname`        | Hostname to use for handshaking. If unset, it will be automatically detected.                                                                    |           |
+| `agentname`       | Name of this agent on the host. Useful if you have multilpe agents bound to different spire servers on the same host and sharing the same port.  | "default" |
+| `port`            | The port to listen on. If unspecified, a random value will be used.                                                                              | random    |
+| `advertised_port` | The port to tell the server to call back on.                                                                                                     | $port     |
+
+If `advertised_port` != `port`, you will need to setup an http proxy between the two ports. This is useful if you already run a webserver on port 80.
+
+A sample configuration:
```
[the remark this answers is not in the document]
kfox1111: Fixed an issue with it. but the intention was for the second example to be specifically for the Proxies section... I can see how that could be confusing though. Maybe include "proxy" in the example string for it?

pkg/common/plugin/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
kfox1111: fixed.

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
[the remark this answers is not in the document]
kfox1111: Chatted about this out of band. Fixed.

elinesterov:
Little feedback on this feature naming:
I just read through all the stuff here, and the name HTTP node attestor is very confusing. Basically, what we attest here is that clients do not lie about the DNS name they send to the server. We resolve the name from the server and contact the client back.
So, DNS is the authority in the current scheme. The HTTP server is just an implementation detail; it could be a gRPC server.

kfox1111:
> Little feedback on this feature naming: I just read through all the stuff here, and the name HTTP node attestor is very confusing. Basically, what we attest here is that clients do not lie about the DNS name they send to the server. We resolve the name from the server and contact the client back. So, DNS is the authority in the current scheme. The HTTP server is just an implementation detail; it could be a gRPC server.

Its currently following the nomenclature of ACME though, which a bunch of people know:
https://letsencrypt.org/docs/challenge-types/
It is very similar to the ACME HTTP-01 challenge type, fairly dissimilar to the ACME DNS-01 challenge type.

I'd expect a gRPC based challenge plugin to be named something like grpc_challenge.

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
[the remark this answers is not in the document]
kfox1111: Fixed

doc/plugin_server_nodeattestor_http_challenge.md
```diff
@@ -0,0 +1,69 @@
...
+The SPIFFE ID produced by the plugin is based on the dns name attested
+The SPIFFE ID has the form:
+
+```xml
+spiffe://<trust_domain>/spire/agent/http_challenge/<hostname>
+```
+
+| Configuration           | Description                                                                                                                                               | Default                             |
+|-------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------|
+| `dns_patterns`          | A list of regular expressions to apply to the hostname being attested. If none match, attestation will fail. If unset, all hostnames are allowed.         |                                     |
```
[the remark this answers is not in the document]
kfox1111: I took a stab at rewording. Please rereview to see if its more clear

doc/plugin_server_nodeattestor_http_challenge.md
```diff
@@ -0,0 +1,69 @@
...
+| `required_port`         | Set to a port number to require clients to listen only on that port. If unset, all port numbers are allowed                                               |                                     |
+| `allow_non_root_ports`  | Set to true to allow ports >= 1024 to be used by the agents with the advertised_port                                                                      | true                                |
+| `agent_path_template`   | A URL path portion format of Agent's SPIFFE ID. Describe in text/template format.                                                                         | "{{ .PluginName }}/{{ .HostName }}" |
```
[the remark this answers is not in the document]
azdagron: I dunno. If there isn't anything meaningful to supply the template, I'd just as soon add this later. It might never be needed.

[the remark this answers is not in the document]
kfox1111: Removed

pkg/common/plugin/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,126 @@
...
+	if strings.Contains(attestationData.AgentName, ".") {
+		return fmt.Errorf("agentname can not contain a dot")
+	}
+	turl := url.URL{
+		Scheme: "http",
+		Host:   net.JoinHostPort(attestationData.HostName, strconv.Itoa(attestationData.Port)),
+		Path:   fmt.Sprintf("/.well-known/spiffe/nodeattestor/http_challenge/%s/challenge", attestationData.AgentName),
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
azdagron, line 81: This timeout is probably better applied by the caller. And instead of using context.Background(), it should use the context from the RPC so that it respects stream cancellation.

pkg/common/plugin/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,126 @@
...
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, "GET", turl.String(), nil)
+	if err != nil {
+		return err
+	}
+
+	client := http.DefaultClient
```
azdagron, line 89: I'd suggest passing the HTTP client into the VerifyChallenge function. This would help not having to do funky things to the DefaultTransport in your unit tests.

pkg/common/plugin/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,126 @@
...
+
+	body, err := io.ReadAll(io.LimitReader(resp.Body, 64))
+	if err != nil {
+		return err
+	}
+	nonce := strings.TrimSpace(string(body))
+	if nonce == challenge.Nonce {
+		return nil
+	}
+	return fmt.Errorf("expected nonce %q but got %q", challenge.Nonce, body)
```
azdagron, line 104: nit: for readability sake I suggest inverting this check to keep the error checking/return pattern consistent
```suggestion
	if nonce != challenge.Nonce {
		return fmt.Errorf("expected nonce %q but got %q", challenge.Nonce, body)
	}
	return nil
```

pkg/common/plugin/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,126 @@
...
+
+const (
+	nonceLen = 32
+
+	// PluginName for http based attestor
+	PluginName = "http_challenge"
+)
+
+var (
+	DefaultRandReader io.Reader
```
azdagron, line 28: Is anybody setting this?

pkg/common/plugin/httpchallenge/httpchallenge.go
```diff
@@ -0,0 +1,126 @@
...
+// MakeAgentID creates an agent ID
+func MakeAgentID(td spiffeid.TrustDomain, hostName string) (spiffeid.ID, error) {
+	agentPath := fmt.Sprintf("/http_challenge/%s", hostName)
+
+	return idutil.AgentID(td, agentPath)
+}
+
+func generateNonce() (string, error) {
+	b := make([]byte, nonceLen)
+	if DefaultRandReader != nil {
```
azdagron, line 116: I might suggest you thread the "rand" reader through GenerateNonce into this function, and if nil, use rand.Read. That would give better control to callers without having to override a package-level global.

pkg/common/plugin/httpchallenge/httpchallenge_test.go
```diff
@@ -0,0 +1,98 @@
...
+			c := &Challenge {
+				Nonce: tt.testNonce,
+			}
+
+			testServer := httptest.NewServer(http.HandlerFunc(func(res http.ResponseWriter, req *http.Request) {
+				res.Write([]byte(tt.nonce))
+			}))
+			defer func() { testServer.Close() }()
+
+			http.DefaultTransport.(*http.Transport).DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
```
azdagron, line 81: I would suggest changing VerifyChallenge to take the http client instead of messing with the default transport.

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge_test.go
```diff
@@ -0,0 +1,423 @@
...
+				require.Nil(t, resp)
+				return
+			}
+			require.NoError(t, err)
+			require.NotNil(t, resp)
+		})
+	}
+}
+
+func TestAttestFailiures(t *testing.T) {
```
azdagron, line 94: Looks like a copy-pasted typo :)

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge_test.go
```diff
@@ -0,0 +1,423 @@
...
+	    }
+	    w.WriteHeader(http.StatusOK)
+	    w.Write([]byte(`MTIzNDU2Nzg5YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnc=`))
+	}))
+	defer server.Close()
+
+	u, _ := url.Parse(server.URL)
+	_, port, _ := net.SplitHostPort(u.Host)
+	oldDialContext := http.DefaultTransport.(*http.Transport).DialContext
+	dialContext := func(ctx context.Context, network, addr string) (net.Conn, error) {
```
azdagron, line 317: Why do we need this special dialing? Can the hostname be "localhost" instead of "foo"?

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge_test.go
```diff
@@ -0,0 +1,423 @@
...
+					Type:  "http_challenge",
+					Value: "hostname:foo",
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			common_httpchallenge.DefaultRandReader = strings.NewReader("123456789abcdefghijklmnopqrstuvwxyz")
```
azdagron, line 378: If possible, it's better to avoid mutating package level variables from tests. Doing so inhibits parallelizing test execution in the future.

Instead of trying to control the generated nonce, can the closure you pass in for the challenge response function update a variable that controls what is returned from the http server?

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge_test.go
```diff
@@ -0,0 +1,209 @@
...
+		serverStream                      nodeattestor.ServerStream
+	}{
+		{
+			name:         "AidAttestation fails if server does not sends a challenge",
+			config:       "",
+			expErr:       "the error",
+			serverStream: streamBuilder.FailAndBuild(errors.New("the error")),
+		},
+		{
+			name:         "AidAttestation fails if agent cannot unmarshall server challenge",
```
azdagron, line 70: typo: unmarshal

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge_test.go
```diff
@@ -0,0 +1,209 @@
...
+				return
+			}
+			require.NoError(t, err)
+		})
+	}
+}
+
+func TestAidAttestationSucceeds(t *testing.T) {
+	port := 9998
+        l, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
```
azdagron, line 94: If possible, we should ideally use an ephemeral port for this.

kfox1111: hmm... yeah. good call.

pkg/common/plugin/httpchallenge/httpchallenge.go
kfox1111: Do you mean, the value of the timeout should be passed to the function, or the whole context.WithTimeout should be lifted out, and its resulting context passed to this function?

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge_test.go
kfox1111: chatted via slack about this... Maybe resolved now. Please rereview.

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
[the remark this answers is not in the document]
kfox1111: Trying to figure out how to do this... Looking at the code, the setConfig/getConfig is about the hcl config. The loadConfigData/struct are objects derived from that config. So, I think both structs are actually needed? How might we restructure it?

doc/plugin_server_nodeattestor_http_challenge.md
```diff
@@ -0,0 +1,55 @@
+# Server plugin: NodeAttestor "http_challenge"
+
+*Must be used in conjunction with the agent-side http_challenge plugin*
+
+The `http_challenge` plugin handshakes via http to ensure the agent is running on a valid
+dns name.
+
+The SPIFFE ID produced by the plugin is based on the dns name attested
```
azdagron, line 8:
```suggestion
The SPIFFE ID produced by the plugin is based on the dns name attested.
```

pkg/common/plugin/httpchallenge/httpchallenge.go
azdagron: I mean the latter. The plugin should pass its stream context to this function. The benefit there is that if the RPC is cancelled ahead of the timeout then the plugin can fail verification immediately and return. The plugin can set the timeout on the context before passing it into this function.

pkg/common/plugin/httpchallenge/httpchallenge_test.go
```diff
@@ -0,0 +1,99 @@
...
+				if addr == "foo.bar:80" {
+					addr = strings.TrimPrefix(testServer.URL, "http://")
+				}
+				dialer := &net.Dialer{}
+				return dialer.DialContext(ctx, network, addr)
+			}
+
+			err := VerifyChallenge(http.DefaultClient, ad, c)
```
azdagron, line 90: The benefit of now passing the http client is that we don't have to mutate with the default transport above. You can do something like the following:
```
transport := http.DefaultTransport.(*http.Transport).Clone()
transport.DialContext = ...

err := VerifyChallenge{&http.Client{Transport: transport}, ad, c)
```

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge_test.go
```diff
@@ -0,0 +1,433 @@
...
+		require.Equal(t, expSel.Type, actual[idx].Type)
+		require.Equal(t, expSel.Value, actual[idx].Value)
+	}
+}
+
+func newClientWithLocalIntercept(url string) *http.Client {
+	u, _ := neturl.Parse(url)
+	_, port, _ := net.SplitHostPort(u.Host)
+	return &http.Client{
+		Transport: &http.Transport{
```
azdagron, line 423: Another option here is to clone the default transport like in my other comment and then replace the dialer.

azdagron (commented):
Looking great! Just the last few nits and I think we're ready to merge. Thanks for the hard work on this, @kfox1111!

pkg/common/plugin/httpchallenge/httpchallenge.go
kfox1111: k. I think I implemented this. but not sure if its doing the right thing with context.Background() still...

azdagron (approved):
\o/

kfox1111:
Thanks for all the help everyone! :)

PR #4951, opened 2024-03-06
Document BundlePublisher

c4rlo:
Mention the BundlePublisher plugin type, and its built-in `aws_s3` implementation, on the SPIRE Server Configuration Reference page.

This plugin type was introduced in spiffe/spire-plugin-sdk#38. See also #2909, #4034.

While I'm touching this, a couple of other tweaks:

- Add CredentialComposer to plugin type list (its `uniqueid` implementation was already mentioned in the built-in plugin table).
- Reorder the tables of plugin types and built-in plugins to hopefully make a little more sense, and to be consistent with each other.
- Change a "sql" to a "SQL".


amartinezfayo (approved):
Thank you @c4rlo!

PR #4961, opened 2024-03-09
Introduce the `gcp_cloudstorage` BundlePublisher plugin

amartinezfayo:
This PR introduces the `gcp_cloudstorage` BundlePublisher plugin, that puts the current trust bundle of the server in a designated Google Cloud Storage bucket, keeping it updated.


doc/plugin_server_bundlepublisher_gcp_cloudstorage.md
```diff
@@ -0,0 +1,54 @@
...
+## Required permissions
+
+The plugin requires the following IAM permissions be granted to the authenticated service account in the configured bucket:
+
+```text
+storage.objects.create
+storage.objects.delete
+```
+
+The `storage.objects.delete` permission is required to overwrite the object when the bundle is updated.
```
MarcosDY, line 40: is it possible to update? instead of deleting? or that is the way gcp api works?

doc/plugin_server_bundlepublisher_gcp_cloudstorage.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Sample configuration
+
+The following configuration uploads the local trust bundle contents to the `example.org` object in the `spire-trust-bundle` bucket. The AWS access key id and secret access key are obtained from the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESSKEY environment variables.
```
MarcosDY, line 44: this example is calling `spire-bundle` can you unify?

doc/plugin_server_bundlepublisher_gcp_cloudstorage.md
```diff
@@ -0,0 +1,54 @@
...
+The plugin requires the following IAM permissions be granted to the authenticated service account in the configured bucket:
+
+```text
+storage.objects.create
+storage.objects.delete
+```
+
+The `storage.objects.delete` permission is required to overwrite the object when the bundle is updated.
+
+## Sample configuration
```
MarcosDY, line 42: NIT: old plugin used to have an example using service_account_file, what do you think about adding the same here?

doc/spire_server.md
```diff
@@ -41,7 +41,8 @@ This document is a configuration reference for SPIRE Server. It includes informa
...
 | Notifier           | [gcs_bundle](/doc/plugin_server_notifier_gcs_bundle.md)              | A notifier that pushes the latest trust bundle contents into an object in Google Cloud Storage.                             |
 | Notifier           | [k8sbundle](/doc/plugin_server_notifier_k8sbundle.md)                | A notifier that pushes the latest trust bundle contents into a Kubernetes ConfigMap.                                        |
-| BundlePublisher    | [aws_s3](/doc/plugin_server_bundlepublisher_aws_s3.md)               | Publishes trust bundles to an Amazon S3 bucket.                                                                             |
+| BundlePublisher    | [aws_s3](/doc/plugin_server_bundlepublisher_aws_s3.md)               | Publishes the trust bundle to an Amazon S3 bucket.                                                                          |
+| BundlePublisher    | [gcp_cloudstorage](/doc/plugin_server_bundlepublisher_gcp_cloudstorage.md)     | Publishes the trust bundle to a Google Cloud Storage bucket.                                                      |
```
MarcosDY, line 45: can you resolve identation issues in this table?

pkg/server/plugin/bundlepublisher/gcpcloudstorage/gcpcloudstorage.go
```diff
@@ -0,0 +1,259 @@
...
+		return nil, status.Error(codes.Internal, "could not initialize storage writer")
+	}
+
+	_, err = storageWriter.Write(bundleBytes)
+	// The number of bytes written can be safely ignored. To determine if an
+	// object was successfully uploaded, we need to look at the error returned
+	// from storageWriter.Close().
+	if err != nil {
+		// Close the storage writer before returning.
+		if closeErr := storageWriter.Close(); err != nil {
```
MarcosDY, line 141: why no to use a defer after writer was created?

pkg/server/plugin/bundlepublisher/gcpcloudstorage/gcpcloudstorage_test.go
```diff
@@ -0,0 +1,427 @@
...
+	"github.com/spiffe/spire/test/plugintest"
+	"github.com/spiffe/spire/test/spiretest"
+	"github.com/spiffe/spire/test/util"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/api/option"
+	"google.golang.org/grpc/codes"
+)
+
+var (
+	writeObjectCount int
```
MarcosDY, line 26: since packages are shared (code with test) can you rename this var to something like `testWriteObjectCount`?

pkg/server/plugin/bundlepublisher/gcpcloudstorage/gcpcloudstorage_test.go
```diff
@@ -0,0 +1,427 @@
...
+		expectMsg    string
+		config       *Config
+		bundle       *types.Bundle
+		writeErr     error
+		closeErr     error
+	}{
+		{
+			name:   "success",
+			bundle: testBundle,
+			config: &Config{
```
MarcosDY, line 163: since config is  pretty much the same except test that validates that plugin was never configured, 
what do you think about adding a bool to tt, something like `noConfigure bool` and just no run configure when that is true, and make this config always default option?

pkg/server/plugin/bundlepublisher/gcpcloudstorage/gcpcloudstorage_test.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+
+	newClient := func(ctx context.Context, opts ...option.ClientOption) (gcsService, error) {
+		return &fakeClient{
+			clientOptions: opts,
+		}, nil
+	}
+	newStorageWriter := getFakeNewStorageWriterFunc(nil, nil)
+	p := newPlugin(newClient, newStorageWriter)
+	p.hooks.wroteObjectFunc = func() { writeObjectCount++ }
```
MarcosDY, line 284: it looks like this var `wroteObjectCount` is only used in this function, may we create it here?

doc/plugin_server_bundlepublisher_gcp_cloudstorage.md
amartinezfayo: This is how the Google Cloud Storage API works, as documented here: https://cloud.google.com/storage/docs/uploading-objects#roles-and-permissions
```
storage.objects.delete
This permission is only required for uploads that overwrite an existing object.
```

pkg/server/plugin/bundlepublisher/gcpcloudstorage/gcpcloudstorage.go
amartinezfayo: if the `_, err = storageWriter.Write(bundleBytes)` call returns an error, we want to first call Close() and then return the error returned by Write, so I don't think that using `defer` would help here.

MarcosDY (approved):

PR #4981, opened 2024-03-14
Change go-metrics dependency

rturner3:
github.com/armon/go-metrics was renamed to
github.com/hashicorp/go-metrics on June 8, 2023 as part of the v0.5.0 release. Update to the latest package name so we can keep getting updates.

Also upgrades the package to the latest release, v0.5.3


evan2645 (approved):
Thanks @rturner3

PR #4989, opened 2024-03-17
Use --silent for make go-bin-path invocation

sorindumitru:
Running integration tests is completely failing for me because PATH is overwritten to an invalid string (make[1] Entering ...).

### DIFF
--- a/test/integration/test-one.sh
+++ b/test/integration/test-one.sh
@@ -17,7 +17,7 @@ REPODIR=$(git rev-parse --show-toplevel)
 
 # Set and export the PATH to one that includes a go binary installed by the
 # Makefile, if necessary.
-PATH=$(cd "${REPODIR}" || exit; make go-bin-path)
+PATH=$(cd "${REPODIR}" || exit; make --silent go-bin-path)
 export PATH
 
 log-info "running \"${TESTNAME}\" test suite..."


amartinezfayo:
Thank you @sorindumitru for this contribution. I would like to understand a little bit more what's happening here.
When does the `test-one.sh` script fail for you? Is it when there is no go cached in the build cache (.build directory)?
What's the result of running `make go-bin-path` in your environment?
Would the `--silent` argument avoid exposing errors when running the script?

sorindumitru:
What was happening was that during some invocation of `make go-bin-path` I'd get something like this:
```
make[1] Entering ...
/path/to/go-bin/ # This was likely the one fetched by the SPIRE build system, bu
make[1] Exiting ...
```

That whole output was used as the path to the go binary. When it then tried to run to build something, it fails because there's no such binary as make[1]. If it helps with anything, I

The weird part was that this was only happening when i ran `make integration`. If I used the script manually it worked ok. So something in the Makefile or somewhere was changing make behaviour.

I don't think this is going to hide errors, here's a simple test I ran:
```
➜  /tmp make test
rm blablalbalba
rm: cannot remove 'blablalbalba': No such file or directory
make: *** [Makefile:4: test] Error 1
➜  /tmp make test --silent
rm: cannot remove 'blablalbalba': No such file or directory
make: *** [Makefile:4: test] Error 1
```

It just removes printing the command it runs.

sorindumitru:
Apparently GNU Make has a concept of [MAKELEVEL](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html), which tells you how deep you are into a recursive make invocation. If that is set (or set to something > 0) you get those "Entering/Leaving directory" (I don't know why, but it's something that GNU Make has decided to  do) which makes the PATH environment variable be:

```
make[1]: Entering directory '/home/sorin/github/spiffe/spire'
/home/sorin/github/spiffe/spire/.build/linux-x86_64/go/1.22.2/bin:/home/sorin/github/spiffe/spire/.build/linux-x86_64/go/1.22.2/bin:/home/sorin/.cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/home/sorin/go/bin
make[1]: Leaving directory '/home/sorin/github/spiffe/spire'
```

You can disable this either with `--silent` or with `--no-print-directory`. You can also do `unset MAKELEVEL` and you get the same effect. You can see this in practice with this small example:

```
$ cat Makefile 
.PHONY: test recursive-test

test:
	@echo test

recursive-test:
	@make test
$ make recursive-test 
make[1]: Entering directory '/tmp'
test
make[1]: Leaving directory '/tmp'
$ make test
test
$ make --version
GNU Make 4.4.1
```

amartinezfayo:
Thank you @sorindumitru for digging into this. I think that using `--no-print-directory` is probably the right fix.

sorindumitru:
While testing `--no-print-directory`, it turns out I can't reproduce the issue anymore without this patch. Not sure what was different the previous time. Maybe we  should just close this.

amartinezfayo:
Thank you @sorindumitru for looking into this. I'm closing this for now. We can reopen and use `--no-print-directory` if needed.

PR #5002, opened 2024-03-20
Remove duplicate github.com/gofrs/uuid v4 dependency

rturner3:
We already import v5. Use v5 in the one spot where v4 was being used. There is no functional change between v4 and v5, only a change to the library being distributed as a Go module.


amartinezfayo (approved):
Thank you @rturner3!

PR #5010, opened 2024-03-22
Try out github.com/christophebedard/dco-check

rturner3:
The DCO app currently used in the repository doesn't support GitHub merge queues. Try out this other DCO check script which will run as our own managed job that we can configure to run for PRs added to the merge queue.

We can try running this check in parallel to the existing check to ensure it's stable before trying to rely on it.

Tested on my fork with a couple sample PRs targeted towards this branch:

Success on DCO present: https://github.com/rturner3/spire/pull/218
Fail on missing DCO: https://github.com/rturner3/spire/pull/216


azdagron (approved):
lgtm

Issue #4963, opened 2024-03-09
Integration with AWS IAM Roles Anywhere


ajay1135:
I'm wondering whether it's worth integrating SPIRE with AWS IAM Roles Anywhere. Roles Anywhere is an AWS service that allows workloads to obtain temporary AWS credentials through PKI trust. On a very high level, you can configure a CA certificate as  root of trust to the service and then use a leaf certificate that forms a certification path to that root of trust in a request to the Roles Anywhere credentialing API, which is called `CreateSession`. The API uses an asymmetric private-key based signature specification for authN (it doesn't use mTLS). The root of trust that is configured with the service is called a trust anchor and needs to be referred to in the request to `CreateSession` along with a few other resources, including the ARN of the role that the workload wants to assume. From a mental model perspective, you can think of the service as providing an `AssumeRoleWithCertificate` API. 

For this integration with SPIRE, I'm thinking that on SPIRE server start up, it can create a trust anchor resource with the Roles Anywhere service based on the trust domain's SPIFFE bundle. This means that the SPIRE server will need to be configured with AWS credentials that have permissions to call `rolesanywhere:CreateTrustAnchor` and `rolesanywhere:UpdateTrustAnchor` (for when the certificate bundle needs to be updated, like in a rotation scenario), along with potentially a few other permissions. Then, this trust anchor ARN should be saved by the SPIRE server so that it can be communicated to the SPIRE agent. The SPIRE agent can then expose the trust anchor ARN information to workloads through the Workload Endpoint, but through a logical service that is independent from the Workload API (as mentioned in the [Extensibility and Services Rendered](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Workload_Endpoint.md#7-extensibility-and-services-rendered)  section of the Workload Endpoint standard). Workloads can obtain the trust anchor ARN for this trust domain from this new logical service and use it to obtain temporary AWS credentials from the Roles Anywhere service. 

There are some details that I'm skipping (like revocation for example - Roles Anywhere currently only supports CRLs as a resource that's configured with the service), but on a high level, does such an integration make sense? If so, how should the integration with SPIRE look (does the flow I provided in the previous paragraph make sense)? If the integration does make sense, I would be happy to implement it (although, I do admit that I need to familiarize myself with the SPIRE codebase a bit more).

evan2645:
Thanks for opening this @ajay1135! Better integration here would be awesome

[...]

PR #5048, opened 2024-04-08
Introduce `aws_rolesanywhere_trustanchor` BundlePublisher plugin

ajay1135:
**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**

Introduces the `aws_rolesanywhere_trustanchor` BundlePublisher plugin. When the server's trust bundle is updated, it will keep the corresponding AWS IAM Roles Anywhere trust anchor updated as well. Users that want to use IAM Roles Anywhere to access AWS using temporary credentials from outside of AWS (on-prem, another Cloud Service Provider, etc.) can consider using this plugin. 

**Description of change**

This plugin needs to be configured with a `trust_anchor_name`, which contains the name of the trust anchor to create or update. If there are multiple with the same name, the first that is found when listing trust anchors (`rolesanywhere:ListTrustAnchors`) is used. And if one can't be found, a new trust anchor is created. 

A few things to be aware of: 
 - Roles Anywhere allows for at most two certificates per trust anchor (if more are provided, I believe only the first two are used (I'm yet to test this, but I will do so))
      - The idea behind supporting two certificates is to support seamless certificate rotation, but I'm not sure how limiting this is from a SPIRE POV
 - Roles Anywhere doesn't support CRL DPs or OCSP endpoints and instead supports a separate CRL resource (revocation is disregarded in this PR)
 - The order in which trust anchors appear when calling `rolesanywhere:ListTrustAnchors` isn't an "inserted first" order (instead it's lexicographical based on UUID (not name) of the trust anchor)

**Which issue this PR fixes**

Not sure if it completely fixes it, but this PR does address part of #4963


pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere.go
```diff
@@ -0,0 +1,257 @@
...
+			},
+			Enabled: func() *bool { b := true; return &b }(),
+		}
+
+		createTrustAnchorOutput, err := p.rolesAnywhereClient.CreateTrustAnchor(ctx, &createTrustAnchorInput)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to create trust anchor: %v", err)
+		}
+		trustAnchorArn = *createTrustAnchorOutput.TrustAnchor.TrustAnchorArn
+	}
```
amartinezfayo, line 181: Giving the responsibility to create the trust anchor to the plugin has a lot of disadvantages in my opinion.
It makes the publishing logic a lot more complex than what would be needed if it assumes that the trust anchor already exists, but more importantly, it's prone to configuration errors that can lead to create a trust anchor with a wrong name, and not update the proper trust anchor. The fact that there could be multiple trust anchors with the same name doesn't help either, and contributes to make this more error prone.
It also requires to grant additional permissions, when we are only interested in keeping a trust anchor updated. 
I think that's better to give the responsibility to create the trust anchor to the user, and just configure the plugin with a specific trust anchor to update. Only the required permissions to update a trust anchor would be required in this case.

amartinezfayo:
Thank you @ajay1135 for this contribution!
My main concern is related with having the plugin configured with a trust anchor name, rather than a specific trust anchor. I see a lot of disadvantages with this approach, the main one being that's prone to configuration errors. This is exacerbated by the fact that there may be multiple trust anchors with the same name, so you can't uniquely identify a trust anchor by its name. Peeking the first of the list returned with the matching name doesn't seem to be a robust solution for this. Updating the wrong trust anchor and have the user trusting an incorrect trust anchor can be pretty bad.
My preference is to configure the plugin with a specific trust anchor ID.
I left some feedback.

amartinezfayo (commented):
Thank you @ajay1135 for this contribution!
My main concern is related with having the plugin configured with a trust anchor name, rather than a specific trust anchor. I see a lot of disadvantages with this approach, the main one being that's prone to configuration errors. This is exacerbated by the fact that there may be multiple trust anchors with the same name, so you can't uniquely identify a trust anchor by its name. Peeking the first of the list returned with the matching name doesn't seem to be a robust solution for this. Updating the wrong trust anchor and have the user trusting an incorrect trust anchor can be pretty bad.
My preference is to configure the plugin with a specific trust anchor ID.
I left some feedback.

ajay1135:
Thanks for the feedback! I think what you said makes sense. I'll update the PR when I get a chance.

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere.go
```diff
@@ -0,0 +1,257 @@
...
+
+	currentBundle := p.getBundle()
+	if proto.Equal(req.GetBundle(), currentBundle) {
+		// Bundle not changed. No need to publish.
+		return &bundlepublisherv1.PublishBundleResponse{}, nil
+	}
+
+	formatter := bundleformat.NewFormatter(req.GetBundle())
+	bundleBytes, err := formatter.Format(bundleformat.PEM)
```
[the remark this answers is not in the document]
ajay1135: Will address in my next revision.

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere.go
```diff
@@ -0,0 +1,257 @@
...
+
+		createTrustAnchorOutput, err := p.rolesAnywhereClient.CreateTrustAnchor(ctx, &createTrustAnchorInput)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to create trust anchor: %v", err)
+		}
+		trustAnchorArn = *createTrustAnchorOutput.TrustAnchor.TrustAnchorArn
+	}
+
+	p.setBundle(req.GetBundle())
+	p.log.Debug("Trust anchor bundle updated", "ARN", trustAnchorArn)
```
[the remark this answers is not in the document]
ajay1135: Will address in my next revision.

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere.go
ajay1135: Yeah, that makes sense. I will change it so that it works as you suggested in my next revision.

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere.go
```diff
@@ -0,0 +1,205 @@
...
+
+	formatter := bundleformat.NewFormatter(req.GetBundle())
+	bundleBytes, err := formatter.Format(bundleformat.PEM)
+	if err != nil {
+		return nil, status.Error(codes.Internal, "could not format bundle to PEM format")
+	}
+	bundleStr := string(bundleBytes)
+
+	// Update the trust anchor that was found
+	updateTrustAnchorInput := rolesanywhere.UpdateTrustAnchorInput{
```
amartinezfayo, line 115: The [maximum length of the SourceData is 8000](https://docs.aws.amazon.com/rolesanywhere/latest/APIReference/API_SourceData.html). Do you think that it would be problematic in your use case?
That would be around 11 certs in the bundle when no upstream authority is used. I think that we should have a validation, and if it's larger than 8000 fail with a descriptive error before trying to update, because the error response from the AWS API will include the SourceData, which can flood the logs.

conf/server/server_full.conf
```diff
@@ -899,6 +899,26 @@ plugins {
...
+    #         # access_key_id: AWS access key id. Default: value of
+    #         # AWS_ACCESS_KEY_ID environment variable.
+    #         # access_key_id = ""
+
+    #         # secret_access_key: AWS secret access key. Default: value of
+    #         # AWS_SECRET_ACCESS_KEY environment variable.
+    #         # secret_access_key = ""
+
+    #         # trust_anchor_name: The AWS IAM Roles Anywhere trust anchor name to which to store the trust bundle. Default: "".
+    #         # trust_anchor_name = "spire-trust-anchor"
```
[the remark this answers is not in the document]
ajay1135: Thanks for the catch. Will update in the next revision.

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	p.rolesAnywhereClient = rolesAnywhere
+
+	p.setConfig(config)
+	p.setBundle(nil)
+	return &configv1.ConfigureResponse{}, nil
+}
+
+// PublishBundle puts the bundle in the first Roles Anywhere trust anchor
+// found with the configured name. If one doesn't exist, it is created.
```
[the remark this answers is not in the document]
ajay1135: Will update in the next revision.

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere.go
ajay1135: Good catch, I can add that check in. But I think what will usually be more limiting is the fact that the maximum bundle size that can be used for a trust anchor is [two currently](https://docs.aws.amazon.com/rolesanywhere/latest/userguide/quotas.html) (see the "Certificates per trust anchor" limit), and it isn't adjustable either. For example, when I try to update a trust anchor with a bundle of size six, I get the following error: 

```
An error occurred (ValidationException) when calling the UpdateTrustAnchor operation: Unable to update a TrustAnchor with a certificate bundle of size 6.
```
At least the error response doesn't contain the bundle contents, so maybe the concern about flooding the logs isn't as relevant here, but it's still an error nonetheless. I think the intent behind this limit being two is to support CA certificate rotation, without loss of availability. So in general, the service is expecting a single CA certificate to be uploaded as a trust anchor (and two when the CA certificate needs to be rotated). 

Given this, in what scenarios will the trust bundle contain more than a single certificate? And in those cases, would it be okay to have this plugin publish just one of those certificates to the Roles Anywhere trust anchor? I'm guessing this behavior would make the plugin inconsistent with the functionality of the other bundle publisher plugins though - I'm not sure if that's okay. If we do go with this approach and additional certificates are needed, they can be provided in the request to `CreateSession` (the Roles Anywhere credentialing API), in the `X-Amz-X509-Chain` HTTP header (I have to test whether there's a limit here, but it should definitely be greater than one). 

What are your thoughts on this?

ajay1135: I think that should work for now, thanks. 

How do you suggest I check that a plugin of a different type has also been configured? Is there any precedence for it? I could potentially "fake" it by adding another configuration field for this plugin, which will allow the user to acknowledge (so a boolean field) that they are also using an `UpstreamAuthority` plugin. It could also be a more general acknowledgement field, so that the user will only configure the plugin if they know for sure they want to use it despite the limitations.

amartinezfayo: Unfortunately, there is no easy way for a plugin to check if other plugin has been configured. Having a setting to explicitly tell that there is an UpstreamAuthority plugin may work, but at the same time it feels a little confusing to me. I would suggest that we start by having a very clear documentation about this plugin being only supported when an UpstreamAuthority plugin is used.

amartinezfayo: Hey @ajay1135 I just wanted to provide a quick summary of what I think is pending:
1) Document that this plugin is only supported when an UpstreamAuthority plugin is used. Probably with a bold note at the very beginning of the .md document.
2) Check that the bundle has 1) no more than 2 certs before is being published and 2) doesn't exceed the maximum length of 8000.
3) Have 2) covered by unit tests.

ajay1135: Thanks for summarizing this. Quick question though - do you think it's necessary to explicitly handle the error case when there are more than two certificates in the bundle, before calling the API? In this case, the error message from the API won't flood the logs with the contents of the bundle. Here's what I get when I call the `UpdateTrustAnchor` API through the CLI, with a bundle with three certificates in it: 

```
An error occurred (ValidationException) when calling the UpdateTrustAnchor operation: Unable to update a TrustAnchor with a certificate bundle of size 3.
```

For now, I'm not explicitly handling this error scenario, but if you think I should, I'll go back and update the PR. But I think my new changes do address the rest of your comments.

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere_test.go
```diff
@@ -0,0 +1,376 @@
...
+	client.updateTrustAnchorErr = errors.New("error calling UpdateTrustAnchor")
+
+	// Call PublishBundle. UpdateTrustAnchor should be called and return an error.
+	resp, err := p.PublishBundle(context.Background(), &bundlepublisherv1.PublishBundleRequest{
+		Bundle: bundle,
+	})
+	require.Error(t, err)
+	require.Nil(t, resp)
+
+	// UpdateTrustAnchor was called, even though it failed, so its counter should be incremented.
```
amartinezfayo, line 272: The `updateTrustAnchorCount` is intended to count the number of times that the bundle was successfully published. If there is an error, this counter should not be increased. The fake implementation of UpdateTrustAnchor should be updated to reflect that.
```suggestion
	// UpdateTrustAnchor was called, even though it failed, so its counter should not be incremented.
```

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere_test.go
```diff
@@ -0,0 +1,376 @@
...
+	require.Equal(t, 1, client.updateTrustAnchorCount)
```
amartinezfayo, line 273: This should remain in 0 due to the error.
```suggestion
	require.Equal(t, 0, client.updateTrustAnchorCount)
```

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+	// Remove the updateTrustAnchorErr and try again.
+	client.updateTrustAnchorErr = nil
+	resp, err = p.PublishBundle(context.Background(), &bundlepublisherv1.PublishBundleRequest{
+		Bundle: bundle,
+	})
+	require.NoError(t, err)
+	require.NotNil(t, resp)
+	require.Equal(t, 2, client.updateTrustAnchorCount)
```
amartinezfayo, line 282:
```suggestion
	require.Equal(t, 1, client.updateTrustAnchorCount)
```

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+	// Call PublishBundle with the same bundle.
+	resp, err = p.PublishBundle(context.Background(), &bundlepublisherv1.PublishBundleRequest{
+		Bundle: bundle,
+	})
+	require.NoError(t, err)
+	require.NotNil(t, resp)
+
+	// The same bundle was used, the counter should be the same as before.
+	require.Equal(t, 2, client.updateTrustAnchorCount)
```
amartinezfayo, line 292:
```suggestion
	require.Equal(t, 1, client.updateTrustAnchorCount)
```

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere_test.go
```diff
@@ -0,0 +1,376 @@
...
+	bundle.SequenceNumber = 2
+	resp, err = p.PublishBundle(context.Background(), &bundlepublisherv1.PublishBundleRequest{
+		Bundle: bundle,
+	})
+	require.NoError(t, err)
+	require.NotNil(t, resp)
+
+	// PublishBundle was called with a different bundle, updateTrustAnchorCount should
+	// be incremented to be 3.
+	require.Equal(t, 3, client.updateTrustAnchorCount)
```
amartinezfayo, line 305:
```suggestion
	require.Equal(t, 2, client.updateTrustAnchorCount)
```

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+	// Try to publish a bundle that's too large, and expect that we receive an error.
+	bundle = getLargeTestBundle(t)
+	bundle.SequenceNumber = 3
+	resp, err = p.PublishBundle(context.Background(), &bundlepublisherv1.PublishBundleRequest{
```
amartinezfayo, line 310: In addition to asserting that there is an error, we should assert the value of resp, which should be nil due to the error.
```
require.Nil(t, resp)
```

amartinezfayo (commented):
Thank you @ajay1135 for addressing the comments, it looks great!
I only have a few comments related with the tests. After addressing those, we should be able to merge.

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+	awsConfig              aws.Config
+	updateTrustAnchorErr   error
+	updateTrustAnchorCount int
+
+	expectTrustAnchorID *string
+}
+
+func (c *fakeClient) UpdateTrustAnchor(_ context.Context, params *rolesanywhere.UpdateTrustAnchorInput, _ ...func(*rolesanywhere.Options)) (*rolesanywhere.UpdateTrustAnchorOutput, error) {
+	c.updateTrustAnchorCount++
```
amartinezfayo, line 327: The updateTrustAnchorCount counter should be incremented only if there is no error.
```suggestion
```

pkg/server/plugin/bundlepublisher/awsrolesanywhere/awsrolesanywhere_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+	if c.updateTrustAnchorErr != nil {
+		return nil, c.updateTrustAnchorErr
+	}
+
+	require.Equal(c.t, c.expectTrustAnchorID, params.TrustAnchorId, "trust anchor id mismatch")
+	trustAnchorArn := "trustAnchorArn"
+	trustAnchorName := "trustAnchorName"
```
amartinezfayo, line 335:
```suggestion
	trustAnchorName := "trustAnchorName"
	c.updateTrustAnchorCount++
```

doc/plugin_server_bundlepublisher_aws_rolesanywhere_trustanchor.md
```diff
@@ -0,0 +1,32 @@
+# Server plugin: BundlePublisher "aws_rolesanywhere_trustanchor"
+
+**Note: This plugin is only supported when an UpstreamAuthority plugin is used.**
```
amartinezfayo, line 3: I think that it may be a good idea to use a Warning note highlighted by markdown: 
```suggestion
> [!WARNING]
> This plugin is only supported when an UpstreamAuthority plugin is used.
```

ajay1135:
Thanks for the comments, and sorry for the delay. I've applied your suggestions and addressed your other comment as well.

amartinezfayo (approved):
Thank you @ajay1135 for this contribution!

PR #5115, opened 2024-04-30
reset timestamps claims to integer when composer plugins are invoked for compatibility with AWS

### LINKED ISSUE #4982: CredentialComposer plugin serializes integer claims as float
* **Version**: b5879e2f4f8f2cc8146f1e07a1ee9ec2a97d6433
* **Platform**: `Linux server-1 5.15.0-1053-aws #58~20.04.1-Ubuntu SMP Mon Jan 22 17:15:01 UTC 2024 x86_64 Linux`
* **Subsystem**: [server\credentialcomposer](https://github.com/spiffe/spire-plugin-sdk/tree/main/templates/server/credentialcomposer)

Problem: Using the composer plugin for modifying JWT-SVIDs yields in timestamp claims like `iat`, `exp` etc. to be formatted as type `float`. While numeric type is valid for timestamps and the the token can be successfully validated using jwt.io, AWS rejects it. I've already reached out to AWS support and they have asked us to fix the issue on our end.

>...based on IAM service design, the float timestamp is currently not supported in token claims. While there might not be any restrictions in JWT specification on timestamps being integers or float, IAM service does not currently support float timestamps, hence the error of "Invalid timestamp in token claims" being subsequently thrown. Kindly update your timestamp accordingly and retry your workflow.

Here's a sample spire JWT-SVID payload when not using the composer plugin
>eyJhdWQiOlsic3RzLmFtYXpvbmF3cy5jb20iXSwiZXhwIjoxNzEwNzA0NDk1LCJpYXQiOjE3MTA0NDUyOTUsImlzcyI6Imh0dHBzOi8vd29ya2xvYWRpZC5pbmRlZWQudGVjaCIsInN1YiI6InNwaWZmZTovL3dvcmtsb2FkaWQuaW5kZWVkLnRlY2gvZm9vLXNhIn0

Note, `exp` and `iat` are integer.

Here's a sample spire JWT-SVID payload when using the credential composer plugin (our plugin implementation never modifies those claims)
>eyJhdWQiOlsic3RzLmFtYXpvbmF3cy5jb20iXSwiZXhwIjoxLjcxMDYxODIxZSswOSwiaHR0cHM6Ly9hd3MuYW1hem9uLmNvbS90YWdzIjp7InByaW5jaXBhbF90YWdzIjp7IkQwIjpbImFiYyJdfSwidHJhbnNpdGl2ZV90YWdfa2V5cyI6WyJEMCJdfSwiaWF0IjoxLjcxMDQ0NTQxZSswOSwiaXNzIjoiaHR0cHM6Ly93b3JrbG9hZGlkLmluZGVlZC50ZWNoIiwicHVycG9zZSI6ImFjY2Vzc190b2tlbiIsInNjb3BlIjoiRDA6YSBEMDpiIEQwOmMiLCJzdWIiOiJzcGlmZmU6Ly93b3JrbG9hZGlkLmluZGVlZC50ZWNoL2Zvby1zYSJ9

Note, `exp` and `iat` are floats. Still valid JWTs and valid timestamps, but AWS SDK will reject the use of such JWT with:
>An error occurred (InvalidIdentityToken) when calling the AssumeRoleWithWebIdentity operation: Invalid timestamp in token claims.

Upon initial investigation is looks like an issue with the proto definition of `Claims`. We believe it’s due to the following representation:
`Claims *structpb.Struct`
The protobuf struct doesn’t support integers and prefers to represent numbers as double.
`double number_value = 2;`

The translations of Claims as protobuf Struct and then further JSON serialization is likely yielding the final format of `float` type.

Goal: Stick to integer representation for well known timestamp claims for compatibility with AWS.

naroraindeed:
Using credentialcomposer plugins forces Claims to be translated as protobuf structs which serializes integers as floats (#4982). AWS rejects validating JWT issued by SPIRE with timestamps that are in scientific notation. AWS STS only accepts integer timestamps as valid. We've discussed this with AWS, and while they agree it's an issue in AWS STS, there's no recourse available with them. This fix helps reset value type for timestamps and also includes unit tests that proves the problem and the limited scope of the fix. This is the minimal change needed for SPIRE to produce verifiable JWT for AWS when using credentialcomposer plugin.

Signed-off-by: narora@indeed.com

**Pull Request check list**

- [Y] Commit conforms to CONTRIBUTING.md?
- [Y] Proper tests/regressions included?
- [N] Documentation updated?

**Affected functionality**
BuildWorkloadJWTSVIDClaims resets data type for certain claims.

**Description of change**
Minimal change needed for SPIRE to produce verifiable JWT for AWS when using credentialcomposer plugin.

**Which issue this PR fixes**
`fixes #4982`

### DIFF
--- a/pkg/server/credtemplate/builder.go
+++ b/pkg/server/credtemplate/builder.go
@@ -345,6 +345,18 @@ func (b *Builder) BuildWorkloadJWTSVIDClaims(ctx context.Context, params Workloa
 		}
 	}
 
+	// AWS will otherwise reject validating timestamps serialized in scientific notation.
+	// Protobuf serializes large integers as float since Claims are represented as google.protobuf.Struct.
+	if len(b.config.CredentialComposers) > 0 {
+		if iat, ok := attributes.Claims["iat"].(float64); ok {
+			attributes.Claims["iat"] = int64(iat)
+		}
+
+		if exp, ok := attributes.Claims["exp"].(float64); ok {
+			attributes.Claims["exp"] = int64(exp)
+		}
+	}
+
 	return attributes.Claims, nil
 }
 
--- a/pkg/server/credtemplate/builder_test.go
+++ b/pkg/server/credtemplate/builder_test.go
@@ -8,6 +8,7 @@ import (
 	"encoding/asn1"
 	"errors"
 	"fmt"
+	"math"
 	"math/big"
 	"net/url"
 	"testing"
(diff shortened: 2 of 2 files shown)


naroraindeed:
fixed lint and DCO workflow failures. Need help to retrigger workflows.

azdagron (approved):
Thanks for this quick fix @naroraindeed. This will be great for the upcoming 1.9.5 release. I think for 1.10.0 we'll do something a little more broadly scoped and this PR lays out some important groundwork. I really appreciate the contribution.

naroraindeed:
Thank you so much @azdagron . I completely agree, this is a tactical fix to make AWS happy. This doesn't fix the generic problem of representing Claims in a way that preserves the integer value type in composer plugins. That's a good next step.
You've reduced the burden on our end to maintain a separate fork for core SPIRE. 👍🏾

liwadman:
@azdagron A review of many common JWT validation libraries seemed to indicate that by type validation alone, timestamps as floats/scientific notation wouldn't be supported. I will say I that I didn't test all these with code, but rather the their code for validating IAT/NBF/EXP claims. Thus it would seem there is extremely mixed support in the JWT ecosystem for non-integer timestamp values, and in time likely it would surface more than just AWS STS would be impacted by this issue.

Libraries I Checked:
Nimbus - java (floats not supported by type validation)
Jose - typescript (floats allowed by type validation)
PyJWT - python(floats not supported by type validation)
Python-jose - python(floats not supported by type validation)
Golang-jwt - go (floats allowed by type validation)
g-jose - go (floats not supported by type validation)
java-jwt - java(floats not supported by type validation)
node-jose javascript(floats allowed by type validation)

PR #5281, opened 2024-07-06
lru cache: Only send notifications to the new subscriber

### LINKED ISSUE #5280: Same SVID is pushed repeatedly on the FetchX509SVID stream by spire-agent with LRU cache
If using a LRU cache the agent will send updates on the streams opened by workload api clients even if there are not changes to propagate. This happens whenever there is a call to `FetchX509SVID` or `FetchX509Bundle` from the same or a different client using the same set of selectors. Both of those APIs end up in https://github.com/spiffe/spire/blob/a62344a4c9ca0bd9709fa815fa08c811407299e8/pkg/agent/manager/cache/lru_cache.go#L555 and will then notify subscribers based on their selectors.

This can be easily reproduced using the spire-agent cli:
1. Start watching for updates using `spire-agent api watch`.
2. Request a SVID using `spire-agent api fetch`. The watcher from step 1 will receive a SVID update again, even if there was no change to warrant it. This behaviour is repeated on each fetch.

This is different from the behaviour of the previous cache. In most case this only causes unnecessary work, but some users of the API can have other side effects which shouldn't happen this often. See for example https://spiffe.slack.com/archives/CBNCC2V17/p1720149402728429

* **Version**: >= 1.9.0 and older if LRU cache was enabled
* **Platform**: any
* **Subsystem**: spire-agent

sorindumitru:
When there's a new subscription to the cache we shouldn't notify all existing subscriptions. That causes them to receive updates even if there was no change to the SVIDs or bundles.

This can cause issues with users that trigger side effects, such as restarting a service, when they receive updates on the X509 stream.

I think this shouldn't be ok, I didn't see any reason for the cache notify all subscribers of a set if it knows the new subscriber already.

**Pull Request check list**

- [x] Documentation updated?

**Which issue this PR fixes**
fixes #5280


amartinezfayo (approved):
LGTM, thank you @sorindumitru!

PR #5324, opened 2024-07-23
unix attestor: fix discover_workload_path on mac os

sorindumitru:
When enabling 'discover_workload_path' on mac os the agent fails to attest any workload due to it trying to access a procfs path:
```
ERRO[0068] Failed to collect all selectors for PID       error="workload attestor \"unix\" failed: rpc error: code = Internal desc = workloadattestor(unix): SHA256 digest: open /proc/16470/exe: no such file or directory" pid=16470 subsystem_name=workload_attestor
DEBU[0068] PID attested to have selectors                pid=16470 selectors="[]" subsystem_name=workload_attestor
ERRO[0068] No identity issued                            method=FetchX509SVID pid=16470 registered=false service=WorkloadAPI subsystem_name=endpoints
```

I assume this is needed to be able to get a path that is accessible by the spire-agent, for example if it runs in a different namespace. I assume it's mostly intended for Linux, for most other OSes `proc.Exe()` would be better. Either way, it's likely to be wrong on various OSes since they either have no procfs, they have no link to the executable in procfs, or they link is named differently (e.g. `a.out` or `file`).

### DIFF
--- a/pkg/agent/plugin/workloadattestor/unix/unix_posix.go
+++ b/pkg/agent/plugin/workloadattestor/unix/unix_posix.go
@@ -176,12 +176,15 @@ func (p *Plugin) Attest(_ context.Context, req *workloadattestorv1.AttestRequest
 		selectorValues = append(selectorValues, makeSelectorValue("path", processPath))
 
 		if config.WorkloadSizeLimit >= 0 {
-			exePath := p.getNamespacedPath(proc)
-			sha256Digest, err := util.GetSHA256Digest(exePath, config.WorkloadSizeLimit)
+			exePath, err := p.getNamespacedPath(proc)
 			if err != nil {
 				return nil, status.Error(codes.Internal, err.Error())
 			}
 
+			sha256Digest, err := util.GetSHA256Digest(exePath, config.WorkloadSizeLimit)
+			if err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
 			selectorValues = append(selectorValues, makeSelectorValue("sha256", sha256Digest))
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/agent/plugin/workloadattestor/unix/unix_posix.go
```diff
@@ -176,13 +176,15 @@ func (p *Plugin) Attest(_ context.Context, req *workloadattestorv1.AttestRequest
 		selectorValues = append(selectorValues, makeSelectorValue("path", processPath))
 
 		if config.WorkloadSizeLimit >= 0 {
-			exePath := p.getNamespacedPath(proc)
-			sha256Digest, err := util.GetSHA256Digest(exePath, config.WorkloadSizeLimit)
-			if err != nil {
-				return nil, status.Error(codes.Internal, err.Error())
+			exePath, err := p.getNamespacedPath(proc)
+			if err == nil {
```
MarcosDY, line 180: error will be ignored, can you return it instead?

sorindumitru: Can do, but wouldn't it more useful to still return the list of selectors that have been attested about the process even if we got some errors along the way? 

We could log the errors so admins are aware of the issue, but return what we have since those selectors might be enough for the workload to function.

amartinezfayo: >Can do, but wouldn't it more useful to still return the list of selectors that have been attested about the process even if we got some errors along the way?

Workload attestors in SPIRE take a conservative approach when there are selectors that can't be discovered, and if there is an unexpected condition that prevents one of the selectors from being discovered, then we usually fail. The important part to consider is if this error would be something expected under normal conditions.
In this case, the `discover_workload_path` configurable is explicitly telling to SPIRE if the user is expecting to be able to collect the `unix:path` and `unix:sha256` selectors or not.
We have this documented in the plugin documentation:
>If configured with discover_workload_path = true, the plugin will discover the workload path to provide additional selectors. If the plugin cannot discover the workload path or gather selectors based on the path, it will fail the attestation attempt.

I suggest to fail the attestation if there is an error here.

MarcosDY (approved):

Issue #3900, opened 2023-02-21
Propagate Taint keys on SPIRE upstream authority


MarcosDY:
SPIRE Upstream authority must obtain the list of tainted keys from NewDownstreamX509CA and GetBundle, to propagates tainted keys to SPIRE Server

Depends on #3885 #3886 #3899

MarcosDY:
to keep this issue alive

MarcosDY:
resolved as part of https://github.com/spiffe/spire/pull/5340

Issue #3893, opened 2023-02-21
Add taint to common Bundle to utils


MarcosDY:
Add taint field to JWT and X509 parsing code

Depends on #3885

MarcosDY:
to keep this issue alive

MarcosDY:
Partially resolved in https://github.com/spiffe/spire/pull/5340

MarcosDY:
Left parts will be solved on #5394

PR #5340, opened 2024-07-30
Add taint upstream authority

### LINKED ISSUE #3899: Update NewDowstreamX509CA to propagate metadata
Propagate Tainted keys when using NewDownstreamX509CA

Depends on #3885

MarcosDY:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
no actual functionalities are affected

**Description of change**
To taint a local authority that is using an upstream authority, we need a way to taint the upstream authority itself and rotate intermediates + SVIDs using that tainted upstream authority.
This PR aims to add a new RPC that allows tainting an upstream authority using SKID as the identifier, update the datastore with that information, and propagate it to downstream servers.
This PR is **not forcing rotation** but only propagating tainted authorities. Forcing rotations will happen in other PRs.

Changes to the SDK can be found [here](https://github.com/spiffe/spire-api-sdk/pull/64).

**Which issue this PR fixes**
fixes #3899, #3893, #3900


pkg/server/ca/manager/slot.go
```diff
@@ -40,7 +40,9 @@ type Slot interface {
 	ShouldPrepareNext(now time.Time) bool
 	ShouldActivateNext(now time.Time) bool
 	Status() journal.Status
+	UpstreamAuthorityID() string
 	AuthorityID() string
+	// TODO: Should we remove public key?
```
amartinezfayo, line 45: Yes, I think that PublicKey() should be removed. It's no longer used.

pkg/server/ca/upstream_client.go
```diff
@@ -139,16 +140,22 @@ func (u *UpstreamClient) runMintX509CAStream(ctx context.Context, csr []byte, tt
 	}
 	defer x509RootsStream.Close()
 
+	// Extract all roots certificates
```
amartinezfayo, line 143:
```suggestion
	// Extract all root certificates
```

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1376,33 +1371,35 @@ func pruneBundle(tx *gorm.DB, trustDomainID string, expiry time.Time, log logrus
...
-	pKey, err := x509.MarshalPKIXPublicKey(publicKeyToTaint)
-	if err != nil {
-		return status.Errorf(codes.InvalidArgument, "failed to marshal public key to taint: %v", err)
+		found = true
+		eachRootCA.TaintedKey = true
 	}
 
-	bundle.X509TaintedKeys = append(bundle.X509TaintedKeys, &common.X509TaintedKey{PublicKey: pKey})
+	if !found {
+		return status.Errorf(codes.NotFound, "no ca found with provided subject key ID")
```
amartinezfayo, line 1401:
```suggestion
		return status.Error(codes.NotFound, "no ca found with provided subject key ID")
```

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1376,33 +1371,35 @@ func pruneBundle(tx *gorm.DB, trustDomainID string, expiry time.Time, log logrus
...
+	}
 
 	_, err = updateBundle(tx, bundle, nil)
```
amartinezfayo, line 1404: I think that we should increment bundle.SequenceNumber before calling updateBundle, otherwise SequenceNumber will remain the same.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1412,58 +1409,36 @@ func taintX509CA(tx *gorm.DB, trustDomainID string, publicKeyToTaint crypto.Publ
...
 		rootCAs = append(rootCAs, ca)
 	}
+
+	if !keyFound {
+		return status.Error(codes.NotFound, "no root CA found with provided subject key ID")
+	}
+
 	bundle.RootCas = rootCAs
 
 	if _, err := updateBundle(tx, bundle, nil); err != nil {
```
amartinezfayo, line 1444: I think that we should increment bundle.SequenceNumber before calling updateBundle, otherwise SequenceNumber will remain the same.

pkg/server/datastore/sqlstore/sqlstore_test.go
```diff
@@ -641,148 +642,116 @@ func (s *PluginSuite) TestBundlePrune() {
...
-	// spiretest.RequireGRPCStatus(t, err, codes.NotFound, "no root CA found with provided public key")
+	t.Run("No root CA is using provided skID", func(t *testing.T) {
+		err := s.ds.RevokeX509CA(ctx, "spiffe://foo", "foo")
+		spiretest.RequireGRPCStatus(t, err, codes.NotFound, "no root CA found with provided subject key ID")
+	})
 
-	// No able to revoke untainted bundles
-	err = s.ds.RevokeX509CA(ctx, "spiffe://foo", s.cert.PublicKey)
-	spiretest.RequireGRPCStatus(t, err, codes.InvalidArgument, "it is not possible to revoke an untainted root CA")
+	t.Run("Unable to revoke untainted bundles", func(t *testing.T) {
```
amartinezfayo, line 734: It would be good to check that the sequence number of the bundle remains the same for those cases that fail.

pkg/server/datastore/sqlstore/sqlstore_test.go
```diff
@@ -641,148 +642,116 @@ func (s *PluginSuite) TestBundlePrune() {
...
 
-	require.Equal(t, expectedRootCAs, fetchedBundle.RootCas)
+		require.Equal(t, expectedRootCAs, fetchedBundle.RootCas)
+	})
 
-	expectedTaintedKeys := []*common.X509TaintedKey{
-		{PublicKey: certPublicKeyRaw},
-	}
-	require.Equal(t, expectedTaintedKeys, fetchedBundle.X509TaintedKeys)
+	t.Run("no bundle with provided skID", func(t *testing.T) {
```
amartinezfayo, line 683: It would be good to check that the sequence number of the bundle remains the same for those cases that fail.

pkg/server/datastore/sqlstore/sqlstore_test.go
```diff
@@ -641,148 +642,116 @@ func (s *PluginSuite) TestBundlePrune() {
...
+	err = s.ds.TaintX509CA(ctx, "spiffe://foo", certID)
 	require.NoError(t, err)
 
-	// Revoke successfully
-	err = s.ds.RevokeX509CA(ctx, "spiffe://foo", s.cert.PublicKey)
-	require.NoError(t, err)
-
-	fetchedBundle, err := s.ds.FetchBundle(ctx, "spiffe://foo")
-	require.NoError(t, err)
+	t.Run("Revoke successfully", func(t *testing.T) {
```
amartinezfayo, line 743: It would be good to check that the sequence number of the bundle is incremented on successful operations.

amartinezfayo:
TestCommonBundleFromProto should capture the proper propagation of the tainted bit.
It doesn't seem like we have explicit tests for ParseX509Authorities and ParseJWTAuthorities?
We should make sure that propagation of the tainted bit is covered by tests in all those functions. We can handle that in a separate PR if that's more convenient.

MarcosDY:
we discuss to add that propagation as part of https://github.com/spiffe/spire/issues/5394 to reduce changes to apply in this PR and it is not required for current PR functionality

amartinezfayo (approved):
🎉 🚀

PR #5345, opened 2024-07-31
Propogate SQLTransactionTimeout

### LINKED ISSUE #5343: [Event-based entry cache] New server config for `sql_transaction_timeout` is not propagated to endpoints config
While generating [endpoints.Config](https://github.com/spiffe/spire/blob/01bedb8e876618e54e0923e0214fd2a60b680a71/pkg/server/server.go#L383) in server, we don't set the value for `SQLTransactionTimeout` reading from `server.Config`, thereby this value being always [set to default](https://github.com/spiffe/spire/blob/01bedb8e876618e54e0923e0214fd2a60b680a71/pkg/server/endpoints/endpoints.go#L125-L127) i.e. [24hrs](https://github.com/spiffe/spire/blob/01bedb8e876618e54e0923e0214fd2a60b680a71/pkg/server/endpoints/endpoints.go#L56)

Should be a simple one line fix for reading and setting the endpoints.Config during server init.

* **Version**: 01bedb8e876618e54e0923e0214fd2a60b680a71
* **Platform**: x86_64 GNU/Linux
* **Subsystem**: server

faisal-memon:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Events based cache

**Description of change**
Fixes bug where SQLTransactionTimeout is not propagated
 
**Which issue this PR fixes**

fixes #5343


pkg/server/server.go
```diff
@@ -397,6 +397,7 @@ func (s *Server) newEndpointsServer(ctx context.Context, catalog catalog.Catalog
 		CacheReloadInterval:          s.config.CacheReloadInterval,
 		EventsBasedCache:             s.config.EventsBasedCache,
 		PruneEventsOlderThan:         s.config.PruneEventsOlderThan,
+		SQLTransactionTimeout:        s.config.SQLTransactionTimeout,
```
amoore877, line 400: it'd be nice to unit test, considering lack thereof caused the issue

MarcosDY (approved):

pkg/server/server.go
faisal-memon: Hey @amoore877 That was my first thought too. Unfortunately I dont know where to work in a unit test here as the fields just end up as unexported fields in the `endpoints` object.

amoore877: ah, and there's no way to easily test with a running module and infer behavior from config I guess. sad times.

the most I've done sometimes in such cases at least verify those unexported fields (which would be accessible in a same-package test) are set as expected. it's one of those kinds of tests that seems silly at first, but catches simple mistakes.

better though would be that behavioral check, but that seems hard compared to the effort of the fix :)

amoore877 (approved):

PR #5400, opened 2024-08-17
Introduce support to localauthority CLI - x509 show

amartinezfayo:
Introduce support to `localauthority` commands in SPIRE Server.
This is the first PR of a series of PRs adding `localauthority` commands, allowing to interact with the LocalAuthority API to perform forced rotation and revocation of X.509 and JWT authorities in SPIRE.


MarcosDY (commented):
looks great just a minor NIT

MarcosDY (approved):

PR #5411, opened 2024-08-20
Add gauge metrics to track missedEvents and cache sizes

stevend-uber:
This PR adds metrics for the event-based cache. This will emit a gauge whenever the cache is updated for missed-events and current cache sizes. 

**Pull Request check list**

- [ ] Documentation updated?

**Which issue this PR fixes**
https://github.com/spiffe/spire/issues/4720


edwbuck (changes requested):
This heavily uses the word "missed" which in my mind implies that we forgot something; because, one only misses items that exist (or existed).

I would prefer using the word "skipped" as there are scenarios where we should skip event ids, because the events associated with them never were created (due to hung transactions that never completed, or autoincrement values that don't increase by one).

edwbuck (approved):
Thank you very much for these changes.  They are a big contribution in monitoring a small value with big implications around the new database event cache.

pkg/server/endpoints/authorized_entryfetcher_attested_nodes.go
```diff
@@ -33,7 +36,7 @@ type attestedNodes struct {
 
 // buildAttestedNodesCache fetches all attested nodes and adds the unexpired ones to the cache.
 // It runs once at startup.
-func buildAttestedNodesCache(ctx context.Context, log logrus.FieldLogger, ds datastore.DataStore, clk clock.Clock, cache *authorizedentries.Cache, sqlTransactionTimeout time.Duration) (*attestedNodes, error) {
+func buildAttestedNodesCache(ctx context.Context, log logrus.FieldLogger, metrics telemetry.Metrics, ds datastore.DataStore, clk clock.Clock, cache *authorizedentries.Cache, sqlTransactionTimeout time.Duration) (*attestedNodes, error) {
```
zmt, line 39: Pre-existing condition (don't address here): we should consider wrapping the large number of args to this function into an arg struct. None are "plain" types like string so it's not really an issue per se, but 7 args is a lot. Similarly for `buildCache` in `authorized_entryfetcher.go`.

pkg/server/endpoints/authorized_entryfetcher_attested_nodes_test.go
```diff
@@ -9,62 +9,83 @@ import (
...
+					Key:    []string{telemetry.Node, telemetry.AgentsByIDCache, telemetry.Count},
+					Val:    1,
+					Labels: nil,
+				},
+			},
 		},
 	} {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.Background()
```
zmt, line 69: Good moving the setup inside `t.Run` to avoid sharing setup between test cases in the table. Perhaps consider extracting to a helper?

It looks like this is the only table where you did this move in the endpoints tests touched. Was this the only test where the changes uncovered an issue with shared setup? Should we be consistent and move all setup inside `t.Run`?

pkg/server/endpoints/authorized_entryfetcher_test.go
```diff
@@ -652,3 +685,75 @@ func TestUpdateAttestedNodesCacheMissedStartupEvents(t *testing.T) {
...
+// AgentsByIDCacheCount
+func agentsByIDMetric(val float32) fakemetrics.MetricItem {
+	return fakemetrics.MetricItem{
+		Type:   fakemetrics.SetGaugeType,
+		Key:    []string{telemetry.Node, telemetry.AgentsByIDCache, telemetry.Count},
+		Val:    val,
+		Labels: nil}
+}
+
+func agentsByIDExpiresMetric(val float32) fakemetrics.MetricItem {
```
zmt, line 698: nit: `agentsByExpiresAtMetric` for consistency

zmt (approved):
LGTM. I left a few minor nits and some general pre-existing style questions.

pkg/common/telemetry/server/server.go
```diff
@@ -7,3 +7,51 @@ import "github.com/spiffe/spire/pkg/common/telemetry"
 func SetEntryDeletedGauge(m telemetry.Metrics, deleted int) {
 	m.SetGauge([]string{telemetry.Entry, telemetry.Deleted}, float32(deleted))
 }
+
+// SetAgentsByIDCacheCountGauge emits a gauge with the number of agents by ID that are
+// currently in the node cache.
+func SetAgentsByIDCacheCountGauge(m telemetry.Metrics, length int) {
```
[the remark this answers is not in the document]
amoore877: the data comes from array lengths, but yes the purpose of the data is it's a count

pkg/server/authorizedentries/cache.go
```diff
@@ -267,7 +267,7 @@ func (c *Cache) removeEntry(entryID string) {
 	}
 }
 
-func (c *Cache) stats() cacheStats {
+func (c *Cache) Stats() cacheStats {
```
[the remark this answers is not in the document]
amoore877: it wasn't clear to me if `btree` package is safe for concurrent accesses, and if not do we have concerns now with that? can there be cache stats updates at the same time a caller is asking for stats?

pkg/server/endpoints/authorized_entryfetcher_attested_nodes_test.go
stevend-uber: Yeah, this was the only test that this was an issue in.

pkg/server/endpoints/authorized_entryfetcher_attested_nodes.go
stevend-uber: I agree. I almost did in this PR but opted against as the PR was already getting long and I wanted to maintain the metrics-centered context.

pkg/server/endpoints/authorized_entryfetcher_test.go
stevend-uber: Fixed :)

pkg/common/telemetry/server/server.go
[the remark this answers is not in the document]
stevend-uber: Fixed

amartinezfayo (approved):
Thank you @stevend-uber!

PR #5471, opened 2024-09-05
Fix test breaking when a proxy is configured on the host

kfox1111:
When the host has HTTP_PROXY settings defined, one of the new http_challege node agent tests will forward the request to the proxy from the local environment, which can't handle the request, rather then redirecting back to the test fixture properly. This patch changes the test to function like the rest of the http_challenge tests and calls back to the test fixture directly.

### DIFF
--- a/pkg/common/plugin/httpchallenge/httpchallenge_test.go
+++ b/pkg/common/plugin/httpchallenge/httpchallenge_test.go
@@ -79,13 +79,14 @@ func TestValidateChallenge(t *testing.T) {
 			}))
 			defer func() { testServer.Close() }()
 
-			transport := http.DefaultTransport.(*http.Transport).Clone()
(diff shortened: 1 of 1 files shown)


pkg/common/plugin/httpchallenge/httpchallenge_test.go
```diff
@@ -79,13 +79,14 @@ func TestValidateChallenge(t *testing.T) {
 			}))
 			defer func() { testServer.Close() }()
 
-			transport := http.DefaultTransport.(*http.Transport).Clone()
```
kfox1111, line 82: The DefaultTransport honors HTTP_PROXY env vars. We should actually start from a new Transport.

kfox1111:
@edwbuck

MarcosDY (approved):

PR #5487, opened 2024-09-10
No longer test against MySQL 5.7

### LINKED ISSUE #4270: Remove support for MySQL 5.7
Oracle is ending extended support for MySQL 5.7 in October 2023, see page 29 of https://www.oracle.com/us/support/library/lifetime-support-technology-069183.pdf

In general, supporting different versions of databases in the project is challenging, especially when the project wants to use database-native features that may not be available across all supported versions. It is also time consuming to run the unit and integration tests against each version of databases, and this results in longer PR CI build times (currently 5-10 minutes per database version).

Furthermore, SPIRE now supports arm64 architecture, but MySQL 5.7 does not, and presumably it never will since it's rapidly approaching EOL. This currently prevents us from running the MySQL 5.7 integration tests for SPIRE on arm64 architecture.

Starting in November 2023, we should strongly consider removing support for MySQL 5.7. This includes some production code that works around the lack of support for common table expressions in MySQL 5.7 as well as testing against MySQL 5.7 in the integration tests.

While the project doesn't yet publish an official matrix of supported database versions (see https://github.com/spiffe/spire/issues/3711), today this is largely implied by the versions that are tested in the repository.

rturner3:
MySQL 5.7 is EOL as of October 2023.

Stop using MySQL 5.7 in integration tests, since it is not easy for the project to support end-of-life database versions. Also log a warning when MySQL 5.7 is used with SPIRE.

Importantly, this commit does not remove any existing support for MySQL 5.7. SPIRE may continue to work for some time with MySQL 5.7, but the project makes no compatibility guarantees with MySQL 5.7 anymore, and any existing compatibility with MySQL 5.7 may break at any time without notice.

It is recommended that all SPIRE users using MySQL as a SQL backend upgrade to MySQL 8.0.

Closes #4270.


pkg/server/datastore/sqlstore/mysql.go
```diff
@@ -61,6 +64,10 @@ func (my mysqlDB) connect(cfg *configuration, isReadOnly bool) (db *gorm.DB, ver
 		return nil, "", false, err
 	}
 
+	if version == "5.7" {
```
amartinezfayo, line 67: Since we are getting the version running a `SELECT VERSION()` query, isn't the version string the complete version number, like 5.7.44?
I'm thinking that this should probably check if the version _starts with_ rather than comparing with a complete string.

rturner3: Yeah, good catch. I changed the logic to check for the version prefix "5.7." so we can handle any 5.7 versions.

amartinezfayo (approved):
Thank you @rturner3!

PR #5500, opened 2024-09-18
Add HashiCorp Vault key manager plugin to SPIRE server

### LINKED ISSUE #5058: Feature Request: Use HashiCorp Vault as a KeyManager
We are currently using HashiCorp Vault as our UpstreamAuthority for our SPIRE setup and we would like to also use the Vault as a KeyManager. Therefore, we propose the addition of a new plugin that integrates Vault into the SPIRE ecosystem via a KeyManager plugin. This plugin could be applied to both the spire-server and the spire-agent. One thing that we should consider is whether Vault supports the appropriate key types that are used by SPIRE.

I would also be willing to implement such a plugin but I'm not quite sure where to get started.

InverseIntegral:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
This MR introduces a new SPIRE server key manager plugin that uses HashiCorp Vault to manage keys.

**Description of change**
This change adds a new key manager plugin to SPIRE server named `hashicorp_vault`. It uses a [transit secrets engine](https://developer.hashicorp.com/vault/docs/secrets/transit) within HashiCorp Vault to manage keys and sign data. The client is largely based on the existing upstream authority plugin for HashiCorp Vault which was introduced in #1611 by @hiyosi.

**Which issue this PR fixes**
This PR fixes #5058

Note that this is my first time implementing such a plugin and I expect there to be a lot of open questions and expected changes. I'm really looking forward to any feedback! :heart:


doc/plugin_server_keymanager_hashicorp_vault.md
```diff
@@ -0,0 +1,162 @@
...
+## Kubernetes Authentication
+
+| key                  | type   | required | description                                                                       | default    |
+|:---------------------|:-------|:---------|:----------------------------------------------------------------------------------|:-----------|
+| k8s_auth_mount_point | string |          | Name of the mount point where the Kubernetes auth method is mounted               | kubernetes |
+| k8s_auth_role_name   | string | ✔        | Name of the Vault role. The plugin authenticates against the named role           |            |
+| token_path           | string | ✔        | Path to the Kubernetes Service Account Token to use authentication with the Vault |            |
+
+```hcl
+    UpstreamAuthority "vault" {
```
amartinezfayo, line 145:
```suggestion
    KeyManager "hashicorp_vault" {
```

amartinezfayo (commented):
Hi @InverseIntegral, thank you very much for this contribution.
I'm still going over this PR, but the first high-level comment that I wanted to provide is that we will need integration tests that exercise this plugin. It should be a new suite added to our integration test framework: https://github.com/spiffe/spire/tree/main/test/integration

I'll be adding more comments as I make progress with the review :) Thank you again for this contribution!

InverseIntegral:
Thank you @amartinezfayo for the initial review and sorry for the delay on my end. I will add integration tests to test the plugin more thoroughly :+1:

doc/plugin_server_keymanager_hashicorp_vault.md
```diff
@@ -0,0 +1,162 @@
...
+|:---------------------|:-------|:---------|:---------------------------------------------------------------------------------------------------------|:---------------------|
+| vault_addr           | string |          | The URL of the Vault server. (e.g., <https://vault.example.com:8443/>)                                   | `${VAULT_ADDR}`      |
+| namespace            | string |          | Name of the Vault namespace. This is only available in the Vault Enterprise.                             | `${VAULT_NAMESPACE}` |
+| transit_engine_path  | string |          | Path of the transit engine that stores the keys.                                                         | transit              |
+| ca_cert_path         | string |          | Path to a CA certificate file used to verify the Vault server certificate. Only PEM format is supported. | `${VAULT_CACERT}`    |
+| insecure_skip_verify | bool   |          | If true, vault client accepts any server certificates                                                    | false                |
```
amartinezfayo, line 16: We should have a note here, that this is only for test environments.

pkg/server/plugin/keymanager/hashicorpvault/hashicorp_vault.go
```diff
@@ -0,0 +1,490 @@
...
+	return nil
+}
+
+func (p *Plugin) genClientParams(method AuthMethod, config *Config) (*ClientParams, error) {
+	cp := &ClientParams{
+		VaultAddr:         p.getEnvOrDefault(envVaultAddr, config.VaultAddr),
+		Namespace:         p.getEnvOrDefault(envVaultNamespace, config.Namespace),
+		TransitEnginePath: p.getEnvOrDefault(envVaultTransitEnginePath, config.TransitEnginePath),
+		CACertPath:        p.getEnvOrDefault(envVaultCACert, config.CACertPath),
+		TLSSKipVerify:     config.InsecureSkipVerify,
```
amartinezfayo, line 243: I think that we should log a warning if InsecureSkipVerify is set to true.

pkg/server/plugin/keymanager/hashicorpvault/hashicorp_vault.go
```diff
@@ -0,0 +1,490 @@
...
+			return nil, err
+		}
+	}
+
+	kt, err := convertToTransitKeyType(keyType)
+	if err != nil {
+		return nil, err
+	}
+
+	err = p.vc.CreateKey(ctx, spireKeyID, *kt)
```
amartinezfayo, line 432: What happens if the key already exists in the specified path?
I'm worried about only the possibility of creating a new key the only the first time, and then always using the same key.

pkg/server/plugin/keymanager/hashicorpvault/vault_client.go
```diff
@@ -0,0 +1,554 @@
...
+	TransitSignatureSignatureAlgorithmPSS      TransitSignatureAlgorithm = "pss"
+	TransitSignatureSignatureAlgorithmPKCS1v15 TransitSignatureAlgorithm = "pkcs1v15"
+)
+
+// CreateKey creates a new key in the specified transit secret engine
+// See: https://developer.hashicorp.com/vault/api-docs/secret/transit#create-key
+func (c *Client) CreateKey(ctx context.Context, spireKeyID string, keyType TransitKeyType) error {
+	arguments := map[string]interface{}{
+		"type":       keyType,
+		"exportable": "false", // TODO: Maybe make this configurable
```
amartinezfayo, line 374: I don't think that we would want to make SPIRE keys exportable.

pkg/server/plugin/keymanager/hashicorpvault/vault_client.go
```diff
@@ -0,0 +1,554 @@
...
+	if !ok {
+		return nil, status.Errorf(codes.Internal, "transit engine get key call was successful but keys are missing")
+	}
+
+	keyMap, ok := keys.(map[string]interface{})
+	if !ok {
+		return nil, status.Errorf(codes.Internal, "expected key map data type %T but got %T", keyMap, keys)
+	}
+
+	// TODO: Should we support multiple versions of the key?
```
amartinezfayo, line 542: I think that SPIRE will always manage a single version of the key, creating a new key when a rotation happens. We need to make sure that keys that have been rotated are being deleted.

pkg/server/plugin/keymanager/hashicorpvault/hashicorp_vault.go
```diff
@@ -0,0 +1,490 @@
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	spireKeyID := req.KeyId
+	newKeyEntry, err := p.createKey(ctx, spireKeyID, req.KeyType)
+	if err != nil {
+		return nil, err
+	}
+
+	p.entries[spireKeyID] = *newKeyEntry
```
amartinezfayo, line 297: It seems that all keys with the same SPIRE Key ID have the same path.
I think that the logic here should be that if we already have this SPIRE Key ID cached, a new key is created and the old one should be deleted.
I've added a comment about my concern about the current implementation that doesn't seem to be creating a new key, since the same path is always used for a given SPIRE Key ID.

amartinezfayo (commented):
Thank you @InverseIntegral for adding the integration tests, and sorry for the delay in the review.
I've added some comments. My main concern is around the handling of the creation of new keys (and the management of the key lifecycle in general), noticing that the path for a given SPIRE Key ID (e.g.: x509-CA-A) is always the same. It seems that the keys may not be rotated?

pkg/server/plugin/keymanager/hashicorpvault/vault_client.go
```diff
@@ -0,0 +1,554 @@
...
+		vc.HttpClient = vapi.DefaultConfig().HttpClient
+	}
+	clientTLSConfig := vc.HttpClient.Transport.(*http.Transport).TLSClientConfig
+
+	var clientCert tls.Certificate
+	foundClientCert := false
+
+	switch {
+	case c.clientParams.ClientCertPath != "" && c.clientParams.ClientKeyPath != "":
+		c, err := tls.LoadX509KeyPair(c.clientParams.ClientCertPath, c.clientParams.ClientKeyPath)
```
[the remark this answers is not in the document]
InverseIntegral: Good idea, fixed it

pkg/server/plugin/keymanager/hashicorpvault/hashicorp_vault.go
InverseIntegral: Added a warning and adjusted the docs

pkg/server/plugin/keymanager/hashicorpvault/vault_client.go
```diff
@@ -0,0 +1,554 @@
...
+
+	return nil
+}
+
+// ConfigureTLS Configures TLS for Vault Client
+func (c *ClientConfig) configureTLS(vc *vapi.Config) error {
+	if vc.HttpClient == nil {
+		vc.HttpClient = vapi.DefaultConfig().HttpClient
+	}
+	clientTLSConfig := vc.HttpClient.Transport.(*http.Transport).TLSClientConfig
```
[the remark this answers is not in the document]
InverseIntegral: Done :+1:

pkg/server/plugin/keymanager/hashicorpvault/vault_client.go
InverseIntegral: In that case I remove the TODO.

pkg/server/plugin/keymanager/hashicorpvault/vault_client.go
InverseIntegral: I see, in that case I will remove this TODO. Thanks for the clarification :slightly_smiling_face:

pkg/server/plugin/keymanager/hashicorpvault/hashicorp_vault.go
amartinezfayo: I think that we should first create the key, and then schedule the old one for deletion if exists. Otherwise, if the deletion happens first and it fails, the key is deleted and there will be no key that replaces it, which is a pretty bad situation.

We need to consider also that in an HA deployment, multiple SPIRE servers will be accessing the same Vault instance, so there should be a way to identify each server. There should be a `KeyIdentifierValue` config attribute as we have in other key managers for this purpose. With that attribute, the server ID can be persisted and retrieved. The plugin needs that to know which keys are managed by this specific instance of SPIRE Server.

InverseIntegral: @amartinezfayo I've implemented it as you suggested. Now, a key is scheduled for deletion and each key has a unique id associated to it.

The `KeyIdentifierValue` config attribute is also there. Unfortunately, HashiCorp Vault doesn't support tags on keys. So we need a different way to associate keys with SPIRE servers. Perhaps a clever naming schema could solve this problem.

amartinezfayo: >The KeyIdentifierValue config attribute is also there. Unfortunately, HashiCorp Vault doesn't support tags on keys.

This is unfortunate, because it makes difficult the identification and disposal of stale keys.
I think that we can embed the KeyIdentifierValue as part of the key name. I also think that it would be useful to be able to identify all the keys belonging to a trust domain, but the user could do that through the use of specific transit engine paths for each trust domain.
Prefixing the key names with "spire-key" is probably a good idea, to easily identify that they are SPIRE keys.

pkg/server/plugin/keymanager/hashicorpvault/hashicorp_vault.go
```diff
@@ -0,0 +1,675 @@
...
+	}
+
+	serverID := config.KeyIdentifierValue
+	if serverID == "" {
+		var err error
+
+		if serverID, err = getOrCreateServerID(config.KeyIdentifierFile); err != nil {
+			return nil, err
+		}
+	}
```
amartinezfayo, line 183: We need to validate that the configuration has a key identifier file or a key identifier value, and that the configuration does not have a key identifier file and a key identifier value at the same time. You can look at how this is validated in other key managers that have those settings.

manuelcoppotelli:
Hello,

this PR was initially scheduled for [milestone 1.13.0](https://github.com/spiffe/spire/milestone/63) milestone (i.e. Jun 2025)
and is costantly being postponed (to Feb 2026)

Is there something missing to test or that the opensourece community can support?

sorindumitru:
> Is there something missing to test or that the opensourece community can support?

Sorry about the late reply here. There's a bit of duplicated code between the two vault plugins, see for example https://github.com/spiffe/spire/pull/5500/changes#diff-15071e1bb2c2bbe9e21d5453eb410b7fc833a27f22b06ad1964bcf5b772d1968. We should move this to some common shared code that we can renew. It has also been observed that the renewal process can be somewhat flaky, it would need some attention to see if it can be replace with something more maintainable.

@manuelcoppotelli would you be interesting in working on that?

sorindumitru:
Closing in favour of https://github.com/spiffe/spire/pull/6889, which is a cleaned up version of this PR

PR #5506, opened 2024-09-20
Return InvalidArgument for invalid input entries

### LINKED ISSUE #5444: spire-server: Better status codes from BatchCreateEntry API
The BatchCreateEntry currently returns some status codes, InvalidArgument, InternalError (for most database issues), AlreadyExists. If I try to create an entry with some invalid parameters, e.g. entry id too long or empty list of selectors, I get back an InternalError status code. This makes it hard to differentiate between user errors (invalid entry) and database entries. I'd like to be able to differentiate between them to know if I need to retry creating a registration entry at a later time or know that it will never work.

* **Version**: 1.10.1
* **Platform**: linux
* **Subsystem**: server

sorindumitru:
- [ ] Documentation updated?

**Affected functionality**
Returned error codes for BatchCreateEntry/BatchUpdateEntry

**Description of change**
Return better error codes in case of invalid entries so users can differentiate between entries that can be retried and those that can't

**Which issue this PR fixes**
fixes #5444


pkg/server/api/entry/v1/service.go
```diff
@@ -658,6 +755,13 @@ func (s *Service) updateEntry(ctx context.Context, e *types.Entry, inputMask *ty
 			Hint:          inputMask.Hint,
 		}
 	}
+
+	if err := validateRegistrationEntryForUpdate(convEntry, mask); err != nil {
```
sorindumitru, line 759: An alternative to this would be to add all the validations here in https://github.com/spiffe/spire/blob/3d158ce5635ab92fc5869b4ecda44c6a153b2a06/pkg/server/api/entry.go#L92 but I wasn't sure that's where the maintainers would want to add these checks.

That function already does some validations, but mostly as a necessity of handling errors. It would also simplify things a bit. Let me know if that's better and I can change this.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -473,12 +474,11 @@ func (ds *Plugin) CreateOrReturnRegistrationEntry(ctx context.Context,
 func (ds *Plugin) createOrReturnRegistrationEntry(ctx context.Context,
 	entry *common.RegistrationEntry,
 ) (registrationEntry *common.RegistrationEntry, existing bool, err error) {
-	// TODO: Validations should be done in the ProtoBuf level [https://github.com/spiffe/spire/issues/44]
```
sorindumitru, line 476: I don't think we'll ever do this, so better to remove the TODO.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -473,12 +474,11 @@ func (ds *Plugin) CreateOrReturnRegistrationEntry(ctx context.Context,
...
-	if err = validateRegistrationEntry(entry); err != nil {
-		return nil, false, err
-	}
-
 	if err = ds.withWriteTx(ctx, func(tx *gorm.DB) (err error) {
+		if err = validateRegistrationEntry(entry); err != nil {
+			return err
+		}
```
sorindumitru, line 480: I've moved this here because errors withing a `withTx` block get wrapped into a `grpc.Status` (I don't know exactly why, maybe it's due to the plugin's history as an actual plugin). Having it in here makes it the handling for create and update be the same. Otherwise one would be wrapped in a grpc Status and one wouldn't so we'd need to handle them differently. I hope this doesn't change observed behaviour by clients too much.

amartinezfayo (approved):
Thank you @sorindumitru!

PR #5537, opened 2024-10-04
Fix undesirable regex pattern compilation approach.

### LINKED ISSUE #5515: Better RegEx pattern complication handling in httpchallenge
In pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go, near line 67 the following code exists

> re := regex.MustCompile(r)

which will cause a runtime panic on a poorly written `hclConfig.AllowedDNSPatterns`, which comes from the httpchallenge user-edited config file.

We need better error handling in this area, to prevent the httpchallenge plugin from runtime panicing.

Discovered during #5303 but not changed within that PR as it is out of scope of adding a Validate gRPC call.

edwbuck:
closes #5515


azdagron (approved):
Thanks, @edwbuck.

PR #5560, opened 2024-10-09
Tainted authorities in the agent cache: use the bundles from the cache update

amartinezfayo:
When processing tainted authorities in the agent cache, use the bundles from the cache update, not the initial bundles.


MarcosDY (approved):
LGTM!!!
And thanks for catching this!!!

Issue #5206, opened 2024-06-10
The Bottom Turtle Reference Architecture(s)


kfox1111:
It should be possible to use SPIRE as the bottom turtle for security. In order to do so, there has to be one or more deploy-able, maintainable, scalable, and documented SPIRE architectures that do not rely on 3rd party roots of trust as part of the network communications/attestations.

The easiest to use multinode setup currently is the helm charts. The helm chart project has multiple documented reference architectures for SPIRE. But all of them rely on the Kubernetes clusters preestablished control plane/node trust. So SPIRE isn't the bottom turtle in those environments. The K8s CA is.

We need:
- [ ] One or more examples, from the ground up, that is able to establish the bottom turtle(s) in an internet disconnected environment
- [ ] Procedures around establishing trust and reestablishing trust when broken (ex, CA rotation too fast and nodes powered off for to long)
- [ ] Procedures for scaling up the number of nodes in that environment without a lot of work
- [ ] A set of issues to raise for features that will make the use case of those architectures better

kfox1111:
For sake of discussion, what could be done with a set of RPI's with some kind of TPM, like:
https://wiki.52pi.com/index.php/EP-0149

kfox1111:
Kubelet is gaining the ability to refresh server certs, merged but not released yet:
https://github.com/kubernetes/kubernetes/pull/124574

client auth can be done via jwt token.

No updating of CA's yet though.

anvega:
Thank you for raising this. I'm currently exploring alongside others the possibilities of using [OpenTitan](https://opentitan.org/) as the silicon root of trust to anchor and bootstrap trust.

Although my exploration is ongoing, I'm eager to collaborate and share my findings.

edwbuck:
> For sake of discussion, what could be done with a set of RPI's with some kind of TPM, like: https://wiki.52pi.com/index.php/EP-0149

I'm confused about the focus of the request, as using Raspberry PI TPMs is a deployment detail, not an architecture (at least in my mind).

If support for the "Infineon Optiga™ SLB 9670 TPM 2.0" is missing, and a pre-requisite for this effort, please consider handling that missing pre-req in a different issue (and linking the two).

kfox1111:
@edwbuck  For example, see:
https://www.hpe.com/psnow/doc/a00020437enw?jumpid=in_pdfviewer-psnow, page 4, "Reference Configuration overview" or page 5, "Hardware"

They go all the way down to an example of workable hardware in their reference.

[...]

PR #5572, opened 2024-10-13
x509pop server plugin support for servers trust bundle

kfox1111:
Enables the x509pop node attestor server plugin to be configured to use the SPIRE Servers own trust bundle.


pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -97,14 +106,51 @@ type Plugin struct {
...
 	return &Plugin{}
 }
 
+func (p *Plugin) BrokerHostServices(broker pluginsdk.ServiceBroker) error {
+	if !broker.BrokerClient(&p.identityProvider) {
+		return status.Errorf(codes.FailedPrecondition, "IdentityProvider host service is required")
+	}
+	go func() {
+		ctx := context.Background()
+		sleep := 1 * time.Second
```
azdagron, line 125: Rather than kick off a goroutine that may or may not be run after the IdentityProvider host service is ready, it would be better to lazily initialize this in the first call to Attest.

kfox1111:
> Using the bundle, without any sort of restriction, seems scary. Doesn't this imply that any workload can just turn around and attest as a node?

Yeah, that I think can be handled like:
```
    NodeAttestor "x509pop" {
      plugin_data {
        spire_trust_bundle = true
        # Only allow spiffe identifiers of the form spiffe://<trustDomain>/k8s-spire-agent-helper/<hostname>
        # Generate identifiers that look like spiffe://<trustDomain>/x509pop/k8s-spire-agent/<hostname>
        agent_path_template = "{{ ${d}p := printf \"spiffe://%s/k8s-spire-agent-helper/\" .TrustBundle }}{{ ${d}s := printf \"%s\" (index .Certificate.URIs 0) }}{{ if gt (len ${d}s) (len ${d}p) }}{{ ${d}ps := slice ${d}s 0 (len ${d}p)
 }}{{ if eq ${d}ps ${d}p }}{{ printf \"/x509pop/k8s-spire-agent/%s\" (slice ${d}s (len ${d}p)) }}{{ end }}{{ end }}"
      }
    }
```

if it agent_path renders out to "", then it should be disallowed.

I did hit:
* https://github.com/spiffe/spire/issues/5574
* https://github.com/spiffe/spire/issues/5573

Trying to get it to work though.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
kfox1111: Done

azdagron:
Hey @kfox1111, do you have an arch diagram or something that spells out the use case clearly that you can share (either here or privately in slack with the maintainer group)? Before we consider taking this we'd like to see if there are alternatives.

kfox1111:
Here's what I have done / working on doing. Got the host -> k8s part working. Still working on the k8s -> vm part.
x509pop

kfox1111:
Related to: https://github.com/spiffe/spire/issues/5206

kfox1111:
Used with the sprig pr, this works:

        agent_path_template = "{{ $$p := printf \"spiffe://%s/k8s-spire-agent-helper/\" .TrustDomain }}{{ $$s := printf \"%s\" (index .Certificate.URIs 0) }}{{ if hasPrefix $$p $$s }}{{ printf \"/x509pop/k8s-spire-agent/%s\" (trimPrefix $$p $$s) }}{{ else }}{{ fail \"Invalid prefix\" }}{{ end }}"

kfox1111:
Updated diagram:
x509pop

kfox1111:
@sorindumitru I think we're pretty much on the same page.

We could make a new x509svidpop plugin but it would end up being almost the same code as the x509pop one. (The patch here is pretty small). So a lot more stuff to maintain.

the svid use case really isn't that different from the normal x509pop one. Either way you probably want some way to filter out which certs are allowed to a subset of all possible ones. That can be done via agent_path_template as done in this pr without any special code in the plugin.

100% agree on the future ability to use some other spire trustDomain too. Was going to propose that in a different patch. The primary use case for me being:

k8s multitenant cluster. Secured with spire. Call this, the resource provider cluster.

Tenant uses resource provider cluster with kubevirt to launch a bunch of vm's for their tenant. Inside, they launch their own spire server, and workloads (maybe even their own k8s cluster in the vms). It would greatly simplify management if they could use x509 certs from the resource provider cluster to node attest to their own spire-server.

That again, would basically be this same x509pop plugin as described here, but getting its trustBundle from disk and keep it refreshed (existing functionality extended to refresh from disk), or get it over vsock (not sure which is best yet)

sorindumitru:
I agree it's going to be very similar to `x509pop`, but dealing with that is an implementation detail, I'm sure some parts can be shared. I just don't think that the template language is a good way of dealing with those differences.

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -18,9 +18,11 @@ spiffe://<trust_domain>/spire/agent/x509pop/<fingerprint>
...
 |-----------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------|
+| `spire_trust_bundle`     | If true, use the spire servers own trust bundle to use for validation.                                                                                                                                                                      |                                         |
+| `svid_prefix`            | The prefix of the SVID to use for matching vaid SVIDS and exchanging them for Node SVIDs                                                                                                                                                    | /spire-exchange                         |
```
azdagron, line 22: Maybe `svid_path_prefix`? Also...
```suggestion
| `svid_prefix`            | The prefix of the SVID to use for matching valid SVIDS and exchanging them for Node SVIDs                                                                                                                                                    | /spire-exchange                         |
```

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -18,9 +18,11 @@ spiffe://<trust_domain>/spire/agent/x509pop/<fingerprint>
...
 | `ca_bundle_paths`     | A list of paths to trusted CA bundles on disk. The files must contain one or more PEM blocks forming the set of trusted root CA's for chain-of-trust verification.                                                                             |                                         |
-| `agent_path_template` | A URL path portion format of Agent's SPIFFE ID. Describe in text/template format.                                                                                                                                                              | `"{{ .PluginName}}/{{ .Fingerprint }}"` |
+| `agent_path_template` | A URL path portion format of Agent's SPIFFE ID. Describe in text/template format.                                                                                                                                                              | `See Agent Path Template for details`   |
```
azdagron, line 25: nit: use a section link to the correct section (https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#section-links)

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -43,9 +45,27 @@ A sample configuration:
 | SHA1 Fingerprint | `x509pop:ca:fingerprint:0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33` | The SHA1 fingerprint as a hex string for each cert in the PoP chain, excluding the leaf. |
 | SerialNumber     | `x509pop:serialnumber:0a1b2c3d4e5f`                               | The leaf certificate serial number as a lowercase hexadecimal string                     |
 
+## SVID Prefix
```
azdagron, line 48:
```suggestion
## SVID Path Prefix
```

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -57,3 +77,4 @@ Some useful values are:
 | .TrustDomain          | The configured trust domain                                                                  |
 | .Subject.CommonName   | The common name field of the agent's x509 certificate                                        |
 | .SerialNumberHex      | The serial number field of the agent's x509 certificate represented as lowercase hexadecimal |
+| .SVIDPath             | The SVID Path after removing the SVID Prefix                                                 |
```
azdagron, line 80: Maybe `SVIDPathSuffix` better implies that a prefix has been stripped?

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -35,15 +38,19 @@ func builtin(p *Plugin) catalog.BuiltIn {
 }
 
 type Config struct {
+	SPIRETrustBundle  bool     `hcl:"spire_trust_bundle"`
+	SPIFFEPrefix      *string  `hcl:"spiffe_prefix"`
```
azdagron, line 42: SVIDPrefix? Also, if we don't need to distinguish between this flag being unset and "" (e.g. if "" can safely imply unset), I'd suggest this be of type `string` to reduce pointery-ness.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -35,15 +38,19 @@ func builtin(p *Plugin) catalog.BuiltIn {
...
 	CABundlePaths     []string `hcl:"ca_bundle_paths"`
 	AgentPathTemplate string   `hcl:"agent_path_template"`
 }
 
 type configuration struct {
-	trustDomain  spiffeid.TrustDomain
-	trustBundle  *x509.CertPool
-	pathTemplate *agentpathtemplate.Template
+	SPIRETrustBundle bool
+	SPIFFEPrefix     string
```
azdagron, line 50: To be consistent, unexport these.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -84,10 +96,22 @@ func buildConfig(coreConfig catalog.CoreConfig, hclText string, status *pluginco
 		pathTemplate = tmpl
 	}
 
+	var SPIFFEPrefix string
```
azdagron, line 99: local variable naming conventions are `lowerCamel`
```suggestion
	var spiffePrefix string
```

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -188,7 +228,20 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
 		return status.Errorf(codes.PermissionDenied, "challenge response verification failed: %v", err)
 	}
 
-	spiffeid, err := x509pop.MakeAgentID(config.trustDomain, config.pathTemplate, leaf)
+	SVIDPath := ""
```
azdagron, line 231:
```suggestion
	svidPath := ""
```

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -188,7 +228,20 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
...
+	if config.SPIRETrustBundle {
+		if len(leaf.URIs) >= 1 {
```
azdagron, line 233: nit: if you invert the condition, you can do the early failure return when there are no URIs and then not need an else (reduces indentation)

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -53,29 +60,34 @@ func buildConfig(coreConfig catalog.CoreConfig, hclText string, status *pluginco
...
-	if len(caPaths) == 0 {
-		status.ReportError("one of ca_bundle_path or ca_bundle_paths must be configured")
-	}
-
 	var trustBundles []*x509.Certificate
-	for _, caPath := range caPaths {
-		certs, err := util.LoadCertificates(caPath)
-		if err != nil {
-			status.ReportErrorf("unable to load trust bundle %q: %v", caPath, err)
+	if !hclConfig.SPIRETrustBundle {
```
azdagron, line 64: Should we be doing sanity checking that folks aren't using both "spire_trust_bundle" and the other ca related flags?

azdagron (commented):
Thanks for this, @kfox1111. The new design eliminates most of my concerns.

I'd like to explore names other than `spire_trust_bundle` and the default `/spire-exchange` prefix value.... I think we can find something a little clearer....

Also, there are no tests for the new functionality. We'll need those before we can merge.

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -18,9 +18,11 @@ spiffe://<trust_domain>/spire/agent/x509pop/<fingerprint>
...
 | `ca_bundle_paths`     | A list of paths to trusted CA bundles on disk. The files must contain one or more PEM blocks forming the set of trusted root CA's for chain-of-trust verification.                                                                             |                                         |
-| `agent_path_template` | A URL path portion format of Agent's SPIFFE ID. Describe in text/template format.                                                                                                                                                              | `"{{ .PluginName}}/{{ .Fingerprint }}"` |
+| `agent_path_template` | A URL path portion format of Agent's SPIFFE ID. Describe in text/template format.                                                                                                                                                              | `See Agent Path Template for details`   |
```
kfox1111, line 25:
```suggestion
| `agent_path_template` | A URL path portion format of Agent's SPIFFE ID. Describe in text/template format.                                                                                                                                                              | `See [Agent Path Template](#agent-path-template) for details`   |
```

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -53,29 +60,44 @@ func buildConfig(coreConfig catalog.CoreConfig, hclText string, status *pluginco
...
-
 	var trustBundles []*x509.Certificate
-	for _, caPath := range caPaths {
-		certs, err := util.LoadCertificates(caPath)
-		if err != nil {
-			status.ReportErrorf("unable to load trust bundle %q: %v", caPath, err)
+	if hclConfig.Mode == "external_pki" {
+		var caPaths []string
+		if hclConfig.CABundlePath != "" && len(hclConfig.CABundlePaths) > 0 {
+			status.ReportError("only one of ca_bundle_path or ca_bundle_paths can be configured, not both")
```
MarcosDY, line 73: is not the same error happens in line 80?
may we return here instead?

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -84,10 +106,22 @@ func buildConfig(coreConfig catalog.CoreConfig, hclText string, status *pluginco
...
 
+	var svidPrefix string
+	if hclConfig.SVIDPrefix == nil {
+		svidPrefix = "/spire-exchange/"
+	} else {
+		svidPrefix = *hclConfig.SVIDPrefix
+		if !strings.HasSuffix(svidPrefix, "/") {
+			svidPrefix += "/"
+		}
+	}
```
MarcosDY, line 117: this can be simplified:
```suggestion
	svidPrefix :=  "/spire-exchange/"
	if hclConfig.SVIDPrefix != nil {
		svidPrefix = *hclConfig.SVIDPrefix
		if !strings.HasSuffix(svidPrefix, "/") {
			svidPrefix += "/"
		}
	}
```

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -226,6 +288,25 @@ func (p *Plugin) Validate(_ context.Context, req *configv1.ValidateRequest) (*co
...
+		certificates, err := x509.ParseCertificates(rawcert.Asn1)
+		if err != nil {
+			return nil, err
+		}
+		trustBundles = append(trustBundles, certificates...)
+	}
+	if len(trustBundles) > 0 {
+		return util.NewCertPool(trustBundles...), nil
+	}
+	return nil, nil
```
MarcosDY, line 307: if this happens we'll reject attestations, but it is going to be hard to understand why it is happening (basically we are not going to have a trusted bundle)
May we at least return wartning or error?

pkg/server/plugin/nodeattestor/x509pop/x509pop_test.go
```diff
@@ -158,7 +160,9 @@ func (s *Suite) TestAttestFailure() {
 
 	s.T().Run("not configured", func(t *testing.T) {
 		attestor := new(nodeattestor.V1)
-		plugintest.Load(t, BuiltIn(), attestor)
+		plugintest.Load(t, BuiltIn(), attestor,
```
MarcosDY, line 163: can you add test case where no trustedbundle is returned? 
and how about defaults?
are

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
kfox1111: Hmm.... The check here checks if both are configured. The check at 80 ensures at least one is configured. Both are needed I think.

I was thinking status.ReportError was fatal. Is it not? Maybe all the ReportError's here then need retruns?

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
kfox1111: Maybe we log a warning about not currently finding a trust bundle? Its nor fatal as it might show up at some point later?

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
MarcosDY: mmm, it is like if both are configured, you must never get the second one right?
status.Report is not a fatal, but just adding a list of errors, than then are used in configure or verify,
but this approach has a good win than you can get the complete list of errors to resolve all together,
but it makes very hard to follow and see dependency errors...
In any case we are fine for now

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
MarcosDY: yeap, I agree but this attestation will fail and mention a verification error
so a warning may help users to understand than that the that attestation failure is cuased by the attestor being unabled to get the trust bundle.

kfox1111: Added a warning.

pkg/server/plugin/nodeattestor/x509pop/x509pop_test.go
kfox1111: Struggling a bit with how to write the tests....

Seems like I need to create a new cert in ./test/fixture/nodeattestor/x509pop/generate.go with the url based spiffe id to test with? There are some magic numbers in that file. Are they just arbitrary, or are there some reason for the specific numbers chosen?

kfox1111: Reading through it some more, it looks like they just have patterns in it. so rather arbitrary.

kfox1111: Ok. I have tests for no trust bundle, along with a valid svid, an invalid svid, and a no svid.

MarcosDY (approved):

PR #5576, opened 2024-10-15
Allow escaping $ in config files that use expand env

kfox1111:
Fixes https://github.com/spiffe/spire/issues/5574.

$$ will match an env var named '$' which resolves to '$', allowing passing through '$'s.


azdagron (commented):
SPIRE for better or worse accommodates a bring-your-own-main approach so `main` isn't quite the right place for this. Further, we can leverage the [os.Expand] function instead of os.ExpandEnv to avoid having to set/unset an environment variable.

I'd suggest adding a function in pkg/common/config or something called ExpandEnv that:
1. Called os.Getenv to gather the current environment
2. Added the '$' entry
3. called os.Expand

This function could then be called in place of os.ExpandEnv in the agent/server run packages.

PR #5565, opened 2024-10-10
Forced rotation:  Remove cached JWT-SVIDs using tainted keys

amartinezfayo:
Remove from the agent cache the JWT-SVIDs using tainted keys

Fixes: #3904.

PR #5583, opened 2024-10-15
Force rotation: adds an integration test to verify that the JWT autho…

### LINKED ISSUE #5582: Integration test: Force rotation of JWT Authorities

MarcosDY:
Adds an integration test to verify that the JWT authority correctly handles forced rotation. 
Ensures that JWT tokens are invalidated and reissued as expected.

Depends on #5565 

**Which issue this PR fixes**
fixes #5582


amartinezfayo (approved):
🎉

PR #5613, opened 2024-10-30
Fixes how upstream-authority-ejbca integration test calls kubectl and helm

nweisenauer-sap:
**Pull Request check list**

- [ x ] Commit conforms to CONTRIBUTING.md?
- [ / ] Proper tests/regressions included?
- [ / ] Documentation updated?

**Affected functionality**
upstream-authority-ejbca test now calls the binaries from the ./bin folder

**Description of change**
On our runners, the new integration test `upstream-authority-ejbca` fails, complaining that `helm` and `kubectl` are not preinstalled. However, this test, similar to the `upstream-authority-cert-manager`, downloads these tools beforehand (inside the `00-setup-kind script`). The issue is that the script downloads these tools to the `./bin` folder, which is fine, but the test itself does mixed calls of `kubectl`, sometimes using `kubectl`, sometimes `./bin/kubectl`. Also, the prerequisite checks do not check for the `./bin` folder.
**Which issue this PR fixes**


amartinezfayo (approved):
Thank you @nweisenauer-sap!

PR #5621, opened 2024-10-31
Resolve flakiness in Evit agent IT

MarcosDY:
This change addresses an issue where the IT attempts to ban an agent that has not yet been initialized. A retry mechanism is now implemented to ensure the agent is fully registered before eviction is attempted.


amartinezfayo (approved):
Thank you @MarcosDY!

amartinezfayo (approved):

PR #5626, opened 2024-11-04
cli command to validate configuration using plugin validation routines.

### LINKED ISSUE #5268: Rework spire-server validate command to launch plugin instances for validation.
Currently the SPIRE server has a command to validate the configuration file.  This request would extend the validation to have per-plugin validation be part of the process, using the newly added "Validate" RPC provided by the spire-plugin-sdk "config" service.

Upon running the command, the logic would launch (in a non-conflicting manner) an instance of each plugin referenced in the configuration and pass the configuration to the plugin's `Validate` func.

After the core configuration an all plugins report valid / invalid status, the results would be combined for feedback to the administrator that validated the configuration.

Validation is limited to ensuring that the correct fields are present, they have values of the correct types, and where they must agree with each other internally, they agree.

Validation doesn't extend to anything that requires the additional reads of config files not related to SPIRE, opening sockets, or communicating to other systems.

edwbuck:
Closes #5268


evan2645:
@edwbuck we are reviewing 1.11.1 release and see that we will need to call a commit for that around thanksgiving time. This PR is tagged for that release, however it is still in draft and seems it needs more work. We will also need time to review once it comes out of draft. When do you think that will happen?

edwbuck:
> @edwbuck we are reviewing 1.11.1 release and see that we will need to call a commit for that around thanksgiving time. ...

Hi Evan.  I'll put some more effort into it today.  I was going to revisit it this weekend, but I have a few hours to start early.  I am still planning to be ready for the Dec 5 deadline (with extra time for the review process) but if it's not in good shape by the end of this weekend, I'll express it in the next contributor's call.

edwbuck:
A forgotten code signoff in the middle of the MR makes this branch difficult, if not impossible to move forward (it's not a matter of just rewriting my history, but also the history of two main branch merge requests).  

Please close this and proceed with its replacement #5665

amartinezfayo:
Closing in favor of #5665.

PR #5657, opened 2024-11-20
Configurable JWT Issuer for the OIDC Discovery Provider

aluthra-37:
**Pull Request check list**
- [x] Documentation updated?

**Affected functionality**
Added a config item to the OIDC provider (jwt_issuer). This config specifies the issuer for the OIDC provider configuration request. Previously, only the request host could be the issuer. This change allows users to define a different JWT issuer with a path.

**Description of change**
- Updating OIDC provider config with new field (jwt_issuer)
- Added validation for new config
- Updated logic in handler to use newly created config in OIDC configuration response if config is defined.

**Which issue this PR fixes**
Fix for https://github.com/spiffe/spire/issues/5480


rturner3 (approved):

kfox1111:
This change has a bug....

When setting the issuer:
```
  "jwt_issuer": "https://oidc-discovery-provider.example.org",
```

The generated oidc config:
```
# curl https://oidc-discovery-provider.example.org:8181/.well-known/openid-configuration  -k
{
  "issuer": "https://oidc-discovery-provider.home.efox.cc",
  "jwks_uri": "https://oidc-discovery-provider.home.efox.cc/keys",
<SNIP>
```

Does not have the contacting port in the jwks_url, so it is invalid. The issuer override should only apply to issuer.

kfox1111:
It also now ignores the hostname contacted in favor of the hostname in jwt_issuer making the domains allowlist only work with the jwt_issuer as the entry. This does not seem to be correct behavior.

I think we need to talk more about the indended use cases. For my use case, I only need issuer set to jwt_issuer and nothing more.

PR #5673, opened 2024-11-28
Enable optional trust domain label for all metrics

### LINKED ISSUE #5652: Enable optional trust domain label for all metrics
From the [telemetry doc](https://github.com/spiffe/spire/blob/v1.10.4/doc/telemetry/telemetry.md), there are some labels that accompany metrics, we would find it useful in our deployment to be able to set a custom label to distinguish different sets of spire-servers from each other.

To this end, we could enable the use of ServiceName and EnableServiceLabel (which is from [go-metric config](https://pkg.go.dev/github.com/hashicorp/go-metrics#Config)) in [metrics.go](https://github.com/spiffe/spire/blob/a8857bafdaee0ad84c8063cfaead2c4e75c0721d/pkg/common/telemetry/metrics.go#L54),  which would give the ability to optionally pass a ServiceName and enable that ServiceName to be turned to a label. Looking at go-metric's [implementation](https://github.com/hashicorp/go-metrics/blob/c5fef7bda84c66401ad068056b262e594c2dcd1c/metrics.go#L36), it looks like it always uses "service" as the label. It would be nice if we could get it to be some kind of custom label, for example support for global labels could be introduced in [metrics.go](https://github.com/spiffe/spire/blob/main/pkg/common/telemetry/metrics.go).

I would be happy to implement this if it gets accepted.

graceajibade:
This PR is based on this [comment](https://github.com/spiffe/spire/issues/5652#issuecomment-2501830354) and the desire for a different tag name than `service` which is what [go-metrics](https://github.com/hashicorp/go-metrics/blob/c5fef7bda84c66401ad068056b262e594c2dcd1c/metrics.go#L36) would use as a label if `ServiceName` and `EnableServiceLabel` were used.

The `trust_domain` label was used to separate it from the `trust_domain_id` label that is already added for some metrics.

Since this label is enabled for all metrics, the best place to add it looks to be in each of the metric function in metric.go.

Fixes #5652


pkg/common/telemetry/metrics.go
```diff
@@ -83,12 +84,19 @@ func NewMetrics(c *MetricsConfig) (*MetricsImpl, error) {
...
 		conf.AllowedLabels = c.FileConfig.AllowedLabels
 		conf.BlockedLabels = c.FileConfig.BlockedLabels
 		conf.AllowedPrefixes = c.FileConfig.AllowedPrefixes
 		conf.BlockedPrefixes = c.FileConfig.BlockedPrefixes
 
+		if c.FileConfig.EnableTrustDomainLabel != nil {
+			impl.enableTrustDomainLabel = *c.FileConfig.EnableTrustDomainLabel
+		} else {
+			impl.enableTrustDomainLabel = false
+		}
```
MarcosDY, line 98: since default is false you can just avoid else, or set 
```
impl.enableTrustDomainLabel = false
if c.FileConfig.EnableTrustDomainLabel != nil {
	impl.enableTrustDomainLabel = *c.FileConfig.EnableTrustDomainLabel
} 
```

pkg/common/telemetry/metrics.go
```diff
@@ -132,41 +148,65 @@ func (m *MetricsImpl) EmitKey(key []string, val float32) {
...
 	for _, s := range m.metricsSinks {
 		s.IncrCounterWithLabels(key, val, sanitizedLabels)
 	}
 }
 
 func (m *MetricsImpl) AddSample(key []string, val float32) {
-	for _, s := range m.metricsSinks {
-		s.AddSample(key, val)
+	if m.enableTrustDomainLabel {
+		m.AddSampleWithLabels(key, val, nil)
```
MarcosDY, line 174: what happens if we always call `m.AddSampleWithLabels(key, val, nil)`? even may that reduce the uses of all repeted ifs?

c4rlo:
Just mentioning here that this addresses issue #5652.

MarcosDY:
@graceajibade can you resolve DCO? 
thanks!

PR #5689, opened 2024-12-07
Enable expand env for oidc discovery provider

kfox1111:
Fixes: https://github.com/spiffe/spire/issues/5688

### DIFF
--- a/support/oidc-discovery-provider/config.go
+++ b/support/oidc-discovery-provider/config.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/hcl"
+	"github.com/spiffe/spire/pkg/common/config"
 	"github.com/zeebo/errs"
 )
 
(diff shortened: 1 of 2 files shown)


support/oidc-discovery-provider/config_test.go
```diff
@@ -23,14 +23,14 @@ func TestLoadConfig(t *testing.T) {
...
-	_, err := LoadConfig(confPath)
+	_, err := LoadConfig(confPath, false)
 	require.Error(err)
 	require.Contains(err.Error(), "unable to load configuration:")
 
 	err = os.WriteFile(confPath, []byte(minimalServerAPIConfig), 0600)
 	require.NoError(err)
 
-	config, err := LoadConfig(confPath)
+	config, err := LoadConfig(confPath, false)
```
rturner3, line 33: Can you also add a test case that expands environment variables defined in the config and verifies that the expansion works as expected?

kfox1111: yes. I'll add a test for it.

rturner3 (approved):

PR #5698, opened 2024-12-10
Update CHANGELOG with v1.11.1 release

rturner3:
Update CHANGELOG with v1.11.1 release.


MarcosDY (approved):

evan2645 (approved):

kfox1111:
I think the issue with the issuer config option for this release is a bigger concern then what is described here. Best option IMO would be to revert it for now. Second best would be to release anyway, but warn people not to use it at all, as it may have different behavior in the near future.

PR #5718, opened 2024-12-14
Svidstore gcp region

### LINKED ISSUE #5717: svidstore gcp_secretmanager Support regions
We need support to set the region the secret is created in due to regulations. 

add selector region. 

* **Version**: 1.11.1
* **Platform**: all
* **Subsystem**: gcp_secretmanager

jenshornan:
**Pull Request check list**

- [X ] Commit conforms to CONTRIBUTING.md?
- [ X] Proper tests/regressions included?
- [ X] Documentation updated?

**Affected functionality**

Now you are able to specify the region the secret gets created in

**Description of change**

Add new selector called region. If no region selector is given it will be the old behavior to create a global secret. 

**Which issue this PR fixes**

fixes #5717


doc/plugin_agent_svidstore_gcp_secretmanager.md
```diff
@@ -71,3 +71,4 @@ Selectors are used on `storable` entries to describe metadata that is needed by
 | `gcp_secretmanager:projectid`      | `gcp_secretmanager:projectid:some-project`                                       | x        | The Google Cloud project ID which the plugin will use Secret Manager       |
 | `gcp_secretmanager:role`           | `gcp_secretmanager:role:roles/secretmanager.viewer`                              | -        | The Google Cloud role id for IAM policy (serviceaccount required when set) |
 | `gcp_secretmanager:serviceaccount` | `gcp_secretmanager:serviceaccount:test-secret@test-proj.iam.gserviceaccount.com` | -        | The Google Cloud Service account for IAM policy (role required when set)   |
+| `gcp_secretmanager:regions`        | `gcp_secretmanager:serviceaccount:europe-north1,europe-west1`                    | -        | List of Google Cloud Region to create the secret in                        |
```
sorindumitru, line 74:
```suggestion
| `gcp_secretmanager:regions`        | `gcp_secretmanager:regions:europe-north1,europe-west1`                    | -        | List of Google Cloud Region to create the secret in                        |
```

jenshornan:
Sorry for failing test. Should be fixed now.

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
```diff
@@ -145,11 +145,7 @@ func (p *SecretManagerPlugin) PutX509SVID(ctx context.Context, req *svidstorev1.
 			Parent:   opt.parent(),
 			SecretId: opt.name,
 			Secret: &secretmanagerpb.Secret{
-				Replication: &secretmanagerpb.Replication{
-					Replication: &secretmanagerpb.Replication_Automatic_{
-						Automatic: &secretmanagerpb.Replication_Automatic{},
-					},
-				},
+				Replication: opt.replication,
```
sorindumitru, line 148: This is only done if the secret is not found. If a secret already exists or if the list of locations changed, then we wouldn't be updating the secret with the new list. We probably need to call `UpdateSecret` in those cases.

sorindumitru: Although I see that the replication cannot be changed after the secret has been created, based on the comments in https://pkg.go.dev/cloud.google.com/go/secretmanager@v1.14.3/apiv1/secretmanagerpb#Secret.

I don't know if deleting and recreating it would work, if it doesn't we'd need to at least log something to inform the user.

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
```diff
@@ -346,11 +343,44 @@ func optionsFromSecretData(selectorData []string) (*secretOptions, error) {
...
+	var replica *secretmanagerpb.Replication
+
+	if !ok {
+		replica = &secretmanagerpb.Replication{
+			Replication: &secretmanagerpb.Replication_Automatic_{
+				Automatic: &secretmanagerpb.Replication_Automatic{},
+			},
+		}
+	} else {
+		regionsSlice := strings.Split(regions, ",")
```
sorindumitru, line 357: Could you also add some validation to make sure there's at least one region in here? It seems that we need to provide at least on replica to `CreateSecret` for user managed replicas.

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
sorindumitru: @MarcosDY What do you think, should we attempt to delete and recreate? That might end up being bad user experience, i imagine the secret with the X509-SVID will dissapear for a short amount of time.

jenshornan: That is a possibility. I do think Replication field is immutable. So in the case someone change it we will need to destroy and then re-create the secret with the new location setting. 

Dont know do if appending more regions to the list of regions would be possible as an update...

Maybe easier to just destory and re-create in both cases? 

I can look in to it.

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
jenshornan: Sure will fix!

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
jenshornan: Sorry did not refresh so I missed you follow up comment will wait for a decision. Also to take in to consideration is that it might create IAM problems for the user.

jenshornan: Ok will update dockumentation.

pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud_test.go
```diff
@@ -460,6 +460,50 @@ func TestPutX509SVID(t *testing.T) {
...
 			},
 		},
+		{
+			name: "Add payload and create regional secret",
+			req: &svidstore.X509SVID{
+				SVID: successReq.SVID,
+				Metadata: []string{
+					"name:secret1",
+					"projectid:project1",
+					"regions:europe-north1",
```
MarcosDY, line 470: can you add a test case with multiple regions?

jenshornan: Fixed!

MarcosDY (approved):

jenshornan:
I ran the failing unit-test locally and it passes.

TestUpstreamProcessTaintedAuthority

Running tool: /opt/homebrew/bin/go test -timeout 30s -run ^TestUpstreamProcessTaintedAuthority$ github.com/spiffe/spire/pkg/server/ca/manager

ok  	github.com/spiffe/spire/pkg/server/ca/manager	0.750s

PR #2121, opened 2021-02-19
Timeout test execution at 60s

bradbl:
Signed-off-by: Brad Blackard <bradb@uber.com>

**Pull Request check list**

- [X] Commit conforms to CONTRIBUTING.md?

**Description of change**
Adds a go test flag to limit the amount of time the test execution of a package can take. 

**Which issue this PR fixes**
Addresses https://github.com/spiffe/spire/issues/2105.

PR #5725, opened 2024-12-17
Remove usage of deprecated grpc-go methods

### LINKED ISSUE #5152: Do not use grpc.Dial and deprecated DialOptions
SPIRE uses grpc.Dial and the deprecated DialOptions FailOnNonTempDialError, WithBlock, and WithReturnConnectionError.

grpc.Dial is a deprecated function that also creates the same virtual connection pool as grpc.NewClient. However, unlike grpc.NewClient, it immediately starts connecting and supports a few additional DialOptions that control this initial connection attempt. These are: WithBlock, WithTimeout, WithReturnConnectionError, and FailOnNonTempDialError.

FailOnNonTempDialError, WithBlock, and WithReturnConnectionError are three DialOptions that are only supported by Dial because they only affect the behavior of Dial itself. WithBlock causes Dial to wait until the ClientConn reports its State as connectivity.Connected. The other two deal with returning connection errors before the timeout (WithTimeout or on the context when using DialContext).

More information can be found in the [Anti-Patterns of Client creation](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md) document.

rturner3:
Replace usage of deprecated `grpc.Dial()`/`grpc.DialContext()` methods with `grpc.NewClient()`. Also remove usage of `grpc.WithBlock()`, `grpc.FailOnNonTempDialError()`, and `grpc.WithReturnConnectionError()` options.

The combination of these changes results in a couple behavioral changes when setting up gRPC clients:

1. gRPC will no longer dial when creating the client. Instead, it will wait until the client is used for the first time with an RPC invocation.

2. gRPC uses the DNS resolver by default when building the `*grpc.ClientConn` using `grpc.NewClient()`, whereas previously it used to resolve addresses the `passthrough` resolver by default. The result of this change in behavior is that for any invocations of `grpc.Dial()`/`grpc.DialContext()` that did not specify a URI scheme, gRPC now implicitly tries to resolve the address passed to `grpc.NewClient()` using DNS. This breaks some assumptions in the code. The workaround to preserve the previous address resolution behavior is to prepend addresses with no scheme defined with the resolver URI scheme `passthrough:`.

Also refactored some test-related code in `cmd/spire-server/cli/common` into a new `test/clitest` package, since it is not intended for use in production code.

Fixes #5152.


pkg/common/health/cache_test.go
```diff
@@ -80,25 +79,98 @@ func TestHealthFailsAndRecover(t *testing.T) {
...
 
-	t.Run("start successfully", func(t *testing.T) {
+	t.Run("fail to start initially", func(t *testing.T) {
 		// Wait for initial calls
-		select {
-		case <-waitFor:
-		case <-ctx.Done():
-			require.Fail(t, "unable to get updates because context is finished")
-		}
+		<-waitFor
```
MarcosDY, line 89: is no possible that we get into a blocked test here?
or the plan is to depends in test timeout instead?

test/clitest/common_windows.go
```diff
@@ -0,0 +1,15 @@
+//go:build windows
+
+package clitest
+
+var (
+	AddrArg         = "-namedPipeName"
+	AddrError       = "rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: open \\\\\\\\.\\\\pipe\\\\does-not-exist: The system cannot find the file specified.\"\n"
```
MarcosDY, line 7: nit: for simplification can you update this string to use "`"?

test/fakes/fakeentryclient/entryclient.go
```diff
@@ -49,7 +49,7 @@ func New(t *testing.T, trustDomain spiffeid.TrustDomain, ds datastore.DataStore,
 	service := entry.New(entry.Config{
 		TrustDomain: trustDomain,
 		DataStore:   ds,
-		//EntryFetcher: authorizedEntryFetcherWithCache,
+		// EntryFetcher: authorizedEntryFetcherWithCache,
```
MarcosDY, line 52: is this required? can you remove this comment instead?

cmd/spire-server/util/util_posix.go
```diff
@@ -18,13 +17,22 @@ func (a *Adapter) addOSFlags(flags *flag.FlagSet) {
...
+	// As a workaround, use the passthrough resolver to prevent using the DNS resolver.
+	// More context can be found in this issue: https://github.com/grpc/grpc-go/issues/1786#issuecomment-2114124036
+	return "unix:" + a.socketPath
 }
 
 func dialer(ctx context.Context, addr string) (net.Conn, error) {
-	return (&net.Dialer{}).DialContext(ctx, "unix", addr)
+	// This is an ugly workaround to circumvent grpc-go needing us to provide the resolver in the address
+	// in order to bypass DNS lookup, which is not relevant in the case of CLI invocation.
+	socketPathAddr := strings.TrimPrefix(addr, "unix:")
```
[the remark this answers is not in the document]
rturner3: Yes, it is related to the same issue as linked in the above method. I can also mention it in this comment too for clarity.

pkg/common/health/cache_test.go
rturner3: In the past we decided to always rely on `go test` timeouts rather than introducing test-specific timeouts, which might be unreliable. https://github.com/spiffe/spire/pull/2121

test/clitest/common_windows.go
rturner3: Sure, it was copy-pasted from another file that I moved here, but this is a good improvement for readability.

test/fakes/fakeentryclient/entryclient.go
rturner3: Not sure why this was commented out in the first place. I think my IDE formatting settings (`gofumpt`) made this change on my behalf, but I can remove the commented out line.

cmd/spire-server/cli/common/common_windows.go
```diff
@@ -1,25 +0,0 @@
...
-package common
-
-import (
-	"net"
-
-	"github.com/spiffe/spire/pkg/common/namedpipe"
-)
-
-var (
-	AddrArg         = "-namedPipeName"
```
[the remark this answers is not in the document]
rturner3: `test/clitest/common_windows.go`

test/clitest/common_windows.go
rturner3: Actually we cannot make this a raw string literal because it has an escape sequence at the end, which will not be interpreted correctly if we use a raw string literal.

MarcosDY (approved):

PR #5738, opened 2024-12-20
golangci-lint: add gofmt

c4rlo:
Note that unlike `goimports` (which we already use and which includes most of the `gofmt` functionality), this `gofmt` linter is equivalent to `gofmt -s` i.e. including some code simplifications.

Bring current code base into compliance with the new linter (by running `gofmt -w -s` on all affected files).


zmt:
We should consider https://github.com/mvdan/gofumpt.

amartinezfayo:
>We should consider https://github.com/mvdan/gofumpt.

gofumpt is an interesting option. I would go ahead and introduce the `gofmt` linter now and consider the use of gofumpt in the future.

amartinezfayo (approved):
Thank you @c4rlo!

PR #5739, opened 2024-12-20
golangci-lint: enable more govet checks and address issues

c4rlo:
It turns out that the `govet` (= `go vet`) linter has a few more tricks up its sleeve, you just need to enable them.

This find a couple of bugs in the tests which are also being fixed in this commit:

1. The spire-server tests for `BatchCreateFederatedBundle` and friends were accidentally not including JWT keys in the bundle they were testing. This ended up only affecting assertions on log message fields.

   The fix for this engendered a bit of refactoring to enable access to the required JWT struct conversion function.
2. The spire-server tests for the CA journal were _almost_ failing in their attempt to list CA journals; it ended up working anyway because a conversion between different struct types happened to be unnecessary because gorm could work with either one due to matching struct field names.


amartinezfayo (approved):
Thank you @c4rlo for this contribution!
Good catch!

PR #5740, opened 2024-12-20
golangci-lint: add intrange & update code

c4rlo:
Add the `intrange` linter, which enforces usage of the hip new-ish "range over int" feature in Go.

Update the code to match.


amartinezfayo (approved):
Thank you @c4rlo, this is good to have!

PR #5743, opened 2024-12-20
golangci-lint: add wastedassign & update code

c4rlo:
This linter caught an oversight that is probably inconsequential, but it shows its potential to catch more serious oversights as well.


amartinezfayo (approved):
Thank you @c4rlo!

PR #5749, opened 2024-12-27
Fix CompareObjectHandles check to run if syscall is available

### LINKED ISSUE #5748: peertracker: CompareObjectHandles check for Windows does not run
* **Version**: v1.11.1
* **Platform**: Windows
* **Subsystem**: peertracker

---

On Windows, if the `CompareObjectHandles` syscall is present, then `procCompareObjectHandlesErr` will be `nil` here.

https://github.com/spiffe/spire/blob/32cc98e140544810b501a4b6e7116f8d3054bac3/pkg/common/peertracker/npipe_windows.go#L15-L20

Further down, `compareObjectHandles()` will then always return `nil` and skip the check, since `isCompareObjectHandlesFound()` = `procCompareObjectHandlesErr == nil` = `true`.

https://github.com/spiffe/spire/blob/32cc98e140544810b501a4b6e7116f8d3054bac3/pkg/common/peertracker/npipe_windows.go#L61-L76

The only call path for these functions is here:

https://github.com/spiffe/spire/blob/32cc98e140544810b501a4b6e7116f8d3054bac3/pkg/common/peertracker/tracker_windows.go#L149-L154

jameshiew:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
`CompareObjectHandles` peertracker check on Windows systems which support it

**Description of change**
Change `if isCompareObjectHandlesFound()` to `if procCompareObjectHandlesErr != nil` in `compareObjectHandles` so that the error is forwarded if it is present, otherwise the check can run (discussed with SPIRE maintainers)

**Which issue this PR fixes**
Fixes #5748


pkg/common/peertracker/npipe_windows.go
```diff
@@ -65,7 +65,7 @@ func isCompareObjectHandlesFound() bool {
 // compareObjectHandles compares two object handles to determine if they
 // refer to the same underlying kernel object
 func compareObjectHandles(firstHandle, secondHandle windows.Handle) error {
-	if isCompareObjectHandlesFound() {
+	if procCompareObjectHandlesErr != nil {
```
jameshiew, line 68: I couldn't see an easy way to unit test `compareObjectHandles` without refactoring, I just included the one line change in this PR. To make it minimally testable could look something like in the direction of below - changing `compareObjectHandles` into a method on some type so that anything used by it e.g. `syscall.SyscallN` could be overridden on the struct used in testing.

```go
// syscaller wraps anything used by `compareObjectHandles` calls
type syscaller struct {
	procCompareObjectHandlesAddr *uintptr // from procCompareObjectHandles.Addr()
	procCompareObjectHandlesErr  error
	syscallN func(trap uintptr, args ...uintptr) (uintptr, uintptr, syscall.Errno)
}

// windowsSyscaller is the only syscaller that should be used at runtime
var windowsSyscaller syscaller

// init sets up the real syscaller
func init() {
	kernelbase := windows.NewLazyDLL("kernelbase.dll")
	procCompareObjectHandles := kernelbase.NewProc("CompareObjectHandles")
	if err := procCompareObjectHandles.Find(); err != nil {
		windowsSyscaller = syscaller{
			procCompareObjectHandlesErr: err,
			syscallN:                    syscall.SyscallN,
		}
	} else {
		syscallAddr := procCompareObjectHandles.Addr()
		windowsSyscaller = syscaller{
			procCompareObjectHandlesAddr: &syscallAddr,
			syscallN:                     syscall.SyscallN,
		}
	}
}

func (s syscaller) isCompareObjectHandlesFound() bool {
	return s.procCompareObjectHandlesAddr == nil
}

// compareObjectHandles compares two object handles to determine if they
// refer to the same underlying kernel object
func (s syscaller) compareObjectHandles(firstHandle, secondHandle windows.Handle) error {
	if s.procCompareObjectHandlesErr != nil {
		return s.procCompareObjectHandlesErr
	}
	r1, _, e1 := s.syscallN(*s.procCompareObjectHandlesAddr, uintptr(firstHandle), uintptr(secondHandle))
	if r1 == 0 {
		return e1
	}
	return nil
}
```

amartinezfayo (approved):
Thank you @jameshiew for this contribution.
I agree that there is no easy way to have this testable without some refactoring.
Maybe we can have an issue created for that so we can look at that separately.
It would be great if you could create an issue with your proposed approach.

Thanks!

PR #5750, opened 2024-12-28
svidstore: ParseMetadata to handle ':' in values

sorindumitru:
It's possible that the metadata stored in entry selectors for the SvidStore plugins will contain ':'. For example the AWS secret manager one can specify an ARN, which can include ':'. Modify the parsing to take this into account.


amartinezfayo (approved):
Thank you @sorindumitru!

PR #5757, opened 2025-01-03
adding support for publishing bundle to Minio Object Storage (#5395)

Dimss:
As a part of #5395 this PR add support to Minio S3 storage. 
The PR does not meant to be merge to main yet.
I do need someone look on the PR and provide initial feedback. 
I wanna make sure this functionality is still needed and overall logic does make sense.
if all good, I'll continue to work on it, add tests, etc...

### DIFF
--- a/doc/plugin_server_bundlepublisher_aws_s3.md
+++ b/doc/plugin_server_bundlepublisher_aws_s3.md
@@ -13,6 +13,7 @@ The plugin accepts the following configuration options:
 | bucket            | The Amazon S3 bucket name to which the trust bundle is uploaded.                                                                                               | Yes.                                                                   |                                                     |
 | object_key        | The object key inside the bucket.                                                                                                                              | Yes.                                                                   |                                                     |
 | format            | Format in which the trust bundle is stored, &lt;spiffe &vert; jwks &vert; pem&gt;. See [Supported bundle formats](#supported-bundle-formats) for more details. | Yes.                                                                   |                                                     |
+| endpoint          | A custom S3 endpoint should be set when using third-party object storage providers, such as Minio.                                                             | No.                                                                    |                                                     |
 
 ## Supported bundle formats
 
@@ -42,6 +43,22 @@ The following configuration uploads the local trust bundle contents to the `exam
     BundlePublisher "aws_s3" {
         plugin_data {
             region = "us-east-1"
+            bucket = "spire-trust-bundle"
+            object_key = "example.org"
+            format = "spiffe"
+        }
+    }
+```
+
+The following configuration uploads the local trust bundle contents to the `example.org` object in the `spire-trust-bundle` bucket on Minio server.
+
+```hcl
+    BundlePublisher "aws_s3" {
+        plugin_data {
+            endpoint = "https://my-org-minio.example.org"
+            region = "minio-sample-region"
+            access_key_id  = "minio-key-id"
+            secret_access_key = "minio-access-key"
             bucket = "spire-trust-bundle"
             object_key = "example.org"
             format = "spiffe"
--- a/pkg/server/plugin/bundlepublisher/awss3/awss3.go
+++ b/pkg/server/plugin/bundlepublisher/awss3/awss3.go
@@ -3,6 +3,7 @@ package awss3
 import (
 	"bytes"
 	"context"
+	"net/url"
 	"sync"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -44,6 +45,7 @@ type Config struct {
 	Bucket          string `hcl:"bucket" json:"bucket"`
 	ObjectKey       string `hcl:"object_key" json:"object_key"`
 	Format          string `hcl:"format" json:"format"`
+	Endpoint        string `hcl:"endpoint" json:"endpoint"`
 
 	// bundleFormat is used to store the content of Format, parsed
 	// as bundleformat.Format.
@@ -69,6 +71,11 @@ func buildConfig(coreConfig catalog.CoreConfig, hclText string, status *pluginco
 	}
 	if newConfig.Format == "" {
 		status.ReportError("configuration is missing the bundle format")
+	}
+	if newConfig.Endpoint != "" {
+		if _, err := url.ParseRequestURI(newConfig.Endpoint); err != nil {
+			status.ReportErrorf("could not parse endpoint url: %v", err)
+		}
 	}
 
 	bundleFormat, err := bundleformat.FromString(newConfig.Format)
--- a/pkg/server/plugin/bundlepublisher/awss3/awss3_test.go
+++ b/pkg/server/plugin/bundlepublisher/awss3/awss3_test.go
@@ -44,6 +44,18 @@ func TestConfigure(t *testing.T) {
 			},
 		},
 		{
+			name: "success with custom endpoint",
+			config: &Config{
+				AccessKeyID:     "access-key-id",
+				SecretAccessKey: "secret-access-key",
+				Region:          "region",
+				Bucket:          "bucket",
+				ObjectKey:       "object-key",
+				Format:          "spiffe",
+				Endpoint:        "http://example.com",
+			},
+		},
+		{
 			name: "no region",
 			config: &Config{
 				Bucket:    "bucket",
@@ -97,6 +109,20 @@ func TestConfigure(t *testing.T) {
 			expectCode:   codes.Internal,
 			expectMsg:    "failed to create client: client creation error",
 			newClientErr: errors.New("client creation error"),
+		},
+		{
+			name: "invalid endpoint url",
+			config: &Config{
+				AccessKeyID:     "access-key-id",
+				SecretAccessKey: "secret-access-key",
+				Region:          "region",
+				Bucket:          "bucket",
+				ObjectKey:       "object-key",
+				Format:          "spiffe",
+				Endpoint:        "endpoint",
+			},
+			expectCode: codes.InvalidArgument,
+			expectMsg:  "could not parse endpoint url",
 		},
 	} {
 		t.Run(tt.name, func(t *testing.T) {
--- a/pkg/server/plugin/bundlepublisher/awss3/client.go
+++ b/pkg/server/plugin/bundlepublisher/awss3/client.go
@@ -17,6 +17,7 @@ func newAWSConfig(ctx context.Context, c *Config) (aws.Config, error) {
 	cfg, err := config.LoadDefaultConfig(ctx,
 		config.WithRegion(c.Region),
 	)
+
 	if err != nil {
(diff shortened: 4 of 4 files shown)


sorindumitru:
Thanks for the contribution @Dimss! I'll try to take a look at it this week.

sorindumitru:
@Dimss Minio is API compatible with AWS S3 API, so it should be possible to use the existing `awss3` bundle publisher here. Have you tried using that? Are there any issues or missing functionality?

Dimss:
`awss3` might be working with `Minio`. 
However, by default when user using `awss3` bundle publisher he doesn't have an option to configure what's `Minio` calls `Endpoint URL`.  We possibly can extend the `awss3` configuration with `Endpoint URL`, but then perhaps it will be not `awss3` but generic `s3` bundle publisher. 
In addition, `Minio` allow insecure connections to buckets (might be useful for dev/test envs) 

I think the main question here, should it be a generic S3 publisher or publisher for each provider with it's backend implementation.

sorindumitru:
Having separate implementations for different platforms may make sense eventually, but I feel that for now it only adds maintenance burden.

If someone does end up needing something from this plugin where it makes sense to have it use platform specific APIs, such as platform specific authentication methods, we should look into splitting off that part into a separate plugin. @Dimss, do you need anything from an S3 publisher that is Minio specific?

For now, it's probably easiest to just add support for specifying an optional endpoint to the awss3 plugin.

Dimss:
@sorindumitru ok by me. 
So then I'll extend existing `awss3` plugin with extra optional parameter `endpoint`  and `insecure`. 
Do you agree?

kfox1111:
Just endpoint for now I think? if you need to load in a custom ca for talking to minio, it can be done via the system trust bundle. If its actually insecure, not sure there is any advantage to funneling it through the s3 service rather then just using the agents `insecure_bootstrap` flag?

Dimss:
@kfox1111  agree. then I'll just add one optional `endpoint` param into existing `awss3` plugin

Dimss:
@kfox1111, @sorindumitru  I applied required changes, pls review.

go.sum
```diff
@@ -2230,4 +2230,4 @@ sigs.k8s.io/structured-merge-diff/v4 v4.4.2/go.mod h1:N8f93tFZh9U6vpxwRArLiikrE5
 sigs.k8s.io/yaml v1.4.0 h1:Mk1wCc2gy/F0THH0TAp1QYyJNzRm2KCLy3o5ASXVI5E=
 sigs.k8s.io/yaml v1.4.0/go.mod h1:Ejl7/uTz7PSA4eKMyQCUTnhZYNmLIl+5c2lQPGR2BPY=
 software.sslmate.com/src/go-pkcs12 v0.4.0 h1:H2g08FrTvSFKUj+D309j1DPfk5APnIdAQAB8aEykJ5k=
-software.sslmate.com/src/go-pkcs12 v0.4.0/go.mod h1:Qiz0EyvDRJjjxGyUQa2cCNZn/wMyzrRJ/qcDXOQazLI=
+software.sslmate.com/src/go-pkcs12 v0.4.0/go.mod h1:Qiz0EyvDRJjjxGyUQa2cCNZn/wMyzrRJ/qcDXOQazLI=
```
amartinezfayo, line 2233: Newline missing.

pkg/server/plugin/bundlepublisher/awss3/client.go
```diff
@@ -2,7 +2,6 @@ package awss3
 
 import (
 	"context"
-
```
amartinezfayo, line 5: It seems that this is not `goimports`-ed

pkg/server/plugin/bundlepublisher/awss3/client.go
```diff
@@ -25,9 +25,20 @@ func newAWSConfig(ctx context.Context, c *Config) (aws.Config, error) {
...
+	if c.Endpoint != "" {
+		cfg.BaseEndpoint = aws.String(c.Endpoint)
+	}
+
 	return cfg, nil
 }
 
 func newS3Client(c aws.Config) (simpleStorageService, error) {
-	return s3.NewFromConfig(c), nil
+	options := func(options *s3.Options) {}
```
amartinezfayo, line 36: Prefer the use of `var` for zero value.
```suggestion
	var options func(options *s3.Options)
```

amartinezfayo (commented):
Thank you @Dimss for this contribution!

We need to update the [documentation](https://github.com/spiffe/spire/blob/main/doc/plugin_server_bundlepublisher_aws_s3.md) of the aws_s3 bundle publisher plugin with the new Endpoint setting, indicating that's an optional setting (not required).

We should also add test coverage for this, making sure to check that the configuration has been properly wired up to the aws config.

amartinezfayo:
Hi @Dimss, I think we’re pretty close to having this merged. Have you had a chance to look at the latest small comments?

Dimss:
Hey @amartinezfayo yeah, I saw the comments, I'll make it done in coming days.

Dimss:
@amartinezfayo I've applied required changes, added tests and updated docs. 
pls review.

kfox1111 (dismissed):
LGTM. Thanks! :)

kfox1111:
Looks like the DCO needs fixing.

Dimss:
@kfox1111 DCO has been added.

pkg/server/plugin/bundlepublisher/awss3/awss3.go
```diff
@@ -70,6 +72,11 @@ func buildConfig(coreConfig catalog.CoreConfig, hclText string, status *pluginco
 	if newConfig.Format == "" {
 		status.ReportError("configuration is missing the bundle format")
 	}
+	if newConfig.Endpoint != "" {
+		if _, err := url.ParseRequestURI(newConfig.Endpoint); err != nil {
+			status.ReportError("bad endpoint url format")
```
sorindumitru, line 77: Maybe something like this to also include the error:
```suggestion
			status.ReportErrorf("could not parse endpoint url: %v", err)
```

pkg/server/plugin/bundlepublisher/awss3/client.go
```diff
@@ -25,9 +26,20 @@ func newAWSConfig(ctx context.Context, c *Config) (aws.Config, error) {
...
 func newS3Client(c aws.Config) (simpleStorageService, error) {
-	return s3.NewFromConfig(c), nil
+	var options func(options *s3.Options)
+	if c.BaseEndpoint != nil && *c.BaseEndpoint != "" {
+		options = func(options *s3.Options) {
+			options.UsePathStyle = true
+			options.BaseEndpoint = c.BaseEndpoint
+		}
+	}
+	return s3.NewFromConfig(c, options), nil
```
sorindumitru, line 44: I think `NewFromConfig` would end up calling the `nil` options in the case that the `BaseEndpoint` is not configured. We could have the options have a default of an empty function.

pkg/server/plugin/bundlepublisher/awss3/awss3_test.go
```diff
@@ -98,6 +110,20 @@ func TestConfigure(t *testing.T) {
 			expectMsg:    "failed to create client: client creation error",
 			newClientErr: errors.New("client creation error"),
 		},
+		{
+			name: "wrong endpoint format",
```
sorindumitru, line 114: I find the format in the test name a bit confusing because there's a `Format` field in the config too. Maybe we can say:
```suggestion
			name: "invalid endpoint url",
```

go.mod
```diff
@@ -1,166 +1,161 @@
 module github.com/spiffe/spire
 
-go 1.23.4
+go 1.24.0
```
sorindumitru, line 3: This might be some leftover from rebasing, could you clean it up?

doc/plugin_server_bundlepublisher_aws_s3.md
```diff
@@ -13,7 +13,7 @@ The plugin accepts the following configuration options:
 | bucket            | The Amazon S3 bucket name to which the trust bundle is uploaded.                                                                                               | Yes.                                                                   |                                                     |
 | object_key        | The object key inside the bucket.                                                                                                                              | Yes.                                                                   |                                                     |
 | format            | Format in which the trust bundle is stored, &lt;spiffe &vert; jwks &vert; pem&gt;. See [Supported bundle formats](#supported-bundle-formats) for more details. | Yes.                                                                   |                                                     |
-
+| endpoint          | A custom S3 endpoint should be set when using third-party object storage providers, such as Minio.                                                             | No.                 |                                                     |
```
sorindumitru, line 16: The `Default` column contains the default value for the configuration. Since there's no default here, could you leave it blank?

Could you also align the column markers with the previous rows? I doesn't matter for the rendered markdown, but it makes it a bit easier when editing the files.

pkg/server/plugin/bundlepublisher/awss3/client.go
Dimss: @sorindumitru I do not understand what you mean. 
can you pls provide an example.

go.mod
Dimss: @sorindumitru 
this looks up to date.. not sure why github show this diff.
in PR, the go.mod have valid go version

pkg/server/plugin/bundlepublisher/awss3/client.go
sorindumitru: The `options` variable defaults to the zero value of a function pointer, which is `nil`. `NewFromConfig` will then try to call that `nil` and get a segfault. See for example https://play.golang.com/p/SPZxT29tKjN

We can avoid this by initializing the variable to a function that doesn't do anything.

Dimss: @sorindumitru, yes you are right. 
@amartinezfayo  suggested to change it `var` [comment](https://github.com/spiffe/spire/pull/5757#discussion_r1945294092)
I'll revert it back to `options := func(options *s3.Options) {}`, is that OK?

go.sum
```diff
@@ -2229,4 +2229,4 @@ sigs.k8s.io/structured-merge-diff/v4 v4.4.2/go.mod h1:N8f93tFZh9U6vpxwRArLiikrE5
...
```
sorindumitru, line 2232: Looks like we still have these new line changes, could you also remove this? The linter is also complaining about them.

pkg/server/plugin/bundlepublisher/awss3/client.go
sorindumitru: Sounds good to me.

Dimss:
@sorindumitru if everything in place, please approve.

amartinezfayo (approved):
Thank you @Dimss for this contribution! 🚀

PR #5762, opened 2025-01-07
Clean up integration test Docker warnings

rturner3:
Docker doesn't like it when Dockerfile keywords on the same line don't match in capitalization, e.g. `FROM foo as bar`.

Also remove deprecated `version` property from a Docker Compose file.


amartinezfayo (approved):
🎉

PR #5764, opened 2025-01-09
SyncAuthorizedEntries: fix issue with entry id reuse and add integration test

### LINKED ISSUE #5522: Exercise SyncAuthorizedEntries in integration tests
We need to exercise SyncAuthorizedEntries in integration tests.

### LINKED ISSUE #5686: sync authorized entries: issues with entry id reuse
When using the experimental `use_sync_authorized_entries = true` configuration users can run into situations where agents don't see entry updates if users recreate entries with the same name:
* spire-server entry create "entryID"
* spire-server entry update "entryID" (optionally, I think you still have the same issue if you don't do this) 
* spire-server entry delete "entryID" followed quickly (shorter than the agent sync interval) by spire-server entry create "entryID"

The agent would receive a revision about the entry "entryID" but will see the RevisionNumber set 0, which is smaller than the previous RevisionNumber it had, and will ignore that update.

This has usually a small chance of happening, due to using UUIDs by default for the entry IDs, but since users are allowed to specify an entry ID it can happen more easily.

Ideas for dealing with this:
* Include the "created_at" field in the entry revision update and use both it and the revision number to decide if an entry needs to be re-synced. This adds a bit more overhead to the communication, but it should cover all situations.
* Replace the two with a single "updated at" field.

* **Version**: >= 1.8.7
* **Platform**: any
* **Subsystem**: server,agent

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Registration entry synchronization between agent and server.

**Description of change**
The agent will now include the created_at field when determining if a registration entry is stale or not.

**Which issue this PR fixes**
fixes #5522
fixes #5686
depends on https://github.com/spiffe/spire-api-sdk/pull/69


test/integration/suites/sync-authorized-entries/04-create-workload-entries
```diff
@@ -0,0 +1,40 @@
...
+for id in $(seq 1 512); do
+    ENTRIES=$(jq \
+        --arg uidselector "uid:${id}"\
+        --arg spiffeID  "spiffe://domain.test/workload${id}"\
+        '.entries[.entries| length] |= . + {
+    "parent_id": "spiffe://domain.test/node",
+    "spiffe_id": $spiffeID,
+    "selectors": [{"type": "unix", "value": $uidselector}]
+}' <<< ${ENTRIES})
+done
```
MarcosDY, line 16: all entry creation can be simplified with something like this:
```
ENTRIES=$(jq -n '{
    entries: [
        {
            parent_id: "spiffe://domain.test/node",
            spiffe_id: "spiffe://domain.test/theworkload",
            selectors: [
                {
                    type: "unix",
                    value: "uid:0"
                }
            ] 
        },
        (
            range(1; 512) | {
                parent_id: "spiffe://domain.test/node",
                spiffe_id: ("spiffe://domain.test/workload" + (. | tostring)),
                selectors: [
                    {
                        type: "unix",
                        value: ("uid:" + (. | tostring))
                    }
                ]
            }
        )
    ]
}')
```

MarcosDY: if you go for this, you will need to change the verified entry

test/integration/suites/sync-authorized-entries/04-create-workload-entries
```diff
@@ -0,0 +1,40 @@
...
+    /opt/spire/bin/spire-server entry create \
+    -parentID "spiffe://domain.test/node" \
+    -spiffeID "spiffe://domain.test/theworkload" \
+    -selector "unix:uid:0"
+
+# Check at most 30 times (with one second in between) that the agent has
+# successfully synced down the workload entry.
+MAXCHECKS=30
+CHECKINTERVAL=1
+for ((i=1;i<=MAXCHECKS;i++)); do
```
MarcosDY, line 31: you can use check-synced-entry than is doing the same

test/integration/suites/sync-authorized-entries/README.md
```diff
@@ -0,0 +1,9 @@
+# Join Token Suite
```
MarcosDY, line 1: looks like this readme is not updated?

test/integration/suites/sync-authorized-entries/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Description
+
+This suite verifies that:
```
MarcosDY, line 5: it looks tome than we must recreate an entry with the same reused id right? 
and verify it is failing as expected, or recreated once we removed the old one, 
maybe you can add more scripts verify that?

test/integration/suites/sync-authorized-entries/docker-compose.yaml
```diff
@@ -0,0 +1,16 @@
...
+  spire-agent:
+    image: spire-agent:latest-local
+    volumes:
+      - ./conf/agent:/opt/spire/conf/agent
+    command: ["-config", "/opt/spire/conf/agent/agent.conf"]
+  bad-spire-agent:
+    image: spire-agent:latest-local
+    volumes:
+      - ./conf/bad-agent:/opt/spire/conf/agent
+    command: ["-config", "/opt/spire/conf/agent/agent.conf"]
```
MarcosDY, line 16: why it is required?

test/integration/suites/sync-authorized-entries/README.md
sorindumitru: I think that ends up being difficult to handle in the integration test because of timing. I've added some more checks to the unit tests to verify those conditions. Do you think those are enough or should I see about checking those in the integration test?

test/integration/suites/sync-authorized-entries/docker-compose.yaml
sorindumitru: Its not, just some leftovers from the test I started from. I thought I cleaned those up, but turns out i didn't

test/integration/suites/sync-authorized-entries/README.md
MarcosDY: yeah, it is going difficult to test the edge scenario within an IT

MarcosDY (approved):

MarcosDY (approved):

PR #5769, opened 2025-01-11
Upgrade to OPA v1

sorindumitru:
The new version is backwards incompatible with the v0 version. This includes the policy language itself, which has become stricter.

The 1.0 API allows specifying the version of the policy at the time of the loading the policy so I'll look into making that configurable to make it easier for users to upgrade. To be decided if we want to default to V1 or V0. Since this is an experimental config, I think we can default to V1.

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**
closes #5754


pkg/server/authpolicy/policy.rego
```diff
@@ -32,15 +32,15 @@ default allow = false
 
 
 # Admin allow check
-allow_if_admin = true {
```
rturner3, line 35: Is this new syntax backwards-compatible or do we need to maintain two separate Rego policies for v0 Rego and v1 Rego?

sorindumitru: For this particular file I don't think we need to care, it's only used if another policy is not specified. It uses an embed file and we can specify the correct version for it.

As for backwards-compatibility, I don't think there is any. RegoV0 seems to load this file just file, but it probably misinterprets it because our auth tests are failing if I do that.

sorindumitru:
@rturner3 I ended up inverting the condition so that if a user provides a policy file, it defaults to rego v0. This should provide a nicer upgrade path at the cost of a bit more changes on our side to remove v0 support completely.

We can still use rego.v1 for the default policy since we are in control of the policy and of the version there.

rturner3:
LGTM, I think there are some merge conflicts now that need to be resolved though.

rturner3 (approved):

PR #5775, opened 2025-01-15
Add support for San selectors in x509pop node attestor plugin

### LINKED ISSUE #5746: Support for optional SAN based label selectors in X509 pop Node Attestor
# Problem
The current join token-based node attestation plugin identifies only a single node. Unlike the k8s/aws_iid node attestation, the join token does not expose any flexibile selectors, which limits us in certain aspects:

1. We need to create a join token request for every single node. We could create a spiffe id while creating join token, but this locks us down to having only one SpiffeId format for all nodes. Therefore, these nodes cannot be part of multiple clusters without managing additional metadata in a different.  
2. Creating and managing multiple aliases for nodes in a dynamic environment becomes operationally challenging especially when new nodes are added/removed.

# Solution
Multiple options were discussed in [this](https://github.com/spiffe/spire/issues/5647) issue and the outcome of that discussion pointed us towards enhancing x509pop node attestor plugin to support additional selectors by extracting SANs on the leaf certificate.

## Specifications

The URI SAN on the leaf certificate could be of the format. 
x509pop://trustdomain:key:value
Some examples of the SAN selectors
- x509pop://example.org/datacenter:us-east-1
- x509pop://example.org/env:prod

Considerations:
- The SAN extraction into selectors should be deliberate, hence need to be a URI SANS beginning with x509pop://

**Server Side:**

Additional Configuration

```
   NodeAttestor "x509pop" {
        plugin_data {
            ca_bundle_path = "/opt/spire/conf/server/agent-cacert.pem"
            enable_san_selector = false  #defaulted to false
            agent_path_template = "/cn/{{ .Subject.CommonName }}/{{.San.datacenter}}/{{.San.env}}"
        }
    }
```

Sample Exposed Selector format:
x509pop:san:datacenter:us-east-1
x509pop:san:env:prod

Subject Alternative Name format : 
**Agent Side Changes:** 
None

**Implementation Options**

- The custom selectors would be passed as a map to the [template function](https://github.com/spiffe/spire/blob/main/pkg/common/plugin/x509pop/x509pop.go#L271C1-L285C2) enabling construction of the Agent ID 
- The map of key, value selectors would be sent to [this](https://github.com/spiffe/spire/blob/main/pkg/server/plugin/nodeattestor/x509pop/x509pop.go#L326) method to expose them as selectors

# Related Discussion
https://github.com/spiffe/spire/issues/5647

snanjundaswamy:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
x509pop Node Attestor plugin now supports SAN based additional selectors. Each san is for the URI format x509pop://<trustdomain>/key:value. Each of these SANs is exposed as a selector with x509pop:san:<key>:<value>. These selectors can also be used as a part of agent template with {{ .San.<key>}} 

**Description of change**
The change contains feature augmentation on x509pop node attestor plugin along with tests. The certificates for testing were generated with test/fixture/nodeattestor/x509pop/generate.go

**Which issue this PR fixes**
`fixes #5746`


test/fixture/nodeattestor/x509pop/generate.go
```diff
@@ -48,6 +48,10 @@ func main() {
 		KeyUsage:     x509.KeyUsageDigitalSignature,
 		NotAfter:     neverExpires,
 		Subject:      pkix.Name{CommonName: "COMMONNAME"},
+		URIs: []*url.URL{
+			{Scheme: "x509pop", Host: "example.org", Path: "/datacenter:us-east-1"},
+			{Scheme: "x509pop", Host: "example.org", Path: "/environment:production"},
```
kfox1111, line 53: just curious (naming things is hard), there a reason for /environment:production over say, /environment/production ?

sorindumitru: Looks like `:` is a [reserved character](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) for URLs so it might not be the best choice. It may work now, but the standard library could add more validation in the future. `/name/value` seems like a reasonable choice to me.

sorindumitru: Although by https://datatracker.ietf.org/doc/html/rfc3986#section-3.3 it seems to allowed in the path.

snanjundaswamy: I'm not too opinionated on ':', I simply chose the character for consistency with separators in selectors. Based on what  I hear, '/' sounds like a natural fit. I will make this change.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -352,5 +361,26 @@ func buildSelectorValues(leaf *x509.Certificate, chains [][]*x509.Certificate) [
...
+	uriSelectorMap := make(map[string]string)
+	sanPrefix := "x509pop://" + trustDomain + "/"
+	for _, uri := range leaf.URIs {
+		if strings.HasPrefix(uri.String(), sanPrefix) {
+			unprefixedUriSan := strings.TrimPrefix(uri.String(), sanPrefix)
+			if strings.Contains(unprefixedUriSan, ":") {
+				lastIndex := strings.LastIndex(unprefixedUriSan, ":")
+				uriSelectorKey := unprefixedUriSan[:lastIndex]
+				uriSelectorValue := unprefixedUriSan[lastIndex+1:]
+				uriSelectorMap[uriSelectorKey] = uriSelectorValue
```
sorindumitru, line 381: Any reason for using `LastIndex` for splitting the string? Something like `x509pop://example.org/datacenter:ny:1` would be split as `datacenter:ny => 1`, but to me it makes more sense as `datacenter => ny:1` (which could be done through `SplitN`).

Definitely a bit of an uncommon edge case, but it would be good to figure out from the beginning.

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -44,6 +44,7 @@ A sample configuration:
 | Common Name      | `x509pop:subject:cn:example.org`                                  | The Subject's Common Name (see X.500 Distinguished Names)                                |
 | SHA1 Fingerprint | `x509pop:ca:fingerprint:0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33` | The SHA1 fingerprint as a hex string for each cert in the PoP chain, excluding the leaf. |
 | SerialNumber     | `x509pop:serialnumber:0a1b2c3d4e5f`                               | The leaf certificate serial number as a lowercase hexadecimal string                     |
+| San     | `x509pop:san.<key>:<value>`                               | The san selectors on the leaf selectors. The expected format of the uri san is `x509pop://<trust_domain>/<key>:<value>`  string                     |
```
sorindumitru, line 47:
```suggestion
| San     | `x509pop:san:<key>:<value>`                               | The san selectors on the leaf selectors. The expected format of the uri san is `x509pop://<trust_domain>/<key>:<value>`  string                     |
```

sorindumitru (commented):
Thanks for the contribution. Generally looks good to me, with some small comments. It would also be good if we had some tests for using the selectors in the path component, do you think that could be done?

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
snanjundaswamy: I thought about this a little bit myself. In my experience, the Key is usually fully qualified as against the values.
Something like  `company.org.service.datacenter = us-east-1`. References on k8s annotations [here](https://kubernetes.io/docs/reference/labels-annotations-taints/). 
Either way, this should be a simple change, lmk what you think

rturner3: We have a convention in the project around how selectors are formatted `key:value`, where `key` cannot contain a `:` delimiter. See this helper method for reference: https://github.com/spiffe/spire/blob/50976c35a45ef278090b6ba5053850c83c606c87/cmd/spire-server/util/util.go#L185

We should be consistent with that convention here as well.

test/fixture/nodeattestor/x509pop/generate.go
rturner3: Since `:` is allowed in URL paths, I would suggest we stick to it as the SAN selector delimiter since it's what we use everywhere else.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
snanjundaswamy: @rturner3 @sorindumitru 
Ok makes sense now. 
But what about the values? Should we also retain the '/' in the values coming from the URI San? This will help us eliminate ambiguity around the usage of ':'  in values. Here's an example for reference
datacenter:company/foo/us-east-1

rturner3: Yes, I think we should keep the SAN value as-is.

test/fixture/nodeattestor/x509pop/generate.go
snanjundaswamy: Done. This would be the convention in the plugin for SAN selectors

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
snanjundaswamy: Apologies for late turn around. I was traveling the past few weeks. 
Updated the PR to include this suggestion. 
Also, to ensure tests pass, I had to [sort the list](https://github.com/spiffe/spire/pull/5775/files#diff-81cafc86e5f19760ffb5641855cf52f352ff1821318edf2247dc8d22d56a86caR446-R453) of expected selectors and the actual selectors. I've kept the method in the same file to keep the scope limited as I do not see a broader/generic usecase at the moment.  I know this is not elegant, but I am open to suggestions.

pkg/server/plugin/nodeattestor/x509pop/x509pop_test.go
```diff
@@ -427,3 +442,12 @@ func unmarshal(t *testing.T, data []byte, obj any) {
 func expectNoChallenge(context.Context, []byte) ([]byte, error) {
 	return nil, errors.New("challenge is not expected")
 }
+
+func sortX509PopSelectors(s []*common.Selector) {
```
sorindumitru, line 446: I think you can use https://github.com/spiffe/spire/blob/1b343d9c4bbd32178c83aeee412dd22916a84fd6/pkg/common/util/sort.go#L45 for this.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -352,5 +361,28 @@ func buildSelectorValues(leaf *x509.Certificate, chains [][]*x509.Certificate) [
...
+
 	return selectorValues
 }
+
+func (p *Plugin) parseUriSanSelectors(leaf *x509.Certificate, trustDomain string) map[string]string {
+	uriSelectorMap := make(map[string]string)
+	sanPrefix := "x509pop://" + trustDomain + "/"
+	for _, uri := range leaf.URIs {
+		if !strings.HasPrefix(uri.String(), sanPrefix) {
+			p.log.Warn(uri.String(), "san does not contain the expected scheme or trust domain")
```
sorindumitru, line 376: We shouldn't warn (or maybe log anything here) since the certs could have URIs for other reasons (for example the spiffe use case in this plugin).

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -352,5 +361,28 @@ func buildSelectorValues(leaf *x509.Certificate, chains [][]*x509.Certificate) [
...
+			continue
+		}
+		segments := strings.Split(strings.Trim(uri.Path, "/"), "/")
+		if len(segments) < 2 {
+			p.log.Warn("cannot extract x509pop san selectors from", uri.String())
+			continue
+		}
+
+		uriSelectorMap[segments[0]] = strings.Join(segments[1:], "/")
```
sorindumitru, line 385: You could use `strings.SplitN(strings.Trim(uri.Path, "/"), "/", 2)` so you don't have to reconstitute the second element.

sorindumitru:
@snanjundaswamy Looks like the DCO check is failing, see https://github.com/spiffe/spire/actions/runs/13219709206/job/36941915403?pr=5775 for the details. I think the signoff has to match the commit author email.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
snanjundaswamy: Yes, cleaner that way. Done.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
snanjundaswamy: Yes, this makes sense. Looks like an obvious one. There could be additional sans that may have a different scheme beside x509pop. I have made this change.

snanjundaswamy:
> @snanjundaswamy Looks like the DCO check is failing, see https://github.com/spiffe/spire/actions/runs/13219709206/job/36941915403?pr=5775 for the details. I think the signoff has to match the commit author email.

Looks like this was just one commit in the chain. Fixed it. Thank you. Please take a look once you get a chance.

snanjundaswamy:
@sorindumitru   I'm still trying to get the integration tests working locally but cannot figure out from the logs if this is a sporadic issue. Are there any pointers for me?

sorindumitru:
> @sorindumitru I'm still trying to get the integration tests working locally but cannot figure out from the logs if this is a sporadic issue. Are there any pointers for me?

Yes, it's possible it's a flaky test. It's re-testing now, let's see if it continues to fail.

For the integration tests, you can run a single one like this:
```
make images # builds docker images required by the tests
./test/integration/test-one.sh ./test/integration/suites/agent-cli
```

snanjundaswamy:
> > @sorindumitru I'm still trying to get the integration tests working locally but cannot figure out from the logs if this is a sporadic issue. Are there any pointers for me?
> 
> Yes, it's possible it's a flaky test. It's re-testing now, let's see if it continues to fail.
> 
> For the integration tests, you can run a single one like this:
> 
> ```
> make images # builds docker images required by the tests
> ./test/integration/test-one.sh ./test/integration/suites/agent-cli
> ```

@sorindumitru  Thank you. Last run has succeeded. 
I've rebased on the latest main, PTAL once you get a chance

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -352,5 +361,25 @@ func buildSelectorValues(leaf *x509.Certificate, chains [][]*x509.Certificate) [
...
 }
+
+func (p *Plugin) parseUriSanSelectors(leaf *x509.Certificate, trustDomain string) map[string]string {
+	uriSelectorMap := make(map[string]string)
+	sanPrefix := "x509pop://" + trustDomain + "/"
+	for _, uri := range leaf.URIs {
+		if strings.HasPrefix(uri.String(), sanPrefix) {
+			segments := strings.SplitN(strings.Trim(uri.Path, "/"), "/", 2)
+			if len(segments) < 2 {
+				p.log.Warn("cannot extract x509pop san selectors from", uri.String())
```
sorindumitru, line 378: The hlog logger expects key value pairs after the message. So you can either do that and adjust the message to match or use `Sprintf` to print a full string.

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -44,6 +44,7 @@ A sample configuration:
 | Common Name      | `x509pop:subject:cn:example.org`                                  | The Subject's Common Name (see X.500 Distinguished Names)                                |
 | SHA1 Fingerprint | `x509pop:ca:fingerprint:0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33` | The SHA1 fingerprint as a hex string for each cert in the PoP chain, excluding the leaf. |
 | SerialNumber     | `x509pop:serialnumber:0a1b2c3d4e5f`                               | The leaf certificate serial number as a lowercase hexadecimal string                     |
+| San     | `x509pop:san:<key>:<value>`                               | The san selectors on the leaf certificate. The expected format of the uri san is `x509pop://<trust_domain>/<key>:<value>`. One selector is exposed per uri san corresponding to x509pop uri scheme. string                     |
```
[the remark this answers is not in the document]
snanjundaswamy: Increased the scope little bit.
Screenshot 2025-02-18 at 11 04 46 AM

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
snanjundaswamy: Done.

sorindumitru (approved):

Issue #5770, opened 2025-01-11
cli socket path env


kfox1111:
When setting up a more traditional system layout, where your systems socket is under (/var)/run, it would be nice to be able to override the default socket lookup path with an environment variable in `spire-agent api` and `spire-server entry` type commands.

rturner3:
This sounds reasonable @kfox1111. We would welcome a PR with those changes.

kfox1111:
SPIFFE_ENDPOINT_SOCKET for the spiffe workloadload socket for spire-agent seems like the thing to do.

for the server api, its less clear.

SPIRE_ADMIN_SOCKET?

rturner3:
> SPIRE_SERVER_ADMIN_SOCKET?

SGTM

kfox1111:
After some discussion, we may want to do this in 2 phases. I would think client changes can be made safely without much of a heads up to behavioral changes? While run is a more disruptive change that may need a minor version bump at least as a heads up?

1. Patch change
    * [ ] spire-agent client changes
    * [ ] spire-server client changes
 2. Minor change
    * [ ] spire-agent run change
    * [ ] spire-server run change

arianvp:
Sounds good to me

arianvp:
Just wanted to let you know that https://github.com/ghostunnel/ghostunnel already uses `SPIFFE_ENDPOINT_SOCKET` but requires the value to be a URI of shape `unix:<path>` or `tcp:<addr>:<port>`

PR #5776, opened 2025-01-15
SPIFFE_ENDPOINT_SOCKET env support for spire-agent

kfox1111:
Enables specifying SPIFFE_ENDPOINT_SOCKET to the non run subcommands to the spire-agent

partially-fixes: https://github.com/spiffe/spire/issues/5770

### DIFF
--- a/cmd/spire-agent/cli/common/defaults_posix.go
+++ b/cmd/spire-agent/cli/common/defaults_posix.go
@@ -2,9 +2,24 @@
 
 package common
 
+import (
+	"os"
+)
+
 const (
-	// DefaultSocketPath is the SPIRE agent's default socket path
-	DefaultSocketPath = "/tmp/spire-agent/public/api.sock"
+	// DefaultRunSocketPath is the SPIRE agent's default socket path
+	DefaultRunSocketPath = "/tmp/spire-agent/public/api.sock"
 	// DefaultAdminSocketPath is the SPIRE agent's default admin socket path
 	DefaultAdminSocketPath = "/tmp/spire-agent/private/admin.sock"
 )
+
+// DefaultSocketPath is the SPIRE agent's default socket path
+var DefaultSocketPath string
+
+func init() {
+	DefaultSocketPath = DefaultRunSocketPath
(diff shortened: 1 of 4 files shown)


cmd/spire-agent/cli/common/defaults_posix.go
```diff
@@ -2,9 +2,24 @@
...
-	// DefaultSocketPath is the SPIRE agent's default socket path
-	DefaultSocketPath = "/tmp/spire-agent/public/api.sock"
+	// DefaultRunSocketPath is the SPIRE agent's default socket path
+	DefaultRunSocketPath = "/tmp/spire-agent/public/api.sock"
 	// DefaultAdminSocketPath is the SPIRE agent's default admin socket path
 	DefaultAdminSocketPath = "/tmp/spire-agent/private/admin.sock"
 )
+
+// DefaultSocketPath is the SPIRE agent's default socket path
+var DefaultSocketPath string
```
MarcosDY, line 17: So, it is going to be used for `api` calls and `healthchecks` except for run command where we are going to keep using 'DefaultRunSocketPath' 
Is it expected to change healthchecks?
And if not may we the code to change default inside `AddOSFlags` instead?

kfox1111: We chatted about this. I believe the current plan is to do everything but run in one pr, and run in a future pr.

kfox1111:
SPIFFE_ENDPOINT_SOCKET needs to support unix:// on the front according to: https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Workload_Endpoint.md#4-locating-the-endpoint

But this patch doesn't work with that currently.

kfox1111:
Have run into a few problems with this, and a bigger config question/issue. Still working through it.

sorindumitru:
@kfox1111, do you still want us to keep this issue around? I think the solution for this will likely be replaced by what was discussed in https://github.com/spiffe/spire/issues/6247.

kfox1111:
I'm good closing this in favor of #6247

Issue #5789, opened 2025-01-16
Flaky test TestUpstreamProcessTaintedAuthority


rturner3:
```
--- FAIL: TestUpstreamProcessTaintedAuthority (60.05s)
    manager_test.go:411: 
        	Error Trace:	/home/runner/work/spire/spire/pkg/server/ca/manager/manager_test.go:411
        	Error:      	deadline reached
        	Test:       	TestUpstreamProcessTaintedAuthority
    manager_test.go:1423: 
        	Error Trace:	/home/runner/work/spire/spire/pkg/server/ca/manager/manager_test.go:1423
        	            				/home/runner/work/spire/spire/pkg/server/ca/manager/manager_test.go:1418
        	            				/home/runner/work/spire/spire/pkg/server/ca/manager/manager_test.go:414
        	Error:      	Received unexpected error:
        	            	datastore-sql: context deadline exceeded
        	Test:       	TestUpstreamProcessTaintedAuthority
FAIL
FAIL	github.com/spiffe/spire/pkg/server/ca/manager	60.670s
```

https://github.com/spiffe/spire/actions/runs/12816442140/job/35737575426?pr=5718

sorindumitru:
Looked a bit into this when I saw this issue before. The error seems to be due to getting a "no such table" error from go-sqlite3 somewhere else in the stack, here it is with some extra logging:
```
2025/01/20 22:12:46 ERROR Could not prepare x509-ca error="failed to fetch bundle: rpc error: code = Unknown desc = datastore-sql: no such table: bundles"
--- FAIL: TestUpstreamProcessTaintedAuthority (60.05s)
    manager_test.go:411:
                Error Trace:    /home/sorin/github/spiffe/spire/pkg/server/ca/manager/manager_test.go:411
                Error:          deadline reached
                Test:           TestUpstreamProcessTaintedAuthority
    manager_test.go:1425:
                Error Trace:    /home/sorin/github/spiffe/spire/pkg/server/ca/manager/manager_test.go:1425
                                                        /home/sorin/github/spiffe/spire/pkg/server/ca/manager/manager_test.go:1420
                                                        /home/sorin/github/spiffe/spire/pkg/server/ca/manager/manager_test.go:414
                Error:          Received unexpected error:
                                datastore-sql: context deadline exceeded
                Test:           TestUpstreamProcessTaintedAuthority
FAIL
FAIL    github.com/spiffe/spire/pkg/server/ca/manager   205.087s
FAIL
```

[...]

PR #5805, opened 2025-01-21
Fix flaky test TestUpstreamProcessTaintedAuthorityBackoff

sorindumitru:
We need to wait for the backoff routine to hit the clock.After() call before moving the clock

**Which issue this PR fixes**
Somewhat related to #5789


MarcosDY:
@sorindumitru looks like it broke another tests?

sorindumitru:
> @sorindumitru looks like it broke another tests?

I think that might be unrelated. The changes here are inside a single test

MarcosDY (approved):

PR #5835, opened 2025-02-02
sdsv3: Don't log error when connection closes

sorindumitru:
If we receive a nil error on the `errch` it means that the connection has been closed cleanly. We shouldn't log an error in that case.


amartinezfayo (approved):
Thank you @sorindumitru!

PR #5842, opened 2025-02-04
Fix events based cache issues introduced in 1.11.0

### LINKED ISSUE #5771: Events based cache issues in SPIRE 1.11
We’ve seen some issues over the weekend with the events based cache. Instances that have restarted have taken a lot of time to populate the cache and become available:

```
{"file":"authorized_entryfetcher.go:40","func":"[github.com/spiffe/spire/pkg/server/endpoints.NewAuthorizedEntryFetcherWithEventsBasedCache](http://github.com/spiffe/spire/pkg/server/endpoints.NewAuthorizedEntryFetcherWithEventsBasedCache)","level":"info","msg":"Building event-based in-memory entry cache","subsystem_name":"endpoints","time":"2025-01-11T14:13:56Z"}
{"file":"authorized_entryfetcher.go:45","func":"[github.com/spiffe/spire/pkg/server/endpoints.NewAuthorizedEntryFetcherWithEventsBasedCache](http://github.com/spiffe/spire/pkg/server/endpoints.NewAuthorizedEntryFetcherWithEventsBasedCache)","level":"info","msg":"Completed building event-based in-memory entry cache","subsystem_name":"endpoints","time":"2025-01-11T14:41:02Z"}

```

From the metrics I can tell this it spent this time fetching registration entries. Hydrating the cache seems to now:
1) Read all entries from the datastore
2) Read all events from the datastore
3) Read all entries associated with those events

In this case we had a large number of events, maybe ~500k, which meant the server was stuck here for ~30 minutes. Looking through the past logs I can see it taking multiple seconds or minutes before.

Additionally, we mark the fetched events as processed in https://github.com/spiffe/spire/blob/a49eaad6276b4e092e25a32c5301dbafa7373cd5/pkg/server/endpoints/authorized_entryfetcher_registration_entries.go#L138, but later on we might completely bail on processing the entries associated with them due to the datastore returning an error in https://github.com/spiffe/spire/blob/a49eaad6276b4e092e25a32c5301dbafa7373cd5/pkg/server/endpoints/authorized_entryfetcher_registration_entries.go#L230. Since we then only fetch changed events from the last processed one, we might skip some entry updates.

The version of the events based cache we had in 1.10.4 doesn’t seem to have these issues. It would be worth considering reverting the latest changes to the cache since they only improve a small edge case (de-duplicating entries that might be created due to long running transactions started before the cache was hydrated; I can’t imagine there being more than a couple and usually 0 of these).  https://github.com/spiffe/spire/pull/5509 is the change that would need to be reverted.

@amoore877, @stevend-uber tagging you here since I know you were also using the events based cache.

* **Version**: >= 1.11.0 
* **Platform**: any
* **Subsystem**: server, agent

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
spire-server events based cache

**Description of change**

**Which issue this PR fixes**
fixes #5771

### DIFF
--- a/pkg/server/endpoints/authorized_entryfetcher_attested_nodes.go
+++ b/pkg/server/endpoints/authorized_entryfetcher_attested_nodes.go
@@ -41,9 +41,6 @@ type attestedNodes struct {
 }
 
 func (a *attestedNodes) captureChangedNodes(ctx context.Context) error {
-	// first, reset what we might fetch
-	a.fetchNodes = make(map[string]struct{})
-
 	if err := a.searchBeforeFirstEvent(ctx); err != nil {
 		return err
 	}
@@ -218,7 +215,7 @@ func (a *attestedNodes) updateCachedNodes(ctx context.Context) error {
 	for spiffeId := range a.fetchNodes {
 		node, err := a.ds.FetchAttestedNode(ctx, spiffeId)
 		if err != nil {
-			return err
+			continue
 		}
 
 		// Node was deleted
@@ -230,7 +227,7 @@ func (a *attestedNodes) updateCachedNodes(ctx context.Context) error {
 
 		selectors, err := a.ds.GetNodeSelectors(ctx, spiffeId, datastore.RequireCurrent)
 		if err != nil {
-			return err
+			continue
 		}
 		node.Selectors = selectors
 
--- a/pkg/server/endpoints/authorized_entryfetcher_attested_nodes.go
+++ b/pkg/server/endpoints/authorized_entryfetcher_attested_nodes.go
@@ -187,12 +187,15 @@ func buildAttestedNodesCache(ctx context.Context, log logrus.FieldLogger, metric
 		},
 	}
 
+	if err := attestedNodes.captureChangedNodes(ctx); err != nil {
+		return nil, err
+	}
+
 	if err := attestedNodes.loadCache(ctx); err != nil {
 		return nil, err
 	}
-	if err := attestedNodes.updateCache(ctx); err != nil {
-		return nil, err
-	}
+
+	attestedNodes.emitMetrics()
 
 	return attestedNodes, nil
 }
--- a/pkg/server/endpoints/authorized_entryfetcher_registration_entries.go
+++ b/pkg/server/endpoints/authorized_entryfetcher_registration_entries.go
@@ -40,9 +40,6 @@ type registrationEntries struct {
 }
 
 func (a *registrationEntries) captureChangedEntries(ctx context.Context) error {
-	// first, reset the entries we might fetch.
-	a.fetchEntries = make(map[string]struct{})
-
 	if err := a.searchBeforeFirstEvent(ctx); err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


pkg/server/endpoints/authorized_entryfetcher_attested_nodes.go
```diff
@@ -215,7 +215,7 @@ func (a *attestedNodes) updateCachedNodes(ctx context.Context) error {
 	for spiffeId := range a.fetchNodes {
 		node, err := a.ds.FetchAttestedNode(ctx, spiffeId)
 		if err != nil {
-			return err
```
amartinezfayo, line 218: Having an error here is really an expected situation?
The FetchAttestedNode function will not return an error if the record was not found (https://github.com/spiffe/spire/blob/main/pkg/server/datastore/sqlstore/sqlstore.go#L1577), so it would only return an error if there was an SQL error. Are there other error situations that we expect as normal behavior?

pkg/server/endpoints/authorized_entryfetcher_attested_nodes.go
```diff
@@ -215,7 +215,7 @@ func (a *attestedNodes) updateCachedNodes(ctx context.Context) error {
...
+			continue
```
amartinezfayo, line 218: Having an error here is really an expected situation?
The FetchAttestedNode function will not return an error if the record was not found (https://github.com/spiffe/spire/blob/main/pkg/server/datastore/sqlstore/sqlstore.go#L1577), so it would only return an error if there was an SQL error. Are there other error situations that we expect as normal behavior?

sorindumitru: Yes, this would be SQL errors, either the something wrong with the request (which wouldn't be a problem) or likely transient errors which we should retry at some point.

faisal-memon:
i can help review

pkg/server/endpoints/authorized_entryfetcher_registration_entries.go
```diff
@@ -40,9 +40,6 @@ type registrationEntries struct {
 }
 
 func (a *registrationEntries) captureChangedEntries(ctx context.Context) error {
-	// first, reset the entries we might fetch.
-	a.fetchEntries = make(map[string]struct{})
```
faisal-memon, line 44: Where is this map being allocated now?

pkg/server/endpoints/authorized_entryfetcher_attested_nodes.go
faisal-memon: It will retry next iteration so this should be ok.

pkg/server/endpoints/authorized_entryfetcher_registration_entries.go
sorindumitru: It's allocated in `buildRegistrationEntriesCache`, which is how the structure is created.

faisal-memon: Thanks.

faisal-memon (approved):
Looks better now.

amartinezfayo (approved):
Thank you @sorindumitru!

PR #5853, opened 2025-02-09
Update database connections defaults

### LINKED ISSUE #5802: Consider changing defaults for database connections
Currently we have the following defaults for connections to the database:
* max open connections - 100
* max idle connections - 2 (I think we just take this default from `db/sql` so it might be good to at least document it as such)

This means that any deployments that has some kind of consistent load is going to end up in the situation where connections to the database are being churned. If we end up needing more connections than max idle connections we will need to create a connection, run the query we need to run, and then close it. That doesn't make for very good performance usually.

Having the number of idle connections be somewhat higher if not equal to max open connections ends with less wasted load. That's also what the [official documentation](https://go.dev/doc/database/manage-connections) recommends for high parallelism apps:
```
 By default an sql.DB keeps two idle connections at any given moment. Raising the limit can avoid frequent reconnects in programs with significant parallelism.
```

To make sure that the connections are not kept open indefinitely we can adjust ConnMaxIdleTime  to some small values, like 10-60s, so the connections are closed if they are not needed anymore, for example during a burst of activity.

jananiarunachalam:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**


pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1061,9 +1061,15 @@ func (ds *Plugin) openDB(cfg *configuration, isReadOnly bool) (*gorm.DB, string,
...
 	}
+	db.DB().SetMaxIdleConns(100) // default value
 	if cfg.MaxIdleConns != nil {
 		db.DB().SetMaxIdleConns(*cfg.MaxIdleConns)
 	}
+	defaultConnMaxLifetime, err := time.ParseDuration("30s")
+	db.DB().SetConnMaxLifetime(defaultConnMaxLifetime) // default value
+	if err != nil {
+		return nil, "", false, nil, fmt.Errorf("failed to parse conn_max_lifetime %q: %w", defaultConnMaxLifetime, err)
+	}
```
amartinezfayo, line 1072: The SetConnMaxLifetime function is already called later, based on the ConnMaxLifetime setting.
We don't need to make any change there. What we need to adjust is the ConnMaxIdleTime property calling SetConnMaxIdleTime, maybe with a value of 30s.
Also note that you will not need to make any parsing for that (i.e. `time.ParseDuration("30s")`), you can just call `SetConnMaxIdleTime(time.Second * 30`). We may define this as a constant in the function, and probably the MaxOpenConns and MaxIdleConns numbers too.

amartinezfayo (commented):
Thank you @jananiarunachalam for working on this contribution!

sorindumitru:
Also, could you update the documentation in https://github.com/spiffe/spire/blob/main/doc/plugin_server_datastore_sql.md with the new defaults?

doc/plugin_server_datastore_sql.md
```diff
@@ -3,16 +3,16 @@
...
-| conn_max_lifetime    | The maximum amount of time a connection may be reused (default: unlimited)                                                                                                                                                                                                         |
+| max_idle_conns       | The maximum number of idle connections in the pool (default: 100)                                                                                                                                                                                                                  |
+| conn_max_lifetime    | The maximum amount of time a connection may be reused (default: 30s)                                                                                                                                                                                                               |
```
jananiarunachalam, line 15: Actual change

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -1052,13 +1052,19 @@ func (ds *Plugin) openDB(cfg *configuration, isReadOnly bool) (*gorm.DB, string,
...
 	if cfg.MaxOpenConns != nil {
 		db.DB().SetMaxOpenConns(*cfg.MaxOpenConns)
 	}
+	const maxIdleConns = 100
+	db.DB().SetMaxIdleConns(maxIdleConns)
 	if cfg.MaxIdleConns != nil {
 		db.DB().SetMaxIdleConns(*cfg.MaxIdleConns)
 	}
+	const connMaxLifetime = time.Second * 30
+	db.DB().SetConnMaxLifetime(connMaxLifetime)
```
amartinezfayo, line 1067: The SetConnMaxLifetime function is already called later, based on the ConnMaxLifetime setting. We shouldn't need to make any change here. What we need to adjust is the `ConnMaxIdleTime` property calling `SetConnMaxIdleTime`, maybe with a value of 30s.

doc/plugin_server_datastore_sql.md
```diff
@@ -11,8 +11,8 @@ The `sql` plugin implements SQL based data storage for the SPIRE server using SQ
...
-| conn_max_lifetime    | The maximum amount of time a connection may be reused (default: unlimited)                                                                                                                                                                                                         |
+| max_idle_conns       | The maximum number of idle connections in the pool (default: 100)                                                                                                                                                                                                                    |
+| conn_max_lifetime    | The maximum amount of time a connection may be reused (default: 30s)                                                                                                                                                                                                         |
```
amartinezfayo, line 15: We don't need to make any change in the `conn_max_lifetime`. We need to set ConnMaxIdleTime to 30s and document it (we will not expose a setting for that).

amartinezfayo (approved):
Thank you @jananiarunachalam!

PR #5880, opened 2025-02-19
Update to Go 1.24.0

rturner3:
- Update RSA keys used in tests to have >= 1024 bits. `rsa.GenerateKey()` in Go 1.24 now throws errors when a key size < 1024 bits is used: https://pkg.go.dev/crypto/rsa#hdr-Minimum_key_size
- Update internal fork of `golang.org/x/crypto/acme/autocert/internal/acmetest` to the version from v0.33.0. There are changes to this file that are required for Go 1.24 compatibility, see [CL](https://go-review.googlesource.com/c/crypto/+/610797) for reference. Existing changes to the fork of this file were preserved.
  - One additional forked change was required to disable a newly added behavior upstream that causes any tests that encounter HTTP errors in the CA to fail. This was required to preserve our existing unit test `TestACMEAuth/new-account-tos-not-accepted` from `pkg/server/endpoints/bundle/server_test.go` that verifies some of the forked behavior around ACME CA Terms of Service being accepted. This deviation was annotated with a comment in the forked `pkg/server/endpoints/bundle/internal/acmetest/ca.go` file.
- Rework the `TestACMEAuth` subtests in `pkg/server/endpoints/bundle/server_test.go` to accommodate newer behavior of the forked `acmetest` CA from `golang.org/x/crypto`. `acmetest.CAServer` now has a `Start()` method that must be called to generate the root CA certificate and URL for the fake server. This method also registers a closer method as a test cleanup method. This resulted in a few changes to the tests:
  - Every subtest now needs its own `acmetest.CAServer` since they are not reusable across tests due to the closer method being registered as a test cleanup method
  - The `cached` subtest needed to be merged with the `initial` subtest because it depended on `acmetest.CAServer` state that can no longer be shared across subtests
- Suppress new `staticcheck` linter warnings related to importing deprecated OPA v0 packages


sorindumitru:
Looks like there some extra changes in the diff from merges, would you mind rebasing again to clean things up a bit?

rturner3:
> Looks like there some extra changes in the diff from merges, would you mind rebasing again to clean things up a bit?

Hmm, one of the merges must have gotten messed up. I just rebased on main, I think it's in a better state now.

pkg/server/api/middleware/authorization_test.go
```diff
@@ -10,8 +10,8 @@ import (
 	"net/url"
 	"testing"
 
-	"github.com/open-policy-agent/opa/v1/ast"
-	"github.com/open-policy-agent/opa/v1/storage/inmem"
+	"github.com/open-policy-agent/opa/ast"           //nolint:staticcheck // Will be removed when https://github.com/spiffe/spire/issues/5887 is closed
```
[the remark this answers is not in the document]
rturner3: Ah, now I see I missed the `/v1` component in the path during the merge. Fixed in latest push.

sorindumitru (approved):

pkg/agent/plugin/nodeattestor/awsiid/iid_test.go
```diff
@@ -38,17 +38,35 @@ const (
...
-aB3i+xyxgZNwkGEkLGRl/Uwasbp7O/sU43wh5ywWp/AG0iFe1RhwMd8LMq5ron6o
-s2eql71hJKsGEwIDAQABAmEAoDa9YcKe8Q68C5TXE8He33z3Ealea3/hET4VxEsI
-p9mfS6kpMQ+qpRSB2aMfVKP1mrAQ4/5TarrG1ZG3T/Mt9Oy1QHbzALvz2XObIvcR
-0cnG353CLQK/nobvWcwAtac5AjEA9k+1a9R6eFaO3grl9yg5XY2+MboV4wjbsDS3
-s4+MivneTPwvK6eHxtoAlYCNOAslAjEA8yy0PJw3TLBK80DryF3r/Q4wd4uYeFhN
-G6EBF0LccLB7GbKpcDHgnNjW/wObx+LXAjBeP4/G6+3U4CIYuojWMvEIaDVPp8m6
-LuiJGxLzxUjc4NF8Gb8e8CLXJxG0IxVmTXUCMQDSPJAG5rgYoUHrVPGEZU8llSLp
-99J2GUFw5Z3f0nprIukKqqA606RxdjdKeoAwLDkCMCptc0jZR3VM4w1wnwvAe0FL
-t61Ol/Q+OqWFX74JwsUU56FqPFm3Y9k7HxDILdedoQ==
+MIIEowIBAAKCAQEAsPnoGUOnrpiSqt4XynxA+HRP7S+BSObI6qJ7fQAVSPtRkqso
```
MarcosDY, line 45: may we start to use testkey.MustRSA for all test like this one? to avoid hardcoded keys?

rturner3: Seems like a good improvement. We will need some more helper methods in `testkey` to represent the keys as PKCS#1 or PKCS#8 PEM blocks to make this more convenient to use in tests. I will add these helpers and update these usages.

rturner3: Actually, taking a closer look at all these tests, it seems that the keys being in PEM format is not really a requirement, just was used as a convenient format to hardcode the keys. So these additional helpers are not really needed.

MarcosDY (approved):

Issue #4624, opened 2023-10-31
spire-agent: allow automatic re-bootstrap


sorindumitru:
If an agent is disconnected from spire-server for a longer period of time it may end up in the situation where the bundle it saved on disk will be out of date. This will lead to the agent being unable to connect to spire-server because the certificate the server presents can not be validated using the outdated bundle the agent has:
```
{"error":"create attestation client: failed to dial dns:///spire-server:14511: context deadline exceeded: connection error: desc = \"transport: authentication handshake failed: x509svid: could not verify leaf certificate: x509: certificate signed by unknown authority (possibly because of \\\"x509: ECDSA verification failure\\\" while trying to verify candidate authority certificate \\\"spireat-server\\\")\"","file":"run.go:207","func":"github.com/spiffe/spire/cmd/spire-agent/cli/run.(*Command).Run","level":"error","msg":"Agent crashed","time":"2023-10-31T06:42:52-04:00"}
```

Even if spire-agent has access to an up to date bundle from other sources (e.g. a config map that spire-server maintains and 
was used for the initial bootstrap) it will not use as long as it has a bundle saved. The operator has to intervene to delete that saved bundle to get the agent to reconnect. I think this was a deliberate design choice, but maybe it doesn't always make sense.

In a lot of cases the initial bootstrap is automated, e.g. the config map example above, so it would likely wouldn't make much of a difference if we allowed the agent to re-bootstrap. It would definitely be more convenient from the perspective of the operator since in some environments it's difficult to get the required access to remove that saved bundle.

This could be configurable and logs and metrics would need to be emitted to make it possible for operators to be notified of the re-bootstrap.

* **Version**: any
* **Platform**: any
* **Subsystem**: agent

MarcosDY:
This can be caused for 2 main reasons:
1- Agent is down for some time, and didn't got SVIDs updates
2- Server is forced to create a new bundle, and agent lost communication.

In both cases we can get into an status where our stored Agent SVID is no longer valid, and we must force a new attestation.
And using current Agent SVID will result in
```
could not verify leaf certificate: x509: certificate signed by unknown authority
```

To solve `1` when loading SVID from disk we can verify if bundle is expired and if that happens use configured bundle to force an attestation.

That is not going to solve `2`  because it is possible that bundle is not expired...
A possible solution is try to communicate with the server presenting our SVID and verify if it is success, but it it fails then try to attest agent to get new SVID.

sorindumitru:
[...]

PR #5892, opened 2025-02-22
Rebootstrap support - Phase 1

kfox1111:
**Pull Request check list**

- [ ] Documentation updated?

**Description of change**
Allows the spire-agent to rebootstrap itself if the server ca goes invalid.

**Which issue this PR fixes**
partially fixes: https://github.com/spiffe/spire/issues/4624

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -32,6 +32,7 @@ import (
 	"github.com/spiffe/spire/pkg/common/telemetry"
 	"github.com/spiffe/spire/pkg/common/uptime"
 	"github.com/spiffe/spire/pkg/common/util"
+	"github.com/spiffe/spire/pkg/common/x509util"
 	"github.com/spiffe/spire/pkg/common/version"
 	_ "golang.org/x/net/trace" // registers handlers on the DefaultServeMux
 	"google.golang.org/grpc"
@@ -105,18 +106,44 @@ func (a *Agent) Run(ctx context.Context) error {
 
 	var as *node_attestor.AttestationResult
 
+//FIXME KMF
+a.c.RetryBootstrap=true
+
 	if a.c.RetryBootstrap {
+		var rebootstrapTime time.Time
+		//FIXME KMF configurable
+		rebootstrapTimeoutSeconds := 10
+		rebootstrapTimeoutUSconds := time.Duration(float64(rebootstrapTimeoutSeconds) * float64(time.Second))
+
 		attBackoffClock := clock.New()
 		attBackoff := backoff.NewBackoff(
 			attBackoffClock,
 			bootstrapBackoffInterval,
 			backoff.WithMaxElapsedTime(bootstrapBackoffMaxElapsedTime),
+			//KMF how to ignore max time if rebootstrapping
 		)
 
 		for {
 			as, err = a.attest(ctx, sto, cat, metrics, nodeAttestor)
 			if err == nil {
 				break
+			}
+
+			if x509util.IsUnknownAuthorityError(err) {
+				if rebootstrapTime.IsZero() {
+					rebootstrapTime = time.Now()
+				}
+				seconds := time.Now().Sub(rebootstrapTime)
+				if seconds < rebootstrapTimeoutUSconds {
+					fmt.Printf("Trust Bandle and Server dont agree.... Ignoring for now. Rebootstrap timeout left: %s\n", rebootstrapTimeoutUSconds - seconds)
+				} else {
+					//FIXME loop and keep retrying syncronize... first with timeout, and then after passed timeout, add clearing bundle and deletesvid.
+					//FIXME double check... if this mechanism work with online update too... its sharing most of the code?
+					fmt.Printf("Trust Bandle and Server dont agree.... rebootstrapping")
+					a.c.TrustBundle = nil
+					sto.StoreBundle(nil)
+					err = nil
+				}
 			}
 
 			if status.Code(err) == codes.PermissionDenied {
@@ -288,20 +315,44 @@ func (a *Agent) newManager(ctx context.Context, sto storage.Storage, cat catalog
 		RotationStrategy:         rotationutil.NewRotationStrategy(a.c.AvailabilityTarget),
 		TLSPolicy:                a.c.TLSPolicy,
 	}
-
+//FIXME KMF
+a.c.RetryBootstrap=true
 	mgr := manager.New(config)
 	if a.c.RetryBootstrap {
+		var rebootstrapTime time.Time
+		//FIXME KMF configurable
+		rebootstrapTimeoutSeconds := 10
+		rebootstrapTimeoutUSconds := time.Duration(float64(rebootstrapTimeoutSeconds) * float64(time.Second))
+
 		initBackoffClock := clock.New()
 		initBackoff := backoff.NewBackoff(
 			initBackoffClock,
 			bootstrapBackoffInterval,
 			backoff.WithMaxElapsedTime(bootstrapBackoffMaxElapsedTime),
+			//KMF how to ignore max time if rebootstrapping
 		)
 
 		for {
 			err := mgr.Initialize(ctx)
 			if err == nil {
 				return mgr, nil
+			}
+			if x509util.IsUnknownAuthorityError(err) {
+				if rebootstrapTime.IsZero() {
+					rebootstrapTime = time.Now()
+				}
+				seconds := time.Now().Sub(rebootstrapTime)
+				if seconds < rebootstrapTimeoutUSconds {
+					fmt.Printf("Trust Bandle and Server dont agree.... Ignoring for now. Rebootstrap timeout left: %s\n", rebootstrapTimeoutUSconds - seconds)
+				} else {
+					//FIXME loop and keep retrying syncronize... first with timeout, and then after passed timeout, add clearing bundle and deletesvid.
+					//FIXME double check... if this mechanism work with online update too... its sharing most of the code?
+					fmt.Printf("Trust Bandle and Server dont agree.... rebootstrapping")
+					sto.DeleteSVID()
+					sto.StoreBundle(nil)
+					return nil, errors.New("Agent needs to rebootstrap. shutting down")
+
+				}
 			}
 
 			if nodeutil.ShouldAgentReattest(err) || nodeutil.ShouldAgentShutdown(err) {
--- a/pkg/agent/manager/manager.go
+++ b/pkg/agent/manager/manager.go
@@ -315,6 +315,9 @@ func (m *manager) FetchJWTSVID(ctx context.Context, entry *common.RegistrationEn
 }
 
 func (m *manager) runSynchronizer(ctx context.Context) error {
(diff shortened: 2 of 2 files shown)


amartinezfayo (commented):
Thank you, @kfox1111, for opening this PR!

Having three different flags (`retry_bootstrap`, `rebootstrap`, and `rebootstrap_delay`) to control bootstrapping/re-bootstrapping behavior seems confusing. I believe we can avoid having the `rebootstrap_delay` flag by choosing a safe default value that works well in most scenarios. This flag is particularly problematic because its validity depends on the `rebootstrap` flag. Instead of exposing it as a user-configurable option, SPIRE should infer an appropriate delay, unless we determine that tweaking this is essential for users.

We discussed making re-bootstrapping with retries the default behavior in SPIRE Agent during our maintainer's sync. One concern raised was that, particularly in Kubernetes environments, it is easier to detect an agent in a crash loop than to monitor logs for failures. However, given that we use the "WithMaxElapsedTime" backoff strategy (with a max elapsed time of one minute), agents would still crash after a minute (which seems like a relatively short duration to me).

I think that ideally, a single flag should enable or disable automatic re-bootstrapping retries, both at startup and when the agent stops recognizing the server’s authority. To move toward that goal, I propose the following plan:

In 1.12.0:

- Introduce a new flag, that may be called `automatic_rebootstrapping`. When enabled, the agent retries bootstrapping at startup and re-bootstraps if it no longer recognizes the server’s authority when syncing authorized entries. We could set a maximum elapsed time before giving up (perhaps 5–10 minutes).
- Deprecate the `retry_bootstrap setting` in favor of the new `automatic_rebootstrapping` setting, logging a warning when `retry_bootstrap` is used.

In 1.13.0:

- Remove the `retry_bootstrap` setting.

What do you think?

@sorindumitru, I know you have thoughts on this, and I’d love to hear your perspective.

sorindumitru:
> @sorindumitru, I know you have thoughts on this, and I’d love to hear your perspective.

I think this all sounds good. We should also think about how to handle non-reattestable agents. There's two situations there:
* The agent doesn't have an identity yet. We should retry, we don't even know if it's reattestable or not at this point.
* The agent already has an identity. Retrying isn't likely to help in this case, but maybe we should retry anyway in case the agent gets evicted?

There's also some cases where we decide to remove the cached agent SVID. We should make sure we don't do that for non-reattestable agents.

kfox1111:
Hi @amartinezfayo. Thanks for the review.

Totally agree on minimizing the flags. I was surprised to see retry_bootstrap in there while coding, and filed https://github.com/spiffe/spire/issues/5896 to discuss removing it in the longer term.

I tried adding just one option, rebootstrap_delay, that enables rebootstrapping. But the cli command line parser library used can't make duration's optional, so had to add a second flag to gate it. Maybe there is a way to do that, that I missed? or maybe we don't support rebootstrap_delay on the cli? config file only would work for me.

As for picking a duration, I think that may be really hard to do for the user. Some folks may want it as an absolute last resort. more then a weekend so a human is involved?) or some may want it extremely fast. First time you see a failure, immediately reboostrap. And probably a lot in between.

There's levels of safety in the system:
1. server cert in the established trust bundle - very very unlikely there is a security issue
2. bootstrapping - happens exceedly infrequently, and often with a sysadmin involved in the process, so there is actively looking for funny business.
3. rebootstrapping - happens potentially at any time. harder to protect against badness. Kind of up to each organization to decide the tradeoff between system unavailability and risk of recovery from a compromised server I think.


As for k8s support, I totally get that. Doing most things in k8s myself. Though I think the same kind of thing can be handled by a k8s readiness probe. Mark the pods unready if not attested/reattested. Then they can be alerted on via normal k8s means (pods stuck unready for more then n minutes) but still leave control in the hands of spire-agent for when to (re)attest. It may require some changes between liveness and readyness probes too. (not ready for a while doesnt mean liveness should fire and kill it.)

To get this working well, I'm guessing a timeout with WithMaxElapsedTime cant be as short as a minute. It would require multiple calls to the external url for trust bundle fetching in short order, and with lots of agents doing it, could lead to a thundering herd issue. I think the readiness thing could cover making the MaxElapsedTime much longer.

@sorindumitru, thank you too for the review

>  There's two situations there:
> * The agent doesn't have an identity yet. We should retry, we don't even know if it's reattestable or not at this point.
> * The agent already has an identity. Retrying isn't likely to help in this case, but maybe we should retry anyway in case the agent gets evicted?

For 1, the pr here does that.

For 2, I'm on the fence. If the server trust is broken, the agent is broken. Reboostrapping might not fully work, but still maybe better then just throwing x509 errors? I think there are probably some ways to Rebootstrap, then Reattest with the node cert anyway, with some extra code. Maybe that discussion should wait for a later phase though? A little worried this PR is already complicated enough without that kind of thing. Maybe we just ignore the rebootstrap flag if reattestable == false for now, or throw an error on start in that case and revist later?

pkg/agent/trustbundlesources/bundle.go
```diff
@@ -8,52 +8,230 @@ import (
...
+		b.startTime = time.Now()
+		err = b.storage.StoreBootstrapState(b.use, b.startTime, b.connectionAttempts)
+		b.updateMetrics()
+	}
+	return b.startTime, err
+}
+
+func (b *Bundle) IsBootstrap() bool {
+	return b.use != UseRebootstrap
+}
```
amartinezfayo, line 138: None of these functions seem to be covered by tests. Could you add test coverage for them?

pkg/agent/trustbundlesources/bundle.go
```diff
@@ -8,52 +8,230 @@ import (
...
+
+func (b *Bundle) SetSuccessIfRunning() error {
+	if !b.startTime.IsZero() {
+		return b.SetSuccess()
+	}
+	return nil
+}
+
+func (b *Bundle) SetSuccess() error {
+	var err error
```
amartinezfayo, line 92: This declaration is probably not needed if we have something like this:
```
	if b.storage != nil {
		if err := b.storage.StoreBootstrapState(b.use, b.startTime, b.connectionAttempts); err != nil {
			return err
		}
		b.updateMetrics()
		return b.storage.StoreBundle(b.lastBundle)
	}
	return nil
```

pkg/agent/trustbundlesources/bundle.go
```diff
@@ -8,52 +8,230 @@ import (
...
-	case bconfig.TrustBundleURL != "":
-		bundleBytes, err = downloadTrustBundle(bconfig.TrustBundleURL, bconfig.TrustBundleUnixSocket)
+	case b.config.TrustBundleURL != "":
+		u, err := url.Parse(b.config.TrustBundleURL)
 		if err != nil {
-			return err
+			return nil, false, fmt.Errorf("unable to parse trust bundle URL: %w", err)
+		}
+		if b.config.TrustBundleUnixSocket != "" {
+			params := u.Query()
```
amartinezfayo, line 165: It doesn't seem like we have these parameters documented anywhere. Is this something specific for your current implementation?

cmd/spire-agent/cli/run/run.go
```diff
@@ -387,8 +391,36 @@ func NewAgentConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool)
...
 	ac.RetryBootstrap = c.Agent.RetryBootstrap
 
+	if c.Agent.RebootstrapDelay == "" {
+		c.Agent.RebootstrapDelay = "10m"
+	}
+	delay, err := time.ParseDuration(c.Agent.RebootstrapDelay)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing rebootstrap delay duration: %w", err)
+	}
+	ac.RebootstrapMode = c.Agent.RebootstrapMode
```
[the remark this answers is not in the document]
kfox1111: Done.

cmd/spire-agent/cli/run/run.go
```diff
@@ -387,8 +391,36 @@ func NewAgentConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool)
...
 
+	switch c.Agent.RebootstrapMode {
+	case agent.RebootstrapNever:
+	case agent.RebootstrapAuto:
+	case agent.RebootstrapAlways:
+	case "":
+		c.Agent.RebootstrapMode = agent.RebootstrapNever
+	default:
+		return nil, fmt.Errorf("unknown rebootstrap mode specified: %s", c.Agent.RebootstrapMode)
+	}
```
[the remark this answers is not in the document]
kfox1111: done

doc/spire_agent.md
```diff
@@ -80,17 +82,46 @@ This may be useful for templating configuration files, for example across differ
 | `use_sync_authorized_entries` | Use SyncAuthorizedEntries API for periodically synchronization of authorized entries | true                    |
 | `require_pq_kem`              | Require use of a post-quantum-safe key exchange method for TLS handshakes            | false                   |
 
-### Initial trust bundle configuration
+### Server Attestation
```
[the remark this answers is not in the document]
kfox1111: I was kind of thinking the opposite.

bootstrapping or rebootstrapping with a trust bundle is kind of hard to articulate how it works. Especially when combined with something like trust_bundle_unix_socket. The trust bundle is an artifact of how the process works, but in the config its called out as the only thing to set, so its easy to loose track of the details of the rest of the process which are probably even more important.

We have a lot of documentation around the concept of Attesting and I think it does make a whole lot of sense in the context of the server too to reuse that terminology so we don't have to teach the same concept twice with different names. We have the server attest the node. The node has to attest the server too. We just havent been calling it that in the past.

Attesting is using additional information provided by a 3rd party to establish trust.

In the case of a trust bundle file, that is done by the human sysadmin safely transferring it across the network somehow, and copying it into place, Attesting that this is trust-able info to use. The integrity of the server attestation relies heavily on the human involved doing it properly, and we need the human to clearly understand their part in the attestation process.

In the case of a trust bundle url, we use the https CAs in the hosts web chain of trust. We're using those CAs, the certificate on the web server, and the stored info on the other end of the url to Attest the server is trustable if it matches this trust bundle. Again, the user shouldn't be too focused on the trust bundle itself, but the integrity of all the rest of that plumbing to safely attest the spire server to the node

With trust_bundle_unix_socket, its even more of a server attestor plugin just like a node attestor plugin. It is completely up to the plugin to determine what the best mechanism is to Attest the server to the node.

In all 3 cases, its not just the trust bundle that is important, its the whole process around getting a valid, up to date bootstrap trust bundle from the server, not directly but through an attestation process that involves other systems or people, to the node safely, and to the most extend possible, having things other then the node and server Attest validity along the way in the process.

Maybe some of this explanation is missing from the docs below and we might want to figure out how to add it so there is less confusion and more clarity for sysadmins to know their responsibilities?

pkg/agent/trustbundlesources/bundle.go
kfox1111: This is related to:
https://github.com/spiffe/spire/pull/5932#discussion_r2015152646

Added them to the pr here so it can be seen in the context of server attestation/rebootstrapping.

PR #5909, opened 2025-02-27
Apply gopls modernize pass

c4rlo:
Recent [gopls](https://pkg.go.dev/golang.org/x/tools/gopls) [versions](https://github.com/golang/tools/releases/tag/gopls%2Fv0.18.0) contain a "[modernize](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize)" pass that can also be run standalone. Run this on the spire repo, with small manual fixups afterwards.

The command I ran was:

    go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix ./...

If this finds favour, it may also make sense to add this as a linter to CI. It seems golangci-lint [won't](https://github.com/golangci/golangci-lint/discussions/5466) be adding support for it directly, so we'd have to do it ourselves.

### DIFF
--- a/pkg/agent/plugin/workloadattestor/docker/docker.go
+++ b/pkg/agent/plugin/workloadattestor/docker/docker.go
@@ -74,7 +74,7 @@ type dockerPluginConfig struct {
 
 	UnusedKeyPositions map[string][]token.Pos `hcl:",unusedKeyPositions"`
 
-	Experimental experimentalConfig `hcl:"experimental,omitempty" json:"experimental,omitempty"`
+	Experimental experimentalConfig `hcl:"experimental,omitempty" json:"experimental"`
 
 	containerHelper *containerHelper
 	dockerOpts      []dockerclient.Opt
--- a/pkg/server/bundle/client/manager.go
+++ b/pkg/server/bundle/client/manager.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"maps"
 	"sync"
 	"time"
 
@@ -350,8 +351,6 @@ func calculateNextUpdate(b *spiffebundle.Bundle) time.Duration {
 
 func cloneTrustDomainConfigs(configs map[spiffeid.TrustDomain]TrustDomainConfig) map[spiffeid.TrustDomain]TrustDomainConfig {
 	clone := make(map[spiffeid.TrustDomain]TrustDomainConfig, len(configs))
-	for k, v := range configs {
-		clone[k] = v
-	}
+	maps.Copy(clone, configs)
 	return clone
 }
--- a/pkg/server/bundle/client/sources.go
+++ b/pkg/server/bundle/client/sources.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"maps"
 	"sync"
 
 	"github.com/sirupsen/logrus"
@@ -55,9 +56,7 @@ func (s *TrustDomainConfigSet) GetTrustDomainConfigs(context.Context) (map[spiff
 
 func duplicateTrustDomainConfigMap(in TrustDomainConfigMap) TrustDomainConfigMap {
 	out := make(TrustDomainConfigMap, len(in))
-	for td, config := range in {
-		out[td] = config
-	}
+	maps.Copy(out, in)
 	return out
 }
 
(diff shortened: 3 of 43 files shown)


pkg/server/credvalidator/validator.go
```diff
@@ -166,20 +167,10 @@ func checkX509CertificateExpiration(cert *x509.Certificate, now time.Time) error
 }
 
 func hasExtKeyUsage(extKeyUsage []x509.ExtKeyUsage, want x509.ExtKeyUsage) bool {
```
sorindumitru, line 169: nit: we could probably just remove these functions now and inline the `slices.Contains` invocation.

pkg/server/bundle/client/manager.go
```diff
@@ -350,8 +351,6 @@ func calculateNextUpdate(b *spiffebundle.Bundle) time.Duration {
 
 func cloneTrustDomainConfigs(configs map[spiffeid.TrustDomain]TrustDomainConfig) map[spiffeid.TrustDomain]TrustDomainConfig {
 	clone := make(map[spiffeid.TrustDomain]TrustDomainConfig, len(configs))
-	for k, v := range configs {
-		clone[k] = v
-	}
+	maps.Copy(clone, configs)
 	return clone
 }
```
sorindumitru, line 356: It think we might be able to replace this whole function with `maps.Clone()`.

pkg/server/bundle/client/sources.go
```diff
@@ -55,9 +56,7 @@ func (s *TrustDomainConfigSet) GetTrustDomainConfigs(context.Context) (map[spiff
 
 func duplicateTrustDomainConfigMap(in TrustDomainConfigMap) TrustDomainConfigMap {
 	out := make(TrustDomainConfigMap, len(in))
-	for td, config := range in {
-		out[td] = config
-	}
+	maps.Copy(out, in)
 	return out
 }
```
sorindumitru, line 61: This could also be replaced by a maps.Clone() call

sorindumitru (commented):
Thanks @c4rlo! Looks like there's a merge conflict, could you rebase your changes on latest main?

pkg/server/bundle/client/manager.go
c4rlo: Doing this actually turns out to cause a test failure, as it changes behaviour: `maps.Clone()` returns `nil` for a `nil` map. So holding off on making those changes.

c4rlo:
Latest force-push is a result of me running modernize again – it resulted in a few more changes.

pkg/agent/plugin/workloadattestor/docker/docker.go
```diff
@@ -74,7 +74,7 @@ type dockerPluginConfig struct {
 
 	UnusedKeyPositions map[string][]token.Pos `hcl:",unusedKeyPositions"`
 
-	Experimental experimentalConfig `hcl:"experimental,omitempty" json:"experimental,omitempty"`
+	Experimental experimentalConfig `hcl:"experimental,omitempty" json:"experimental"`
```
c4rlo, line 77: I think modernize removed `omitempty` here because that doesn't actually do anything for fields of struct type (it can never be "empty"). Possibly what was actually meant was `omitzero`, in which case the field would be omitted if the struct has the "zero" value; but that would be a change in behaviour.

Likely the same applies for the `hcl` annotation, but modernize didn't touch that so neither did I.

pkg/server/plugin/upstreamauthority/certmanager/internal/v1/types_certificaterequest.go
```diff
@@ -49,7 +49,7 @@ const (
 // point in time request for a certificate and cannot be re-used.
 type CertificateRequest struct {
 	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
+	metav1.ObjectMeta `json:"metadata"`
```
sorindumitru, line 52: This file is copied from `cert-manager` because they discourage importing from their module. We should avoid making modifications to this, so could you revert the change in this file? 

This pattern seems to be common in general with k8s, see for example https://pkg.go.dev/k8s.io/api@v0.27.2/core/v1#Pod

sorindumitru (approved):
LGTM, thanks @c4rlo!

PR #5917, opened 2025-03-03
Default 'use_legacy_downstream_x509_ca_ttl' to false

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Downstream spire-server X509 CA TTL

**Description of change**
We were supposed to default  'use_legacy_downstream_x509_ca_ttl' to false in 1.11.0 but we forgot. Switching it now and scheduling the removal of the flag in 1.13.0


cmd/spire-server/cli/run/run.go
```diff
@@ -575,10 +575,9 @@ func NewServerConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool
 			sc.Log.Info("Using preferred downstream X509 CA TTL calculation")
 		}
 	} else {
-		// The default value should be false in SPIRE 1.11.0 and the flag
-		// removed in SPIRE 1.12.0.
-		sc.UseLegacyDownstreamX509CATTL = true
-		sc.Log.Info("Using legacy downstream X509 CA TTL calculation by default; this default will change in a future release")
+		// The flag should be removed in SPIRE 1.13.0.
+		sc.UseLegacyDownstreamX509CATTL = false
+		sc.Log.Info("Using preferred downstream X509 CA TTL calculation by default; this option will be removed in a future release")
```
amartinezfayo, line 580: I think that this message is a little confusing, because it may not be clear what "this option" means (you may think that's referring to "Using preferred downstream X509 CA TTL calculation by default".

I'm thinking that if the user doesn't set the `use_legacy_downstream_x509_ca_ttl` setting (as it's in this flow), we may just log "Using preferred downstream X509 CA TTL calculation".

On the other hand, I think that if c.Server.UseLegacyDownstreamX509CATTL != nil we should always log a warning saying that this setting will be removed in a future release (currently we are logging this only if it's set to true).

What do you think?

amartinezfayo (approved):
🎉

PR #5936, opened 2025-03-10
Update to Ubuntu 24.04 base dev image image.

### LINKED ISSUE #5935: Spire101 Dockerfile.dev example uses a very old version of Ubuntu
The "Spire101 Quickstart" uses Ubuntu LTS version 16.04 ("Xenial") which has been out of standard support since April 2021 (See: Ubuntu Release Cycle.) Updating it to a modern version of Ubuntu will provide updated OS tools ensuring ongoing support.

There is no functionality changes needed or any current problem with the current version other than age, however why wait for it to break?

h110hawk:
**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
No change.

**Description of change**
Updates Dockerfile.dev to use Ubuntu LTS 24.04 and other necessary workflow changes.

1. uid 1000 -> 1001 because the default `ubuntu` user is uid 1000.
2. git configuration of the /spire directory as safe is required.
3. Update example certificate write directory to `/tmp` as the current directory doesn't have permission for `workload` to write to it.

**Which issue this PR fixes**
fixes #5935

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM ubuntu:xenial
+FROM ubuntu:24.04
 WORKDIR /spire
 RUN apt-get update && apt-get -y install \
     curl unzip git build-essential ca-certificates libssl-dev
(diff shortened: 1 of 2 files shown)


sorindumitru (approved):
Thanks for the contribution @h110hawk!

PR #5939, opened 2025-03-10
BundlePublisher exec plugin

kfox1111:
**Pull Request check list**

- [x] Documentation updated?

**Description**
Adds a BundlePublisher that execs a command with the bundle on stdin

Fixes: https://github.com/spiffe/spire/issues/5944


sorindumitru:
@kfox1111 Could you also open an issue for this so we can have some discussion about this there?

kfox1111:
https://github.com/spiffe/spire/issues/5944

pkg/server/plugin/bundlepublisher/exec/exec.go
```diff
@@ -0,0 +1,203 @@
...
+		status.ReportErrorf("unable to decode configuration: %v", err)
+		return nil
+	}
+
+	if len(newConfig.Cmd) < 1 {
+		status.ReportError("configuration is missing cmd")
+	}
+	if newConfig.Format == "" {
+		status.ReportError("configuration is missing the bundle format")
+	}
```
faisal-memon, line 57: Maybe have default format?

pkg/server/plugin/bundlepublisher/exec/exec.go
```diff
@@ -0,0 +1,203 @@
...
+	_, notes, err := pluginconf.Build(req, buildConfig)
+
+	return &configv1.ValidateResponse{
+		Valid: err == nil,
+		Notes: notes,
+	}, err
+}
+
+// PublishBundle puts the bundle in the configured S3 bucket name and
+// object key.
```
faisal-memon, line 118:
```suggestion
// PublishBundle runs the specified command and passes the bundle through stdin
```

doc/plugin_server_bundlepublisher_exec.md
```diff
@@ -0,0 +1,38 @@
+# Server plugin: BundlePublisher "exec"
+
+The `exec` plugin executes the specified command giving it the current trust bundle of the server
+on its stdin
```
faisal-memon, line 4:
```suggestion
The `exec` plugin executes the specified command and passes the current trust bundle of the server
on its stdin
```

pkg/server/plugin/bundlepublisher/exec/exec.go
```diff
@@ -0,0 +1,203 @@
...
+func buildConfig(coreConfig catalog.CoreConfig, hclText string, status *pluginconf.Status) *Config {
+	newConfig := new(Config)
+
+	if err := hcl.Decode(newConfig, hclText); err != nil {
+		status.ReportErrorf("unable to decode configuration: %v", err)
+		return nil
+	}
+
+	if len(newConfig.Cmd) < 1 {
+		status.ReportError("configuration is missing cmd")
```
faisal-memon, line 53: If there is no cmd should we exit and return error?

pkg/server/plugin/bundlepublisher/exec/exec.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	formatter := bundleformat.NewFormatter(req.Bundle)
+	bundleBytes, err := formatter.Format(config.bundleFormat)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "could not format bundle: %v", err.Error())
+	}
+
+	// Push trust bundle to user configured process
+	// We use gosec -- the annotation below will disable a security check that users didnt specify the command
```
faisal-memon, line 142:
```suggestion
	// We use gosec -- the annotation below will disable a security check that users didn't specify the command
```

pkg/server/plugin/bundlepublisher/exec/exec.go
```diff
@@ -0,0 +1,203 @@
...
+	return builtin(New())
+}
+
+func New() *Plugin {
+	return newPlugin()
+}
+
+// Config holds the configuration of the plugin.
+type Config struct {
+	Cmd    []string `hcl:"cmd" json:"cmd"`
```
faisal-memon, line 36: Should the Args be a separate configurable?

doc/plugin_server_bundlepublisher_exec.md
```diff
@@ -0,0 +1,39 @@
...
+| cmd               | Command to run                                                                                                                                                 | Yes.                                                                   |                                                     |
+| args              | Arguments to the command to run                                                                                                                                | No.                                                                    |                                                     |
+| format            | Format in which the trust bundle is stored, &lt;spiffe &vert; jwks &vert; pem&gt;. See [Supported bundle formats](#supported-bundle-formats) for more details. | No.                                                                    | spiffe                                              |
```
faisal-memon, line 12:
```suggestion
| cmd               | Command to run                                                                                                                                                 | Yes                                                                    |                                                     |
| args              | Arguments to the command to run                                                                                                                                | No                                                                     |                                                     |
| format            | Format in which the trust bundle is stored, &lt;spiffe &vert; jwks &vert; pem&gt;. See [Supported bundle formats](#supported-bundle-formats) for more details. | No                                                                     | spiffe                                              |
```

pkg/server/plugin/bundlepublisher/exec/exec.go
```diff
@@ -0,0 +1,204 @@
...
+	bundleBytes, err := formatter.Format(config.bundleFormat)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "could not format bundle: %v", err.Error())
+	}
+
+	// Push trust bundle to user configured process
+	// We use gosec -- the annotation below will disable a security check that users didn't specify the command
+	// Its their command.
+	/* #nosec G204 */
+	cmd := exec.Command(config.Cmd, config.Args...)
```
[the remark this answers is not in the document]
kfox1111: Doesnt go reap child processes automatically?

We could put a timeout around it perhaps.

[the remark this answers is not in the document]
kfox1111: updated to use exec.CommandContext

pkg/server/plugin/bundlepublisher/exec/exec_test.go
```diff
@@ -0,0 +1,185 @@
...
+		expectCode   codes.Code
+		expectMsg    string
+		config       *Config
+		bundle       *types.Bundle
+		putObjectErr error
+	}{
+		{
+			name:   "success",
+			bundle: testBundle,
+			config: &Config{
```
MarcosDY, line 93: This test is not verifying if we are really sending bundles to some place, what do you think about saving the response in a temporal file? or create a fake for `exec.CommandContext` using hook and store whatever command is send to it?
with this we can be sure we are executing what is expected

pkg/server/plugin/bundlepublisher/exec/exec.go
```diff
@@ -0,0 +1,204 @@
...
+	bundleBytes, err := formatter.Format(config.bundleFormat)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "could not format bundle: %v", err.Error())
+	}
+
+	// Push trust bundle to user configured process
+	// We use gosec -- the annotation below will disable a security check that users didn't specify the command
+	// Its their command.
+	/* #nosec G204 */
+	cmd := exec.CommandContext(ctx, config.Cmd, config.Args...)
```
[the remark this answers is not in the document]
kfox1111: This one, I pretty strongly disagree with. We can not protect the user from all possible commands a shell can execute. There are infinite possibilities. We can take a stab at it, but will always miss things. and then we are more liable, as "we didn't project me from doing this other unwise thing". As we accepted that we will protect them at all, then not fully protecting them is on us, not them. :/

Its completely under the sysadmins control.  They know their systems well and what is appropriate. They configure the exec plugin, they can test it on their test cluster to ensure proper functioning before deploying to production, its up to them exactly what it will do.

They may write the bundle to a temporary file, scp it somewhere else, then remove the temporary file, for example. It would be exceptionally hard to parse bash or powershell or whatever and guess all the ways things could break. Or, if we only allow a few commands possible to run, it completely defeats the purpose of the plugin I think.

Security wise, forking off and running in the same context as the server is basically how the plugins themselves work now. So, not sure how much it buys us to add settings like "switch user to this before execing". could do that in the command itself I think.

An exec plugin allows certain things to be written quickly/easily by a sysadmin. But it doesn't preclude those things from eventually being made regular BundlePublisher plugins at some point too.

doc/plugin_server_bundlepublisher_exec.md
```diff
@@ -0,0 +1,39 @@
...
+
+### JWKS format
+
+The trust bundle is encoded as an RFC 7517 compliant JWK Set, omitting SPIFFE-specific parameters. Both the JWT authorities and the X.509 authorities are included.
+
+### PEM format
+
+The trust bundle is formatted using PEM encoding. Only the X.509 authorities are included.
+
+## Sample configuration
```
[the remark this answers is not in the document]
kfox1111: It should work on all platforms.

I can take a rough stab, but don't have a windows box to test on. Do you?

pkg/server/plugin/bundlepublisher/exec/exec.go
kfox1111: Should be fixed

pkg/server/plugin/bundlepublisher/exec/exec_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+			// Check that the plugin has the expected configuration.
+			tt.config.bundleFormat, err = bundleformat.FromString(tt.config.Format)
+			require.NoError(t, err)
+			require.Equal(t, tt.config, p.config)
+		})
+	}
+}
+
+func TestPublishBundle(t *testing.T) {
```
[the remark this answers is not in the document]
kfox1111: I took a stab at it.

pkg/server/plugin/bundlepublisher/exec/exec.go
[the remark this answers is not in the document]
MarcosDY: Can we achieve this with a BundlePublisher that writes bundle updates to disk?

Instead of adding an exec command, we could update this plugin to store updates on disk. Then, another process could watch for file updates. This approach would avoid the need to add the capability to run any command that consumes standard input.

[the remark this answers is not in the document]
kfox1111: what if (I do), want to simply
cmd = "bash"
args = ["-c", "ssh othermachine 'cat > trustbundle.spiffe"]

No disk writes needed locally.

Inotify would mean more daemons for a sysadmin to manage, and have to carefully secure communications on. I'd rather pull the data right from a pipe and push it to gpg, openssl, or ssh or other tools as part of my scripting, rather then let it land on disk in the middle and have to reconcile the local copy my own code. The idea is let the sysadmin be able to simply use trust bundles without a lot of work. Otherwise, writing a custom BundlePublisher plugin is probably worth the pain.

sorindumitru:
We discussed this during the maintainer sync yesterday. We can see how this can be useful, but we think this kind of plugin puts too much configuration burden on the users which are going against our secure by default approach. It would also be a plugin that would be very hard for us to provide support for since most of logic would be in the command being executed.

We're going to close this for now. Some alternatives that you can consider:
* A dedicated plugin for your specific use case.
* Keeping this as an external plugin.

PR #5942, opened 2025-03-10
valid_csr test: move mock clock initialisation before any operations

### LINKED ISSUE #5841: Flaky test - TestMintX509CA/valid_CSR
```
--- FAIL: TestMintX509CA (0.06s)
    --- FAIL: TestMintX509CA/valid_CSR (0.01s)
        load.go:90: log: time="2025-02-03T17:56:19Z" level=error msg="Failed to watch the Workload API: EOF"
            
        spire_test.go:305: 
            	Error Trace:	/Users/runner/work/spire/spire/pkg/server/plugin/upstreamauthority/spire/spire_test.go:305
            	Error:      	Not equal: 
            	            	expected: 1h0m0s
            	            	actual  : 59m59s
            	Test:       	TestMintX509CA/valid_CSR
FAIL
FAIL	github.com/spiffe/spire/pkg/server/plugin/upstreamauthority/spire	0.113s
```

https://github.com/spiffe/spire/actions/runs/13119600185/job/36602222871?pr=5838

sorindumitru:
Otherwise by the time we run the test the seconds of the clock might have rolled and cause us to think that the CA is valid for 1 hour - 1 second instead of 1 hour.

**Affected functionality**
tests

**Description of change**
move mock clock initialisation before any operations, otherwise by the time we run the test the seconds of the clock might have rolled and cause us to think that the CA is valid for 1 hour - 1 second instead of 1 hour.

**Which issue this PR fixes**
fixes #5841


amartinezfayo (commented):
Thank you @sorindumitru for looking at this.
I don't think that clock.Mock advances the time unless that the Add method is called? In that case, I don't think that this change would have any effect.

sorindumitru:
> I don't think that clock.Mock advances the time unless that the Add method is called? In that case, I don't think that this change would have any effect.

The problem is that the CA is created at the start of the function, because it's needed for some test expectations. It may be valid from 13:25:51.000 to 14:25:51.000, for example. This will also cap (or at least that's what it seemed it did) the lifetime of the issued intermediate CAs. The mock clock gets initialized for each subtest, using the current time. If the test takes longer  than a second, the mock clock will have the now() as 14:25:52.000 as opposed to  (14:25:51).

During the test we verify that that the difference between the `NotAfter` of the certs, which is capped at 14:25:51.000, and now() matches the TTL. But if we are in the situation where the seconds of `now()` has changed, we will be one second short (because `NotAfter` is capped based on the start of the test).

I was able to consistently reproduce this issue using `-count 128` and it seemed to fix the issue with this patch. Another alternative would be to record `now()` at the start of test and in each subtest we create the mock clock using `NewMockAt()`.

PR #5967, opened 2025-03-22
integration tests: wait for spire-server to come up

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Integration tests

**Description of change**
We don't wait for spire-server to be up before trying to make use of it. This sometimes leads to failures in integration tests, for example when trying to read the bundle because it might not be available yet.

### DIFF
--- a/test/integration/suites-windows/windows-service/01-start-server-service
+++ b/test/integration/suites-windows/windows-service/01-start-server-service
@@ -1,7 +1,7 @@
 #!/bin/bash
 source ./common
 
-docker-spire-server-up spire-server
+docker-up spire-server
 
 create-service spire-server C:/spire/bin/spire-server.exe
 start-service spire-server run -config C:/spire/conf/server/server.conf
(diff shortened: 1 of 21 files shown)


amartinezfayo (dismissed):
Thank you for this, @sorindumitru!

test/integration/suites-windows/windows-service/01-start-server-service
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 source ./common
 
-docker-up spire-server
+docker-spire-server-up spire-server
```
amartinezfayo, line 4: It seems that the `check-server-started()` check is not working well on Windows.

sorindumitru: Seems like this docker-compose suite doesn't bring the server up with the container. It  ends up being with the following commands.

I'll just leave this one like this for now, if it ends up being a problem we can revisit. It asserts the windows service is up, so better than nothing, but doesn't actually check that it is ready to serve requests.

amartinezfayo (approved):

PR #6021, opened 2025-04-20
Extract the trust bundle code from agent/cli/run

kfox1111:
**Pull Request check list**

- [x] Documentation updated?

**Description of change**
The trust bundle fetching code is only available from the cli run file. It will need to be in a more common location for reattestation code to come. We move the existing code to a common location with as few changes as possible. This will make future reviews easier as changes will not involve moving large blocks of code at the same time as changing behavior for reattestation support.


amartinezfayo (commented):
Thank you @kfox1111 for this, it looks good.
Since SetupTrustBundle is now a public function in the trustbundlesources package, I think that we should probably have test coverage for it in the trustbundlesources package. Could you add tests there?

amartinezfayo (approved):
Thank you @kfox1111!

PR #6034, opened 2025-04-28
Improve performance for fetching authorized entries

sorindumitru:
This does a bunch of speedups to authorized entries lookup (used for syncing agent entries but also for signing SVIDs):
* Make the caches return a new type, ReadOnlyEntry. This delays the clone of the entry from the cache until we actually need to clone it. Useful for SyncAuthorizedEntries where we retrieve all entries that the agent is authorized for but would only need to clone a few of them that we actually send to the agent. Also useful for signing SVIDs which can clone just the fields of the entry that are used.
* Switch from proto.Clone to our own Clone method. This allows us to combine cloning with output mask application, which avoid cloning some fields. This is also 1.5-2x faster than proto.Clone
* The entry cache can use just the path component of the entry when storing entries in the cache. This avoid some unnecessary work, which was seen in spire-server cpu profile data.

Some of the benchmarks we have in the tests show good improvements. Seen improvements also while running this, although somewhat smaller than the benchmark since some of the work still needs to be done in other places. Before:
```
BenchmarkBuildInMemory-16                             55          21637647 ns/op        14817260 B/op     100311 allocs/op
BenchmarkGetAuthorizedEntriesInMemory-16            5833            205069 ns/op          291966 B/op       3005 allocs/op
BenchmarkEntryLookup-16                               46          25513459 ns/op         5123503 B/op     136200 allocs/op
```

and after:
```
BenchmarkBuildInMemory-16                             54          19256267 ns/op         9915763 B/op     100310 allocs/op
BenchmarkGetAuthorizedEntriesInMemory-16           64569             17356 ns/op            9338 B/op         10 allocs/op
BenchmarkEntryLookup-16                              100          11445476 ns/op          156498 B/op       1536 allocs/op
```

For the events based cache. Before:
```
BenchmarkGetAuthorizedEntriesInMemory-16            9115            131016 ns/op          205207 B/op       1513 allocs/op
BenchmarkEntryLookup-16                               57          18919070 ns/op          209983 B/op       2461 allocs/op
```

and after:
```
BenchmarkGetAuthorizedEntriesInMemory-16           23606             50479 ns/op           68260 B/op         13 allocs/op
BenchmarkEntryLookup-16                               62          17778846 ns/op           91850 B/op       1180 allocs/op
```

### DIFF
--- a/pkg/server/api/entry.go
+++ b/pkg/server/api/entry.go
@@ -11,11 +11,121 @@ import (
 	"github.com/spiffe/spire/pkg/common/protoutil"
 	"github.com/spiffe/spire/pkg/common/x509util"
 	"github.com/spiffe/spire/proto/spire/common"
+	"google.golang.org/protobuf/proto"
 )
 
 const (
 	hintMaximumLength = 1024
 )
+
+type ReadOnlyEntry struct {
+	entry *types.Entry
+}
+
+func NewReadOnlyEntry(entry *types.Entry) ReadOnlyEntry {
+	return ReadOnlyEntry{
+		entry: entry,
+	}
+}
+
+func (e *ReadOnlyEntry) GetSpiffeId() *types.SPIFFEID {
+	return &types.SPIFFEID{
+		TrustDomain: e.entry.SpiffeId.TrustDomain,
+		Path:        e.entry.SpiffeId.Path,
+	}
+}
+
+func (e *ReadOnlyEntry) GetX509SvidTtl() int32 {
+	return e.entry.X509SvidTtl
+}
+
+func (e *ReadOnlyEntry) GetJwtSvidTtl() int32 {
+	return e.entry.JwtSvidTtl
+}
+
+func (e *ReadOnlyEntry) GetDnsNames() []string {
+	return slices.Clone(e.entry.DnsNames)
+}
+
+func (e *ReadOnlyEntry) GetRevisionNumber() int64 {
+	return e.entry.RevisionNumber
+}
+
+// Manually clone the entry instead of using the protobuf helpers
+// since those are two times slower.
+func (e *ReadOnlyEntry) Clone(mask *types.EntryMask) *types.Entry {
+	if mask == nil {
+		return proto.Clone(e.entry).(*types.Entry)
+	}
+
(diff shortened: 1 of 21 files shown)


pkg/server/api/entry.go
```diff
@@ -11,12 +11,130 @@ import (
...
+	return slices.Clone(e.entry.DnsNames)
+}
+
+func (e *ReadOnlyEntry) GetRevisionNumber() int64 {
+	return e.entry.RevisionNumber
+}
+
+func (e *ReadOnlyEntry) GetCreatedAt() int64 {
+	return e.entry.CreatedAt
+}
```
amartinezfayo, line 60: It doesn't seem like we there is test coverage for this, could you add that?

amartinezfayo (commented):
Thank you @sorindumitru for this, it looks great!

sorindumitru:
One thing I realized was that it's somewhat possible to switch the trust domain of a deployment. I've added some changes to filter the entries with the wrong trust domain from the cache and a test for that too.

pkg/server/api/entry.go
```diff
@@ -11,12 +11,130 @@ import (
...
+
+// Manually clone the entry instead of using the protobuf helpers
+// since those are two times slower.
+func (e *ReadOnlyEntry) Clone(mask *types.EntryMask) *types.Entry {
```
amartinezfayo, line 64: I'm a bit concerned about the potential for field skew here. If the types.Entry struct is updated in the future (e.g., new fields are added), the manual Clone implementation will also need to be updated. Otherwise, newly added fields won't be cloned, which could lead to subtle, hard-to-detect bugs.

It might be helpful to add a comment warning about this, or ideally, include a test (perhaps using reflection) that ensures all fields are properly cloned.

sorindumitru: There's a test for this, which indeed uses reflection to make sure all fields are cloned: https://github.com/spiffe/spire/blob/d73d8a7b11c8152811cb4de843fe2173d55d0c16/pkg/server/api/entry_test.go#L727

amartinezfayo: I missed that, thanks!

Issue #5896, opened 2025-02-25
Consider removing retry_bootstrap=false


kfox1111:
With retry_bootstrap, we have two different bootstrap codepaths to test/maintain. It also complicates the code for https://github.com/spiffe/spire/issues/4624 having two different codepaths.

Consider simplifying the code base by removing support for retry_bootstrap=false

kfox1111:
I think we may want to do the following:

In 1.12.2 deprecate the retry_bootstrap flag

In 1.13.0 change default from retry_bootstrap=false to rebootstrap=true

In 1.14.0 remove the flag and the code path for retry_bootstrap=false

kfox1111:
Based on usage in the chart, I don't think we want to deprecate the flag until the default is changed.

PR #6050, opened 2025-05-07
Deprecate retry_bootstrap

kfox1111:
Partially-fixes: https://github.com/spiffe/spire/issues/5896

### DIFF
--- a/cmd/spire-agent/cli/run/run.go
+++ b/cmd/spire-agent/cli/run/run.go
@@ -426,6 +426,10 @@ func NewAgentConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool)
 		ac.LogReopener = log.ReopenOnSignal(logger, reopenableFile)
 	}
 
+	if ac.RetryBootstrap {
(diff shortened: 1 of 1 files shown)


cmd/spire-agent/cli/run/run.go
```diff
@@ -426,6 +426,10 @@ func NewAgentConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool)
 		ac.LogReopener = log.ReopenOnSignal(logger, reopenableFile)
 	}
 
+	if c.Agent.RetryBootstrap {
+		ac.Log.Warn("The 'retry_bootstrap' configuration is deprecated. It will be enabled by default in SPIRE 1.13 and the option removed in SPIRE 1.14.")
+	}
+
```
MarcosDY, line 432: We are unable to change default behaviors without a complete minor rotation, so we need to move this to 1.14, before removing, 
And need to display this to users that are not using it.
Maybe something like this?

```suggestion
	if !c.Agent.RetryBootstrap {
		ac.Log.Warn("The 'retry_bootstrap' configuration is going to be default in SPIRE 1.14 and the option removed in SPIRE 1.15.")
	}

```

PR #6058, opened 2025-05-08
Delete some dead code

c4rlo:
I ran

    go run golang.org/x/tools/cmd/deadcode@latest -test ./...

and manually reviewed the result.

I selected a subset of "dead code" (never-called functions) to delete. I did not delete some other functions that seemed like they could still be useful in future.

One function flagged as dead was `DeltaSecretsTest` in `pkg/agent/endpoints/sdsv3/handler_test.go`, which seems like it was a misspelled function that was intended as a test case. I tried renaming it to `TestDeltaSecrets` but it turned out to be broken and hard to fix. Since it only tests that an RPC method is unimplemented, I opted to just remove the function instead.


c4rlo:
The CI failure is Windows only and looks unrelated to my changes.

sorindumitru (approved):

sorindumitru:
thanks, @c4rlo!

nikotih:
Hi @sorindumitru 

Just to let you know that this removed fakeentryclient that we used in our unit tests. Not a big deal, we just copied over what you had here

But just wanted to clarify that it was intentional and you dont plan to bring it back to existing fakes:
https://github.com/spiffe/spire/tree/main/test/fakes

PR #6059, opened 2025-05-09
Add ARM64 support for integration tests in GitHub workflows

### LINKED ISSUE #3740: Run integration tests for arm64 on release builds
#3607 introduces support for arm64-based release container images. Integration tests should be run against these artifacts in the release pipeline, at a minimum.

amartinezfayo:
- Add ARM64 support to integration tests in GitHub workflows.
- Ensure integration-k8s is included as a required check for success (was previously missing).

Fixes #3740.


.github/workflows/pr_build.yaml
```diff
@@ -239,10 +239,9 @@ jobs:
     outputs:
       test: ${{ steps.set-matrix.outputs.test }}
   
-
   integration:
-    name: integration (linux)
-    runs-on: ubuntu-22.04
+    name: integration (${{ matrix.arch }})
```
sorindumitru, line 243: One small nit I have for this is that the name of the integration steps have lost the job index:
```
- PR Build / integration (linux) (5, 4)
+ PR Build / integration (x64)
```

I wonder if that's something that we can bring back, maybe something like:
```suggestion
    name: integration (${{ matrix.arch }}) (${{ strategy.job-index}}/${{ strategy.job-total }})
```

sorindumitru (approved):
LGTM, thanks @amartinezfayo!

Issue #6051, opened 2025-05-07
retry_bootstrap delays metrics/health checks


kfox1111:
When retry_bootstrap is true, during the startup/retrying, health checks / metrics are not started. This time is currently is capped at 5 minutes, requiring fairly long health check delays on Kubernetes, and no ability to collect metrics on what its doing. In the future, the length of time will be greater then 5 minutes, so especially important to fix by then.

PR #6063, opened 2025-05-09
Fix metrics/health checks when retry_bootstrap=true

kfox1111:
**Pull Request check list**

- [ ] ~~Documentation updated?~~

**Affected functionality**
When retry_bootstrap is true, during the startup/retrying, health checks / metrics are not started. This time is currently is capped at 5 minutes, requiring fairly long health check delays on Kubernetes, and no ability to collect metrics on what its doing.

**Which issue this PR fixes**
fixes: https://github.com/spiffe/spire/issues/6051

### DIFF
--- a/pkg/common/health/health.go
+++ b/pkg/common/health/health.go
@@ -20,6 +20,10 @@ const (
 
 // State is the health state of a subsystem.
 type State struct {
+	// Started is whether the subsystem is finished starting.
+	// if undefined, it is treated as started=true.
+	Started *bool
+
 	// Live is whether the subsystem is live (i.e. in a good state
 	// or in a state it can recover from while remaining alive). Global
 	// liveness is only reported true if all subsystems report live.
@@ -134,27 +138,37 @@ func (c *checker) ListenAndServe(ctx context.Context) error {
 	return nil
 }
 
+// StartedState returns the global startup state.
+func (c *checker) StartedState() bool {
+	startup, _, _, _, _ := c.checkStates()
+
+	return startup
+}
+
 // LiveState returns the global live state and details.
 func (c *checker) LiveState() (bool, any) {
-	live, _, details, _ := c.checkStates()
+	_, live, _, details, _ := c.checkStates()
 
 	return live, details
 }
 
 // ReadyState returns the global ready state and details.
 func (c *checker) ReadyState() (bool, any) {
-	_, ready, _, details := c.checkStates()
+	_, _, ready, _, details := c.checkStates()
 
 	return ready, details
 }
 
-func (c *checker) checkStates() (bool, bool, any, any) {
-	isLive, isReady := true, true
+func (c *checker) checkStates() (bool, bool, bool, any, any) {
+	isStarted, isLive, isReady := true, true, true
 
 	liveDetails := make(map[string]any)
 	readyDetails := make(map[string]any)
 	for subsystemName, subsystemState := range c.cache.getStatuses() {
 		state := subsystemState.details
+		if state.Started == nil || !*state.Started {
+			isStarted = false
+		}
 		if !state.Live {
 			isLive = false
 		}
@@ -167,7 +181,7 @@ func (c *checker) checkStates() (bool, bool, any, any) {
 		readyDetails[subsystemName] = state.ReadyDetails
 	}
 
-	return isLive, isReady, liveDetails, readyDetails
+	return isStarted, isLive, isReady, liveDetails, readyDetails
 }
 
 func (c *checker) liveHandler(w http.ResponseWriter, _ *http.Request) {
--- a/pkg/common/util/task.go
+++ b/pkg/common/util/task.go
@@ -7,6 +7,68 @@ import (
 	"sync"
 )
 
+type TaskRunner struct {
+	wg        sync.WaitGroup
+	ctx       context.Context
+	cancels   []context.CancelFunc
+	errch     chan error
+	taskCount int
+}
+
+func NewTaskRunner(ctx context.Context) *TaskRunner {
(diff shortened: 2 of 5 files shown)


kfox1111:
Fix broken out of https://github.com/spiffe/spire/pull/5892

pkg/common/health/health.go
```diff
@@ -20,6 +20,10 @@ const (
 
 // State is the health state of a subsystem.
 type State struct {
+	// Started is whether the subsystem is finished starting.
+	// if undefined, it is treated as started=true.
```
amartinezfayo, line 24: Could you elaborate on why the undefined state is needed?

kfox1111: For backwards compatibility with existing code. For example,  the spire-server code in this pr still doesn't set whether it is started or not, so it implies started=true so the spire-server behaves as it always has.

Wasn't sure all the places it may be used. We could update them all if that was preferable.

pkg/common/util/task.go
```diff
@@ -7,46 +7,58 @@ import (
...
+	cancels   []context.CancelFunc
+	errch     chan error
+	taskCount int
+}
 
-	errch := make(chan error, len(tasks))
+func NewTaskRunner(ctx context.Context) *TaskRunner {
+	return &TaskRunner{
+		ctx:   ctx,
+		errch: make(chan error, 1),
```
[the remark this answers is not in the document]
kfox1111: I don't think so?

The tasks are entering the queue only when they are exiting. The wait at the end is the main task waiting for them all to exit, so its ok if they block before they exit? They can't block each other, other then to return their exit state I think, and thats not deadlockable?

The problem is difficult to solve otherwise, as the channel size depends on how many tasks there can be, but with the new interface, groups of tasks can be started at different times, so you cant just make chan with the size of the len(tasks) as that would have to happen beforehand. Still maybe possible, but feels like maybe more work then needed if the above is true?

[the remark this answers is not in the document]
amartinezfayo: I'm thinking that probably the main problem is not in the size of the buffer for errors, but in the logic of the Wait function. My main concern is that when there is an error, the function just returns and does not process other potential errors, so if there is an additional error in one of the other pending tasks, you will be waiting forever to be able to send that error to the channel (this seems to be a problem also in the current implementation, it's not introduced here). The same will happen if the context is done, because the function immediately returns.
A simple test leveraging the existing functions in task_test.go seems to confirm this:
```
func TestRunTaskMultipleFailures(t *testing.T) {
	in1, _, t1 := newFakeTask()
	in2, _, t2 := newFakeTask()

	testRunTasks(ctx, t1, t2)

	in1 <- errors.New("some error 1!")
	in2 <- errors.New("some error 2!")
}
```
I could make it block pretty easily. It would be great if we can address this. We may do that in a separate PR if needed.
Does this make sense?

pkg/common/util/task.go
```diff
@@ -7,46 +7,58 @@ import (
...
-func RunTasks(ctx context.Context, tasks ...func(context.Context) error) error {
-	var wg sync.WaitGroup
-	ctx, cancel := context.WithCancel(ctx)
-	defer func() {
-		cancel()
-		wg.Wait()
-	}()
+type TaskRunner struct {
+	wg        sync.WaitGroup
+	ctx       context.Context
```
amartinezfayo, line 12: Having a context in a struct is generally discouraged in Go, as it can lead to confusion about context lifetimes and responsibilities: https://go.dev/blog/context-and-structs

From what I can see, the context is stored in the TaskRunner and appears to serve as a global context for all operations. This approach can make it difficult understand when cancellation signals propagate.

Have you considered passing contexts explicitly to methods that need them?

pkg/common/util/task.go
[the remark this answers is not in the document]
kfox1111: Hmm.... I think I understand.

I don't think the problem is the putting in of results, but the taking of them out.

Something needs to be pulling results back out in a timely manner or else things will block.

"If a task fails half way through a long init, how quickly can we report that failure back?"

I can think of two options to solve it...

1. the brute force approach. have a go routine for the wait happening all the time, and if it sees an error, os.Exit the whole app. This is somewhat undesirable.....
2. There are two cases I think. the async case, and the sync case. We have the sync case handled. While things are started, but before there is nothing more to do in init, we need a way of processing and safely returning errors but not ok task exits... The code would need to check in each loop iteration during init at the current error states of tasks and if an error is seen, return from the init loop early. Also, if in that async handler, a failure is seen, it probably should cancel the context of the overall init processing if possible, to speed up exiting of whatever processing is happening, so it can return the error faster in that case

Actually, maybe the context thing could be a 3rd option? make an explicit cancelable context for the tasks back from util.NewTaskRunner(). start the wait function in a go routine. Cancel the context if an error is seen? use that as the async handler? Not sure that would work with all the other code during init though...

pkg/common/util/task.go
kfox1111: May need to discuss fixing the other issue first... I'm thinking we may want ctx here to be able to cancel things on failure...

That said, assuming we don't,
currently it treats all started tasks in the same context. they are all started/canceled in the same context. This makes all tasks run through the same TaskRunner the same.

If we dont do that, we would need to accept a context per set of tasks, and then keep track of all of those contexts in the struct so wait can cancel them all? I figured one context rather then multiple would be simpler/more understandable for this case then multiple.

amartinezfayo: That's fair. Moving the context out of the struct would likely introduce more complexity than it's worth.

amartinezfayo (approved):
Thank you @kfox1111!

PR #6074, opened 2025-05-16
Use Policy field instead of PolicyIdentifiers in cert templates

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
CredentialComposer plugins that use the PolicyIdentifiers field in the X509CAAttributes structure.

**Description of change**
Since go 1.24, to which we updated in the 1.12.0 release of SPIRE, the crypto/x509 package no longer applies the certificate policies from the PolicyIdentifier field when using the `CreateCertificate` method. Instead of this the Policies field is used.

The effect is that since the 1.12.0 release of SPIRE the CAs minted by deployments using a CredentialComposer plugin which specifies PolicyIdentifiers no longer have those fields populated in the CAs.

This can be mitigated on the affected releases (1.12.0 and 1.12.1) by setting the GODEBUG environment variable to ❯ `x509usepolicies=0`


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [1.12.1] - 2025-05-16
+
+### Fixed
+
+- PolicyCredentials set by CredentialComposer plugins are now applied again to CA certificates (#6074)
```
MarcosDY, line 7: how about something like:

```
- Regression where PolicyCredentials set by CredentialComposer plugins were not correctly applied to CA certificates. (#6074)
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [1.12.1] - 2025-05-16
```
MarcosDY, line 3:
```suggestion
## [1.12.2] - 2025-05-16
```

amartinezfayo (approved):
LGTM, thank you so much @sorindumitru!

amartinezfayo (approved):

Issue #6055, opened 2025-05-08
Gracefully stop spire-server


sorindumitru:
Currently the spire-server gRPC servers are stopped using [Stop()](https://github.com/spiffe/spire/blob/662da58da33d5befb2afc23eb994e12916b0d4eb/pkg/server/endpoints/endpoints.go#L297) which forcefully closes connections without waiting for the RPCs to finish.

We should change this to do a GracefulStop with some kind of timeout. The function doens't take in a timeout, so we might have to do it with a goroutine, channel and timer.

tjons:
@sorindumitru I'd be happy to take this one if you'll assign it to me. I'm trying to get my feet wet with contributing to SPIRE while working on the Cassandra datastore implementation I mentioned in #5995.

sorindumitru:
This is now done

PR #6076, opened 2025-05-18
implement gracefulstop with a configurable timeout in spire-server

tjons:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
`spire-server` shutdown now uses `GracefulStop()` with a default wait time of 30s, replacing the current `Stop()` which will terminate in-flight RPCs.

**Description of change**
1. Refactor `pkg/server/endpoints/config.go` to use grpc's `GracefulStop` shutdown functionality.
2. Add a configurable timeout, `graceful_stop_wait`, to the `spire-server` config to allow users to specify a configurable duration to wait before forcefully disconnecting clients.
3. Add test cases for `graceful_stop_wait` configuration parameter.
4. Update documentation with the new configuration parameter.

**Which issue this PR fixes**
https://github.com/spiffe/spire/issues/6055


pkg/server/endpoints/endpoints.go
```diff
@@ -294,7 +296,21 @@ func (e *Endpoints) runTCPServer(ctx context.Context, server *grpc.Server) error
...
+		go func() {
+			log.Info("Attempting graceful stop")
+			server.GracefulStop()
+			close(stopComplete)
+		}()
+
+		shutdownDeadline := time.After(e.GracefulStopWait)
+		select {
+		case <-shutdownDeadline:
+			log.Infof("Graceful stop unsuccessful, forced stop after %s", e.GracefulStopWait.String())
```
sorindumitru, line 310: I think we should also call `server.Stop()` in this case since the `server.GracefulStop()` might be still be doing some work.

pkg/server/endpoints/endpoints.go
```diff
@@ -338,7 +354,20 @@ func (e *Endpoints) runLocalAccess(ctx context.Context, server *grpc.Server) err
...
+		case <-stopComplete:
+			log.Info("Graceful stop successful")
+		}
```
sorindumitru, line 370: Can we have a helper function for this, since it's done in 2 different places?

pkg/server/endpoints/endpoints.go
tjons: ahh! I had done this and then removed it erroneously in a refactor :\

pkg/server/endpoints/endpoints.go
tjons: added... I had considered this originally but didn't like the idea of having to pass the parameterized logger around. what do you think of the implementation? an alternative would be to reconstruct a new logger in the helper but that would require passing the listener/information about the listener into the helper

pkg/server/endpoints/endpoints_test.go
```diff
@@ -240,7 +240,6 @@ func TestListenAndServe(t *testing.T) {
 	prepareDataStore(t, ds, []*testca.CA{ca, federatedCA}, agentSVID)
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
-	defer cancel()
```
sorindumitru, line 243: Was thi removed accidentally?

pkg/server/endpoints/endpoints.go
```diff
@@ -337,14 +338,36 @@ func (e *Endpoints) runLocalAccess(ctx context.Context, server *grpc.Server) err
...
+	go func() {
+		log.Info("Attempting graceful stop")
+		server.GracefulStop()
+		close(stopComplete)
+	}()
+
+	shutdownDeadline := time.After(defaultGracefulStopTimeout)
+	select {
+	case <-shutdownDeadline:
+		log.Infof("Graceful stop unsuccessful, forced stop after %s", defaultGracefulStopTimeout.String())
```
faisal-memon, line 362:
```suggestion
		log.Infof("Graceful stop unsuccessful, forced stop after %s", gracefulStopTimeout.String())
```

pkg/server/endpoints/endpoints_test.go
tjons: no, the context is explicitly cancelled on ~L391 so this appeared duplicative.

pkg/server/endpoints/endpoints.go
```diff
@@ -56,6 +56,10 @@ const (
 
 	// This is the default amount of time to wait for an event before giving up
 	defaultEventTimeout = 15 * time.Minute
+
+	// This is the time to wait for graceful termination of the gRPC server
+	// before forcefully terminating.
+	defaultGracefulStopTimeout = 10 * time.Second
```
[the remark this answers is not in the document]
tjons: good point, thanks @faisal-memon

faisal-memon (approved):

pkg/server/endpoints/endpoints_test.go
sorindumitru: I think it's still better to keep. This makes sure it gets canceled  if the test is terminated early due to an assertion.

It also matches how you'd normally use the context. A defer to cancel it when exiting the function + optionally manually calling cancel early when that makes sesne, like here to stop the server.

tjons: done!

tjons:
> LGTM, thanks @tjons for the contribution!

Thanks for the help!

PR #6090, opened 2025-05-28
Use UpstreamAuthority.SubscribeToLocalBundle RPC

### LINKED ISSUE #6083: Trust bundle missing active signing keys on some clusters
Hi, coming from this thread https://spiffe.slack.com/archives/C7XDP01HB/p1747864066874729 where @evan2645 and @sorindumitru asked me to raise an issue.

We have a root cluster and 8 regional clusters, all of them running on EC2 instances and AWS Postgres as datastore. A few days ago we terminated/created all the EC2 instances in a rolling fashion, one node at a time, as we’ve done before. DBs are always untouched. The /data directory is gone when servers startup as we don’t persist it, so servers can’t find keys and they create new ones.

We've noticed that some clusters are missing signing keys:

on root cluster:
$ spire-server bundle show -format spiffe | grep kid | wc -l
58

on 5 out of 8 regional clusters:
$ spire-server bundle show -format spiffe | grep kid | wc -l
58

and on the other 3:
$ spire-server bundle show -format spiffe | grep kid | wc -l
53

We are running 1.10.4

Thank you.

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
UpstreamAuthority plugins.

**Description of change**
If the UpstreamAuthority plugin implements the GetTrustBundle RPC use it to keep the trust bundle of the trust domain in sync. This allows us to get bundle updates before we create a new CA or key.

**Which issue this PR fixes**
fixes #6083

### DIFF
--- a/pkg/server/ca/rotator/rotator.go
+++ b/pkg/server/ca/rotator/rotator.go
@@ -37,6 +37,8 @@ type CAManager interface {
 	ActivateJWTKey(ctx context.Context)
 	RotateJWTKey(ctx context.Context)
 
+	SubscribeToLocalBundle(ctx context.Context) error
+
 	PruneBundle(ctx context.Context) error
 	PruneCAJournals(ctx context.Context) error
 }
@@ -77,9 +79,24 @@ func (r *Rotator) Run(ctx context.Context) error {
 	if err := r.c.Manager.NotifyBundleLoaded(ctx); err != nil {
 		return err
 	}
+
 	err := util.RunTasks(ctx,
 		func(ctx context.Context) error {
 			return r.rotateEvery(ctx, rotateInterval)
+		},
+		func(ctx context.Context) error {
+			var lastError error
+			for {
+				select {
+				case <-ctx.Done():
+					return lastError
+				case <-time.After(5 * time.Second):
+					lastError = r.c.Manager.SubscribeToLocalBundle(ctx)
+					if lastError == nil {
+						return nil
+					}
+				}
+			}
 		},
 		func(ctx context.Context) error {
 			return r.pruneBundleEvery(ctx, pruneBundleInterval)
--- a/pkg/server/ca/rotator/rotator.go
+++ b/pkg/server/ca/rotator/rotator.go
@@ -85,18 +85,7 @@ func (r *Rotator) Run(ctx context.Context) error {
 			return r.rotateEvery(ctx, rotateInterval)
 		},
 		func(ctx context.Context) error {
-			var lastError error
-			for {
-				select {
-				case <-ctx.Done():
-					return lastError
-				case <-time.After(5 * time.Second):
-					lastError = r.c.Manager.SubscribeToLocalBundle(ctx)
-					if lastError == nil {
-						return nil
-					}
-				}
-			}
+			return r.c.Manager.SubscribeToLocalBundle(ctx)
 		},
 		func(ctx context.Context) error {
 			return r.pruneBundleEvery(ctx, pruneBundleInterval)
--- a/pkg/server/ca/upstream_client_test.go
+++ b/pkg/server/ca/upstream_client_test.go
@@ -151,6 +151,43 @@ func TestUpstreamClientPublishJWTKey_NotImplemented(t *testing.T) {
 	require.Nil(t, jwtKeys)
 }
 
+func TestUpstreamClientSubscribeToLocalBundle(t *testing.T) {
+	client, updater, ua := setUpUpstreamClientTest(t, fakeupstreamauthority.Config{
+		TrustDomain:               trustDomain,
+		UseSubscribeToLocalBundle: true,
+	})
+
+	err := client.SubscribeToLocalBundle(t.Context())
+	require.NoError(t, err)
+
+	// We should get an update with the initial CA and a list of empty JWT keys since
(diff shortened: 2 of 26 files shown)


pkg/server/ca/rotator/rotator.go
```diff
@@ -77,10 +79,25 @@ func (r *Rotator) Run(ctx context.Context) error {
 	if err := r.c.Manager.NotifyBundleLoaded(ctx); err != nil {
 		return err
 	}
+
 	err := util.RunTasks(ctx,
 		func(ctx context.Context) error {
 			return r.rotateEvery(ctx, rotateInterval)
 		},
+		func(ctx context.Context) error {
+			var lastError error
```
MarcosDY: may we move this to Subscribe?
and only start when we really need to subscribe

pkg/server/ca/upstream_client_test.go
```diff
@@ -151,6 +151,43 @@ func TestUpstreamClientPublishJWTKey_NotImplemented(t *testing.T) {
 	require.Nil(t, jwtKeys)
 }
 
+func TestUpstreamClientSubscribeToLocalBundle(t *testing.T) {
+	client, updater, ua := setUpUpstreamClientTest(t, fakeupstreamauthority.Config{
```
MarcosDY:
```suggestion
	client, updater, ua := setupUpstreamClientTest(t, fakeupstreamauthority.Config{
```

pkg/server/ca/upstream_client_test.go
```diff
@@ -151,6 +151,43 @@ func TestUpstreamClientPublishJWTKey_NotImplemented(t *testing.T) {
...
+		UseSubscribeToLocalBundle: true,
+	})
+
+	err := client.SubscribeToLocalBundle(t.Context())
+	require.NoError(t, err)
+
+	// We should get an update with the initial CA and a list of empty JWT keys since
+	// the fakeupstreamauthority does not create one by default.
+	require.Equal(t, ua.X509Roots(), updater.WaitForAppendedX509Roots(t))
+	spiretest.RequireProtoListEqual(t, []*common.PublicKey{}, updater.WaitForAppendedJWTKeys(t))
```
MarcosDY: require.Empty?

pkg/server/plugin/upstreamauthority/upstreamauthority.go
```diff
@@ -32,6 +32,16 @@ type UpstreamAuthority interface {
 	// the upstream authority does not support streaming updates, the stream
 	// will return io.EOF when called.
 	PublishJWTKey(ctx context.Context, jwtKey *common.PublicKey) (jwtAuthorities []*common.PublicKey, stream UpstreamJWTAuthorityStream, err error)
+
+	// GetUpstreamAuthorities can be used to sync the local trust bundle with
```
MarcosDY:
```suggestion
	// SubscribeToLocalBundle can be used to sync the local trust bundle with
```

pkg/server/plugin/upstreamauthority/v1.go
```diff
@@ -208,3 +241,41 @@ func (s *v1UpstreamJWTAuthorityStream) RecvUpstreamJWTAuthorities() ([]*common.P
...
+		case errors.Is(err, io.EOF):
+			// This is expected if the plugin does not support streaming
+			// authority updates.
+			return nil, nil, err
+		case err != nil:
+			return nil, nil, s.v1.WrapErr(err)
+		}
+
+		x509Authorities, err := s.v1.parseX509Authorities(resp.UpstreamX509Roots)
+		if len(resp.UpstreamX509Roots) != 0 && err != nil {
```
MarcosDY: why to verify roots length? parse may return no error only when there was something to parse, right?

test/integration/suites/nested-rotation/00-setup
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+pwd
```
MarcosDY: can you remove this?

pkg/server/ca/rotator/rotator.go
sorindumitru: Yes, I think moving this inside of Subscribe makes it a bit clearer. Not sure what you mean about starting it when we really need to subscribe. We always need to subscribe at the start of spire-server.

test/integration/suites/nested-rotation/00-setup
sorindumitru: ups, yes.

pkg/server/plugin/upstreamauthority/v1.go
sorindumitru: `parseX509Authorities` checks to see that there's at least one authority there. I was thinking it may still be ok to receive the JWT ones in this case.

But I think you're right to say this is a bit weird and there should always be some x509 roots, so I removed these checks.

pkg/server/ca/rotator/rotator.go
MarcosDY: based in code it is only required when upstreamAuthority is set 

```
func (m *Manager) SubscribeToLocalBundle(ctx context.Context) error {
	if m.upstreamClient == nil {
		return nil
	}
```

PR #6091, opened 2025-05-28
Forward original grpc authority host to downstream plugins

nikotih:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

When Spire agent connects via grpc to Spire server it is using configuration e.g.:
server_address = "spire-server-xyz.spiffe.io"
server_port = "8081"

This information is naturally present in grpc context as ":authority" metadata in spire server core. Further, when spire server does node attestation it initiates downstream grpc calls to plugins. This overrides original ":authority" header with "localhost" since all plugins are next to the server.

For our tenancy model we use spire domains as selectors in node attestation plugins so we would like to "preserve" original authority context. 

See example usage in unit test

Feel free to suggest alternative places for this e.g. middleware or let me know if you think its not a good idea 

**Description of change**

Spire server forwards original ":authority" grpc context metadata to outgoing node attestation plugins under another context key

**Which issue this PR fixes**

https://github.com/spiffe/spire/issues/6101

### DIFF
--- a/pkg/server/plugin/nodeattestor/v1.go
+++ b/pkg/server/plugin/nodeattestor/v1.go
@@ -39,7 +39,10 @@ func (v1 *V1) Attest(ctx context.Context, payload []byte, challengeFn func(ctx c
 	defer cancel()
 
 	// forward original request host to downstream plugins
-	originalHost := getOriginalHost(ctx)
+	originalHost, err := getOriginalHost(ctx)
+	if err != nil {
+		v1.Log.WithError(err).Warn("Failed to extract ':authority' header from gRPC metadata")
+	}
 	ctx = metadata.AppendToOutgoingContext(ctx, XForwardedHostKey, originalHost)
 
 	stream, err := v1.NodeAttestorPluginClient.Attest(ctx)
@@ -116,16 +119,16 @@ func (v1 *V1) streamError(err error) error {
 	return v1.WrapErr(err)
 }
 
-func getOriginalHost(ctx context.Context) string {
+func getOriginalHost(ctx context.Context) (string, error) {
 	authority := metadata.ValueFromIncomingContext(ctx, ":authority")
 	if len(authority) == 0 {
-		return ""
+		return "", errors.New("empty :authority header")
 	}
 	// should be just one in a slice
 	// example value: spire-server-xyz.spiffe.io:8081
 	host, _, err := net.SplitHostPort(authority[0])
 	if err != nil {
-		return ""
+		return "", err
 	}
-	return host
+	return host, nil
 }
--- a/pkg/server/plugin/nodeattestor/base/base_test.go
+++ b/pkg/server/plugin/nodeattestor/base/base_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/spiffe/spire/test/spiretest"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 )
 
@@ -33,6 +34,10 @@ func TestBaseAssessTOFU(t *testing.T) {
 	const errorID = "spiffe://domain.test/spire/agent/error"
 
 	log, hook := test.NewNullLogger()
+	ctx := metadata.NewIncomingContext(
+		context.Background(),
+		metadata.New(map[string]string{":authority": "spire-server:8081"}),
+	)
 
 	agentStore := fakeagentstore.New()
 	agentStore.SetAgentInfo(&agentstorev1.AgentInfo{AgentId: attestedID})
@@ -49,7 +54,7 @@ func TestBaseAssessTOFU(t *testing.T) {
 
 	t.Run("with unattested agent", func(t *testing.T) {
 		hook.Reset()
-		result, err := na.Attest(context.Background(), []byte(unattestedID), failOnChallenge)
+		result, err := na.Attest(ctx, []byte(unattestedID), failOnChallenge)
 		require.NoError(t, err)
 		require.Equal(t, &nodeattestor.AttestResult{
 			AgentID: unattestedID,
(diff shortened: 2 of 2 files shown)


sorindumitru:
@nikotih Could you open an issue with some more details about what you're doing? For example what do you mean by "spire domains"? Is it the trust domain name(this is already available in the Configure plugin call) or something else?

Some more details on how this ends up being used would also be useful.

nikotih:
Hi @sorindumitru 

No its not trust domain. I've created https://github.com/spiffe/spire/issues/6101 to describe what we are doing. Let me know if you want me to clarify something more

nikotih:
i've renamed it to `x-forwarded-authority` and made a self contained unit test that is based on allowed authority in plugin (can be defined e.g. in plugin config)

I dont think you have intentionally wanted to hide original authority from plugins and this was rather a side effect due to usage of go-plugins

@rturner3 please let me know if this fits better and can be accepted. Thank you

nikotih:
Changed to `x-forwarded-host` which is standardized https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-Host

pkg/server/plugin/nodeattestor/v1.go
```diff
@@ -4,14 +4,20 @@ import (
...
 	nodeattestorv1 "github.com/spiffe/spire-plugin-sdk/proto/spire/plugin/server/nodeattestor/v1"
 	"github.com/spiffe/spire/pkg/common/plugin"
 	"github.com/spiffe/spire/proto/spire/common"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 )
 
+const (
+	XForwardedHostKey = "X-Forwarded-Host"
```
MarcosDY: Can we change this value to "X-Unsafe-Forwarded-Host" or something similar to be clear than it may not be secure to use? 
and can you add some comments to explain why it is not safe to use for attestation, 
So users are not using it for that end.

nikotih: Hi @MarcosDY 

yes can change the name, just let me know which one you prefer. Naming and caching are too hardest CS problems

few additional options (with AI assist):

- `X-Unsafe-Forwarded-Host` - Suggested by you
- `X-Untrusted-Forwarded-Host` - Alternative wording (picked this one)
- `X-Client-Reported-Host` - Clearly indicates client-controlled, not authoritative (AI option)
- `X-Client-Provided-Host` - Alternative wording
- `X-Client-Authority` - If going back to HTTP/2 terms (AI option)
- `X-Untrusted-Authority` - Combination of above

internally we use `x-spire-server-by-agent` (not great i know)

let me know if you want me to change var name as well

sorindumitru: `X-Unsafe-Forwarded-Host` or `X-Untrusted-Forwarded-Host` both sound ok to us.

pkg/server/plugin/nodeattestor/v1.go
```diff
@@ -28,6 +38,10 @@ func (v1 *V1) Attest(ctx context.Context, payload []byte, challengeFn func(ctx c
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	// forward original request host to downstream plugins
+	originalHost := getOriginalHost(ctx)
+	ctx = metadata.AppendToOutgoingContext(ctx, XForwardedHostKey, originalHost)
```
sorindumitru: We get no feedback here if we're unable to parse original host from the metadata. Could we log something if we fail to parse the metadata, with any any err that results from that?

nikotih: In theory we should never see error, but maybe grpc lib will change something and it breaks

extra changes:
- getOriginalHost now returns errors
- errors are logged with logger from context
- due to above then had to modify all plugin unit tests to define logger in context before passing to Attest (no logger in context is panicking https://github.com/spiffe/spire/blob/main/pkg/common/api/rpccontext/logger.go#L20
- tests that were already using logger from context needed authority metadata header to not log errors related to it (some tests assert specific log lines)

Sample
```
spire-server-1  | time="2025-09-24T10:58:22Z" level=warning msg="Failed to extract ':authority' header from gRPC metadata" authorized_as=nobody authorized_via= caller_addr="172.19.0.3:43520" error="empty :authority header" method=AttestAgent request_id=5b248547-7eab-4ebf-b505-e5b3776c5404 service=agent.v1.Agent subsystem_name=api
```

Let me know if this fits, if you want to improve error message further or if you things its too invasive now

pkg/server/plugin/nodeattestor/v1.go
```diff
@@ -28,6 +39,14 @@ func (v1 *V1) Attest(ctx context.Context, payload []byte, challengeFn func(ctx c
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	// forward original request host to downstream plugins
+	log := rpccontext.Logger(ctx)
+	originalHost, err := getOriginalHost(ctx)
+	if err != nil {
+		log.WithError(err).Warn("Failed to extract ':authority' header from gRPC metadata")
```
sorindumitru: Looks like the `plugin.Facade` already has a log, so this should work:

```suggestion
		v1.Log.WithError(err).Warn("Failed to extract ':authority' header from gRPC metadata")
```

That might also remove the need for some of the changes in the tests.

nikotih: my bad, did not spot it. Thank you!

adjusted code. base plugin test still needed adjustment

```
spire-server-1  | time="2025-09-25T14:40:46Z" level=warning msg="Failed to extract ':authority' header from gRPC metadata" error="empty :authority header" external=false plugin_name=x509pop plugin_type=NodeAttestor subsystem_name=catalog
```

Issue #6024, opened 2025-04-21
k8s notifier / bundle publisher support for 'format' option


kfox1111:
Unlike most of the other notifiers/bundle publishers, it can not currently upload in spiffe format. Just pem.

Maybe the notifier needs to be ported to the new bundle publisher interface first.

kfox1111:
https://kubernetes.io/blog/2025/04/23/kubernetes-v1-33-release/#clustertrustbundles might be interesting to add to the bundle publisher too...

amartinezfayo:
Thank you @kfox1111 for opening this issue.
Since the plan is to deprecate the Notifier interface in favor of the BundlePublisher interface, we will be adding support to a `k8s_configmap` BundlePublisher plugin, that will support the SPIFFE format.

>https://kubernetes.io/blog/2025/04/23/kubernetes-v1-33-release/#clustertrustbundles might be interesting to add to the bundle publisher too...

Yes, looks like a good fit for the BundlePublisher. We have #4034 to track proposed BundlePublisher destinations and formats. You may comment on that issue so we keep it in the radar.

Issue #5872, opened 2025-02-14
spire-server bundle notifier with multiple clusters and failure


kfox1111:
The Kubernetes bundle notifier will fail the spire-server when there are multiple clusters defined and any one of the Kubernetes clusters go unresponsive.

We may need to make this behavior configurable, or we might just need to make it more robust. Say, try our best to stay running, run a thread per k8s cluster and when the cluster becomes available again, re-upload the bundle then.

Related to https://github.com/spiffe/helm-charts-hardened/issues/528

rturner3:
Thanks for raising this, @kfox1111. I agree we should add some more resiliency to the notifier integration with Kubernetes.

amartinezfayo:
Since the Notifier plugin is being deprecated, this functionality has been addressed in the `k8s_configmap` BundlePublisher plugin (#6105).

Issue #4976, opened 2024-03-14
K8s bundle creation


kfox1111:
Currently the k8s bundler assumes a configmap already exists. This means things like helm has to precreate it, but then spire manages it. Having multiple owners of an object in k8s can lead to some problems with gitops / race conditions.

Instead, the k8s bundler should try and create the configmap if it fails to modify an existing one.

MarcosDY:
This is an interesting topic, so, how is the workflow you suggest here?
Try to get CM:
- if exists udpate
- if no exists create and insert bundle

What happens if server stops? are we going to remove CM? (I think response here is no, because another services that requires this CM are no longer able to get CM)

When creating services, right now, since CM already exists (no matter if it is empty or no)  services can mount this CM and read, but now with current approach, users will need to wait until spire-server is up  (and create CM) to populate services, is that ok?

kfox1111:
Yeah. Algorithm looks about like I was thinking.

+1 to leaving the configmap in place. other things will need it. It currently doesn't remove the CM on exit so keeping that behavior is probably best.

Kubernetes pods depending on the configmap will just pause the pod in creating state until the configmap exists if they are in volumes. So k8s natively will take care of it. In the chart we have:
```
volumes:
  - configMap:
      defaultMode: 420
      name: spire-bundle
    name: spire-bundle
```

kfox1111:
stale but still useful

amartinezfayo:
Since the Notifier plugin is being deprecated, this functionality has been addressed in the `k8s_configmap` BundlePublisher plugin (#6105).

PR #6105, opened 2025-06-03
Add `k8s_configmap` BundlePublisher plugin

amartinezfayo:
This PR introduces the `k8s_configmap` BundlePublisher plugin, which adds the ability to push the server's current trust bundle to a designated Kubernetes ConfigMap. It supports multiple cluster configurations.

Fixes: https://github.com/spiffe/spire/issues/6024


pkg/server/plugin/bundlepublisher/k8sconfigmap/k8sconfigmap.go
```diff
@@ -0,0 +1,271 @@
...
+
+	for _, cluster := range config.Clusters {
+		bundleBytes, err := formatter.Format(cluster.bundleFormat)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "could not format bundle: %v", err.Error())
+		}
+
+		cm, err := cluster.k8sClient.GetConfigMap(ctx, cluster.Namespace, cluster.ConfigMapName)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to get ConfigMap: %v", err)
```
sorindumitru: Two plugin behaviour questions here:
* Should we continue instead of returning? An error with one of the clusters shouldn't affect any other clusters. See https://github.com/spiffe/spire/issues/5872 for a report of this being an issue.
* Should we create the ConfigMap if it doesn't exist? See https://github.com/spiffe/spire/issues/4976 for a request for that.

amartinezfayo: >Should we continue instead of returning? An error with one of the clusters shouldn't affect any other clusters. See https://github.com/spiffe/spire/issues/5872 for a report of this being an issue.

I think that makes sense. I've addressed that.

>Should we create the ConfigMap if it doesn't exist? See https://github.com/spiffe/spire/issues/4976 for a request for that.

The posture that we have adopted in BundlePublisher plugins is that the lifecycle of the resources is managed by the users (e.g. aws_s3 and gcp_cloudstorage plugins assume that the bucket has been created). I think that has the benefit of ensuring that the user has a conscious decision about the resources that are being created and the access controls around them.
In the case of the k8s_configmap plugin, I still think that the same reasoning applies, but I also recognize that it may be a different scenario. ConfigMaps don't have the same access control complexity as S3 buckets or Cloud Storage. They inherit namespace-level RBAC, which is typically already well-defined. Also, I believe that the k8s ecosystem often favors declarative, self-healing approaches where controllers create resources as needed and  users might expect this behavior.
Thoughts?

kfox1111: I'd love to get the configmap object out of the chart. k8s really doesnt do well when there are two owners of an object, and having helm be one owner and the publisher be another has caused issues.

kfox1111:
Tried it with the chart. got:
time="2025-06-08T15:44:59Z" level=error msg="Fatal run error" error="failed to configure plugin \"k8s_configmap\": rpc error: code = InvalidArgument desc = configuration must have at least one cluster"
time="2025-06-08T15:44:59Z" level=error msg="Server crashed" error="failed to configure plugin \"k8s_configmap\": rpc error: code = InvalidArgument desc = configuration must have at least one cluster"

doc/plugin_server_bundlepublisher_k8s_configmap.md
```diff
@@ -0,0 +1,57 @@
...
+| configmap_key   | The key within the ConfigMap for the bundle.                                                                                                                                                           | Yes.                                        |         |
+| namespace       | The namespace containing the ConfigMap.                                                                                                                                                                | Yes.                                        |         |
+| kubeconfig_path | The path on disk to the kubeconfig containing configuration to enable interaction with the Kubernetes API server. If unset, in-cluster credentials will be used. Only one in-cluster can be configured.| Required when configuring a remote cluster. |         |
+| format          | Format in which the trust bundle is stored, &lt;spiffe &vert; jwks &vert; pem&gt;. See [Supported bundle formats](#supported-bundle-formats) for more details.                                         | Yes.                                        |         |
```
sorindumitru: These should be nested in the `clusters` structure. Maybe we should specify it similarly to [k8s_psat](https://github.com/spiffe/spire/blob/main/doc/plugin_server_nodeattestor_k8s_psat.md)?

sorindumitru:
> Tried it with the chart. got: time="2025-06-08T15:44:59Z" level=error msg="Fatal run error" error="failed to configure plugin "k8s_configmap": rpc error: code = InvalidArgument desc = configuration must have at least one cluster" time="2025-06-08T15:44:59Z" level=error msg="Server crashed" error="failed to configure plugin "k8s_configmap": rpc error: code = InvalidArgument desc = configuration must have at least one cluster"

The configuration is slightly different from the old notifier plugin. You need to specify clusters in the `clusters` field, including the cluster running spire-server. It's a change, but the configuration ends up being more consistent.

pkg/server/plugin/bundlepublisher/k8sconfigmap/client.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil, err
+	}
+	return configMap, nil
+}
+
+// CreateConfigMap creates a new ConfigMap.
+func (c *k8sClient) CreateConfigMap(ctx context.Context, configMap *corev1.ConfigMap) error {
+	_, err := c.clientset.CoreV1().ConfigMaps(configMap.Namespace).Create(ctx, configMap, metav1.CreateOptions{})
+	return err
+}
```
sorindumitru, line 50: This doesn't seem to be used at the moment.

pkg/server/plugin/bundlepublisher/k8sconfigmap/k8sconfigmap.go
```diff
@@ -0,0 +1,281 @@
...
+	newConfig := new(Config)
+
+	if err := hcl.Decode(newConfig, hclText); err != nil {
+		status.ReportErrorf("unable to decode configuration: %v", err)
+		return nil
+	}
+
+	if len(newConfig.Clusters) == 0 {
+		status.ReportError("configuration must have at least one cluster")
+	}
```
sorindumitru: I think it's useful to allow empty Clusters just in case. We could log an error, so the user notices it, but still continue with normal operations.

This is helpful in the case where you want to register clusters dynamically and may start with an empty list (you put the plugin configuration in `plugin_data_file` and use a signal to trigger a reload when you change it)

pkg/server/plugin/bundlepublisher/k8sconfigmap/k8sconfigmap.go
```diff
@@ -0,0 +1,281 @@
...
+		case bundleformat.PEM:
+		default:
+			status.ReportErrorf("bundle format %q is not supported", cluster.Format)
+			return nil
+		}
+		cluster.bundleFormat = bundleFormat
+
+		if cluster.KubeConfigPath == "" {
+			if inClusterUsed {
+				status.ReportError("only one cluster can use in-cluster configuratio (empty kubeconfig_path)")
```
kfox1111:
```suggestion
				status.ReportError("only one cluster can use in-cluster configuration (empty kubeconfig_path)")
```

kfox1111:
During testing, I found a case where one service was expecting the bundle in `spiffe` format, and another one where it was expecting it in `pem` format, and both make sense...

So, maybe loosening the check that says you cant have more then 1 configuration without kubeconfig would be good.

amartinezfayo:
Thank you @sorindumitru and @kfox1111 for the feedback.
I made the changes to:
- Have the configuration as a map of clusters, keyed by an arbitrary ID 
- Allow to have no clusters configured
- Create the ConfigMap if it doesn't exist
- Allow to have more than one in-cluster configuration (this allows to publish multiple formats to an in-cluster configured, in the same way that publishing multiple formats to a cluster configured through kubeconfig_path is allowed)

conf/server/server_full.conf
```diff
@@ -941,6 +941,29 @@ plugins {
     #         # trust_anchor_id = "153d3e58-cab5-4a59-a0a1-3febad2937c4"
     #     }
     # }
+
+    # BundlePublisher "k8s_configmap": A bundle publisher that puts the current trust
+    # bundle of the server in a designated Kubernetes ConfigMap, keeping it updated.
+    # BundlePublisher "k8s_configmap" {
+    #     plugin_data {
+    #         clusters = [
```
kfox1111: This needs updating I think.

doc/plugin_server_bundlepublisher_k8s_configmap.md
```diff
@@ -38,6 +38,53 @@ The trust bundle is encoded as an RFC 7517 compliant JWK Set, omitting SPIFFE-sp
...
+
+```yaml
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: spire-server-cluster-role
+rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["create", "get", "patch"] # create is not needed if the ConfigMap already exists.
```
kfox1111: I had to add update as well.

amartinezfayo: I think it should be "update" instead of "patch".

kfox1111 (approved):
LGTM. Thanks! :)

sorindumitru (approved):
LGTM. One optional thing we could do is to switch one or all k8s based integration tests to use the new bundlepublisher, but also fine to leave that for when we deprecate/remove the notifier plugin.

amartinezfayo:
>LGTM. One optional thing we could do is to switch one or all k8s based integration tests to use the new bundlepublisher, but also fine to leave that for when we deprecate/remove the notifier plugin.

That's a great suggestion. I've opened https://github.com/spiffe/spire/issues/6127 to track this.

PR #6126, opened 2025-06-11
fix: subject_types_supported cannot be an empty list

### LINKED ISSUE #6125: OIDC provider returning illegal subject_types_supported
The OIDC provider currently returns `subject_types_supported: []` in the discovery document. This causes [this library](https://github.com/hidglobal/oauth-2.0-sdk-with-openid-connect-extensions) to fail with the exception below

```
Caused by: java.lang.IllegalArgumentException: At least one supported subject type must be specified
	at com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata.ensureAtLeastOneSubjectType(OIDCProviderMetadata.java:421)
	at com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata.<init>(OIDCProviderMetadata.java:358)
	at com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata.parse(OIDCProviderMetadata.java:1362)
	at com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata.resolve(OIDCProviderMetadata.java:1863)
	at com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata.resolve(OIDCProviderMetadata.java:1820)
	at com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata.resolve(OIDCProviderMetadata.java:1783)
	at com.indeed.trino.header.auth.TrinoJWTProcessor.getJwtProcessor(TrinoJWTProcessor.java:40)
```

The [spec](https://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes) says

> A Subject Identifier is a locally unique and never reassigned identifier within the Issuer for the End-User, which is intended to be consumed by the Client. Two Subject Identifier types are defined by this specification:
> 
> public
> This provides the same sub (subject) value to all Clients. It is the default if the provider has no subject_types_supported element in its discovery document.
> pairwise
> This provides a different sub value to each Client, so as not to enable Clients to correlate the End-User's activities without permission.
> 
> The OpenID Provider's Discovery document MUST list its supported Subject Identifier types in the subject_types_supported element. If there is more than one type listed in the array, the Client MAY elect to provide its preferred identifier type using the subject_type parameter during Registration.

I think the spec is saying if you have a `subject_types_supported` element, the value cannot be empty, but the element itself is not required. However, if that's true, not all libraries are implemented with that interpretation. For example, this library which failed has an implementation of
https://github.com/hidglobal/oauth-2.0-sdk-with-openid-connect-extensions/blob/0a7fa5785ce8ac17ac95ce5294a24c8743794a6f/src/main/java/com/nimbusds/openid/connect/sdk/op/OIDCProviderMetadata.java#L1731-L1735
```java
		List<SubjectType> subjectTypes = new ArrayList<>();
		
		for (String v: JSONObjectUtils.getStringArray(jsonObject, "subject_types_supported")) {
			subjectTypes.add(SubjectType.parse(v));
		}
```
https://github.com/hidglobal/oauth-2.0-sdk-with-openid-connect-extensions/blob/0a7fa5785ce8ac17ac95ce5294a24c8743794a6f/src/main/java/com/nimbusds/openid/connect/sdk/op/OIDCProviderMetadata.java#L363-L364
```java
		if (subjectTypes.size() < 1)
			throw new IllegalArgumentException("At least one supported subject type must be specified");
```
Which would require this element. Therefore, my recommended fix would be that we populate the list value with `"public"`.

keeganwitt:
Closes #6125

**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
OIDC provider discovery document.

**Description of change**
Populates the `subject_types_supported` value in the OIDC provider discovery document.

**Which issue this PR fixes**
#6125

### DIFF
--- a/support/oidc-discovery-provider/handler.go
+++ b/support/oidc-discovery-provider/handler.go
@@ -135,7 +135,7 @@ func (h *Handler) serveWellKnown(w http.ResponseWriter, r *http.Request) {
 
 		AuthorizationEndpoint:            "",
 		ResponseTypesSupported:           []string{"id_token"},
-		SubjectTypesSupported:            []string{},
+		SubjectTypesSupported:            []string{"public"},
 		IDTokenSigningAlgValuesSupported: []string{"RS256", "ES256", "ES384"},
 	}
 
--- a/support/oidc-discovery-provider/handler_test.go
+++ b/support/oidc-discovery-provider/handler_test.go
@@ -40,7 +40,9 @@ func TestHandlerHTTPS(t *testing.T) {
   "response_types_supported": [
     "id_token"
   ],
-  "subject_types_supported": [],
+  "subject_types_supported": [
+    "public"
+  ],
   "id_token_signing_alg_values_supported": [
     "RS256",
     "ES256",
@@ -208,7 +210,9 @@ func TestHandlerHTTPInsecure(t *testing.T) {
   "response_types_supported": [
     "id_token"
   ],
-  "subject_types_supported": [],
+  "subject_types_supported": [
+    "public"
+  ],
   "id_token_signing_alg_values_supported": [
     "RS256",
     "ES256",
@@ -321,7 +325,9 @@ func TestHandlerHTTP(t *testing.T) {
   "response_types_supported": [
     "id_token"
   ],
-  "subject_types_supported": [],
+  "subject_types_supported": [
+    "public"
+  ],
   "id_token_signing_alg_values_supported": [
     "RS256",
     "ES256",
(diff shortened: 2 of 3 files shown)


keeganwitt:
Someone can correct me if I'm mistaken, but I don't think the CI failures are related to my changes.

amartinezfayo (approved):
>Someone can correct me if I'm mistaken, but I don't think the CI failures are related to my changes.

They are not related.
Thank you @keeganwitt for this contribution!

PR #6133, opened 2025-06-15
Use sha of branch for caching windows executables

### LINKED ISSUE #6044: Windows integration test broken
I submitted a change that is not being reflected in the tests. It looks like the integration test is using a previous build of the pr for testing newer versions of it. There is no way the newer version of the pr can generate the string that it is outputting. Seems like its maybe caching builds incorrectly.

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Windows CI runs.

**Description of change**
Using the executables does not work because they do not exist when we try to fetch from the cache. This means we may end up fetching executables from previous runs of the workflows on the branch.

**Which issue this PR fixes**
fixes #6044

### DIFF
--- a/.github/workflows/pr_build.yaml
+++ b/.github/workflows/pr_build.yaml
@@ -207,7 +207,7 @@ jobs:
         uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
         with:
           path: ./bin/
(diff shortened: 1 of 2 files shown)


amartinezfayo (approved):
Thank you @sorindumitru for looking into this!

PR #6151, opened 2025-06-24
Full events cache refresh

### LINKED ISSUE #4973: Entry Update and Event Creation Can Possibly Drift. Consider periodic full-db reconciliation of cache
In the event that there occurs drift between in-mem event-based cache and entries/agents tables, we should consider a configurable periodic method that does a full reconciliation of the in-mem cache with the db.

faisal-memon:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Events based cache

**Description of change**

Sets up a full reload of the cache every 6 hours as a safety net to catch any potential drift. Can change frequency through `full_cache_reload_interval`

**Which issue this PR fixes**

fixes #4973 
partially addresses #5624


pkg/server/endpoints/authorized_entryfetcher.go
```diff
@@ -35,94 +40,109 @@ type eventsBasedCache interface {
...
 
-	attestedNodes, err := buildAttestedNodesCache(ctx, log, metrics, ds, clk, cache, cacheReloadInterval, eventTimeout)
+	attestedNodes, err := buildAttestedNodesCache(ctx, a.c.log, a.c.metrics, a.c.ds, a.c.clk, cache, a.c.cacheReloadInterval, a.c.eventTimeout)
 	if err != nil {
-		return nil, nil, nil, err
+		return err
 	}
 
-	return cache, registrationEntries, attestedNodes, nil
+	a.cache = cache
```
sorindumitru, line 153: I think we'd need to have some kind of mutex around setting these otherwise it may race with other goroutines accessing the cache.

sorindumitru: Looks like it's visible in the tests: https://github.com/spiffe/spire/actions/runs/15843210407/job/44659666995?pr=6151

faisal-memon: thanks, good catch.

faisal-memon: Added the locks, also tuned back the frequency to 24 hours which i think is sufficient.

doc/spire_server.md
```diff
@@ -89,15 +89,16 @@ This may be useful for templating configuration files, for example across differ
...
+| experimental                 | Description                                                                                                                                                                                                            | Default                            |
+|:-----------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------|
+| `cache_reload_interval`      | The amount of time between two reloads of the in-memory entry cache. Increasing this will mitigate high database load for extra large deployments, but will also slow propagation of new or updated entries to agents. | 5s                                 |
+| `full_cache_reload_interval` | How often to a full reload of the cache from the database when using the events based cache.                                                                                                                           | 6h                                 |
```
sorindumitru:
```suggestion
| `full_cache_reload_interval` | How often to a full reload of the cache from the database when using the events based cache.                                                                                                                           | 24h                                |
```

pkg/server/endpoints/authorized_entryfetcher.go
```diff
@@ -14,115 +15,153 @@ import (
...
 	return errors.Join(updateRegistrationEntriesCacheErr, updateAttestedNodesCacheErr)
 }
 
-func buildCache(ctx context.Context, log logrus.FieldLogger, metrics telemetry.Metrics, ds datastore.DataStore, clk clock.Clock, cacheReloadInterval, eventTimeout time.Duration) (*authorizedentries.Cache, *registrationEntries, *attestedNodes, error) {
-	cache := authorizedentries.NewCache(clk)
+func (a *AuthorizedEntryFetcherEvents) buildCache(ctx context.Context) error {
+	cache := authorizedentries.NewCache(a.c.clk)
 
-	registrationEntries, err := buildRegistrationEntriesCache(ctx, log, metrics, ds, clk, cache, pageSize, cacheReloadInterval, eventTimeout)
+	registrationEntries, err := buildRegistrationEntriesCache(ctx, a.c.log, a.c.metrics, a.c.ds, a.c.clk, cache, pageSize, a.c.cacheReloadInterval, a.c.eventTimeout)
```
sorindumitru, line 142: Since `buildCache` is now a member of  `AuthorizedEntryFetcherEvents`, maybe we can also make `buildRegistrationEntriesCache` a member to avoid passing in so many parameters to it? Similarly for `buildAttestedNodesCache`.

pkg/server/endpoints/authorized_entryfetcher_test.go
```diff
@@ -205,6 +229,56 @@ func TestBuildCacheSavesSkippedEvents(t *testing.T) {
...
+	// First iteration, cache should not be rebuilt
+	clk.Add(2 * time.Second)
+	ef.mu.RLock()
+	require.Equal(t, initialCache, ef.cache)
+	ef.mu.RUnlock()
+
+	// Second iteration, cache should be rebuilt
+	clk.Add(2 * time.Second)
+	ef.mu.RLock()
+	require.NotEqual(t, initialCache, ef.cache)
```
sorindumitru, line 274: This is maybe enough but I wonder if it's possible to test the case where the full rebuild actually helps? For example creating a registration entry and deleting the event (or however these unit tests manage to make skipped events) and then waiting for the full reload to read that registration entry.

pkg/server/endpoints/authorized_entryfetcher.go
```diff
@@ -14,115 +15,153 @@ import (
...
+		return err
 	}
 
-	return cache, registrationEntries, attestedNodes, nil
+	a.mu.Lock()
+	a.cache = cache
+	a.mu.Unlock()
+
+	a.registrationEntries = registrationEntries
+	a.attestedNodes = attestedNodes
```
sorindumitru, line 157: I think these also need to be under the lock, they are accessed from the `updateCache` function.

pkg/server/endpoints/authorized_entryfetcher_test.go
faisal-memon: That's a good idea.

pkg/server/endpoints/authorized_entryfetcher.go
faisal-memon: Yes but that is single threaded.

pkg/server/endpoints/authorized_entryfetcher.go
faisal-memon: Yes. Plan to do a cleanup pr after and can address that issue then.

pkg/server/endpoints/authorized_entryfetcher.go
faisal-memon: access to these vars is single threaded.

pkg/server/endpoints/authorized_entryfetcher_test.go
faisal-memon: added

sorindumitru (approved):

PR #6164, opened 2025-07-02
Fix metrics/health checks when retry_bootstrap=true

kfox1111:
Re-submission of 6063 with fix.

**Affected functionality**
When retry_bootstrap is true, during the startup/retrying, health checks / metrics are not started. This time is currently is capped at 5 minutes, requiring fairly long health check delays on Kubernetes, and no ability to collect metrics on what its doing.

**Which issue this PR fixes**
fixes: https://github.com/spiffe/spire/issues/6051


pkg/common/health/cache.go
```diff
@@ -104,12 +104,14 @@ func (c *cache) start(ctx context.Context) error {
 		return errors.New("no health checks defined")
 	}
 
+	time.Sleep(8 * time.Second)
```
amartinezfayo: I think that we should avoid calling time.Sleep.
We should probably have a context-based delay with cancellation support, and leverage the clk clock from the cache struct. One consequence of having this sleep is that tests also execute and are delayed while running.

kfox1111: the sleep is still needed for the case when bootstrapping takes a while and we dont want it to show an alert too quickly.

I can add channel/function to call once startup has been completed successfully that can cancel the sleep?

pkg/common/health/cache.go
```diff
@@ -96,7 +100,20 @@ func (c *cache) getStatuses() map[string]checkState {
...
+func (c *cache) StartupComplete() {
+	c.startupComplete <- struct{}{}
+}
+
 func (c *cache) start(ctx context.Context) error {
+	if c.delayedStart {
+		select {
+		case <-c.startupComplete:
+			// Even when notified, it may take a tiny bit for the service to become responsive.
+			time.Sleep(200 * time.Millisecond)
```
amartinezfayo: This approach seems a bit arbitrary and might be prone to race conditions depending on the environment. Could we instead signal through a channel when it's safe to assume the services are up?

For example, in the agent, this might correspond to when the endpoints are actively serving. We already have a signaling mechanism used for testing (see: [endpoints.go#L128](https://github.com/spiffe/spire/blob/main/pkg/agent/endpoints/endpoints.go#L128)). Perhaps we could expose something like that.

An approach based on explicit signaling might allow us to eliminate the delayedStart attribute and avoid the fixed 8-second wait altogether.

amartinezfayo: After discussing this with @kfox1111, I explored an alternative solution to avoid using the delayedStart flag and the time.Sleep call. The new approach uses a signaling mechanism through channels as a deterministic way to know when the services are listening and responsive.

I’ve added a commit implementing this change (82027be).

pkg/agent/agent.go
```diff
@@ -239,19 +252,17 @@ func (a *Agent) Run(ctx context.Context) error {
...
-
-	if err := healthChecker.AddCheck("agent", a); err != nil {
-		return fmt.Errorf("failed adding healthcheck: %w", err)
-	}
+	agentEndpoints := a.newEndpoints(metrics, manager, workloadAttestor)
+	go func() {
+		agentEndpoints.WaitForListening(readyForHealthChecks)
+	}()
 
+	a.started = true
```
sorindumitru: Could we do this in the go routine, after we've finished waiting for the endpoints to start listening? I think that's kind of the intention with this flag.

kfox1111: the one up a couple lines? Yeah, I think that makes sense.

kfox1111: Done

pkg/server/endpoints/endpoints.go
```diff
@@ -421,3 +435,13 @@ func (e *Endpoints) makeInterceptors() (grpc.UnaryServerInterceptor, grpc.Stream
...
+
+func (e *Endpoints) triggerListeningHook() {
+	if e.hooks.listening != nil {
+		e.hooks.listening <- struct{}{}
+	}
+}
+
+func (e *Endpoints) WaitForListening() {
+	<-e.hooks.listening
+}
```
sorindumitru: Should this also check for the hook being non-nil?

pkg/agent/endpoints/endpoints.go
```diff
@@ -147,3 +152,8 @@ func (e *Endpoints) triggerListeningHook() {
 		e.hooks.listening <- struct{}{}
 	}
 }
+
+func (e *Endpoints) WaitForListening(listening chan struct{}) {
+	<-e.hooks.listening
+	listening <- struct{}{}
+}
```
sorindumitru: Similarly to the server side, should this check for non-nil of `e.hooks.listening`? I don't think it's worth checking fro `listening` to be non-nil.

pkg/common/health/health.go
```diff
@@ -134,27 +138,37 @@ func (c *checker) ListenAndServe(ctx context.Context) error {
...
-	isLive, isReady := true, true
+func (c *checker) checkStates() (bool, bool, bool, any, any) {
+	isStarted, isLive, isReady := true, true, true
 
 	liveDetails := make(map[string]any)
 	readyDetails := make(map[string]any)
 	for subsystemName, subsystemState := range c.cache.getStatuses() {
 		state := subsystemState.details
+		if state.Started == nil || !*state.Started {
+			isStarted = false
```
sorindumitru: Should we default to `isStarted = true` for the case where `state.Started == nil`? At least that's what the comment for the field says.

pkg/common/health/health.go
```diff
@@ -134,27 +138,37 @@ func (c *checker) ListenAndServe(ctx context.Context) error {
...
+func (c *checker) checkStates() (bool, bool, bool, any, any) {
+	isStarted, isLive, isReady := true, true, true
 
 	liveDetails := make(map[string]any)
 	readyDetails := make(map[string]any)
 	for subsystemName, subsystemState := range c.cache.getStatuses() {
 		state := subsystemState.details
+		if state.Started == nil || !*state.Started {
+			isStarted = true
+		}
```
sorindumitru: Maybe something like this?
```suggestion
		if state.Started != nil  {
			isStarted = *state.Started
		}
```

I think we want it to stay true if it's `nil` and otherwise to take the value of the flag.

sorindumitru (approved):
LGTM, thanks everyone!

PR #6176, opened 2025-07-09
Reduce node fetch dependancy for RPCs

### LINKED ISSUE #5954: Reduce node fetch dependancy for RPCs
Currently for every RPC the agent calls, the server verifies that it is actually an agent by fetching the attested node information (a "node fetch" operation). The reason for this is so we can verify that the agent is in possession of either the current or the next (immediately after re-attestation or renewal) SVID and verify that the agent wasn't banned or evicted. We also need to have up to date information about the agent, so we fetch this from the rw connection, so from the primary node. This makes every RPC called by the agent dependant on the primary node of the database being available.

For most operations that the agent does, this is maybe not that big of an issue. It can delay synchronization of authorized entries and renewal of X509-SVIDs, but as long as the datastore recovers it will also automatically recover.

For JWT-SVID issuing this can be a problem. SVIDs are not preloaded so the first time one is requested, it will be fetched from the server. If the primary node changes, for example due to database maintenance, this can lead to visible spikes in failures to fetch JWT-SVIDs. It would be good if we could go over these periods of short unavailability without any visible failures.

I propose we change the current algorithm, of fetching the latest node information from the rw connction, to the following:
* Fetch node information from some potentially stale data and verify the agent against that. If the caller is attested as an agent based on that information, let the request through.
* Otherwise, fetch node information from the rw connection and verify it against that.

This has the downside the agent is potentially trusted for a bit longer than it should be, for example eviction operations don't have immediate effect. As long as the period of time we trust the stale information for is bounded it may be an acceptable compromise.

For the source of the stale node information we could use either:
* an in-memory cache of the node attestation data. This would allow us to better control how long we trust the stale information for. For example we could consider the information in the cache useable for 30 seconds, but needs to be refreshed after 10-15 seconds.
* the read-only datastore connection. This allows us to read this information from any of the read-replicas that are still available, in the hope that they can still serve the data. We don't have that much control over how stale the data is, but we will usually have up to date information since replication is quick.

Some operations still need up to date attested node information, such as `RenewAgent`, so they will have to fetch that information themselves. They currently already do that.

There were some other people complaining about this in https://github.com/spiffe/spire/issues/4484#issuecomment-2392493031 as well as some discussion of it a long time ago in the contributors sync.

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Verifying attested nodes

**Description of change**
This PR adds a cache of attested nodes, with short TTLs, to be used when validating nodes during RPC processing. The cache can be used to avoid frequent round trips to the database, for example during JWT-SVID signing. This can help with dealing with short periods of unavailability of the database.

**Which issue this PR fixes**
fixes #5954


doc/spire_server.md
```diff
@@ -84,6 +84,7 @@ This may be useful for templating configuration files, for example across differ
...
 | `trust_domain`                      | The trust domain that this server belongs to (should be no more than 255 characters)                                                                                                                                                            |                                                                |
 | `use_legacy_downstream_x509_ca_ttl` | Use the downstream spire-server registration entry TTL as the downstream CA TTL. This is deprecated and will be removed in a future version.                                                                                                    | false                                                          |
+| `max_attested_node_info_staleness`  | How long to cache and use attested node information before requiring fetchign up to date data from the datastore.                                                                                                                               | 0s                                                             |
```
MarcosDY:
```suggestion
| `max_attested_node_info_staleness`  | How long to cache and use attested node information before requiring fetching up to date data from the datastore.                                                                                                                               | 0s                                                             |
```

pkg/server/cache/nodecache/cache.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	return node, nodeRefreshTime, nil
+}
+
+func (c *Cache) RefreshAttestedNode(ctx context.Context, id string) (*common.AttestedNode, error) {
+	node, err := c.ds.FetchAttestedNode(ctx, id)
+	if err != nil {
+		c.mtx.Lock()
+		defer c.mtx.Unlock()
```
MarcosDY: looks like lock is applied always on error or success, can you move this above?

MarcosDY: or you can call `RemoveAttestedNode()`

pkg/server/endpoints/endpoints.go
```diff
@@ -149,14 +154,23 @@ func New(ctx context.Context, c Config) (*Endpoints, error) {
...
 	ds := c.Catalog.GetDataStore()
 
+	var nodeCache *nodecache.Cache
+
+	var err error
 	var ef api.AuthorizedEntryFetcher
-	var cacheRebuildTask, pruneEventsTask func(context.Context) error
+	var cacheRebuildTask, nodeCacheRebuildTask, pruneEventsTask func(context.Context) error
 	if c.EventsBasedCache {
+		nodeCache, err = nodecache.New(ctx, c.Log, ds, c.Clock, true, c.MaxAttestedNodeInfoStaleness != 0)
```
MarcosDY: since it is shared  may we initialize it before if?

pkg/server/endpoints/middleware.go
```diff
@@ -67,7 +68,25 @@ func AgentAuthorizer(ds datastore.DataStore, clk clock.Clock) middleware.AgentAu
 			return errorutil.PermissionDenied(types.PermissionDeniedDetails_AGENT_EXPIRED, "agent %q SVID is expired", id)
 		}
 
-		attestedNode, err := ds.FetchAttestedNode(ctx, id)
+		cachedAgent, agentCacheTime, err := nodeCache.FetchAttestedNode(id)
+		switch {
+		case err != nil:
+			// AttestedNode not found in local cache, will fetch from the datastore
```
MarcosDY: can we verify if it is really not found issue? I'm worried about we getting real errors and never noticed

pkg/server/endpoints/middleware.go
```diff
@@ -67,7 +68,25 @@ func AgentAuthorizer(ds datastore.DataStore, clk clock.Clock) middleware.AgentAu
...
+		case cachedAgent == nil:
+			// AttestedNode not found in local cache, will fetch from the datastore
+		case clk.Now().Sub(agentCacheTime) >= maxAttestedNodeInfoStaleness:
+			// Cached AttestedNode is stale, will attempt to refresh from the database
+		case cachedAgent.CertSerialNumber == "":
+			// Attested node was not found in the cache, will fetch from the datastore
+		case cachedAgent.CertSerialNumber == agentSVID.SerialNumber.String():
+			// AgentSVID matches the current serial number, access granted. We
```
MarcosDY: can you complete the comment?

pkg/server/endpoints/middleware.go
```diff
@@ -67,7 +68,25 @@ func AgentAuthorizer(ds datastore.DataStore, clk clock.Clock) middleware.AgentAu
...
+			return nil
+		default:
+			// Could not validate the agent using the cache attested node information
+			// so we'll try fetching the up to date data from the datastore.
+		}
+
+		attestedNode, err := nodeCache.RefreshAttestedNode(ctx, id)
```
MarcosDY: may we call it FetchAttestedNode?

pkg/server/endpoints/middleware.go
sorindumitru: I think I don't even need the error there since this is only supposed to check the cache, probably a leftover from something. I'll remove it.

MarcosDY (approved):

PR #6179, opened 2025-07-10
Include selectors in 'no identity issued' logs

### LINKED ISSUE #2651: Log workload selectors in same log as SVID return result
Today SPIRE Agent has logs for FetchX509SVID results for:
- [No Identity Issued](https://sourcegraph.com/github.com/spiffe/spire@626a0ce28f4ad852a3f06b7e244a8d3541a51962/-/blob/pkg/agent/endpoints/workload/handler.go?L311), [Identity Issued](https://sourcegraph.com/github.com/spiffe/spire@626a0ce28f4ad852a3f06b7e244a8d3541a51962/-/blob/pkg/agent/endpoints/workload/handler.go?L337), and errors in between.
- [Discovered selectors](https://sourcegraph.com/github.com/spiffe/spire@626a0ce28f4ad852a3f06b7e244a8d3541a51962/-/blob/pkg/agent/attestor/workload/workload.go?L77)

However the discovered selectors and identity decision are separate log entries, which makes debugging a bit more difficult than it needs it be, especially in the case of "No Identity Issued":
- two separate logs must be found to see the result and what was discovered about the caller
- if there are many parallel callers, it is difficult to associate results with selectors

Adding selectors as a log field to the logger passed in [here](https://sourcegraph.com/github.com/spiffe/spire@626a0ce28f4ad852a3f06b7e244a8d3541a51962/-/blob/pkg/agent/endpoints/workload/handler.go?L232) (and potentially other parts of this package) could add clarity for operators.

Potentially this could mean that the existing [discovered selectors](https://sourcegraph.com/github.com/spiffe/spire@626a0ce28f4ad852a3f06b7e244a8d3541a51962/-/blob/pkg/agent/attestor/workload/workload.go?L77) log becomes redundant and could be removed, though it is already Debug level.

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Agent logging when no identity could be issued

**Description of change**
The agent will include the previously attested selectors for the workload when logging that no identity could be issued

**Which issue this PR fixes**
fixes #2651


pkg/agent/endpoints/workload/handler_test.go
```diff
@@ -95,9 +95,10 @@ func TestFetchX509SVID(t *testing.T) {
 					Level:   logrus.ErrorLevel,
 					Message: "No identity issued",
 					Data: logrus.Fields{
-						"registered": "false",
-						"service":    "WorkloadAPI",
-						"method":     "FetchX509SVID",
+						"previously_attested_selectors": "[]",
```
MarcosDY: can you add a test that contains selectors? so we can control returned format?

pkg/agent/endpoints/workload/handler.go
```diff
@@ -103,7 +103,10 @@ func (h *Handler) FetchJWTSVID(ctx context.Context, req *workload.JWTSVIDRequest
 	}
 
 	if len(resp.Svids) == 0 {
-		log.WithField(telemetry.Registered, false).Error("No identity issued")
+		log.WithFields(logrus.Fields{
+			telemetry.Registered:                  false,
+			telemetry.PreviouslyAttestedSelectors: selectors,
```
sorindumitru: One thing @zmt pointed out here was that these selectors could have secret/private information in them that we may leak into logs by logging them. This shouldn't happen, but people are free to put whatever they want in things such as pod/container labels. Since this information is readable by a lot of processes on a node, I doubt this is a common situation.

We already log such selectors at debug level in a bunch of places and at error level in https://github.com/spiffe/spire/blob/eeec5782edf4ebb67954b4f4166088b2bff92dee/pkg/agent/attestor/workload/workload.go#L79, so I don't think we're making this any worse by logging them in this case too.

One thing that was proposed is to have some support for filtering these selectors. I think that can be done as a separate enhancement.

MarcosDY: we are already sending this logs in debug, so actually users can filter using that, but now we are introducing each in info level

MarcosDY: here: https://github.com/MarcosDY/spire/blob/b5beef4006d62ae206bfc7aceab8e818e9b926a2/pkg/agent/attestor/workload/workload.go#L90

sorindumitru: Yes, we are already logging them in debug. But debug level logging is usually a lot of logs, so might not be enabled. If we log it here, we only log it when there is an issue so when the logs are the most useful.

I could remove it from the debug log, but that's probably still useful to be logged since it gets logged for each attestation.

zmt: We were forced to disable debug logging, which makes it very difficult to operate without any selector logs. It's like flying blind when workloads get rejected. Even with the debug selector logging, it was difficult to correlate with the no identity issued by PID at scale. I think this does belong at info level aggregated to the "no identity issued" log.

We are in the process of developing our own patch similar to this, but it also adds selector logging to the "Fetched X.509 SVID" message, which we also set to Info level. For that, it should be sufficient to just log the matched selectors that attested the workload via the entry although at present our implementation just logs them all.

In the same patch, we're doing similar cleanup to delegated identity logging by making "[Subscribing to cache changes](https://github.com/spiffe/spire/blob/main/pkg/agent/api/delegatedidentity/v1/service.go#L183)" Debug level and adding selectors to the "[Fetched X.509 SVID for delegated identity](https://github.com/spiffe/spire/blob/main/pkg/agent/api/delegatedidentity/v1/service.go#L239)" message at Info level.

You might consider some/all of those changes as part of this PR, although it does expand the initial scope. It would be nice to resync with upstream on logging details when we get to code that includes this PR. We have not deployed our flavor of this yet pending redactor work.

As a separate patch, we're going to implement a simple configurable selector redactor. We're trying to make it lightweight and only applied to logging with a sensible default list of redactions (like "secret", "key", etc.) while accepting a configurable list. So, configurables should be bool to enable default and optionally specify a custom list. This work is in flight.

pkg/common/telemetry/names.go
```diff
@@ -435,6 +435,9 @@ const (
 	// PreferredServiceName tags the preferred service name
 	PreferredServiceName = "preferred_service_name"
 
+	// Selectors tags some group of registration selector
+	PreviouslyAttestedSelectors = "previously_attested_selectors"
```
zmt: Why not just use the [Selectors label](https://github.com/spiffe/spire/blob/main/pkg/common/telemetry/names.go#L510-L511) ("selectors")? I don't see the need for this customizaiton. It doesn't hurt per se, just seems unnecessary (and adds 20 additional characters per instance).

zmt (approved):
This is essentially identical to a subset of our planned patch changes with better tests. I offered some additional suggestions. I'll sync back once we progress on a redactor implementatiuon and probably try to work it up as a PR adapted from our patch.

pkg/agent/endpoints/workload/handler.go
amartinezfayo: Thank you @zmt. We are thinking that ideally this PR should be in the same release as the separate patch including a configurable selector redactor that your team is working on. Do you have any estimation of the timeline when that could be done? That can help us to plan better how this is introduced. Thanks!

sorindumitru: @zmt, we discussed this today. We think the selectors in those log messages would be useful, but it may lead to a large number of extra logs.

There's no auditing support in the agent yet, but if there was that information could be added as part of the audit logs. If you think that would be useful, would you mind opening an issue for that so we can start discussing it?

pkg/common/telemetry/names.go
sorindumitru: This was suggested on related issue to this a while ago. Discussing it again today, we do think that it would be less confusing overall to use "selectors" or "discovered_selectors" as we do in other places. I'll update the patch with that.

MarcosDY (approved):
LGTM!!!

pkg/agent/endpoints/workload/handler.go
zmt: I have finally gotten time to cycle back to re-working our log adjustments, including implementation of a lightweight redactor for selector logs with configurable. Since this PR has landed, I reverted our local patches to instead cherry-pick this.

I just put up a draft PR with the delta of what we originally added to workload api logging: #6254 as a conversation starter. Next up will be adjustments that we made to delegated identity logging as a separate draft PR.

After that, I will tackle the redactor, but I think I will have to start over from scratch to make it contributable.

PR #6184, opened 2025-07-13
Un-flake datastore-mysql-replication test

sorindumitru:
This test relies on 100ms sleeps for the replication to happen. This is hard to guarantee in a CI environment, so I'm replacing these sleeps with require.Eventually, waiting for up to 1 second for the changes to replicate.


pkg/server/datastore/sqlstore/sqlstore_test.go
```diff
@@ -2934,12 +2930,6 @@ func (s *PluginSuite) testListRegistrationEntries(dataConsistency datastore.Data
 					entryIDMap[entryOut.EntryId] = entryIn.EntryId
 				}
 
-				// Optionally sleep to give time for the entries to propagate to
-				// the replicas.
-				if dataConsistency == datastore.TolerateStale && s.readOnlyDelay > 0 {
-					time.Sleep(s.readOnlyDelay)
-				}
```
sorindumitru, line 3053: This requires some bigger changes to the test to be able to also test using `Eventually`. It didn't seem to be required though, probably because we first check with `RequireCurrent` and by the time that finishes the data has replicated already.

pkg/server/datastore/sqlstore/sqlstore_test.go
```diff
@@ -1676,41 +1665,48 @@ func (s *PluginSuite) TestNodeSelectors() {
 	}
 
 	// assert there are no selectors for foo
-	selectors := s.getNodeSelectors("foo", datastore.TolerateStale)
+	selectors := s.getNodeSelectors("foo", datastore.RequireCurrent)
 	s.Require().Empty(selectors)
-	selectors = s.getNodeSelectors("foo", datastore.RequireCurrent)
+	selectors = s.getNodeSelectors("foo", datastore.TolerateStale)
```
sorindumitru: The order here seemed a bit more natural. First check that the function does what it's supposed to do and then check that the data is replicated.

amartinezfayo (approved):
Thank you @sorindumitru!

PR #6238, opened 2025-08-09
fix(cli): avoid printing 'but failed to delete 0 entries' when all de…

### LINKED ISSUE #6225: spire-server entry delete:  "but failed to delete 0 entries"
As per example below:

```sh
$ bin/spire-server entry delete -entryID d90ba760-c5af-4dd9-8c87-56b7c7204d00
Deleted entry with ID: d90ba760-c5af-4dd9-8c87-56b7c7204d00


Deleted 1 entries successfully, but failed to delete 0 entries
```

_"but failed to delete 0 entries"_ is both un-necessary and confusing. 😉

vishalanarase:
…letions succeed

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**
fixes #6225

### DIFF
--- a/cmd/spire-server/cli/entry/delete.go
+++ b/cmd/spire-server/cli/entry/delete.go
@@ -145,7 +145,11 @@ func (c *deleteCommand) prettyPrintDelete(env *commoncli.Env, results ...any) er
 		return errors.New("failed to delete one or more entries")
 	}
 
-	env.Printf("Deleted %d entries successfully, but failed to delete %d entries", len(succeeded), len(failed))
+	if len(failed) > 0 {
+		env.Printf("Deleted %d entries successfully, but failed to delete %d entries", len(succeeded), len(failed))
+	} else {
+		env.Printf("Deleted %d entries successfully", len(succeeded))
+	}
 
 	return nil
 }


cmd/spire-server/cli/entry/delete.go
```diff
@@ -145,7 +145,11 @@ func (c *deleteCommand) prettyPrintDelete(env *commoncli.Env, results ...any) er
 		return errors.New("failed to delete one or more entries")
 	}
 
-	env.Printf("Deleted %d entries successfully, but failed to delete %d entries", len(succeeded), len(failed))
+	if len(failed) > 0 {
```
sorindumitru: There's already a a check like this just before this, which ends up returning so we never enter this if block. Could you combine the two together?

vishalanarase: okay

vishalanarase: upated the PR

cmd/spire-server/cli/entry/delete.go
```diff
@@ -140,12 +141,20 @@ func (c *deleteCommand) prettyPrintDelete(env *commoncli.Env, results ...any) er
 			result.Status.Message)
 	}
 
+	var summaryMsg string
 	if len(failed) > 0 {
 		env.Printf("\n\n")
+		summaryMsg = fmt.Sprintf("Deleted %d entries successfully, but failed to delete %d entries", len(succeeded), len(failed))
```
sorindumitru: We can move the variable to be inside the if condition:
```suggestion
if len(failed) > 0 {
env.Printf("\n\n")
		summaryMsg := fmt.Sprintf("Deleted %d entries successfully, but failed to delete %d entries", len(succeeded), len(failed))
```

cmd/spire-server/cli/entry/delete.go
```diff
@@ -142,10 +143,17 @@ func (c *deleteCommand) prettyPrintDelete(env *commoncli.Env, results ...any) er
 
 	if len(failed) > 0 {
 		env.Printf("\n\n")
+		summaryMsg := fmt.Sprintf("Deleted %d entries successfully, but failed to delete %d entries", len(succeeded), len(failed))
+
+		if len(succeeded) == 0 {
+			summaryMsg = fmt.Sprintf("Failed to delete %d entries", len(failed))
+		}
+
+		env.Printf("%s\n\n", summaryMsg)
```
sorindumitru: One last thing. I think this would leave us with 2 empty lines after the output which wouldn't look very nice so probably best tonot have them.

```suggestion
		env.Printf("%s", summaryMsg)
```

sorindumitru (approved):
Thanks for the contribution, @vishalanarase!

PR #6249, opened 2025-08-18
Bump SPIRE to 1.13.1

sorindumitru:


pkg/server/datastore/sqlstore/migration.go
```diff
@@ -259,6 +259,7 @@ import (
 // | v1.12.2 |        |                                                                           |
 // | v1.12.3 |        |                                                                           |
 // | v1.12.4 |        |                                                                           |
+// | v1.13.0 |        |                                                                           |
```
MarcosDY, line 264: can you add the minor version separator?
```suggestion
// |*********|********|***************************************************************************|
// | v1.13.0 |        |                                                                           |
```

sorindumitru: Thanks, changed it

Issue #2666, opened 2022-01-03
Periodically re-attest workloads


evan2645:
Workload attestation currently occurs when a workload first connects to the workload api. This is fine for most use cases, but causes a problem for selectors that may describe runtime state that can change over the lifetime of a workload.

Workload attestation is relatively expensive, and we must block on it when the workload first calls us, but it is not necessary to block on further re-attestations which can be done in the background.

rturner3:
It would probably be a good idea to evaluate the current selectors supported by the builtin WorkloadAttestor plugins to see which are immutable for the lifetime of a process. One edge case scenario I could think of is for the `unix:user`, `unix:group`, and `unix:supplementary_group` selectors.

Example:

1. Entry `spiffe://domain/foo` is registered in SPIRE Server with selector `unix:user:bob`
2. Workload is launched on a host as the `bob` user, and it receives the identity `spiffe://domain/foo` identity from SPIRE Agent
3. The `bob` user is renamed to `robert` on the host

In this case, the `robert` user will retain the same uid as `bob`, but it will fail to re-attest to SPIRE.

I'm not sure how realistic this scenario is in practice, but I will say it is generally more convenient to register workloads by `unix:user` rather than `unix:uid` since the user's uid is not always known and may not be stable across all hosts where the workload runs.

PR #6254, opened 2025-08-19
Workloadapi FetchX509SVID logs selectors at info on success

zmt:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Agent logging on successful FetchX509SVID.

**Description of change**
Agent logs at info level with selectors on successful FetchX509SVID.

Initial follow-up to discussion in #6179, relates to #2651, and probably depends on #2666 (closed unplanned).


zmt:
TBH: I'm not entirely convinced of the value of logging full selectors here, but it might still be useful to log the matching selectors.

amartinezfayo:
We discussed this in the SPIRE Contributor call and concluded that logging a “Fetched X.509 SVID” message at the Info level each time an X509-SVID is fetched could overwhelm the agent logs.

This might be a good opportunity to introduce Audit Logging in the agent, similar to what we already have in the server. That way, this event could be captured as an audit log, allowing users to opt in if they want to track fetched X.509 SVIDs.

Even though selectors may change after attestation, I think it’s still valuable to include selectors in the audit log entry. Users should be able to have the selectors redacted.

zmt:
I'm going to abandon this for now. I like the suggestion of adding audit log capability to spire-agent.

PR #6272, opened 2025-08-26
Add option to disable JWT-SVID profile on spire server

### LINKED ISSUE #4473: Provide the ability to disable JWT-SVID support in SPIRE
There are some scenarios where JWT-SVID support is not needed, and servers could avoid having to maintain JWT authorities and the rotation of JWT keys.

There is consensus that having the ability to disable JWT-SVID issuance in SPIRE would be a good thing to have, but we need to find answers to some open questions before starting the implementation.

- The [SPIFFE Workload API](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Workload_API.md) standard mandates that both the X.509-SVID and the JWT-SVID profiles must be supported by SPIFFE implementations, but operators may disable a specific profile in their deployment. This means that there is no limitation from the spec perspective. But the error code in case that a profile is disabled is not defined by the specification. We need to address that at the specification level.
- Name and semantic of the config setting to be added in the server (e.g.: disable_jwt defaulting to false vs enable_jwt defaulting to true).
- How is this handled by agents? Servers can inform agents about the disabled profiles. Requests to the Workload API related with the JWT-SVID profile should return the code defined by the specification.
- What's the behavior in nested SPIRE deployments? The error code returned should be consistent across SPIRE.
- Considerations when federating with other trust domains. Define the behavior when validating JWT-SVIDs.

nikotih:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Spire server JWT-SVID

**Description of change**

New option to disable JWT-SVID in spire server. JWT keys are no longer managed. Server and Agent api endpoints related to JWT return precondition failed grpc code

Answers from the issue:
- returning FailedPrecondition grpc code in JWT related endpoints

- decided to go with disable_jwt so dont have to go through all unit tests setting enable_jwt to true (but let me know what you prefer)

- example calls:
```
time="2025-08-26T13:35:22Z" level=error msg="JWT functionality is disabled" authorized_as=local authorized_via=transport method=MintJWTSVID request_id=5f535de6-d5e5-466a-beb4-ceab4f32e88f service=svid.v1.SVID subsystem_name=api
time="2025-08-26T13:35:22Z" level=error msg="JWT functionality is disabled" authorized_as=local authorized_via=transport method=PrepareJWTAuthority request_id=a756696c-c986-4e10-8f41-7c10a769f9e3 service=localauthority.v1.LocalAuthority subsystem_name=api
time="2025-08-26T13:35:23Z" level=error msg="JWT functionality is disabled" authorized_as=local authorized_via=transport method=GetJWTAuthorityState request_id=4bdbcaf6-cfbd-4a6b-abe4-0982d2502f88 service=localauthority.v1.LocalAuthority subsystem_name=api
time="2025-08-26T13:35:23Z" level=error msg="JWT functionality is disabled" authorized_as=local authorized_via=transport method=ActivateJWTAuthority request_id=8e6191f2-dd52-48d5-a16b-3ff17149fba7 service=localauthority.v1.LocalAuthority subsystem_name=api
time="2025-08-26T13:35:23Z" level=error msg="JWT functionality is disabled" authorized_as=local authorized_via=transport method=TaintJWTAuthority request_id=5963ec52-dca2-4d90-89af-0d7b0a3ab0ec service=localauthority.v1.LocalAuthority subsystem_name=api
time="2025-08-26T13:35:23Z" level=error msg="JWT functionality is disabled" authorized_as=local authorized_via=transport method=RevokeJWTAuthority request_id=aad432cf-6f46-4cea-a983-89916b14f1be service=localauthority.v1.LocalAuthority subsystem_name=api
time="2025-08-26T13:35:23Z" level=error msg="JWT functionality is disabled" authorized_as=agent authorized_via=datastore caller_addr="10.244.0.1:61487" caller_id="spiffe://example.org/spire/agent/k8s_psat/example-cluster/eb94ef4d-e54b-4783-8165-ecb99d8806b7" method=NewJWTSVID request_id=b4697234-c82e-4d36-8789-1440235980a3 service=svid.v1.SVID subsystem_name=api
```

- i tested this with nested deployment and got mixed results. Servers where jwt is disabled cannot issue JWT but can validate JWT from other nested servers. I suppose this is working as expected?

- i did not find any federated trust domain integration test that uses JWT. I tried some things in ghosttunnel-federation test and in general came to opinion it wont be an issue. Trust bundles are exchanged and jwt validation was not disabled. (but feel free to suggest test cases to verify this properly)

Overall this fulfils our requirement:
- no jwt keys are managed
- jwt endpoints are disabled

Before i proceed with unit/integration tests and **documentation** please let me know if this is in general correct direction. 
I was a little lost with double abstraction levels - there is CA and CA manager, there is also CA rotator
Maybe you prefer disablement config in endpoints directly (however, CA manager would still need it to not manage JWT keys)

**Which issue this PR fixes**

fixes #4473

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -94,7 +94,7 @@ require (
 	k8s.io/client-go v0.34.1
 	k8s.io/kube-aggregator v0.34.1
 	k8s.io/mount-utils v0.34.1
-	sigs.k8s.io/controller-runtime v0.22.3
+	sigs.k8s.io/controller-runtime v0.22.4
 )
 
 require (
--- a/go.sum
+++ b/go.sum
@@ -2560,8 +2560,8 @@ rsc.io/binaryregexp v0.2.0/go.mod h1:qTv7/COck+e2FymRvadv62gMdZztPaShugOCi3I+8D8
 rsc.io/pdf v0.1.1/go.mod h1:n8OzWcQ6Sp37PL01nO98y4iUCRdTGarVfzxY20ICaU4=
 rsc.io/quote/v3 v3.1.0/go.mod h1:yEA65RcK8LyAZtP9Kv3t0HmxON59tX3rD+tICJqUlj0=
 rsc.io/sampler v1.3.0/go.mod h1:T1hPZKmBbMNahiBKFy5HrXp6adAjACjK9JXDnKaTXpA=
-sigs.k8s.io/controller-runtime v0.22.3 h1:I7mfqz/a/WdmDCEnXmSPm8/b/yRTy6JsKKENTijTq8Y=
-sigs.k8s.io/controller-runtime v0.22.3/go.mod h1:+QX1XUpTXN4mLoblf4tqr5CQcyHPAki2HLXqQMY6vh8=
+sigs.k8s.io/controller-runtime v0.22.4 h1:GEjV7KV3TY8e+tJ2LCTxUTanW4z/FmNB7l327UfMq9A=
+sigs.k8s.io/controller-runtime v0.22.4/go.mod h1:+QX1XUpTXN4mLoblf4tqr5CQcyHPAki2HLXqQMY6vh8=
 sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 h1:gBQPwqORJ8d8/YNZWEjoZs7npUVDpVXUUOFfW6CgAqE=
 sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8/go.mod h1:mdzfpAEoE6DHQEN0uh9ZbOCuHbLK5wOm7dK4ctXE9Tg=
 sigs.k8s.io/randfill v1.0.0 h1:JfjMILfT8A6RbawdsK2JXGBR5AQVfd+9TbzrlneTyrU=


keeganwitt:
Maybe a dumb question, but should we update the logic here too?
https://github.com/spiffe/spire/blob/eefb529d8782c3a4eb607d976826b39c249ec3cc/pkg/common/bundleutil/bundle.go#L235-L238

keeganwitt:
Capturing an idea @amartinezfayo had in the sync today: we might want to have the SPIRE server return a 501 UNIMPLEMENTED response when the OIDC provider tries to get JWKS, so that it can error out appropriately.

nikotih:
thank you @keeganwitt 

WIP thoughts

had a look at oidc provider code. found 2 usages:

a) talking to server directly via socket and calling getBundle and output mask JWT keys only
https://github.com/spiffe/spire/blob/main/support/oidc-discovery-provider/server_api.go#L112

Can set an exit branch for that if server has JWT disabled and get bundles is only asking for JWT keys
```
func (s *Service) GetBundle(ctx context.Context, req *bundlev1.GetBundleRequest) (*types.Bundle, error) {
	rpccontext.AddRPCAuditFields(ctx, logrus.Fields{telemetry.TrustDomainID: s.td.Name()})
	log := rpccontext.Logger(ctx)

	// if only JWKs are requested and we have JWT disabled
	if s.IsJWTDisabled() && s.req.OutputMask != nil && req.OutputMask.JwtAuthorities {
		return nil, api.MakeErr(log, codes.Unimplemented, "JWKs are not managed by this server", nil)
	}
```

Not the best solution but would work.
Another problem here is that bundles endpoint does not have CA or CA manager it uses datastore
So again might have to bring "disable_jwt" flag to endpoints or CA / CA manager

b) talking to agent via workload api and calling get bundle on local agent

this is more complicated, basically agent needs to know if jwt is disabled on server

can of course assume that if JwtAuthorities are empty in DB then its disabled? but that wont be the case for nested deployments or federated trust domains.

so agent needs to distinguish local server has jwt disabled but JwtAuthorities from nested / federated still in bundle

amartinezfayo:
Thank you @nikotih for working on this contribution.
I think that the approach in general looks good. I also explored if we could disable JWT support in the middleware, but I think that it will end up being a more complex implementation.

A couple of quick considerations:
- I think that we should return `codes.Unimplemented` as opposed to `codes.FailedPrecondition` when JWT-SVID support is disabled. I think that consumers receiving `codes.Unimplemented` can identify better that JWT-SVID support has been disabled in SPIRE.
- I would also suggest to name the setting `disable_jwt_svids`, to be more specific.
- We should log a message when the server has started with JWT SVID issuance disabled (probably at info level).

Have you looked at how this would work in the context of a nested SPIRE deployment? Particularly, if there are more than one level and one of the SPIRE servers in the middle has JWT-SVID disabled. It would be good to have this considered and clarity on the behavior.

nikotih:
Hi @amartinezfayo 

I've changed to codes.Unimplemented, renamed to disable_jwt_svids and added log line.

For nested setup i've modified original integration test and disabled jwt on intermediate B - it cannot issue JWTs but agent connected to it can still validate against other bundles

Apart from that i am more uncertain about parts of code that Keegan mentioned.
a) PruneBundle will need to know 0 jwtkeys might be ok - either flag that JWT is disabled have to be available there or it should not error on no keys in bundle
First will require datastore to know if jwt is disabled and propagate it all the way down (a lot of changes in interfaces), second might have side effects i am not aware of (0 keys in bundle is ok?)

b) Get Bundle issues mentioned above. If you want to simplify this in oidc provider to no JWT keys in bundle = jwt disabled that can work. Otherwise get bundle api becomes unpredictable and high cognitive complexity design:
- get all returns CAs and JWKs including from nested and federated
- get only JWKs will return unimplemented (despite bundle having JWKs from nested of federated)
- agent caching bundles will also need to know jwt is disabled - it gets all bundles from server (CA+JWT) so when e.g. oidc provider asks for bundles form agent - agent has no idea if its disabled on server

I would maybe think of some alternative way of propagating "JWT disabled" via grpc, some config sync between server<->agent, agent<->oidc provider and server<->oidc provider. This will keep bundle endpoints as is

If you have any ideas or suggestions please share

amartinezfayo:
Hey @nikotih, these are my thoughts:
>a) PruneBundle will need to know 0 jwtkeys might be ok - either flag that JWT is disabled have to be available there or it should not error on no keys in bundle
First will require datastore to know if jwt is disabled and propagate it all the way down (a lot of changes in interfaces), second might have side effects i am not aware of (0 keys in bundle is ok?)

I think that we should just need to update the condition that returns the error, so it only returns an error if the current bundle already has JWT keys and the resulting bundle after pruning has no JWT keys. Something like this:

```go
	if len(bundle.JwtSigningKeys) > 0 && len(newBundle.JwtSigningKeys) == 0 {
		log.Warn("Pruning halted; all known JWT signing keys have expired")
		return nil, false, errors.New("would prune all JWT signing keys")
	}
```

This way, a server with JWT SVIDs disabled should not return an error when pruning the bundle.

>b) Get Bundle issues mentioned above. If you want to simplify this in oidc provider to no JWT keys in bundle = jwt disabled that can work. Otherwise get bundle api becomes unpredictable and high cognitive complexity design:

In my opinion, the GetBundle RPC should just return the empty set of JWT authorities when JWT SVIDs are disabled. RPCs that are JWT-specific (e.g. MintJWTSVID) should return an error with the Unimplemented code.
Since the SPIRE OIDC Discovery Provider is a separate tool, it doesn't directly know about the `disable_jwt_svids` setting. I think that as it own logic could return Unimplemented in the `/keys` endpoint when SPIRE Server returned no JWT keys.

nikotih:
Thanks @amartinezfayo 

adjusted as you described. I presume this covers it then - marked as ready for review
let me know which kind of tests you would like - unit tests, integration tests (i've extended nested suite) or something else

cmd/spire-server/cli/run/run.go
```diff
@@ -79,6 +79,7 @@ type serverConfig struct {
 	DefaultJWTSVIDTTL            string             `hcl:"default_jwt_svid_ttl"`
 	Experimental                 experimentalConfig `hcl:"experimental"`
 	Federation                   *federationConfig  `hcl:"federation"`
+	DisableJWT                   bool               `hcl:"disable_jwt_svids"`
```
amartinezfayo: I think that it would be good to have consistency with the variable names. This probably needs to be renamed to DisableJWTSVIDs everywhere.

support/oidc-discovery-provider/handler.go
```diff
@@ -161,6 +161,11 @@ func (h *Handler) serveKeys(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if len(jwks.Keys) == 0 {
+		http.Error(w, "jwt not supported/enabled in this service", http.StatusInternalServerError)
```
amartinezfayo: I think that this should return Not Implemented as opposed to an Internal error.
```suggestion
		http.Error(w, "jwt not supported/enabled in this service", http.StatusNotImplemented)
```

nikotih:
Hi @amartinezfayo 

- renamed to `DisableJWTSVIDs` everywhere
- adjusted oidc helper to `http.StatusNotImplemented`
- created new integration suite `fetch-jwt-svids`
    - used `fetch-x509-svids` as base so it also checks cache pruning - let me know if you want to simplify and remove that part
    - switches to `disable_jwt_svids = true` and uses asserts on all apis that should fail
    - had to use disk persistence for server to not re-bootstrap bundle for agent - but can also change that
    - adjusted `check-x509-svid-count` -> `check-svid-count`

pkg/server/ca/ca.go
```diff
@@ -391,6 +393,14 @@ func (ca *CA) signJWTSVID(jwtKey *JWTKey, claims map[string]any) (string, error)
 	return signedToken, nil
 }
 
+func (ca *CA) IsJWTSVIDsDisabled() bool {
+	return ca.c.DisableJWTSVIDs
+}
+
+func (ca *CA) SetDisableJWTSVIDs(disableJWTSVIDs bool) {
```
amartinezfayo: It seems like SetDisableJWTSVIDs is only used by tests?
If that's the case, I recommend removing this method entirely to prevent confusion and potential misuse in production code. The `DisableJWTSVIDs` setting should be configured at CA construction time only, and not provide the ability to be mutated at runtime.

Could we just have an attribute in the serviceTest struct indicating if JWTSVIDs are enabled or disabled? We can update that attribute as needed in the tests to exercise JWTSVIDs disabled scenarios.

doc/spire_server.md
```diff
@@ -68,6 +68,7 @@ This may be useful for templating configuration files, for example across differ
...
 | `experimental`                      | The experimental options that are subject to change or removal (see below)                                                                                                                                                                      |                                                                |
 | `federation`                        | Bundle endpoints configuration section used for [federation](#federation-configuration)                                                                                                                                                         |                                                                |
+| `disable_jwt_svids`                 | If true, disables JWT-SVID profile                                                                                                                                                                                                              | false                                                          |
```
amartinezfayo: Maybe we could be a little more descriptive here to help users to understand when to use this configuration?

Something like:
If true, completely disables JWT-SVID functionality. The server will not generate JWT keys, sign JWT-SVIDs, or implement JWT-related API calls. This is useful for deployments that don't need JWT-SVIDs support.

pkg/server/ca/ca.go
nikotih: from what i remember i had to add it due to `test/fakes/fakeserverca/serverca.go` overlay over CA. if there is no setter then that fake serverca has to store its own flag and ignore what is set in the actual CA

anyways i will have a look again tomorrow

doc/spire_server.md
nikotih: Adjusted

pkg/server/ca/ca.go
nikotih: i've removed SetDisableJWTSVIDs from ca.go but kept it on `fakeserverca/serverca.go` 
that fake ca then stores local flag and does not go to check value of real CA

that was smallest change to make this work, let me know if you want to pursue adjustment to `serviceTest struct`. I've tried to sketch it but it would require big change in tests

amartinezfayo (approved):
Thank you again, @nikotih, for this valuable contribution.
The ability to disable JWT-SVIDs is something we’ve heard requests for from the community for quite some time. This will be included in the 1.14.0 release 🚀

PR #6275, opened 2025-08-27
Test against currently supported Postgres versions

rturner3:
Test against the latest supported versions: https://www.postgresql.org/support/versioning/

Fix one config that was deprecated in an older version of Postgres.


sorindumitru (approved):
LGTM

PR #6276, opened 2025-08-27
Support setting refresh hint on bundle publisher plugins

### LINKED ISSUE #6256: Set refresh hint with bundle publisher plugins
With the spire server federation bundle endpoint you can [set the refresh hint](https://github.com/spiffe/spire/blob/main/doc/spire_server.md?plain=1#L223) but with the bundle publisher plugins this is not possible to configure. It would be great to be able to set this when publishing spiffe format bundles.

hennersz:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

The bundle publisher plugins that support the spiffe format now also allow you to set/override the refresh hint for the bundle.

**Description of change**

With the federation bundle endpoint you can set the [refresh hint](https://github.com/spiffe/spire/blob/main/doc/spire_server.md?plain=1#L223) to control how often federating servers attempt to refresh the bundle. This PR adds similar functionality to the 3 bundle plugins that support the spiffe format, namely GCP, k8s and S3 bundle publishers. If the refresh hint is configured for the plugin it will set/overwrite the refresh hint on the bundle when publishing it. If the refresh hint is unchanged the behaviour of the plugins is unchanged.

**Which issue this PR fixes**

fixes #6256


amartinezfayo:
Nicely done, @hennersz! Thank you so much for this contribution ❤️
It looks great. Could you fix the linter issue? It's just about a non-wrapping format verb used in fmt.Errorf, where `%w` should be used.

hennersz:
heya @amartinezfayo should be sorted now

amartinezfayo (approved):
Thank you @hennersz!

PR #6293, opened 2025-09-02
doc: fix `x509pop` URI SAN selector documentation

r-bk:
The san selector uri should be of the form

`x509pop://<trust-domain>/<key>/<value>`

as is implemented at [1].

[1] https://github.com/spiffe/spire/blob/c68c6a452d1fdb9e482eeaad56c1a61248609a3a/pkg/server/plugin/nodeattestor/x509pop/x509pop.go#L377


sorindumitru (approved):
Thanks for fixing this, @r-bk!

sorindumitru:
@r-bk, you also need to apply the DCO to your commits. You can do that using:
```
git commit --amend --signoff
git push --force
```

PR #6294, opened 2025-09-02
Configurable HashAlgorithm For Generating Subject Key Identifiers

### LINKED ISSUE #6292: [FIPS Compliance] Configurable HashAlgorithm (x509util/keyid.go)
Go 1.25.0 already makes it relatively easy to run the spire-server in a FIPS 140-3 mode by setting the `fips140=only` `GODEBUG` flag.
However, SHA-1 usage of spire is a problem in this mode. SHA-1 usage can be found in the following places (#4009):

- ./pkg/server/plugin/nodeattestor/tpmdevid/selectors.go
- ./pkg/server/plugin/keymanager/gcpkms/gcpkms.go
- ./pkg/agent/plugin/svidstore/gcpsecretmanager/gcloud.go
- ./pkg/common/plugin/x509pop/x509pop.go
- ./pkg/common/x509util/keyid.go

To work towards a spire-server that is able to run in the `fips140=only` mode, we should be able to restrict the hash algorithms used. Therefore I suggest to introduce a config entry `HashAlgorithm` that can be set to `sha256`, with `sha1` being the default value to preserve the behaviour of currently deployed spire-servers.

This issue should fix the most central usage of SHA-1, i.e. the one in `./pkg/common/x509util/keyid.go`. By using custom plugins or plugins other than the ones listed above, one could already start running a spire-server in FIPS 140-3 mode when setting the `HashAlgorithm` config entry to "sha256".

The other usages of SHA-1 can be addressed in follow-up issues.

nweisenauer-sap:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
`./pkg/common/x509util/keyid.go`
`GetSubjectKeyID` now receives a parameter to decide whether it should use SHA-1 or SHA-256 to compute the subject key identifier. The implementation follows the go standard library [crypto/x509](https://cs.opensource.google/go/go/+/master:src/crypto/x509/x509.go;l=1732?q=RFC%207093&ss=go%2Fgo).

**Description of change**
To work towards a spire-server that is able to run in the `fips140=only` mode, we should be able to restrict the hash algorithms used. Therefore I introduced a config entry `HashAlgorithm` that can be set to `sha256`, with `sha1` being the default value to preserve the behaviour of currently deployed spire-servers.

This issue should fix the most central usage of SHA-1, i.e. the one in `./pkg/common/x509util/keyid.go`. By using custom plugins or plugins other than the ones listed above, one could already start running a spire-server in FIPS 140-3 mode when setting the `HashAlgorithm` config entry to `sha256`.

The other usages of SHA-1 can be addressed in follow-up issues.

**Which issue this PR fixes**
fixes #6292


doc/spire_server.md
```diff
@@ -68,6 +68,7 @@ This may be useful for templating configuration files, for example across differ
...
 | `experimental`                      | The experimental options that are subject to change or removal (see below)                                                                                                                                                                      |                                                                |
 | `federation`                        | Bundle endpoints configuration section used for [federation](#federation-configuration)                                                                                                                                                         |                                                                |
+| `hash_algorithm`                    | Specifies the algorithm used to generate the SubjectKeyID                                                                                                                                                                                                  | sha1                                                           |
```
sorindumitru: Two notes on this:

* We can call it something like `fips_compatibility_mode`. This makes it have a clearer purpose and also allow us to keep further changes under the same configuration option.
* We should move this under the experimental section until we're done with the fips compatible changes.

nweisenauer-sap: Hey,
I actually had it like that as a boolean before but I didn't really like it. I think naming it `fips_compatibility_mode` would currently give the wrong impression that you can run SPIRE in a FIPS compliant mode by just setting this to true, while actually there are a lot more strings attached to that.
Additionally, maybe in the future we don't want to toggle between SHA-1 and SHA-256 but have some new hash algorithm that can be used XYZ-123. With this string config we could simply add it in.

So in my opinion, this is just a small and even independent step to achieve a `fips_compatibility_mode` which could still have its own config entry, as I implemented it. But I am also open to change it to a `bool` and call it `fips_compatibility_mode`, adding some description that this does not achieve FIPS compliance and rather is just a step towards that.

sorindumitru: Another option would be to us [fips140.Enabled()](https://pkg.go.dev/crypto/fips140#Enabled) to check if fips140 mode is enabled and use that to check if we should use sha-1 or sha-256. Our understanding is that using sha-1 in fips140 mode would panic so it would be safe to use it. This way we don't need any configuration for this.

We'd also want to limit the changes we have to support this special mode and transitions from fips mode to non-fips mode and vice-versa. We'd like to have a global flag that says once the server is started in tips mode it's always expected to be started in fips mode. We can do this with the datastore, with a new key:value pair table.

nweisenauer-sap: Ah, I wasn't aware of this function! Sounds great, that way we don't need any configs spread across server and plugins and can use this global Go flag - much simpler. When `fips140.Enabled()` reports true, which is only the case if the user made the decision to set the proper GODEBUG flag, we use SHA-256, otherwise we use the "standard" SHA-1 for compatibility with existing deployments.

And about the datastore entry, do you think we can deal with it in a separate PR? So I will just replace the config entry here with the usage of `fips140.Enabled()`?

nweisenauer-sap: Any updates?

sorindumitru:
Looked a bit a this today, the main side effect I can see when using `GODEBUG=fips140=on` you change the subject key id of the server CA, which means that it can't find its CA in the journal. The server will then create a new CA and start using it immediately. Probably not that big of an issue, but maybe something to note.

sorindumitru:
Sorry for the delay here @nweisenauer-sap. Looked a bit more at this today; I think we should still make sure that we only apply these changes for the case where the fips140-3 mode is set to `only`, not `on`. Unfortunately the Enabled function doesn't make a distinction between the two.

Would it be possible to look at the environment variable to make this decision?

nweisenauer-sap:
Hey @sorindumitru,
unfortunately reading the env variable would not work out, as there are at least two other ways of setting the godebug values. One of them is to declare a `godebug(fips140=only)` block in the mod file, the other way is to add a comment `//go:debug fips140=only` preceding the package statement in the source file, both of them would not appear in `os.Getenv("GODEBUG")`.
The `fips140.Enabled()` function is the only public way of querying whether fips140 is enabled, unfortunately it does not give us any information about whether it is set to "on" or "only". The godebug pkg itself is internal, so we cannot use `godebug.Value("#fips140")` to query the exact value.

However, I think setting fips140 mode for SPIRE is a conscious decision and users who want to run this mode do care about not using SHA-1, so I would still suggest we go for the proposed solution. What do you think?

sorindumitru:
> However, I think setting fips140 mode for SPIRE is a conscious decision and users who want to run this mode do care about not using SHA-1, so I would still suggest we go for the proposed solution. What do you think?

Our main concern here is not making any potentially breaking behaviour changes. It's possible that someone is already using `fips140=on` and in that case they may encounter some surprising things. In particular this may cause spire-server to not be able to use it's previous X509 and JWT signing keys, because it uses the subject key id for lookup, and having to create new ones and start using them immediately. This can cause issues due to time it takes for the new keys to propagate in bundles to all workloads.

How about having our own `FIPS140OnlyEnabled` function that:
* returns false if `fips140.Enabled()` is false
* returns strings.Contains(os.Getenv("GODEBUG"), "fips140=only")

How does that sound to you?

nweisenauer-sap:
Ok, yeah I understand your concern. I can build this `FIPS140OnlyEnabled ` function as suggested. Basically this would ignore setting the GODEBUG flag in any other way **but** the env var, but it should be safe and exclude the mentioned scenario of not  being able to use previous X509 and JWT signing keys, since anybody who sets the env var to "only" currently experiences panics.

sorindumitru:
> Ok, yeah I understand your concern. I can build this `FIPS140OnlyEnabled ` function as suggested. Basically this would ignore setting the GODEBUG flag in any other way **but** the env var, but it should be safe and exclude the mentioned scenario of not being able to use previous X509 and JWT signing keys, since anybody who sets the env var to "only" currently experiences panics.

Yes, sounds good. Sorry about the multiple iterations on this, it took a while to fully understand how this fips140 mode works :).

nweisenauer-sap:
No worries, that is how it should be.

sorindumitru (approved):
LGTM, thanks @nweisenauer-sap !

PR #6303, opened 2025-09-06
Add a variant of the X.509 CA metric with absolute expiration time, not TTL

vandry:
- [✓] Commit conforms to CONTRIBUTING.md?
- [✓] Proper tests/regressions included?
- [✓] Documentation updated?

**Affected functionality**
Add a variant of the X.509 CA metric with absolute expiration time, not TTL

**Description of change**
- Add support for float64 gauge metrics (needed for a time_t-valued metric)
- Add a variant of the X.509 CA metric with absolute expiration time, not TTL.
    
    In Prometheus at least, a metric with the TTL is not usable because it is not known when the metric was set and thus what time the TTL is relative to (it is possible to convey the timestamp when the metric was set using an optional extended scrape format, but Spire does not export in that format and even if it did the resulting metric would more difficult to work with in Prometheus). It is better to publish the absolute expiration time and let the monitoring system subtract the current time to get a TTL if it wants.
    
    I left the existing TTL-based metric in place rather than deleting it in case that way is useful or preferred by one of the other supported telemetry exporters, but if it isn't then I would recommend deprecating it.


sorindumitru:
Could you give us some more details about your setup and how you're using these metrics? These metrics are usually ingested into some time series storage which should also include a timestamp. If not the timestamp at which the metric was emitted, at least the one when it was read.

vandry:
> usually ingested into some time series storage which should also include a timestamp

The [Prometheus metric exposition format](https://prometheus.io/docs/instrumenting/exposition_formats/) includes an optional timestamp:

```
metric_name [
  "{" label_name "=" `"` label_value `"` { "," label_name "=" `"` label_value `"` } [ "," ] "}"
] value [ timestamp ]
```

but Spire doesn't set it, so when Prometheus ingests it it substitutes the current time (the scrape time). The TTL is relative to the current time at the time that the metric was set, which is an unknown amount of time in the past. If the value of the metric were calculated on demand each time a request for the metrics was received then it would be different but it's not.

sorindumitru:
> when Prometheus ingests it it substitutes the current time (the scrape time

Doesn't Prometheus scrape often enough that this is not an issue? If it's off by a minute, I don't think it really matters that much for an expiry. Are you scraping infrequently?

vandry:
> Doesn't Prometheus scrape often enough that this is not an issue?

Yes but it continually scrapes the same value that was set whenever the key was activated (when `SetGauge` was called)

Consider the timeline:
- New CA is generated, good for 86400 seconds (for example)
- Let's say it takes a couple of seconds' precessing before it is activated
- Activation occurs. Metric gets set to the TTL "right now" which is 86398
- Prometheus scrapes some few seconds later, gets 86398 for the value
- A minute later Prometheus scrapes again, gets 86398 again
- and so on
- Next day a new CA is generated and activated again
  - Maybe rotation is faster this time and the metric gets set to 86399
  - Or maybe it's exactly the same and it gets set to 86398 again (unchanged)

Either way the value is useless for observing the age of the certificate.

sorindumitru:
> Yes but it continually scrapes the same value that was set whenever the key was activated (when `SetGauge` was called)

Thanks for this detail, I understand what's happening now. I think this isn't an issue for other metric sinks because this metric is published only once when the key is created/activated. With Prometheus the metric seems to be stored and read later on too.

This might be an issue for other metrics too, making it hard to tell what's happening with a deployment. I see that [go-metrics](https://github.com/hashicorp/go-metrics/blob/8f13a6de82d154f76b4831a5a0e1a823eb58d113/prometheus/prometheus.go#L34) exposes a configuration for  expiring metrics. Maybe that's something that we should be using. We'll discuss among the maintainers and come back to you.

vandry:
Sure, Ill wait for your response. But just as a heads up I would caution against expirong the metric: I suspect that will just cause gaps in the timeseries which is hard to work with and would mean that upon first scrape or after a scraping outage the monitoring system can see no value at all for a long while until the CA is rotated again and the metric appears again (but only until it expires again).

You're probably right that the other sink types deal with the TTL style of metric better.

sorindumitru:
We discussed this today and we think it would be better to periodically publish the TTL of the CA to make sure an up to date metric is published. We can do this from a go-routine that publishes it every couple of minutes.

We feel that using the TTL as opposed to the NotAfter makes it a bit easier to use in some cases, such as when generating alerts (e.g. alert when TTL < 86400).

Would this work ok for your case or do you think that having the NotAfter would still be useful in some cases?

I also think that we should make the changes to the prometheus sink to specify an expiry and also make that configurable for the sink for the cases where someone configured a different scraping interval. Doesn't have to be set to 1 minute, it could be set to something like `default_scraping_interval * N` to deal with some of the cases where prometheus didn't manage to scrape the metrics yet. Those can be left for a separate PR though.

vandry:
I think that the periodic refresh is a bit inferior since you never know upon scraping how much out of date the received value is: it could be anywhere from 0 seconds out of date to several minutes out of date. A graph of the timeseries might look strange as well: say you scrape every 15s and do the periodic update every 60s, then the value plots a straight horizontal line for 4 data points in a row then jumps down by a minute (you would be able to see that in Grafana: when you zoom in enough on the time axis it draws individual dots for scraped datapoints).

However as you said yourself upthread a difference of a few minutes doesn't matter for this metric and for any alerts that are likely to be configured on it. If your solution is applied, that is satisfactory, it will do the job.

sorindumitru:
@vandry Yes, we'd like to go with a configuration option for expiring the in-memory metrics with a default of 1 minute mostly because that's going to improve other metrics that are not published often, not just this one. Would you be open to implementing it?

vandry:
That would roll back https://github.com/spiffe/spire/commit/83c4243a35cef1a1d14fea56334f0203994b1d9d and I do not think it's what you want. Plenty of the metrics are updated less frequently than once per minute and you wouldn't want them to disappear after a minute. Consider in particular:

- Most event counters on a lightly loaded server. Nothing happens for a minute (meaning no counter increments) and the counter disappears. Worse, when an event later does happen, it reappears with a starting value of zero.
- Gauges that track magnitudes like the number of database entries. If the database size happens to not change for a minute or more, the metric is supposed to stick around with an unchanged value, not disappear.
- There is even [one gauge metric](https://github.com/spiffe/spire/blob/2a38ff929c5dbd4e43526777da97075f7e571ae3/pkg/common/telemetry/started.go) that is set exactly once at startup and never updated; also meant to remain with its constant value, not expire.

sorindumitru:
>     * Most event counters on a lightly loaded server. Nothing happens for a minute (meaning no counter increments) and the counter disappears. Worse, when an event later does happen, it reappears with a starting value of zero.

Maybe a counterpart to this is that on a server that doesn't do anything for a few minutes you will have the impression that it is doing something because of the cached metrics; this can be confusing. 

>     * There is even [one gauge metric](https://github.com/spiffe/spire/blob/2a38ff929c5dbd4e43526777da97075f7e571ae3/pkg/common/telemetry/started.go) that is set exactly once at startup and never updated; also meant to remain with its constant value, not expire.

I think it's good to have in mind that we support multiple telemetry sinks and not all of them have this behaviour of caching metrics. Others only publish metrics when they are emitted by the application. If that gauge you mention is worth publishing more often (and it may very well be) I don't think having that behaviour just in the promtheus sink is a good user experience. We should make sure that all sinks behave similarly and I think the way to do that is by periodically emitting that metric.

vandry:
I don't think that's how either counters or gauges work, no matter the sink or its protocol.

Counters are expected to start at 0 and increment each time some "event" occurs. If no events occur, they just don't increment. They just maintain their previous value. A non-incrementing counter does not

> the impression that it is doing something because of the cached metrics

Rather, the flat constant value indicates that nothing is happening. Note that you need at least 2 samples of a counter to tell whether or not something is happening.

Gauges are expected to always reflect the current (or last known) value of something. They might change over time, they might not.

**Neither type of metric should ever disappear, unless the thing being measured stops existing.**

You are of course right that all types of sinks and their differing protocols must be considered. But I'm pretty sure the basic properties and definitions outlined above hold in a sink-independent way.

I can speak for Prometheus and its protocol. It's a stateless polling protocol. The monitoring system calls the target and asks for a dump of all metrics. All metrics should be reported. It doesn't make sense to omit any: with the stateless nature of the protocol you can't know whether or not the caller has any previous known values for any of them anyway. Every scrape can be treated as if the caller is scraping you for the first time. You can't omit a metric just because it hasn't changed since the last scrape.

I don't have personal experience with any of the other sinks and protocols, so I looked 2 of them up. Dogstatsd appears to be an evolution of statsd so they are similar. It looks like a push protocol (in contrast with Prometheus) and I found some [documentation](https://github.com/statsd/statsd/blob/master/docs/metric_types.md). For counters it says

> At each flush the current count is sent and reset to 0.

and for gauges it says

> If the gauge is not updated at the next flush, it will send the previous value.

So, neither counters nor gauges require constant refresh in order to avoid being "forgotten" by the monitoring system. They just get zero values[*] (for counters) or repeated values (for gauges) whenever not updated. They definitely aren't expected to expire or disappear if not updated.

I expect that the remaining sinks and protocols work that way too.

[*] As a stateful protocol statsd can afford to zero counters between samples. This remains equivalent to Prometheus' stateless protocol's monotonic flat value. For counters it's only the derivative that matters.

sorindumitru:
> Rather, the flat constant value indicates that nothing is happening. Note that you need at least 2 samples of a counter to tell whether or not something is happening.

I can now see how this works with Prometheus and probably also why it doesn't necessarily make sense to reset the metrics in memory. I imagine the counter itself is not that useful in a lot of situations, e.g. the call counters we use. There you're mostly interested in the `rate()` or the `increase()` which Prometheus can give you.

All in all, I think that there might not be that much difference between the different metic sinks to the end users and how they use these metrics. They both have points where the counters can reset to 0 it's just a difference of how often they happen so whatever system is used to view the time series needs to account for it. I doubt someone will use the raw counter value for metrics such as `$rpc.call.c`.

> You are of course right that all types of sinks and their differing protocols must be considered. But I'm pretty sure the basic properties and definitions outlined above hold in a sink-independent way.

Not based on what I observe and reading the go-metrics code. [Prometheus](https://github.com/hashicorp/go-metrics/blob/8f13a6de82d154f76b4831a5a0e1a823eb58d113/prometheus/prometheus.go#L407) stores the metrics in memory, consistent with the pull model. StatsD [pushes the metrics to a queue](https://github.com/hashicorp/go-metrics/blob/8f13a6de82d154f76b4831a5a0e1a823eb58d113/statsd.go#L120) from where they are later [written to a UDP socket](https://github.com/hashicorp/go-metrics/blob/8f13a6de82d154f76b4831a5a0e1a823eb58d113/statsd.go#L146).

So if you do `IncrCounter(1); IncrCounter(1);` with prometheus you get a single counter polled with value 2 while statsd publishes 2 different metrics with counter 1. At the next interval prometheus would read 2 again while statsd would see nothing because there's nothing new to publish.

You can easily see the StatsD/DogStatsD metrics being published with `nc -k -u -l localhost 8125`:
```
// We see the version you mentioned only at the start of spire-server.
spire_server.started.1_13_1_dev_unk.example_org:1.000000|g
spire_server.datastore.registration_entry.list.OK:1.000000|c
// after 5 seconds
spire_server.datastore.registration_entry.list.OK:1.000000|c
```

With Prometheus you get the cumulated value instead of just 1. 

This discussion maybe is not that relevant to the initial issue. That the TTL we publish isn't useful because it has no timestamp associated with it. It does have a timestamp, at least when seen as part of a time series not as a single data point. What maybe doesn't make that much sense is only publishing that TTL when creating the metrics. It would be useful to publish that TTL periodically and I think that would fix your initial issue and also make the metric more useful to users of the other sinks. It also helps with the case where metrics are not stored forever for example due to volume.

I'm don't think I'd be opposed to having your other metric as well, it's just that we already have a metric for it with the TTL. If that metric isn't useful because it's not published very often it should

sorindumitru:
@vandry Do you think you'll have the time to implement the change where the TTL is published periodically? Publishing it every minute or every few minutes seems like a good interval to us.

sorindumitru:
@vandry We discussed this a bit more. The documentation for this metric ties it to a specific rotation event so might be confusing to periodically. So to fix it we'd have to introduce a new metric and if that's the case maybe we can just introduce the one here.

pkg/common/telemetry/server/ca_manager.go
```diff
@@ -30,11 +32,20 @@ func StartServerCAManagerPrepareX509CACall(m telemetry.Metrics) *telemetry.CallC
...
-// TTL of CA for a specific TrustDomain
-func SetX509CARotateGauge(m telemetry.Metrics, trustDomain string, val float32) {
+// expiration time and TTL of CA for a specific TrustDomain
+func SetX509CARotateGauge(m telemetry.Metrics, trustDomain string, expiration, now time.Time) {
+	m.SetPrecisionGaugeWithLabels(
+		[]string{telemetry.Manager, telemetry.X509CA, telemetry.Rotate, telemetry.Expiration},
+		// There is little point in using UnixNano: in 2025 float64 only
+		// barely has enough precision for microseconds.
+		// (for the curious: we lose microsecond precision around the year 2112.)
+		float64(expiration.UnixMicro())/1e6,
```
sorindumitru: Is it not possible to user `expiration.Unix()` directly?

vandry: Metrics are floats anyway so  included the fractional part. You're right though, it's overkill: nobody cares if the certificate expires at noon + 0.1s or noon + 0.9s. Probably certificate expirations are only to the nearest second anyway!

Changed.

vandry:
> @vandry Do you think you'll have the time to implement the change where the TTL is published periodically? Publishing it every minute or every few minutes seems like a good interval to us.

Sorry for the silence. Priorities "happened". I did also talk about how gauge metrics with some folks on my side and it looks as though we might agree there's a case to be made to republlish them (but not also counter metrics) periodically, in light of your netcat trace that shows that doesn't happen on its own, plus some other evidence we found. It would only made a difference for push-style sinks (like statsd), not pull-style sinks (like Prometheus) and I don't think the Prometheus options should be changed to expire by default. Let's talk about that and other topics separately to this PR.

sorindumitru (commented):
@vandry I think this is looking good. Would you mind fixing up the DCO on the commits missing it?

vandry:
DCO signoff added

PR #6152, opened 2025-06-24
Implemented periodic purging of expired agents

ValFadeev:
**Pull Request check list**

- [x] Commit conforms to CONTRIBUTING.md?
- [x] Proper tests/regressions included?
- [x] Documentation updated?

**Affected functionality**
Adds the capability to enable periodic pruning of attested agents with certificates expired for longer than a configurable duration of time. 

**Description of change**
Introduces new configuration options `prune_attested_nodes_expired_for` and `prune_non_reattestable_nodes`. When `prune_attested_nodes_expired_for` is set to a given duration value SPIRE server will start a periodic task deleting attested nodes and the corresponding selectors, where agent certificate had been expired for longer than the specified duration. Non-reattestable nodes are not pruned by default. In order to consider the latter for pruning `prune_non_reattestable_nodes` must be set to `true`. Banned nodes are not pruned by this task.

In the interests of limiting the scope of the change, and to require explicit opt-in at this stage, the implementation of the recently introduced `agent purge` subcommand (https://github.com/spiffe/spire/pull/3982) is left unchanged.

**Which issue this PR fixes**

Closes [#5938](https://github.com/spiffe/spire/issues/5938)

PR #6304, opened 2025-09-08
Some improvements to registration manager test

ValFadeev:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

N/A

**Description of change**

This change aims to make improvements to the unit tests of the registration entry manager, partly incorporating the approach and feedback from an earlier change in https://github.com/spiffe/spire/pull/6152:
* uses context from test suite instead of `Background`
* avoids calling a private method directly (`m.prune`)
* tests against the actual pruning cadence, rather than triggering at arbitrary moments


pkg/server/registration/manager_test.go
```diff
@@ -91,50 +92,55 @@ func (s *ManagerSuite) TestPruning() {
...
 	// prune first entry
-	s.clock.Add(_pruningCadence + time.Second)
-	s.NoError(s.m.prune(context.Background()))
-	listResp, err = s.ds.ListRegistrationEntries(context.Background(), &datastore.ListRegistrationEntriesRequest{})
-	s.NoError(err)
-	s.Equal([]*common.RegistrationEntry{registrationEntry2, registrationEntry3}, listResp.Entries)
+	s.clock.Add(_pruningCadence)
+	s.Require().Eventuallyf(func() bool {
+		listResp, err := s.ds.ListRegistrationEntries(ctx, &datastore.ListRegistrationEntriesRequest{})
+		s.NoError(err)
```
sorindumitru: I think that `s.NoError(err)` will abort the whole test, not just one iteration of `Eventually`. You can use `EventuallyWithT` which gives you an object that can be used to use in the `NoError` call.

pkg/server/registration/manager_test.go
```diff
@@ -91,50 +92,55 @@ func (s *ManagerSuite) TestPruning() {
...
+	}, 1*time.Second, 100*time.Millisecond, "Expected no entries to have been pruned")
 
 	// prune first entry
-	s.clock.Add(_pruningCadence + time.Second)
-	s.NoError(s.m.prune(context.Background()))
-	listResp, err = s.ds.ListRegistrationEntries(context.Background(), &datastore.ListRegistrationEntriesRequest{})
-	s.NoError(err)
-	s.Equal([]*common.RegistrationEntry{registrationEntry2, registrationEntry3}, listResp.Entries)
+	s.clock.Add(_pruningCadence)
+	s.Require().Eventuallyf(func() bool {
```
sorindumitru: It's probably sufficient to use the version without `f` at the end since you use a static string, no formatting.
```suggestion
	s.Require().Eventually(func() bool {
```

pkg/server/registration/manager_test.go
ValFadeev: Thanks! This also helps avoid the use of reflection.

PR #6309, opened 2025-09-09
Improve k8s_psat documentation a bit

sorindumitru:
Define the permissions need by spire-server to attest the agent. I don't think we specify these in any place apart from the examples.

No need to mention the APIs we use, they're available on all versions we support now.


doc/plugin_server_nodeattestor_k8s_psat.md
```diff
@@ -61,6 +61,17 @@ A sample configuration for SPIRE server running outside of a Kubernetes cluster:
     }
 ```
 
+The kube config file needs to grant at least the following access to `spire-server`:
```
rturner3: The way this is worded makes it sound like these roles belong in the kubeconfig file, but they are not defined in kubeconfig files. Can we reword this to mention that the Kubernetes user defined in the kubeconfig file needs to have ClusterRoleBindings assigned to ClusterRoles containing the following permissions?

rturner3 (approved):

Issue #6310, opened 2025-09-09
Periodic purging of expired agents background job might never run


nikotih:
#6152 introduced background job for periodic purging of expired agents including TOFU

it allows `prune_attested_nodes_expired_for` and `prune_non_reattestable_nodes` configurations

However, first ticker job is also bound to `prune_attested_nodes_expired_for`. So if one sets `prune_attested_nodes_expired_for` e.g. for 720h (30 days) it might never run because likely server will be restarted due to maintenance / deployment etc.

I see 2 options to solve it:

a) run initial job on the start (might not be desirable if it starts removing stuff before server is loaded)

```
	ticker := m.c.Clock.Ticker(m.c.ExpiredFor)
	defer ticker.Stop()

	for {
		// run right away
		if err := m.prune(ctx, m.c.Clock.Now().Add(-m.c.ExpiredFor), m.c.IncludeNonReattestable); err != nil && ctx.Err() == nil {
			m.log.WithError(err).Error("Failed during periodic pruning of expired nodes")
		}

		select {
		case a := <-m.pruneRequestedCh:
			if err := m.prune(ctx, m.c.Clock.Now().Add(-a.ExpiredFor), a.IncludeNonReattestable); err != nil && ctx.Err() == nil {
				m.log.WithError(err).Error("Failed during on-demand pruning of expired nodes")
			}
		case <-ctx.Done():
			return nil
		// wait for ticket later
		case <-ticker.C:
		}
	}
```
b) disconnect `prune_attested_nodes_expired_for` from job cadence and allow configuring it separately
```
	ticker := m.c.Clock.Ticker(m.c.Interval)
```

Another concern is e.g. if we run 6 spire server replicas and job is executed at the same time. I know its transaction based so probably they will conflict with each other. But maybe it makes sense to also allow jitter interval config

Lastly, hardcoded limit of 1000 entries in https://github.com/spiffe/spire/blob/main/pkg/server/datastore/sqlstore/sqlstore.go#L1787
might also not work well with long expired_for. I suppose faster job cadence can help here, but maybe consider allowing to configure that one too

sorindumitru:
Thanks @nikotih for bringing this to our attention! We discussed this a bit today, we could try:

* Having a much faster cadence, the pruning can run every hour.
[...]

PR #6315, opened 2025-09-10
Using fixed pruning job interval instead of expired for configuration

nikotih:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

Background job that purges attested note entries

**Description of change**

Instead of using "expired for" for job interval, using fixed 1 hours intervals instead

**Which issue this PR fixes**

#6310
This does not include a fix for handling conflicts between spire server replicas


faisal-memon:
the linked issue mentions having a jitter but i dont see that in here. is that still desired?

nikotih:
yes i did not add solution for conflicts as stated in the description. Jitter ticker is not trivial implementation

i can see 3 possibilities here from simple to complicated:
- introduce once random offset from static 1 hour. Something like 1 hour +-[0-15]minute. Job then runs always X duration
- introduce ticker with jitter support e.g. https://github.com/mroth/jitter. Every cycle re-calculates timer
- introduce something like https://github.com/go-co-op/gocron that supports random job runs. But spire already has task runner so this feels like an overkill

I went with first option now, samples:
```
INFO[0000] Node manager periodic prune job started       retry_interval=1h8m53s subsystem_name=node_manager

INFO[0000] Node manager periodic prune job started       retry_interval=57m41s subsystem_name=node_manager
```

Just do mind that jitter does not solve this problem 100% and someone can still run into conflicts if very unlucky

amartinezfayo:
I think that the implemented jitter strategy is good enough. It's a great starting point to avoid the races/collisions; we can see how that works and improve it if needed.

pkg/server/node/manager.go
```diff
@@ -58,7 +69,9 @@ func (m *Manager) Prune(ctx context.Context, expiredFor time.Duration, includeNo
 }
 
 func (m *Manager) pruneEvery(ctx context.Context) error {
-	ticker := m.c.Clock.Ticker(m.c.ExpiredFor)
+	m.log.Info("Node manager periodic prune job started")
```
amartinezfayo: The log already includes the subsystem name (subsystem_name=node_manager), so “Node manager” feels redundant. Instead, the message could be more specific about what is being pruned, for example: “Periodic prune of expired nodes started.”
Since this runs every hour, I think Debug level would be more appropriate.

amartinezfayo (commented):
Thank you for this contribution @nikotih! It's looking great.

pkg/server/node/manager.go
nikotih: adjusted log message. also added `expired_for` and `include_tofu`

```
DEBU[0000] Signed X509 SVID                              expiration="2025-09-17T15:32:07Z" spiffe_id="spiffe://domain.test/spire/server" subsystem_name=svid_rotator
INFO[0000] Building in-memory entry cache                subsystem_name=endpoints
INFO[0000] Completed building in-memory entry cache      subsystem_name=endpoints
INFO[0000] Logger service configured                     launch_log_level=debug
INFO[0000] Periodic prune of expired nodes started       expired_for=24h0m0s include_tofu=false retry_interval=57m0s subsystem_name=node_manager
DEBU[0000] Initializing API endpoints                    subsystem_name=endpoints
INFO[0000] Starting Server APIs                          address="[::]:8081" network=tcp subsystem_name=endpoints
```

This log line is outside of ticker for loop, so it only appears once during startup - so i think info is fine

FYI: cleanup job itself omits log lines from sqlstore:
```
INFO[0071] Pruned expired agents                         count=0 subsystem_name=sql
INFO[0142] Pruned expired agents                         count=0 subsystem_name=sql
```

Issue #6218, opened 2025-08-01
Add the capability to prevent further modifications to entries


daescha:
There is no way to disable the entry modification via cli once you have registered what you need. This seems like a useful feature to have, so you don't have to worry about attackers registering additional workloads if they gain exec access to the spire-server container.

MarcosDY:
Seems related with https://github.com/spiffe/spire/issues/1842 and https://github.com/spiffe/spire/issues/2626

daescha:
@MarcosDY thanks for linking to the interesting discussion on those issues. I was thinking it would probably be better to split this issue into two parts, one for disabling entry additions/updates, the other for adding initial entries. adding the initial entries would just be part of #2626 I guess, so Ill remove that part of this issue. 

e: removed that part.

MarcosDY:
Thanks @daescha!! we saw your PR, can you provide an example of the entry file?
How about hardcoding entry id to those "static" entries?

daescha:
@MarcosDY further discussion should probably take place on the PR (or alternatively #2626) since I changed the subject of this issue to be the other part of my initial issue, but Ill respond to what you said here this time at least.

example is same format as the the file passed to `spire-server entry create -data`:

```json
    {
      "entries": [
        {
          "parent_id": "spiffe://example.com/spire/server",
          "selectors": [
            {
              "type": "k8s_psat",
              "value": "agent_ns:spire-system"
            },
            {
              "type": "k8s_psat",
              "value": "cluster:minikube"
            },
            {
              "type": "k8s_psat",
              "value": "agent_sa:spire-agent"
            }
          ],
          "spiffe_id": "spiffe://example.com/ns/spire-system/sa/spire-agent"
        },
        {
          "parent_id": "spiffe://example.com/ns/spire-system/sa/spire-agent",
          "selectors": [
            {
              "type": "k8s",
              "value": "container-name:envoy-control-plane"
            },
            {
              "type": "k8s",
              "value": "ns:default"
            },
            {
              "type": "k8s",
              "value": "pod-label:app.kubernetes.io/name:gateway"
            },
            {
              "type": "k8s",
              "value": "sa:gateway"
            }
          ],
          "spiffe_id": "spiffe://example.com/ns/default/sa/gateway/gateway"
        },
        {
          "parent_id": "spiffe://example.com/ns/spire-system/sa/spire-agent",
          "selectors": [
            {
              "type": "k8s",
              "value": "container-name:gateway"
            },
            {
[...]

PR #6320, opened 2025-09-12
Add capability to filter on caller path in rego

daescha:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

`caller_path` is added as a referenceable field in the rego policy, when the `caller` is empty and the address is able to be resolved.

**Description of change**

`caller_path` field added to Input struct passed to rego. This field is set in authorization_opa.go:opaAuth, before the Input object is `Eval`d.

**Which issue this PR fixes**

related to https://github.com/spiffe/spire/issues/6218#issuecomment-3281464887


daescha:
@sorindumitru @MarcosDY this is a relatively simple change. My org believes this to be important for an imminent release. We would like to know if we can count on this being available. Would it be possible to get a review on this soon?

pkg/server/authpolicy/policy.go
```diff
@@ -40,7 +40,8 @@ type LocalOpaProviderConfig struct {
 type Input struct {
 	// Caller is the authenticated identity of the actor making a request.
 	Caller string `json:"caller"`
-
+	// CallerPath is the file path of a local actor making a request.
+	CallerPath string `json:"caller_path"`
```
faisal-memon:
```suggestion
	CallerPath string `json:"caller_path"`
	
```

pkg/server/api/middleware/authorization_opa.go
```diff
@@ -32,6 +34,15 @@ func (m *authorizationMiddleware) opaAuth(ctx context.Context, req any, fullMeth
 		Req:        req,
 	}
 
+	if input.Caller == "" {
+		// ignore result if error occurs
+		if watcher, ok := peertracker.WatcherFromContext(ctx); ok {
+			if p, err := process.NewProcess(watcher.PID()); err == nil {
+				input.CallerPath, _ = getAddr(p)
```
faisal-memon: what is the return value being ignored?

daescha: what I meant by that was if its not ok, if err is not nil, carry on and don't set the callerpath or throw an error. Maybe I should remove the comment.

e: removed

pkg/server/api/middleware/authorization_opa.go
```diff
@@ -32,6 +34,15 @@ func (m *authorizationMiddleware) opaAuth(ctx context.Context, req any, fullMeth
 		Req:        req,
 	}
 
+	if input.Caller == "" {
+		// ignore result if error occurs
```
daescha:
```suggestion
```

pkg/server/authpolicy/policy.go
```diff
@@ -40,7 +40,9 @@ type LocalOpaProviderConfig struct {
 type Input struct {
 	// Caller is the authenticated identity of the actor making a request.
 	Caller string `json:"caller"`
-
+	// CallerPath is the file path of a local actor making a request.
```
daescha:
```suggestion

	// CallerPath is the file path of a local actor making a request.
```

daescha:
@amartinezfayo I ran fmt on the file to remove the erroneous spacing so the lint check should pass now

daescha:
@amartinezfayo is there anything I need to do for this?

doc/authorization_policy_engine.md
```diff
@@ -88,6 +88,7 @@ The inputs that are passed into the policy are:
 | input field | Description                                                                                                                      | Example                                    |
 |-------------|----------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------|
 | caller      | The SPIFFE ID (if available) of the caller                                                                                       | spiffe://example.org/workload1             |
+| caller_path | The binary path (if available) of the caller                                                                                     | /spire-controller-manager                  |
```
amartinezfayo: Maybe `caller_file_path` is a better name to be more explicit?

amartinezfayo (commented):
Sorry for the delay @daescha.
This is looking good. I think I have only one comment, that's about the setting name.

doc/authorization_policy_engine.md
daescha: fine with me, done @amartinezfayo. 

though I will note, the original name was mimicking the audit log field of the same name, since they are resolved the same way.

amartinezfayo (approved):
Thank you @daescha!

Issue #6313, opened 2025-09-10
Flaky tests in pkg/common/health


sorindumitru:
I've seen this time out from time to time in CI:

```
panic: test timed out after 1m30s
	running tests:
		TestHealthFailsAndRecover (1m30s)
		TestHealthFailsAndRecover/start_successfully_after_initial_failure (1m30s)

goroutine 21 [running]:
testing.(*M).startAlarm.func1()
	/opt/hostedtoolcache/go/1.24.6/x64/src/testing/testing.go:2484 +0x605
created by time.goFunc
	/opt/hostedtoolcache/go/1.24.6/x64/src/time/sleep.go:215 +0x45

goroutine 1 [chan receive]:
testing.(*T).Run(0xc0002a6700, {0x10ffeff, 0x19}, 0x112e5b8)
	/opt/hostedtoolcache/go/1.24.6/x64/src/testing/testing.go:1859 +0x91e
testing.runTests.func1(0xc0002a6700)
	/opt/hostedtoolcache/go/1.24.6/x64/src/testing/testing.go:2279 +0x86
testing.tRunner(0xc0002a6700, 0xc000287ae0)
	/opt/hostedtoolcache/go/1.24.6/x64/src/testing/testing.go:1792 +0x226
testing.runTests(0xc000010948, {0x17d6a80, 0x7, 0x7}, {0x7?, 0x1?, 0x17e6ca0?})
	/opt/hostedtoolcache/go/1.24.6/x64/src/testing/testing.go:2277 +0x96d
testing.(*M).Run(0xc0003a2140)
	/opt/hostedtoolcache/go/1.24.6/x64/src/testing/testing.go:2142 +0xeeb
main.main()
	_testmain.go:59 +0x165

goroutine 20 [chan receive]:
testing.(*T).Run(0xc000102c40, {0x110e016, 0x28}, 0xc00018a0f0)
	/opt/hostedtoolcache/go/1.24.6/x64/src/testing/testing.go:1859 +0x91e
github.com/spiffe/spire/pkg/common/health.TestHealthFailsAndRecover(0xc000102c40)
	/home/runner/work/spire/spire/pkg/common/health/cache_test.go:153 +0x985
testing.tRunner(0xc000102c40, 0x112e5b8)
	/opt/hostedtoolcache/go/1.24.6/x64/src/testing/testing.go:1792 +0x226
created by testing.(*T).Run in goroutine 1
[...]

PR #6332, opened 2025-09-25
Fix flaky test for the CI  

manute:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

**Description of change**
Refactored the tests inside table driven tests for better readability and added on each test run a possible timeout to not panic, esp sometimes when running inside the CI and takes more time. 

**Which issue this PR fixes**
https://github.com/spiffe/spire/issues/6313


rturner3:
Hi @manute, could you add a description in the PR of how this change fixes the flakiness of the affected test? On first glance, I could not really pick out any differences in behavior in the test with these changes other than how the subtests are defined and orchestrated.

rturner3:
The approach introduced here adds a timeout in the test execution. This will just cause the test to fail faster, but not affect whether or not the test passes. I don't think this will address the underlying source of flakiness of the test.

manute:
> The approach introduced here adds a timeout in the test execution. This will just cause the test to fail faster, but not affect whether or not the test passes. I don't think this will address the underlying source of flakiness of the test.

Thanks for the feedback. I'm kind lost in the error, as locally it doesn't panic so it is hard fix without knowing the proper error. Do you think the issue is more in the sense of how the test is doing the logic or with this approach it would be fixed?  

```go
for _, tt := range tests {
	hook.Reset()
	if tt.setup != nil {
		tt.setup()
	}
	if tt.advance > 0 {
		clockMock.Add(tt.advance)
	}

	// Wait for state to converge
        require.Eventually(t, func() bool {
		return reflect.DeepEqual(c.getStatuses(), tt.expectState())
	}, time.Second, 10*time.Millisecond)

	// Wait until we have *at least* as many logs as expected
	require.Eventually(t, func() bool {
		return len(hook.AllEntries()) >= len(tt.expectLogs)
	}, time.Second, 10*time.Millisecond)

	// log assertion
	spiretest.AssertLogsContainEntries(t, hook.AllEntries(), tt.expectLogs)
```

rturner3:
> > The approach introduced here adds a timeout in the test execution. This will just cause the test to fail faster, but not affect whether or not the test passes. I don't think this will address the underlying source of flakiness of the test.
> 
> Thanks for the feedback. I'm kind lost in the error, as locally it doesn't panic so it is hard fix without knowing the proper error. Do you think the issue is more in the sense of how the test is doing the logic or with this approach it would be fixed?
> 
> ```go
> for _, tt := range tests {
>   t.Run(tt.name, func(t *testing.T) {
> 	hook.Reset()
> 	if tt.setup != nil {
> 		tt.setup()
> 	}
> 	if tt.advance > 0 {
> 		clockMock.Add(tt.advance)
> 	}
> 
> 	// Wait for state to converge
>         require.Eventually(t, func() bool {
> 		return reflect.DeepEqual(c.getStatuses(), tt.expectState())
> 	}, time.Second, 10*time.Millisecond)
> 
> 	// Wait until we have *at least* as many logs as expected
> 	require.Eventually(t, func() bool {
> 		return len(hook.AllEntries()) >= len(tt.expectLogs)
> 	}, time.Second, 10*time.Millisecond)
> 
> 	// log assertion
> 	spiretest.AssertLogsContainEntries(t, hook.AllEntries(), tt.expectLogs)
> ```

Sleeps/waits in tests are usually a recipe for flakiness, since you never really know exactly how long a test is going to take to execute, especially in resource-constrained environments like on GitHub Actions.

Based on the test log failures in https://github.com/spiffe/spire/issues/6313, it appears like the test is failing due to being stuck waiting for a signal over the channel shared between the test and a goroutine that is spawned in the health checker: https://github.com/spiffe/spire/blob/bf235c1b0e601920739d71a3ad1acc5b0829f8cb/pkg/common/health/cache.go#L182-L199

I suspect there may be some subtle issue with how that channel is managed that is resulting in this test nondeterminism.

sorindumitru:
Closing in favour of https://github.com/spiffe/spire/pull/6403

PR #6355, opened 2025-10-05
Validate server cli

### LINKED ISSUE #5268: Rework spire-server validate command to launch plugin instances for validation.

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Description of change**
`spire-server validate` now shows errors reported while validating the server's plugin configuration. Example output:
```
SPIRE server configuration file is invalid.
Validation errors:
        x509pop:
                one of ca_bundle_path or ca_bundle_paths must be configured
        disk:
                'cert_file_path' and 'key_file_path' must be set and not empty
```

**Which issue this PR fixes**
fixes #5268


cmd/spire-server/cli/validate/validate.go
```diff
@@ -32,11 +39,38 @@ func (c *validateCommand) Synopsis() string {
 }
 
 func (c *validateCommand) Run(args []string) int {
-	if _, err := run.LoadConfig(commandName, args, nil, c.env.Stderr, false); err != nil {
+	config, err := run.LoadConfig(commandName, args, []log.Option{log.WithOutputFile(os.DevNull)}, c.env.Stderr, false)
+
+	if err != nil {
```
MarcosDY:
```suggestion
	config, err := run.LoadConfig(commandName, args, []log.Option{log.WithOutputFile(os.DevNull)}, c.env.Stderr, false)
	if err != nil {
```

cmd/spire-server/cli/validate/validate.go
```diff
@@ -32,11 +39,38 @@ func (c *validateCommand) Synopsis() string {
...
 		// Ignore error since a failure to write to stderr cannot very well be reported
 		_ = c.env.ErrPrintf("SPIRE server configuration file is invalid: %v\n", err)
 		return 1
 	}
-	_ = c.env.Println("SPIRE server configuration file is valid.")
-	return 0
+
+	s := server.New(*config)
+
+	pluginNotes, err := s.ValidateConfig(context.Background())
```
MarcosDY: NIT: may we create a context with timeout to be used here?

cmd/spire-server/cli/validate/validate.go
```diff
@@ -32,11 +39,38 @@ func (c *validateCommand) Synopsis() string {
...
+	if err != nil && status.Code(err) != codes.Unimplemented {
+		_ = c.env.ErrPrintf("Could not validate configuration file: %v", err)
+		return 1
+	}
+
+	if len(pluginNotes) == 0 {
```
MarcosDY: it may be a little confusing, here we are:
1-> error if error exists and retun 1
2-> if no error and notes success retunr 0
3-> if notes > 0 print errors  return 1

What do you think about move 3 to 2, so we have success at the end, 
or even create a function so we can reduce the amount of return 1

cmd/spire-server/cli/validate/validate_posix_test.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+plugins {
+` + pluginConfig + `
+}
+`
+}
+
+func TestValidateConfig(t *testing.T) {
+	configDir := t.TempDir()
```
MarcosDY: this is going to create a single folder for all tests, that is going to prevent us to run tests in parallel, can you move this one to be inside each testcase?

cmd/spire-server/cli/validate/validate_posix_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+	testCases := []struct {
+		name                   string
+		config                 string
+		expectedReturn         int
```
MarcosDY: looks like we can calculate this depending if error is expected right?

cmd/spire-server/cli/validate/validate_posix_test.go
```diff
@@ -0,0 +1,259 @@
...
+}
+KeyManager "memory" {
+    plugin_data = {}
+}
+
+UpstreamAuthority "disk" {
+	plugin_data = {
+		cert_file_path = "some/file/some/where"
+	}
+}
```
MarcosDY, line 124: It is a more general ocmment, 
in this approach we are going to depends on a plugin to fail in order for this one to fail, 
it soiunds like we are testing another function but not CLI,
May we fake this to have more control and not be affected by another funcion changes?

pkg/common/catalog/catalog.go
```diff
@@ -215,6 +218,101 @@ func Load(ctx context.Context, config Config, repo Repository) (_ *Catalog, err
 	}, nil
 }
 
+func ValidatePluginConfigs(ctx context.Context, config Config, repo Repository) (pluginNotes map[string][]string, err error) {
+	closers := make(closerGroup, 0)
+	defer func() {
+		// If loading fails, clear out the catalog and close down all plugins
+		// that have been loaded thus far.
+		if err != nil {
```
MarcosDY, line 225: can not this be solved with the context being canceled?

pkg/common/catalog/catalog.go
```diff
@@ -215,6 +218,101 @@ func Load(ctx context.Context, config Config, repo Repository) (_ *Catalog, err
...
+	serviceRepos, err := makeBindableServiceRepos(repo.Services())
+	if err != nil {
+		return nil, err
+	}
+
+	pluginCounts := make(map[string]int)
+
+	pluginNotes = make(map[string][]string)
+	for _, pluginConfig := range config.PluginConfigs {
+		pluginLog, _ := test.NewNullLogger()
```
MarcosDY: this is imos conceptual, but is can we have somehting else here? using a test package for final code, does not looks fine

pkg/common/catalog/catalog.go
```diff
@@ -215,6 +218,101 @@ func Load(ctx context.Context, config Config, repo Repository) (_ *Catalog, err
...
+		pluginRepo, ok := pluginRepos[pluginConfig.Type]
```
MarcosDY, line 249: this is going to retrurn a partial value, maybe another plugins are wrong too, may we add this as an entry of configuration issues?

pkg/common/catalog/catalog.go
```diff
@@ -215,6 +218,101 @@ func Load(ctx context.Context, config Config, repo Repository) (_ *Catalog, err
...
+
+func GetPluginConfigString(c PluginConfig) (string, error) {
+	var dataSource DataSource
+	if c.DataSource == nil {
+		dataSource = FixedData("")
+	} else {
+		dataSource = c.DataSource
+	}
+
+	return dataSource.Load()
```
MarcosDY: we may simplify this way
```suggestion
    	if c.DataSource == nil {
    	   return FixedData("").Load()
    	}
    	
    	return c.DataSource.Load()
```

pkg/server/catalog/catalog.go
```diff
@@ -180,6 +195,64 @@ func Load(ctx context.Context, config Config) (_ *Repository, err error) {
 	return repo, nil
 }
 
+func ValidateConfig(ctx context.Context, config Config) (pluginNotes map[string][]string, err error) {
+	if c, ok := config.PluginConfigs.Find(nodeAttestorType, jointoken.PluginName); ok && c.IsEnabled() && c.IsExternal() {
+		return nil, errors.New("the built-in join_token node attestor cannot be overridden by an external plugin")
+	}
+
+	log, _ := test.NewNullLogger()
```
MarcosDY: can we replace this with no test code?

pkg/server/catalog/catalog.go
```diff
@@ -180,6 +195,64 @@ func Load(ctx context.Context, config Config) (_ *Repository, err error) {
...
+
+		ds := ds_sql.New(log)
+		resp, err := ds.Validate(ctx, coreConfig, dsConfigString)
+		if resp != nil && len(resp.Notes) != 0 {
+			pluginNotes[datastorePluginId] = append(pluginNotes[datastorePluginId], resp.Notes...)
+		}
+
+		if err != nil {
+			pluginNotes[datastorePluginId] = append(pluginNotes[datastorePluginId], err.Error())
+		}
```
MarcosDY: NIT: keep error just affter it is emmited 
```suggestion
	    if err != nil {
			pluginNotes[datastorePluginId] = append(pluginNotes[datastorePluginId], err.Error())
		}
		
		if resp != nil && len(resp.Notes) != 0 {
			pluginNotes[datastorePluginId] = append(pluginNotes[datastorePluginId], resp.Notes...)
		}
```

pkg/server/server.go
```diff
@@ -70,6 +71,18 @@ func (s *Server) Run(ctx context.Context) error {
 	return nil
 }
 
+func (s *Server) ValidateConfig(ctx context.Context) (map[string][]string, error) {
+	return catalog.ValidateConfig(ctx, catalog.Config{
+		Log:              s.config.Log.WithField(telemetry.SubsystemName, telemetry.Catalog),
+		Metrics:          fakemetrics.New(),
```
MarcosDY: are this metirics required?

sorindumitru: No, I'll switch it to the blackhole metric sink.

pkg/server/catalog/catalog.go
sorindumitru: Yes, we can.  We already set `config.Log` to log to `/dev/null` so we don't actually need this test logger. I've also switched it to use `io.Discard` which should work in more places.

cmd/spire-server/cli/validate/validate_posix_test.go
sorindumitru: It's a good point, I'll see what can be done. I don't think we can do that here, but maybe it can be done lower down in the stack.

sorindumitru: Looks like we do the same thing in the [catalog tests](https://github.com/spiffe/spire/blob/9e5e3ab9e453edbab5e690ca554aba0dc274386a/pkg/server/catalog/catalog_test.go#L56). I can move these tests there, but it's still going to use some of the built-in plugins. I think I chose some that are unlikely to cause much breakage over time.

Let me know if you still want me to look into this some more.

pkg/common/catalog/catalog.go
sorindumitru: I'm not sure, do you have something in mind for this? I think this would happen even if the context is valid.

pkg/common/catalog/catalog.go
sorindumitru: Looks like it was reporting this kind of error already, added a test for it.

pkg/common/catalog/catalog.go
```diff
@@ -215,6 +217,97 @@ func Load(ctx context.Context, config Config, repo Repository) (_ *Catalog, err
...
+		pluginRepo, ok := pluginRepos[pluginConfig.Type]
+		if !ok {
+			return nil, fmt.Errorf("unsupported plugin type %q", pluginConfig.Type)
+		}
+
+		if pluginConfig.Disabled {
+			continue
+		}
+
+		pluginNotesId := fmt.Sprintf("%s \"%s\"", pluginConfig.Type, pluginConfig.Name)
```
MarcosDY: NIT: 
```suggestion
		pluginNotesID := fmt.Sprintf("%s %q", pluginConfig.Type, pluginConfig.Name)
```

pkg/server/catalog/catalog.go
```diff
@@ -202,11 +273,8 @@ func loadSQLDataStore(ctx context.Context, config Config, coreConfig catalog.Cor
 
 	dsLog := config.Log.WithField(telemetry.SubsystemName, sqlConfig.Name)
 	ds := ds_sql.New(dsLog)
-	configurer := catalog.ConfigurerFunc(func(ctx context.Context, _ catalog.CoreConfig, configuration string) error {
-		return ds.Configure(ctx, configuration)
-	})
-
-	if _, err := catalog.ConfigurePlugin(ctx, coreConfig, configurer, sqlConfig.DataSource, ""); err != nil {
+	dsConfigurer := &dsConfigurer{ds: ds}
```
MarcosDY: NIT:  to avoid using the struct name as variable can you update var name?
```suggestion
	dsConf := &dsConfigurer{ds: ds}
```

MarcosDY (approved):

Issue #6129, opened 2025-06-13
Flag for jwt preference entry


kfox1111:
At the entry level, it would be good if you could mark the entry likely to be JWT only. This way, SPIRE can prevent spending effort precreating x509 certs for workloads that wont use them.

tjons:
@kfox1111 I'd be happy to work on this one

sorindumitru:
This sounds like a good improvement. We need to figure out how to best represent this in the datastore; we'll discuss this some more and come back with a proposal.

amartinezfayo:
@kfox1111 @tjons If a workload tries to fetch an X509-SVID for an entry that is marked as JWT-only, would you expect the call to fail or have the agent requesting the SVID from the server (since it's not in the cache)? If the agent requests the X509-SVID to the server, would you also expect that the agent keeps it updated in the cache (handling rotation)?

kfox1111:
That is a good question....

Depends on if its a preference flag, or a hard 'no' flag.

if it was an enum, it could have more complicated logic like, "nevercert" which could return no and "nocache" which is don't ever prebuild a cert, its rare to ever ask for a cert.

a middle ground might be:
* dont preallocate certs. if one is asked for, give it. if the connection is left open, cache and refresh it. if the connection is closed, its unlikely they will ask again and just drop it from the cache and dont refresh?

I'm ok with just a hard no too. That is probably easier.

amartinezfayo:
Thank you, @kfox1111, for the feedback.
I think the middle-ground alternative is reasonable and aligns with what we’ve discussed within the maintainers group.

As an implementation detail, I suggest adding a field to the registered_entries table to store a data blob containing an encoded protobuf message where we can specify this and other flags or options. This approach would allow us to extend the available options in the future without requiring schema changes (though future changes would still need to maintain backward compatibility).

rturner3:
If someone is willing to take this up, it would be great to first work on a draft PR with the main logic changes since there may be some more complexity in the implementation.

@tjons are you still interested in working on this?

ValFadeev:
I am happy to have a go at this one, unless work has already started @tjons  @kfox1111 @rturner3

ValFadeev:
It's taken me a while, this being my first pass at updating the datastore. As suggested here, I have opened a draft PR to invite initial feedback. Thank for you for your patience.

Issue #6069, opened 2025-05-14
Entry labels


kfox1111:
I would like to be able to add labels (arbitrary key/value pairs) onto entries. They would be stored in the db along with the entry, and otherwise would be normally ignored by spire.

Kubernetes popularized this kind of thing and has proven quite useful over time.

One particular use case of interest would be to be able to use them in conjunction with custom CredentialComposer plugins. For example, when using JWTs with Minio, the JWT needs a custom "policy" property. (https://github.com/spiffe/club-zero/issues/7). An entry could have a label like 'minio.io/policy=readwrite' added to the entry, and the MinioCredentialComposer could use the saved labels to affect the generated JWTs.

kfox1111:
This would be much cleaner then having to add the needed information into the spiffeid, then scrape it back out to add the property. For example:
https://github.com/spiffe/spire-credentialcomposer-cel?tab=readme-ov-file#minio

Also, there currently isn't a way not to have that info still in the resulting spiffeid as a credential composer cant rewrite the subject.

sorindumitru:
Thanks @kfox1111 for opening this issue. This has been considered in the past, for example https://github.com/spiffe/spire/issues/1336. At that point we've decided not to added them, because:
* We didn't have a lot of good examples of how they could be used.
* They might not be the right abstraction to solve the issue.

I think the minio example is possibly a good example for making use of the metadata, but I'm not sure it's the right abstraction to use:
* It assumes whoever configured the Minio policies has access to configure entries (and have a way to deal with conflicting changes). In some ways it mixes authentication with authorization.
* It forces thinking the Minio policy configuration in terms of workload selectors and even more of that in terms of the `ParentID` relationships in the registration entries.

I think an ideal way of configuring these  would be driven by the SPIFFE ID, i.e. this SPIFFE ID has this kind of access to these buckets. Some ways of dealing with this would be:
* Credential exchange. The workload can swap a SVID for a MinIO specific token, which contains the policies the workload has acess to.
* A CredentialComposer plugin that fetches the list of MinIO policies based on the SPIFFE ID from some other service that is used to definte these policies.

kfox1111:
> Thanks [@kfox1111](https://github.com/kfox1111) for opening this issue. This has been considered in the past, for example [#1336](https://github.com/spiffe/spire/issues/1336). At that point we've decided not to added them, because:
> 
> * We didn't have a lot of good examples of how they could be used.
> * They might not be the right abstraction to solve the issue.
> 
[...]

PR #6360, opened 2025-10-06
[WIP] - x509 cache hint for registration entries

ValFadeev:
This is a draft PR aiming to implement some of the suggestions discussed in https://github.com/spiffe/spire/issues/6129. It adds a new field to `RegisteredEntry` represented by an opaque protobuf-encoded message. The message is meant to contain flags and other fields guiding the agent's behavior with respect to pre-fetching and caching of X509 SVIDs. The intention is to reduce unnecessary work performed by the agent, in case the client is unlikely to ever ask for an X509 SVID and prefers JWT instead. For now only one boolean flag `JWTOnly` is introduced. Other fields may be added later to fine-tune the logic and cover various edge-cases. 
The bulk of this draft is to define the new field in the message definition, update the datastore and have the existing tests pass. The actual usage of the field in `fetchEntries` is purely indicative and requires further consideration.

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

* adds a new field to `RegisteredEntry` and the underlying datastore table
* allows flagging registration entries as preferring JWT SVID over X509
**Description of change**

**Which issue this PR fixes**

Fixes https://github.com/spiffe/spire/issues/6129

### DIFF
--- a/pkg/agent/manager/sync.go
+++ b/pkg/agent/manager/sync.go
@@ -345,7 +345,9 @@ func (m *manager) fetchEntries(ctx context.Context) (_ *cache.UpdateEntries, _ *
 		case entry.StoreSvid:
 			storeEntries[entryID] = entry
 		default:
-			cacheEntries[entryID] = entry
+			if !entry.X509SvidCacheHint.GetJwtOnly() {
+				cacheEntries[entryID] = entry
+			}
 		}
 	}
 
--- a/pkg/server/datastore/sqlstore/migration.go
+++ b/pkg/server/datastore/sqlstore/migration.go
@@ -267,12 +267,12 @@ import (
 
 const (
 	// the latest schema version of the database in the code
-	latestSchemaVersion = 23
+	latestSchemaVersion = 24
 
 	// lastMinorReleaseSchemaVersion is the schema version supported by the
 	// last minor release. When the migrations are opportunistically pruned
 	// from the code after a minor release, this number should be updated.
-	lastMinorReleaseSchemaVersion = 23
+	lastMinorReleaseSchemaVersion = 24
 )
 
 // the current code version
@@ -498,6 +498,8 @@ func migrateVersion(tx *gorm.DB, currVersion int, log logrus.FieldLogger) (versi
 	// }
 	//
 	switch currVersion { //nolint: gocritic,revive // No upgrade required yet, keeping switch for future additions
+	case 24:
+		err = migrateToV24(tx)
 	default:
 		err = newSQLError("no migration support for unknown schema version %d", currVersion)
 	}
@@ -506,6 +508,13 @@ func migrateVersion(tx *gorm.DB, currVersion int, log logrus.FieldLogger) (versi
 	}
 
 	return nextVersion, nil
+}
+
+func migrateToV24(tx *gorm.DB) error {
+	if err := tx.AutoMigrate(&RegisteredEntry{}).Error; err != nil {
+		return newWrappedSQLError(err)
+	}
+	return nil
 }
 
 func addFederatedRegistrationEntriesRegisteredEntryIDIndex(tx *gorm.DB) error {
--- a/pkg/server/datastore/sqlstore/models.go
+++ b/pkg/server/datastore/sqlstore/models.go
@@ -112,6 +112,10 @@ type RegisteredEntry struct {
 
 	// TTL of JWT identities derived from this entry
 	JWTSvidTTL int32 `gorm:"column:jwt_svid_ttl"`
+
+	// X509SvidCacheHint contains a set of options and flags that inform the
+	// agent behaviour with respect to pre-fetching and refreshing X509 SVIDs
+	X509SvidCacheHint []byte `gorm:"size:255,column:x509_svid_cache_hint"` // corresponds to TINYBLOB in MySQL
 }
 
 // RegisteredEntryEvent holds the entry id of a registered entry that had an event
--- a/pkg/server/datastore/sqlstore/sqlstore.go
+++ b/pkg/server/datastore/sqlstore/sqlstore.go
@@ -2504,17 +2504,23 @@ func createRegistrationEntry(tx *gorm.DB, entry *common.RegistrationEntry) (*com
 		return nil, err
 	}
 
+	X509SvidCacheHint, err := proto.Marshal(entry.X509SvidCacheHint)
+	if err != nil {
+		return nil, err
+	}
+
 	newRegisteredEntry := RegisteredEntry{
-		EntryID:    entryID,
-		SpiffeID:   entry.SpiffeId,
-		ParentID:   entry.ParentId,
-		TTL:        entry.X509SvidTtl,
-		Admin:      entry.Admin,
-		Downstream: entry.Downstream,
-		Expiry:     entry.EntryExpiry,
-		StoreSvid:  entry.StoreSvid,
-		JWTSvidTTL: entry.JwtSvidTtl,
-		Hint:       entry.Hint,
+		EntryID:           entryID,
+		SpiffeID:          entry.SpiffeId,
+		ParentID:          entry.ParentId,
+		TTL:               entry.X509SvidTtl,
+		Admin:             entry.Admin,
+		Downstream:        entry.Downstream,
+		Expiry:            entry.EntryExpiry,
+		StoreSvid:         entry.StoreSvid,
+		JWTSvidTTL:        entry.JwtSvidTtl,
+		Hint:              entry.Hint,
+		X509SvidCacheHint: X509SvidCacheHint,
 	}
 
 	if err := tx.Create(&newRegisteredEntry).Error; err != nil {
@@ -2629,7 +2635,8 @@ SELECT
 	NULL AS dns_name_id,
 	NULL AS dns_name,
 	revision_number,
-	jwt_svid_ttl AS reg_jwt_svid_ttl
+	jwt_svid_ttl AS reg_jwt_svid_ttl,
+	x509_svid_cache_hint
 FROM
 	registered_entries
 WHERE id IN (SELECT id FROM listing)
@@ -2637,7 +2644,7 @@ WHERE id IN (SELECT id FROM listing)
 UNION
 
 SELECT
-	F.registered_entry_id, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, B.trust_domain, NULL, NULL, NULL, NULL
+	F.registered_entry_id, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, B.trust_domain, NULL, NULL, NULL, NULL, NULL
 FROM
 	bundles B
 INNER JOIN
@@ -2650,7 +2657,7 @@ WHERE
 UNION
 
 SELECT
-	registered_entry_id, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, id, value, NULL, NULL
+	registered_entry_id, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, id, value, NULL, NULL, NULL
 FROM
 	dns_names
 WHERE registered_entry_id IN (SELECT id FROM listing)
@@ -2658,7 +2665,7 @@ WHERE registered_entry_id IN (SELECT id FROM listing)
 UNION
 
 SELECT
-	registered_entry_id, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, id, type, value, NULL, NULL, NULL, NULL, NULL
+	registered_entry_id, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, id, type, value, NULL, NULL, NULL, NULL, NULL, NULL
 FROM
 	selectors
 WHERE registered_entry_id IN (SELECT id FROM listing)
@@ -2693,7 +2700,8 @@ SELECT
 	NULL ::integer AS dns_name_id,
 	NULL AS dns_name,
 	revision_number,
-	jwt_svid_ttl AS reg_jwt_svid_ttl
+	jwt_svid_ttl AS reg_jwt_svid_ttl,
+	x509_svid_cache_hint
 FROM
 	registered_entries
 WHERE id IN (SELECT id FROM listing)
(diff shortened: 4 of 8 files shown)


proto/spire/common/common.proto
```diff
@@ -96,6 +96,14 @@ message RegistrationEntry {
...
     /** Time of creation, in seconds from epoch */
     int64 created_at = 15;
+    /** Represents a set of options informing the agent behaviour with respect to
+    pre-fecthing and caching X509 SVIDs. This is meant to prevent unnecessary effort
+    spent on providing X509 SVIDs to workloads, which are likely to ask only for JWT SVID. */
+    message X509SvidCacheHint {
+        /** Flag indicating whether the workload is likely to require only JWT SVID. */
+        bool jwt_only = 1;
+    }
+    X509SvidCacheHint x509_svid_cache_hint = 16;
```
sorindumitru: We discussed this today and we think this would be more useful if it was a generic flag field that we can use in the future for non x509-svid cache flags.

For the flag itself, we think `disable_x509_svid_prefetch` would be better since it allows us to specify more easily which SVID types should be prefetched. This might come in handy if we get a new SVID type, for example WIT-SVID.

amartinezfayo:
Hi @ValFadeev, are you planning to go back to this PR?

ValFadeev:
Hi @amartinezfayo, apologies for the delay. I had a few conflicting priorities lately, but I do plan to pick this back up over the coming week.

ValFadeev:
@amartinezfayo offering this for another review. I am not sure I implemented the semantics of an optional field the right way when [loading from row](https://github.com/spiffe/spire/pull/6360/files#diff-a972d23257a60257623830219f33a521a3c95b8223c6e3f80d840d699aa81d3dR3963). I also understand this change may have to be split in two, with the schema migration going first. But hopefully, this gives an outline of the functionality, as it was last described.

pkg/agent/manager/sync.go
```diff
@@ -345,7 +345,9 @@ func (m *manager) fetchEntries(ctx context.Context) (_ *cache.UpdateEntries, _ *
 		case entry.StoreSvid:
 			storeEntries[entryID] = entry
 		default:
-			cacheEntries[entryID] = entry
+			if !entry.CacheHintFlags.GetDisableX509SvidPrefetch() {
+				cacheEntries[entryID] = entry
+			}
```
sorindumitru: This would mean that the entry is not added to the cache, which means that no workload would be able to request a SVID for it. I think the flag should end up in the lru cache where it can inform if the entry needs to prefetched or not but still allow a workload to request an x509-svid if it so desires.

I think [syncSVIDsWithSubscribers](https://github.com/spiffe/spire/blob/5512c9bbb7f167b0e56d0b6fbfb2cbd2e69ecebb/pkg/agent/manager/cache/lru_cache.go#L781) might be the place to do this.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -3944,6 +3960,15 @@ func fillEntryFromRow(entry *common.RegistrationEntry, r *entryRow) error {
 		entry.CreatedAt = roundedInSecondsUnix(r.CreatedAt.Time)
 	}
 
+	if r.CacheHintFlags.Valid {
+		if len(r.CacheHintFlags.V) > 0 {
+			entry.CacheHintFlags = &common.RegistrationEntry_CacheHintFlags{}
+			if err := proto.Unmarshal(r.CacheHintFlags.V, entry.CacheHintFlags); err != nil {
+				return newSQLError("invalid value for X.509 cache hint: %s", err)
```
sorindumitru:
```suggestion
				return newSQLError("could not parse cache hint flags: %s", err)
```

ValFadeev:
@srikalyan mind if I bring this up to date and actually add the proposed logic making use of the flag, with the deadline by the end of the week. Then we can see how the work can be potentially reused/generalised. Thanks.

srikalyan:
@ValFadeev please go ahead and let me know if I can help in anyway.

sorindumitru:
I think you'll also need to make changes to the Entry type in [spire-api-sdk](https://github.com/spiffe/spire-api-sdk/blob/main/proto/spire/api/types/entry.proto) for the agent to get and be able to use the new field.

proto/spire/common/common.proto
```diff
@@ -99,6 +99,15 @@ message RegistrationEntry {
     string hint = 14;
     /** Time of creation, in seconds from epoch */
     int64 created_at = 15;
+    /** Represents a set of flags informing the agent behaviour with respect to
+    pre-fecthing and caching SVIDs. This is meant to prevent unnecessary effort
+    spent on generating SVIDs of types, which are unlikely to be needed. */
+    message CacheHintFlags {
```
sorindumitru: Since we're likely going to use this even for non-flag fields, could we call it something else? Nothing better than `OptionalFields` comes to me right now, let me know if you have other suggestions.

pkg/server/datastore/sqlstore/models.go
```diff
@@ -100,6 +100,10 @@ type RegisteredEntry struct {
 
 	// TTL of JWT identities derived from this entry
 	JWTSvidTTL int32 `gorm:"column:jwt_svid_ttl"`
+
+	// CacheHintFlags contains a set of options and flags that inform the
+	// agent behaviour with respect to pre-fetching and refreshing X509 SVIDs
+	CacheHintFlags []byte `gorm:"size:255,column:cache_hint_flags"` // corresponds to TINYBLOB in MySQL
```
sorindumitru: Since we may end up having other bigger fields in there, maybe we can go one size up to have some more space? I think that would be `size:65535`

proto/spire/common/common.proto
ValFadeev: I would say something like AdditionalAttributes. This is to distinguish them from optional fields in the containing struct, such as Expiry.

pkg/server/datastore/sqlstore/migration.go
```diff
@@ -517,8 +517,7 @@ func migrateVersion(tx *gorm.DB, currVersion int, log logrus.FieldLogger) (versi
 }
 
 func migrateToV24(tx *gorm.DB) error {
-	// Add agent_version column to attested_node_entries table
-	if err := tx.AutoMigrate(&AttestedNode{}).Error; err != nil {
+	if err := tx.AutoMigrate(&AttestedNode{}, &RegisteredEntry{}).Error; err != nil {
```
ValFadeev: Since there is currently no released version between my changes and https://github.com/spiffe/spire/pull/6542, I am temporarily combining migrations here to have tests pass. Once 1.14.2 is released, I will only leave my table and increment the version.

sorindumitru: Usually we seemed to have created independent migrations for different data store changes. I think they're not strictly speaking needed, but they might be useful for the case of local development.

sorindumitru (commented):
Thanks for these changes, @ValFadeev. Are you also planning on adding support for this to the CLI?

cmd/spire-server/cli/entry/create.go
```diff
@@ -209,8 +214,12 @@ func (c *createCommand) parseConfig() ([]*types.Entry, error) {
 
 		selectors = append(selectors, cs)
 	}
-
 	e.Selectors = selectors
+
+	additionalAttributes := &types.Entry_AdditionalAttributes{}
+	additionalAttributes.DisableX509SvidPrefetch = c.disableX509SVIDPrefetch
+	e.AdditionalAttributes = additionalAttributes
```
sorindumitru: Could we only create the additional attributes if one of the attributes in it is defined? Would avoid having to parse the whole field.

pkg/server/api/entry.go
```diff
@@ -179,23 +187,35 @@ func RegistrationEntryToProto(e *common.RegistrationEntry) (*types.Entry, error)
...
+	return entry, nil
+}
+
+func ProtoFromAdditionalAttributes(in *common.RegistrationEntry_AdditionalAttributes) *types.Entry_AdditionalAttributes {
+	if in != nil {
+		return &types.Entry_AdditionalAttributes{
+			DisableX509SvidPrefetch: in.DisableX509SvidPrefetch,
+		}
+	}
+	return nil
```
sorindumitru, line 222: I think we would also need a similar item for doing the reverse, and it should be used in `ProtoToRegistrationEntryWithMask`.

pkg/server/datastore/sqlstore/migration.go
sorindumitru: This still needs to be addressed.

pkg/server/datastore/sqlstore/migration.go
```diff
@@ -506,6 +506,8 @@ func migrateVersion(tx *gorm.DB, currVersion int, log logrus.FieldLogger) (versi
 	switch currVersion {
 	case 23:
 		err = migrateToV24(tx)
```
sorindumitru, line 511: We'll need to do both of them if we're on 23, so maybe:
```suggestion
		err = migrateToV24(tx)
		fallthrough
```

sorindumitru (commented):
Didn't get a chance to look through all of it yet, so there may be more comments coming.

pkg/server/api/entry.go
```diff
@@ -179,23 +187,49 @@ func RegistrationEntryToProto(e *common.RegistrationEntry) (*types.Entry, error)
...
+		Admin:                e.Admin,
+		Downstream:           e.Downstream,
+		ExpiresAt:            e.EntryExpiry,
+		DnsNames:             slices.Clone(e.DnsNames),
+		RevisionNumber:       e.RevisionNumber,
+		StoreSvid:            e.StoreSvid,
+		JwtSvidTtl:           e.JwtSvidTtl,
+		Hint:                 e.Hint,
+		CreatedAt:            e.CreatedAt,
+		AdditionalAttributes: ProtoFromAdditionalAttributes(e.AdditionalAttributes),
```
sorindumitru: Looks like we set the `AdditionalAttributes` twice. Once here and once in a few lines down.

sorindumitru (commented):
Thanks @ValFadeev, this is shaping up nicely. Only one small comment.

sorindumitru:
Hi @ValFadeev , sorry for the delays here. We've merged the API changes, would you mind repointing the dependency to the `github.com/spiffe/spire-api-sdk@next`?

cmd/spire-server/cli/entry/update.go
```diff
@@ -204,6 +209,13 @@ func (c *updateCommand) parseConfig() ([]*types.Entry, error) {
 	}
 
 	e.Selectors = selectors
+
+	if c.disableX509SVIDPrefetch {
```
sorindumitru, line 225: I think this makes it impossible to remove the flag once set. We need a way to determine if the flag was set or not. Maybe we can use BoolFunc so that we know if it was set or not? We could even have a more general bool "setAdditionalAttributes" which is set to true if the BoolFunc is called?

pkg/server/datastore/sqlstore/models.go
```diff
@@ -100,6 +100,11 @@ type RegisteredEntry struct {
 
 	// TTL of JWT identities derived from this entry
 	JWTSvidTTL int32 `gorm:"column:jwt_svid_ttl"`
+
+	// AdditionalAttributes may contain a number of optional fields controlling
+	// the various aspects of the agent's behaviour with respect to a given
+	// registration entry
+	AdditionalAttributes []byte `gorm:"size:65535,column:additional_attributes"`
```
sorindumitru, line 107: Maybe we could also enforce this in the data store. It would at least make for better error messages, if nothing else.

cmd/spire-server/cli/entry/update.go
ValFadeev: That is not what I observe - this is run on a clean build of the server:
```
./bin/spire-server entry show -entryID prefetch-disabled -socketPath=/tmp/spire-server/private/api.sock
Found 1 entry
Entry ID                : prefetch-disabled
SPIFFE ID               : spiffe://example.org/workload-with-prefetch-disabled
Parent ID               : spiffe://example.org/cluster/test
Revision                : 6
X509-SVID TTL           : default
JWT-SVID TTL            : default
Selector                : unix:uid:1004
DisableX509SvidPrefetch : true
```
Now if an update performed without the flag set it is cleared in the entry:
```
./bin/spire-server entry update -entryID prefetch-disabled -selector "unix:uid:1004" -parentID spiffe://example.org/cluster/test -spiffeID spiffe://example.org/workload-with-prefetch-disabled -socketPath=/tmp/spire-server/private/api.sock
Entry ID                : prefetch-disabled
SPIFFE ID               : spiffe://example.org/workload-with-prefetch-disabled
Parent ID               : spiffe://example.org/cluster/test
Revision                : 7
X509-SVID TTL           : default
JWT-SVID TTL            : default
Selector                : unix:uid:1004
```
and set again:
```
./bin/spire-server entry update -entryID prefetch-disabled -selector "unix:uid:1004" -parentID spiffe://example.org/cluster/test -spiffeID spiffe://example.org/workload-with-prefetch-disabled -disableX509SVIDPrefetch -socketPath=/tmp/spire-server/private/api.sock
Entry ID                : prefetch-disabled
SPIFFE ID               : spiffe://example.org/workload-with-prefetch-disabled
Parent ID               : spiffe://example.org/cluster/test
Revision                : 8
X509-SVID TTL           : default
JWT-SVID TTL            : default
Selector                : unix:uid:1004
DisableX509SvidPrefetch : true
```
Tha said, it could indeed be useful to introduce an overarching bool, such as "setAdditionalAttributes", to indicate whether any one of the additional attributes has been provided, to set it up for generalisation. I suppose the flag definition could similar to the below one (for every new additional attribute implemented"
```go
	f.BoolFunc("disableX509SVIDPrefetch", "A boolean value that, when set, disables prefetching X509 SVID for this entry", 
	func(_ string) error {
		c.additionalAttributesSet = true
		c.disableX509SVIDPrefetch = true
		return nil
	})
```
So that the entry field would be populated like this
```go
	if c.additionalAttributesSet {
		e.AdditionalAttributes = &types.Entry_AdditionalAttributes{}
	}

	if c.disableX509SVIDPrefetch {
		e.AdditionalAttributes.DisableX509SvidPrefetch = true
	}
```
If this makes sense, I can make the adjustment.

pkg/server/datastore/sqlstore/migration.go
```diff
@@ -509,6 +509,12 @@ func migrateVersion(tx *gorm.DB, currVersion int, log logrus.FieldLogger) (versi
...
+		if err != nil {
+			break
+		}
+		fallthrough
```
sorindumitru: Since version 24 was release in 1.14.3/1.14.4, we no longer need to fallthrough. Sorry about that, should have realised that would be the case when I asked you to do this.

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -4497,6 +4532,22 @@ func modelToBundle(model *Bundle) (*common.Bundle, error) {
 	return bundle, nil
 }
 
+func validateAditionalAttributes(additionalAttributes *common.RegistrationEntry_AdditionalAttributes) error {
+	if additionalAttributes == nil {
+		return nil
+	}
+
+	marshaledAdditionalAttributes, err := proto.Marshal(additionalAttributes)
```
sorindumitru, line 4540: This looks ok, but it means we marsh the attributes twice on update/creation. Could we instead apply these validations at that point? Maybe we can make this function to also return the bytes.

kfox1111:
I'm curious. would the AdditionalAttributes support in this pr make it easier to add https://github.com/spiffe/spire/issues/6069

PR #6362, opened 2025-10-07
Upgrade to go 1.24.8 and release SPIRE 1.12.6

sorindumitru:
Address a set of CVEs that have been fixed in [go 1.25.2](https://groups.google.com/g/golang-announce/c/4Emdl2iQ_bI?pli=1).


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/spiffe/spire
 
-go 1.24.6
```
amartinezfayo, line 3: We should update to 1.24.8 to prevent jumping to the next minor release in a patch release.

amartinezfayo (approved):

PR #6392, opened 2025-10-23
Add certificate_id field for for GCP CAS Enterprise tier

### LINKED ISSUE #6180: GCP CAS upstreamAuthority plugin does not work with "ENTERPRISE" tier CA pools
Background: GCP Certificate Authority Service has Certificate Authorities grouped into "pools". CAs and the pools they are in share a property called the "tier", which relates to the (expected) management of the issued certificates: the `DEVOPS` tier does not track issued certificates or allow for revoking them, while the `ENTERPRISE` tier does. https://cloud.google.com/certificate-authority-service/docs/tiers

To enable tracking and revoking certificates, certificates issued by an `ENTERPRISE` tier CA must have an additional field set in the certificate request, `certificate_id`:	

> Optional. It must be unique within a location and match the regular expression [a-zA-Z0-9_-]{1,63}. This field is required when using a [CertificateAuthority](https://cloud.google.com/certificate-authority-service/docs/reference/rpc/google.cloud.security.privateca.v1#google.cloud.security.privateca.v1.CertificateAuthority) in the Enterprise [CertificateAuthority.tier](https://cloud.google.com/certificate-authority-?service/docs/reference/rpc/google.cloud.security.privateca.v1#google.cloud.security.privateca.v1.CertificateAuthority.FIELDS.google.cloud.security.privateca.v1.CaPool.Tier.google.cloud.security.privateca.v1.CertificateAuthority.tier), but is optional and its value is ignored otherwise.

https://cloud.google.com/certificate-authority-service/docs/reference/rpc/google.cloud.security.privateca.v1#google.cloud.security.privateca.v1.CreateCertificateRequest

The SPIRE GCP CAS upstreamAuthority does not fill in this field in its certificate request:
https://github.com/spiffe/spire/blob/1ff99fa96d68e3b1ed0d0614fe2dbca9e689bb18/pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go#L288
and thus fails to obtain its intermediate CA certificate when pointed at an `ENTERPRISE`-tier pool:

```
spire-server time="2025-07-08T21:04:59Z" level=info msg="Minting X509 intermediate CA" ca-certificate=projects/sandbox-k8s-458220/locations/us-central1/caPools/dev-ca-pool/certificateAuthorities/dev-ca external=false plugin_name=gcp_cas plugin_type=UpstreamAuthority subsystem_name=catalog ttl=24h0m0s                               
spire-server time="2025-07-08T21:04:59Z" level=error msg="Unable to rotate X509 CA" error="rpc error: code = InvalidArgument desc = upstreamauthority(gcp_cas): `certificate_id` is required when using a CA in the Enterprise tier." subsystem_name=ca_manager                                                                             
```

The GCP CAS plugin should fill in this field, or provide an option to fill in this field, so that it can use existing CA pools that are in the `ENTERPRISE` tier. (It may be worth noting that GCP CAS bills monthly for each pool, so "create another pool" will not always be an option to handle this situation.)

akos011221:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
"gcp_cas" UpstreamAuthority plugin

**Description of change**
Add certificate_id field to CreateCertificateRequest to enable ENTERPRISE tier CA pool compatibility. The certificate_id is generated deterministically from the CSR public key hash for consistent resource naming.

**Which issue this PR fixes**
fixes #6180


pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -284,9 +286,16 @@ func (p *Plugin) mintX509CA(ctx context.Context, csr []byte, preferredTTL int32)
 	// after the path.Split call above.  We need to trim off the /certificateAuthorities/ part for the request below
 	chosenPool = strings.TrimSuffix(chosenPool, "/certificateAuthorities/")
 
+	// CertificateId is a unique identifier required for Enterprise tier CAs. We derive it
+	// from a hash of the public key, ensuring the same certificate request always yields the same resource name.
+	certIDHash := sha256.Sum256(csrParsed.RawSubjectPublicKeyInfo)
+	// Truncate the 32-byte hash to 28 bytes (56 hex characters) to comply with the limit: 7 ("spiffe-") + 56 = 63.
+	certificateID := "spiffe-" + hex.EncodeToString(certIDHash[:28])
```
sorindumitru: I'm not sure why we need this "spiffe-" prefix here. Keeping as much of the hash should make it easier for the id to be unique (although it's likely unique either way).

You could also use base32 instead of hex. I think either way it should be fine, but with base32 the sha256 hash will be encoded to less than 63 chars and will have the expected format. Either way, it might nicer to do the truncation at the end, maybe when adding the ID to the request object? `certificateID[:63]` makes it a bit easier to reason about than the `[:28]`, in my opinion.

akos011221: Thanks for the feedback. I've made changes to the ID generation, it should be simpler now with Base64 RawURLEncoding. It should produce 43 character long strings from the 32 byte hash.

akos011221:
> Would it be possible to update the tests? Maybe the `fakeClient` can verify that a `certificateID` is passed in, has the right format and length, and hasn't been used before?

Added verification for the regex part for the format & length verification. To cover the "has it been used before" part of your request, I could came up with a solution to call the certificate minting multiple times, each with different CSR inside `TestGcpCAS()`.  Added a map to the fakeClient in which each Certificate ID is stored that the client seen so far, and for each certificate creation, the request's Certificate ID is verified against this map.

Also, switched all `require` inside `CreateCertificate` to `assert`, because I've noticed that if the `require` fails, the test will hang. https://pkg.go.dev/github.com/stretchr/testify/require?#hdr-Assertions

Please let me know if I should change anything.

pkg/server/plugin/upstreamauthority/gcpcas/gcpcas.go
```diff
@@ -284,9 +286,16 @@ func (p *Plugin) mintX509CA(ctx context.Context, csr []byte, preferredTTL int32)
 	// after the path.Split call above.  We need to trim off the /certificateAuthorities/ part for the request below
 	chosenPool = strings.TrimSuffix(chosenPool, "/certificateAuthorities/")
 
+	// certificate_id is required when using CertificateAuthority Enterprise tier. We generate a unique ID
+	// from the CSR public key. Same CSR will always produce the same ID.
+	csrSum := sha256.Sum256(csrParsed.RawSubjectPublicKeyInfo)
+	// Convert the hash into a string that matches the `[a-zA-Z0-9_-]{1,63}` requirement for GCP API.
+	certID := base64.RawURLEncoding.EncodeToString(csrSum[:])
```
sorindumitru, line 293: I think `base64` isn't going to match that since it also allows `+` and `/`. `base32` should, though. It's a bit bigger but should still fit inside of the 63 bytes available to us.

akos011221: Hi @sorindumitru, this one uses raw URL base64, so `+`, `/` and the padding are excluded. Let me know if I still should switch to `base32.`

PR #6399, opened 2025-10-27
Add dummy implementations for WIT-SVID RPCs

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Add WIT-SVID RPCs that return not implemented. Depends on https://github.com/spiffe/spire-api-sdk/pull/80

**Description of change**
Some boilerplate changes for the new WIT-SVID RPCs.


pkg/common/fflag/fflag.go
```diff
@@ -26,6 +26,10 @@ type RawConfig []string
 const (
 	// FlagTestFlag is defined purely for testing purposes.
 	FlagTestFlag Flag = "i_am_a_test_flag"
+
+	// FlatWITSVID controls if WIT-SVID and the APIs for it are enabled. When set
```
amartinezfayo:
```suggestion
	// FlagWITSVID controls if WIT-SVID and the APIs for it are enabled. When set
```

pkg/server/api/svid/v1/service.go
```diff
@@ -160,6 +160,10 @@ func (s *Service) MintJWTSVID(ctx context.Context, req *svidv1.MintJWTSVIDReques
 	}, nil
 }
 
+func (s *Service) MintWITSVID(ctx context.Context, req *svidv1.MintWITSVIDRequest) (*svidv1.MintWITSVIDResponse, error) {
+	return nil, status.Error(codes.Unimplemented, "RPC is not implemented")
```
amartinezfayo: I think that we should use `api.MakeErr` here (and in all the places similar to this). We have `api.MakeErr` as a way to consistently return errors while logging the error.
I think that it would also be beneficial to be more specific about the fact that the WIT-SVID functionality is not yet implemented.

```suggestion
	return nil, api.MakeErr(log, codes.Unimplemented, "WIT-SVID functionality is not yet implemented", nil)
```

pkg/common/fflag/fflag.go
```diff
@@ -26,6 +26,10 @@ type RawConfig []string
...
+	// to false all WIT-SVID APIs will return Unimplemented.
+	FlagWITSVID Flag = "wit-svid"
```
amartinezfayo, line 32: The flag is introduced, but we don't use it yet. Is that intentional?

sorindumitru: Yes, we'll definitely need this later during the implementation and thought we should have it defined to be ready to use. I can remove it though if you want to wait for an actual use to appear.

amartinezfayo: I think it’s appropriate to introduce it in this PR. I just wanted to confirm that we’re not unintentionally missing any related changes (e.g., an updated file).

amartinezfayo (approved):

PR #6403, opened 2025-10-29
test: add `waitForAfter()` to fix `TestHealthFailsAndRecover`

HankStat:
**Which issue this PR fixes**
Fixes [#6313](https://github.com/spiffe/spire/issues/6313)

### Problem
`TestHealthFailsAndRecover` is flaky due to a race condition between `clockMock.Add()` in `cache_test.go` and background goroutine calling `c.clk.After()` in `cache.go` that causes a deadlock.

If `Add()` executes before `After()`, the clock tick signal is missed, which leads to the test function block on `<-waitFor` and background goroutine blocks on `c.clk.After()` waiting for the clock tick.

### Solution
The solution is to `clockMock.WaitForAfter()` before every `clockMock.Add()` in the test. This ensures a deterministic order that calls `c.clk.After()` before `clockMock.Add()`. Then, the mock clock sends a signal to the buffered channel `m.afterC`. For the test function, `clockMock.WaitForAfter()` blocks until it receives the signal from `m.AfterC`.

I found `health start to fail`, `health still failing`, and `health recovered` could also have same race condition when running over 1,000 times. Thus, I applied the same fix to these sub tests.

### Verification
The test now correctly and consistently passes on my local machine after running it 10000 times using a loop to simulate the flaky behavior within 162 seconds. 
```bash
go test -v -race ./pkg/common/health -run ^TestHealthFailsAndRecover$ -count=10000
```

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -86,7 +86,7 @@ type Config struct {
 	TrustDomain spiffeid.TrustDomain
 
 	// Sources for getting Trust Bundles
-	TrustBundleSources *trustbundlesources.Bundle
+	TrustBundleSources trustbundlesources.Bundle
 
 	// Join token to use for attestation, if needed
 	JoinToken string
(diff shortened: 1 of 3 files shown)


sorindumitru (approved):
Thanks @HankStat!

Issue #6396, opened 2025-10-25
Missing support for KMS key tagging (KeyManager "aws_kms" plugin)


arber-salihi:
# Current issue
Organizations using SPIRE with the `aws_kms` KeyManager plugin for key management currently have no way to apply tags to the keys created. This limitation creates challenges for organizations in several key areas: cost allocation and tracking, resource organization and management, compliance and governance requirements.

# Proposed solution
Add support for user-defined tags in the plugin configuration, with built-in validation against AWS KMS tagging constraints to ensure everything works without issues. The solution is prioritizing backward compatibility by making this configuration optional, so existing deployments won't be affected, while new users can take advantage of the tagging functionality as needed.

Validation of tags would occur at configuration time, following a "fail-fast" approach.

As we would like Kubernetes-based deployments to benefit this feature, we plan to extend the [SPIRE Helm charts](https://github.com/spiffe/helm-charts-hardened) accordingly once the feature has been merged into SPIRE itself (this repository).

# Implementation overview
## Configuration schema
Add `key_tags` map to plugin configuration:

```hcl
KeyManager "aws_kms" {
    plugin_data {
        region = "us-east-1"
        key_identifier_file = "/path/to/file"

        key_tags = {
            Environment = "production"
            Team             = "security@example.com"
        }
    }
}
```

## Validation rules
Tags are validated during plugin configuration (fail-fast approach). The following validation checks would be performed, to fully aligned with restrictions/conditions from AWS.

- **Tag keys**: 1-128 characters, cannot start with `aws:` (case-insensitive), valid characters only
- **Tag values**: 0-256 characters, valid characters only
- **Total limit**: Maximum 50 tags

Valid characters include unicode letters, numbers, spaces, `+ - = . _ : / @`.

## Changes: core
1. **`aws_kms` KeyManager Plugin (`pkg/server/plugin/keymanager/awskms/awskms.go`)**
   Configuration layer
   - Add support for accepting key-value tag pairs in the plugin configuration
   - Parse tags during plugin initialization
   - Validate tags against AWS KMS constraints (see above)
   
   Runtime behavior
   - Pre-process and store tags during plugin configuration
   - Include the configured tags when creating new KMS keys

2. **Documentation (`doc/plugin_server_keymanager_aws_kms.md`)**
   - Document the new tagging capability
   - Recall AWS-specific tagging constraints and limits
   - Provide at least one short example showing how to specify tags
[...]

Issue #6269, opened 2025-08-25
AWS KMS Plugin - Code scan all the KMS keys in the account after every 48 hours


ggund:
Issue - Spire Server log show this error after every 48 hours . The spire is going through all the KMS that is generated by Spire ,but as it doesn't have any KMS policy on the KMS created by Spire system hence it is failing with below error.

Code  - The code is listing all the keys in the account as shown [here](https://github.com/spiffe/spire/blob/3e7102304c807229f7d26540d7cf3d5dd2840144/pkg/server/plugin/keymanager/awskms/awskms.go#L673)
and then getting the description of the keys as shown [here](https://github.com/spiffe/spire/blob/3e7102304c807229f7d26540d7cf3d5dd2840144/pkg/server/plugin/keymanager/awskms/awskms.go#L694)

Concerns - The code runs every [48 hours](https://github.com/spiffe/spire/blob/3e7102304c807229f7d26540d7cf3d5dd2840144/pkg/server/plugin/keymanager/awskms/awskms.go#L45) and attempts to scan ALL KMS keys in the AWS account [1](https://github.com/spiffe/spire/blob/3e7102304c807229f7d26540d7cf3d5dd2840144/pkg/server/plugin/keymanager/awskms/awskms.go#L673), then calls DescribeKey on each key found [2](https://github.com/spiffe/spire/blob/3e7102304c807229f7d26540d7cf3d5dd2840144/pkg/server/plugin/keymanager/awskms/awskms.go#L694).
This is problematic because:
It requires broader permissions than SPIRE should need
It's costly due to API calls for every key in the account
It violates the principle of least privilege
The code should be modified to only scan keys that SPIRE manages, rather than scanning the entire AWS account.

Error 

`time="2025-08-20T21:21:30Z" level=error msg="Failed to describe key to dispose" external=false key_arn=0xc0011d6350 plugin_name=aws_kms plugin_type=KeyManager reason="operation error KMS: DescribeKey, https response error StatusCode: 400, RequestID: 52a01bb6-f69f-40f1-9681-aa12fa2fd7c6, api error AccessDeniedException: User: arn:aws:sts::xxxxxxxxxxxx:assumed-role/iam-role/1755721406440771733 is not authorized to perform: kms:DescribeKey on resource: arn:aws:kms:us-east-1:xxxxxxxxxxxx:key/0ec2996f-xxx because no resource-based policy allows the kms:DescribeKey action" subsystem_name=catalog
[...]

PR #6410, opened 2025-11-02
Add KMS key tagging support to AWS KMS plugin (aws_kms)

arber-salihi:
**PR check list**

- [x] Documentation updated?

All tests complete successfully.
The application builds correctly.

**Affected functionality**
"aws_kms" plugin (`pkg/server/plugin/keymanager/awskms`)

**Description of change**
This pull request adds support for user-defined tags on AWS KMS keys created by the `aws_kms` plugin. This feature primarily focuses on supporting organizations that may require key tagging for different reasons (cost tracking, resource management, compliance).

More details can be found in the related issue (see below).

**Which issue this PR fixes**
#6396


arber-salihi:
Once this pull request is approved and merged, I plan to open a new pull request to the [`SPIFFE Helm Charts` repository](https://github.com/spiffe/helm-charts-hardened), so users who rely on Helm charts to deploy SPIRE on their clusters can also benefit this functionality.

amartinezfayo:
Thank you @arber-salihi for working on this contribution.

The lack of an efficient way to identify which keys are managed by the aws_kms plugin has been a problem since the introduction of the plugin. This has many impacts, one is not being able to (easily) track the keys for cost / resource / compliance management purposes, but also it makes it very inefficient for the plugin itself to identify which keys belong to the plugin, having to scan all the KMS keys in the account (#6269 has been opened to find a solution for that).

Other plugins, like gcp_kms and azure_key_vault, assign tags/labels to the keys that are managed by the plugin in order to keep track of them, and are used for efficient filtering when performing the listing operations in the service.
We have been looking for a similar approach in the aws_kms plugin. The AWS KMS API does not provide filtering capability based on tags when listing keys, but we have realized that we could use the [GetResources API](https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_GetResources.html) for that. This approach is explained in this [comment](https://github.com/spiffe/spire/issues/6269#issuecomment-3478702776).

I'm thinking that if the plugin automatically adds SPIRE-specific tags to the resources that it manages, maybe that solves the main motivation of this PR? That would allow users to track the AWS KMS keys managed by SPIRE without any additional configurations.
Do you think implementing that solution would fully address #6396, or would users still need the ability to define their own custom tags?

arber-salihi:
Hey @amartinezfayo!

I can only speak for our organization, but we would still need those custom tags. AWS resources are mapped to products and applications, so we need a way to ensure those identifiers are added properly. I believe other organizations would share a similar requirement as well.

That said, having default tags that the server adds automatically definitely makes sense. This is actually what I initially considered. However, I dropped this support for now to maintain backward compatibility.

The main issue with server-added tags is that if the IAM role assigned to the server doesn't have the TagResource permission, the server would keep crashing.

One solution I was considering for automated tagging would be to attempt creating keys with tags by default, and if the server rejects the request due to insufficient permissions, retry the request without tags.

On my end, I'd suggest focusing on custom tags first with this pull request, and then extending it to include default tags (the changes should be quite straightforward based on my internal testing so far). This approach would work if the solution above makes sense, or at least if we can clarify the official stance: is backward compatibility mandatory, or can we assume a breaking change?

What are your thoughts?

amartinezfayo:
> I can only speak for our organization, but we would still need those custom tags. AWS resources are mapped to products and applications, so we need a way to ensure those identifiers are added properly. I believe other organizations would share a similar requirement as well.

Thanks for clarifying. Just to make sure I understand: your requirement is that you need to define specific tag keys/values for your organization's resource mapping, and the plugin-defined tags wouldn't be sufficient for that purpose?

> The main issue with server-added tags is that if the IAM role assigned to the server doesn't have the TagResource permission, the server would keep crashing.

Right. If we go the built-in tags route, we'd need to introduce it as an opt-in feature with a proper deprecation path to avoid breaking existing deployments.

> On my end, I'd suggest focusing on custom tags first with this pull request, and then extending it to include default tags (the changes should be quite straightforward based on my internal testing so far).

That makes sense. Custom tags is definitely a reasonable feature. I mainly wanted to confirm whether built-in tags alone (per #6269) would be sufficient, or if custom tags are truly needed. Based on your use case, it sounds like custom tags are necessary.

amartinezfayo:
While discussing this within the maintainers group, it was raised that if we introduce support to have custom tagging, it would be good to have a reserved set of tags, so the plugin can safely use it in the future.
Reserving the prefix `spire-` seems like the natural choice.

arber-salihi:
> > I can only speak for our organization, but we would still need those custom tags. AWS resources are mapped to products and applications, so we need a way to ensure those identifiers are added properly. I believe other organizations would share a similar requirement as well.
> 
> Thanks for clarifying. Just to make sure I understand: your requirement is that you need to define specific tag keys/values for your organization's resource mapping, and the plugin-defined tags wouldn't be sufficient for that purpose?

Exact!

> While discussing this within the maintainers group, it was raised that if we introduce support to have custom tagging, it would be good to have a reserved set of tags, so the plugin can safely use it in the future. Reserving the prefix `spire-` seems like the natural choice.

That makes sense. I'd personally recommend `spire:` prefix, to align with what AWS uses, but whatever makes most sense to you. I'll try to update the pull request in the very coming days to support this use-case.

P.-S.: I see that two checks are failing. For the documentation, I'll take care of the linter recommendation in my next commit. The second failure (unit tests) does not seem to be related to this pull request. Is my observation correct, or am I missing something?

arber-salihi:
@amartinezfayo Updated! Currently, the reserved tags are the ones starting with `spire-`, but if you would prefer `spire:` to align with what AWS does (cf. my comment above), let me know.

I've also fixed the documentation so the linter does not complain anymore.

doc/plugin_server_keymanager_aws_kms.md
```diff
@@ -40,6 +40,22 @@ The plugin attempts to detect and prune stale aliases. To facilitate stale alias
 
 The plugin also attempts to detect and prune stale keys. All keys managed by the plugin are assigned a `Description` of the form `SPIRE_SERVER/{TRUST_DOMAIN}`. The plugin periodically scans the keys. Any key with a `Description` matching the proper form, that is both unassociated with any alias and has a `CreationDate` older than 48 hours, is removed.
 
+### Key tagging
```
amartinezfayo, line 43: Although reflected in the  "AWS KMS Access"section, I think that it may be good to be explicit here about the permissions requirement. Maybe something like:

When using key tagging, you **must** add the `kms:TagResource` permission to the IAM policy. Key creation will fail without this permission.

pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -368,6 +387,10 @@ func (p *Plugin) createKey(ctx context.Context, spireKeyID string, keyType keyma
 		Policy:      p.keyPolicy,
 	}
 
+	if len(p.keyTags) > 0 {
```
amartinezfayo, line 390: Should we add a note to the documentation about tag lifecycle?

The `key_tags` config is only applied during key creation, which I think is the right approach.

Changing `key_tags` configuration will only affect newly created keys, not existing ones. This is the intended behavior, but I think it's worth documenting to avoid user confusion.

arber-salihi:
Thank you @amartinezfayo for your review. All your comments make sense. Sorry for having missed the full configuration example actually. I've just pushed the requested changes. Let me know if there's anything else worth updating!

Issue #5994, opened 2025-04-07
Can't fetch cached JWT-SVIDs when spire-server is unresponsive


sorindumitru:
spire-agent maintains a cache of JWT-SVIDs (and X509-SVIDs, but those work differently) to serve repeated requests for the same JWT-SVID and to deal with spire-server unavailability.

When a cached JWT-SVID is past 50% of its TTL the agent will first try to request a new one from the server before falling back to the cached SVIDs. All (or most) RPCs the agent makes have a global timeout of [30 seconds](https://github.com/spiffe/spire/blob/ceee48c14e9978fc860bf1db551323e345e88865/pkg/agent/client/client.go#L46) applied. So the request to fetch a JWT-SVID may take up to 30 seconds to timeout if the server is unresponsive.

This is not the best, but I think there's a bigger issue here. If a client connects with a timeout smaller than 30 seconds the whole request ends up being cancelled by the client disconnecting so the agent doesn't get a chance to respond with the client. So a client connecting with a timeout of 5 seconds, for example, will just see all the requests timing out even though the cached SVID has enough lifetime left to be useful.

Some ways to deal with this:
* Apply a smaller timeout to the NewJWTSVID request when a cached entry exists and is valid, for example 1 second. It won't help every client, but it should help a lot of them. (small change)
* Always return the cached SVID and schedule the JWT-SVID for asynchronous refresh. (bigger change)

amartinezfayo:
Thank you @sorindumitru for raising this.
>Apply a smaller timeout to the NewJWTSVID request when a cached entry exists and is valid, for example 1 second. It won't help every client, but it should help a lot of them. (small change)

I agree this could be a good improvement. I personally think that a 1-second timeout might be a bit too aggressive in some environments. I'd suggest something in the range of 3–5 seconds instead.

PR #6454, opened 2025-11-21
fix: NewJWTSVID has shorter timeout when a cached JWT exists

jsnctl:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Cache behaviour for JWT-SVIDs on the Agent

**Description of change**

As described in https://github.com/spiffe/spire/issues/5994, the Agent will attempt to get a new JWT-SVID if a cached JWT-SVID has an impending expiry. This request has a default timeout of 30s, which can and does cause problems with other clients using much shorter timeouts.

This PR shortens the timeout of this cache hit case down to 1s, so the cached SVID can be made use of when the NewJWTSVID RPC on spire-server is unresponsive

**Which issue this PR fixes**

#5994 

---

:warning: I did not add new unit test functionality to this as I think this might need some thinking on how to properly assert the difference in the RPC request. Open to suggestions on how to best achieve this.

### DIFF
--- a/doc/spire_agent.md
+++ b/doc/spire_agent.md
@@ -77,12 +77,13 @@ This may be useful for templating configuration files, for example across differ
 | `x509_svid_cache_max_size`        | Soft limit of max number of X509-SVIDs that would be stored in LRU cache                                                                                                                                                                          | 1000                             |
 | `jwt_svid_cache_max_size`         | Hard limit of max number of JWT-SVIDs that would be stored in LRU cache                                                                                                                                                                           | 1000                             |
 
-| experimental                  | Description                                                                          | Default                 |
-|:------------------------------|--------------------------------------------------------------------------------------|-------------------------|
-| `named_pipe_name`             | Pipe name to bind the SPIRE Agent API named pipe (Windows only)                      | \spire-agent\public\api |
-| `sync_interval`               | Sync interval with SPIRE server with exponential backoff                             | 5 sec                   |
-| `use_sync_authorized_entries` | Use SyncAuthorizedEntries API for periodically synchronization of authorized entries | true                    |
-| `require_pq_kem`              | Require use of a post-quantum-safe key exchange method for TLS handshakes            | false                   |
+| experimental                  | Description                                                                                         | Default                 |
+|:------------------------------|-----------------------------------------------------------------------------------------------------|-------------------------|
+| `named_pipe_name`             | Pipe name to bind the SPIRE Agent API named pipe (Windows only)                                     | \spire-agent\public\api |
+| `sync_interval`               | Sync interval with SPIRE server with exponential backoff                                            | 5 sec                   |
+| `use_sync_authorized_entries` | Use SyncAuthorizedEntries API for periodically synchronization of authorized entries                | true                    |
+| `require_pq_kem`              | Require use of a post-quantum-safe key exchange method for TLS handshakes                           | false                   |
+| `jwt_svid_cache_hit_timeout`  | Custom gRPC timeout (between 5 and 30s) when retrieving a NewJWTSVID when a valid JWT-SVID in cache | 30s                     |
 
 ### Server Attestation
 


pkg/agent/client/client.go
```diff
@@ -43,7 +43,12 @@ var (
 	}
 )
 
-const rpcTimeout = 30 * time.Second
+const (
+	rpcTimeout = 30 * time.Second
+	// We can be more aggressive with timeouts in cases where a valid SVID
+	// exists in the cache but is old enough to try for a new SVID quickly
+	rpcTimeoutWithCacheHit = 1 * time.Second
```
sorindumitru: We'll want this to be a bit higher, 5 or at most 3 seconds. The higher we make the timeout the more likely it will be that someone will be hit by thy issue, but we also don't want it to be too low to make it impossible to get a SVID.

pkg/agent/manager/manager.go
```diff
@@ -298,7 +298,7 @@ func (m *manager) FetchJWTSVID(ctx context.Context, entry *common.RegistrationEn
 		return cachedSVID, nil
 	}
 
-	newSVID, err := m.client.NewJWTSVID(ctx, entry.EntryId, audience)
+	newSVID, err := m.client.NewJWTSVID(ctx, entry.EntryId, audience, ok)
```
sorindumitru: As you hinted in the text, we should include more things in the decision to use the lower timeout. If the JWT-SVID is expired we should use the full timeout since there's nothing for us to return to the user in that case.

pkg/agent/client/client.go
jsnctl: I've increased to 5 seconds

pkg/agent/manager/manager.go
jsnctl: I've added an explicit check to see if the JWT-SVID is expired and added this to the existing `ok` conditional - along with some commentary at the point this is used in the method

pkg/agent/client/client.go
```diff
@@ -43,7 +43,12 @@ var (
 	}
 )
 
-const rpcTimeout = 30 * time.Second
+const (
+	rpcTimeout = 30 * time.Second
+	// We can be more aggressive with timeouts in cases where a valid SVID
+	// exists in the cache but is old enough to try for a new SVID quickly
+	rpcTimeoutWithCacheHit = 5 * time.Second
```
sorindumitru: The changes look good in general, but to avoid breaking any existing users without any workaround and to allow some tuning for specific timeouts could we add an experimental config option to change this secondary timeout? It should allow setting it to any value smaller than the 30 seconds of the original timeout.

Setting the value to anything non-zero should also log a warning that this config is only temporary and that it's recommended to log an issue if you require a non-default timeout here. We can leave that option around for a few releases to see if anybody complaints and gather feedback and then we can remove it.

jsnctl: This makes sense. Should we introduce an `Experimental` struct to the Agent's `Config` for this purpose? It doesn't appear to exist for the Unix Agent Config object as yet

https://github.com/spiffe/spire/blob/46bd7ba1c226318362e679d249deef39b8f966d9/pkg/agent/config.go#L24-L118

sorindumitru: That `Config` is generated from the file based config, which has an [experimental section](https://github.com/spiffe/spire/blob/46bd7ba1c226318362e679d249deef39b8f966d9/cmd/spire-agent/cli/run/run.go#L116). It's ok to add the field to the config from `pkg/agent/config.go`, even if it's exported. We treat it as an internal detail and not something that users should import and use.

So it just need to be in the experimental section in the config file, but internally we can do whatever we want.

jsnctl: Sorry for the delay on this. https://github.com/spiffe/spire/pull/6454/commits/7ab2cdd6bab2cc958f20cc66b827b9764aec0437

I had a go trying to pass this configuration value through the various Agent-related parts of the Config stacks, but it seemed to result in an overly verbose change. Instead, I've had a go setting any timeout override stipulated by the Experimental config as a `var` that is used in place of the default `rpcTimeout`. 

I do have an inkling this might not be idiomatic Go (and passing the value through each stack would be preferred) but thought I'd push this to see what your thoughts were.

I've also added some basic unit test cases to cover the allowed values

jsnctl:
> This looks good to me, but could you also add some words about the flag in the [docs](https://github.com/spiffe/spire/blob/main/doc/spire_agent.md#agent-configuration-file)?
> 
> There also seem to be some merge conflicts that need to be handled.

Addressed these here https://github.com/spiffe/spire/pull/6454/commits/6766e20dfe2aec70e0ce667e252e292d40703395

PR #6465, opened 2025-11-28
Resolved 6464-Windows-Services-Start-binPath-Args-issue

### LINKED ISSUE #6464: Windows service entrypoint only sees service name in Execute args, rejects valid run arguments
When running spire-agent as a Windows service, the service wrapper’s Execute method is only receiving a single argument (the service name). The code in service_windows.go expects args[1] == "run" and returns ERROR_BAD_ARGUMENTS if that isn’t true, which makes the service fail to start even though the command line is valid.

Example service definition:

```
sc create spire-agent binpath= "C:\Install\Spire\spire-agent.exe run -config C:\Install\Spire\spire\conf\agent\agent.conf" start= auto
sc start spire-agent
```

At runtime:

args passed to Execute is just ["spire-agent"]

The full command line (["C:\\Install\\Spire\\spire-agent.exe", "run", "-config", "C:\\Install\\Spire\\spire\\conf\\agent\\agent.conf"]) only appears in os.Args

Because the check uses args[1] instead of os.Args[1], it always fails with ERROR_BAD_ARGUMENTS unless extra start parameters are passed via sc start.

xdefaulter:
**Pull Request check list**

- [ ] Commit conforms to CONTRIBUTING.md?y
- [ ] Proper tests/regressions included?y
- [ ] Documentation updated?n

**Affected functionality**
Instead of using non-persistent Start Parameters in windows services as we do using code below.
```
 sc.exe create spire-server binpath=c:\spire\bin\spire-server.exe
sc.exe start spire-server run -config c:\spire\conf\server\server.conf
```
We can use just create service command that will create the service with all the arguments. The arguments are persistent and the service can start automatically.
` sc.exe create spire-server binpath= "c:\spire\bin\spire-server.exe run -config c:\spire\conf\server\server.conf"`
**Description of change**
In service_windows enterypoint, I have used os.Args that will accept Arguments form binPath. The code has condition to check if os.Args are valid if not then it will check Start Parameters Args if they are also invalid then it will throw error same as before.

**Which issue this PR fixes**
fixes #6464


RodargDev:
@xdefaulter any idea when you will be able to resolve the merge issues? Don't want to be a prick and push it onto you but currently i can only use the sc start command which doesnt work with recovery so we are very interested in this fix. @amartinezfayo could this be hotfixed to 1.14.1 whenever it is merged?

amartinezfayo:
This has been merged and will be included in the 1.14.1 release.
Thanks for the contribution!

PR #6501, opened 2025-12-16
Use returned SPIFFE ID returned with JWT-SVID for caching

sorindumitru:
We use our cached entry for requesting a new svid which may be different from the one that the server has. It may happen that entry gets updated around the same time that a cached SVID is renewed, including changing the SPIFFE ID. In this case the SPIFFE ID of the returned SVID may be different from the one of the entry we have.

Because of the above we shouldn't use the SPIFFE ID of our entry for caching the SVIDs. The NewJWTSVID API also returns the SPIFFE ID of the issued SVID so we can use that for caching purposes.

X509-SVID doesn't have this issue since there we cache using the entry ID.


pkg/agent/client/client.go
```diff
@@ -77,7 +77,7 @@ type Client interface {
 	SyncUpdates(ctx context.Context, cachedEntries map[string]*common.RegistrationEntry, cachedBundles map[string]*common.Bundle) (SyncStats, error)
 	RenewSVID(ctx context.Context, csr []byte) (*X509SVID, error)
 	NewX509SVIDs(ctx context.Context, csrs map[string][]byte) (map[string]*X509SVID, error)
-	NewJWTSVID(ctx context.Context, entryID string, audience []string) (*JWTSVID, error)
+	NewJWTSVID(ctx context.Context, entryID string, audience []string) (*JWTSVID, string, error)
```
amartinezfayo: Have you considered returning spiffeid.ID directly instead of string to avoid the double parsing and improve type safety?

pkg/agent/manager/manager.go
```diff
@@ -310,6 +310,11 @@ func (m *manager) FetchJWTSVID(ctx context.Context, entry *common.RegistrationEn
 		return cachedSVID, nil
 	}
 
+	spiffeID, err = spiffeid.FromString(spiffeIdString)
```
amartinezfayo: Overwriting the spiffeID var may be confusing. The initial parsing is still used for cache lookup, but it's replaced with the server's SPIFFE ID. If the two differ, the cache lookup might miss?
We should probably have two different var names.

pkg/agent/manager/manager.go
```diff
@@ -310,6 +310,11 @@ func (m *manager) FetchJWTSVID(ctx context.Context, entry *common.RegistrationEn
...
+	if err != nil {
+		return nil, errors.New("Invalid SPIFFE ID: " + err.Error())
```
amartinezfayo: I think we should have fmt.Errorf with %w for error wrapping here.
```suggestion
		return nil, fmt.Errorf("invalid SPIFFE ID: %w", err)
 ```

pkg/agent/client/client.go
sorindumitru: Yeah, that's probably best. I think I had in mind that the SPIFFE ID gets converted to string when it's used in the cache, so string is probably going to be better for this. But I can see how that goes in a separate PR.

PR #6507, opened 2025-12-19
Fix flaky Windows tests by using GracefulStop in ServeGRPCServerOnListener

amartinezfayo:
Fixes a race condition in test cleanup that causes flakiness on Windows with named pipes.

Refer to https://github.com/spiffe/spire/actions/runs/20377861641/job/58560948733?pr=6506 as an example.

The issue occurred during gRPC server shutdown when `server.Stop()` was called while `server.Serve()` was still running. On Windows, named pipes use I/O completion ports (IOCP) which require coordination between multiple goroutines during cleanup.


sorindumitru (approved):
LGTM

PR #6514, opened 2025-12-26
Add JWT-SVID audience policy configuration with JTI claim support

### LINKED ISSUE #6043: JWT JTI Attribute
When federating with some other IDP's, they treat the optional JTI field as non optional. It looks like a very easy field we could add for compatibility. It just needs to be unique per token. Looks like some folks just use a sufficiently sized random number.

srikalyan:
## Summary

Implements issue #6043 - JWT JTI Attribute support with per-audience policy configuration.

This PR introduces a flexible JWT-SVID policy system that allows operators to control:
- Whether JTI (JWT ID) claims are included in tokens
- Whether agent-side caching is used

### New `JWTSVIDAudiencePolicy` enum with three modes:
- **DEFAULT (0)**: No JTI claim, caching enabled. Backwards compatible behavior.
- **AUDITABLE (1)**: JTI claim included for audit trails, caching still enabled. Verifiers can optionally enforce replay protection.
- **UNIQUE (2)**: JTI claim included, caching disabled. Each request gets a fresh token for strict replay protection.

### Configuration options:
- `jwt_svid_default_audience_policy`: Default policy for audiences not explicitly configured
- `jwt_svid_audience_policies`: Per-audience policy overrides (map of audience → policy)

### Key behaviors:
- **Most restrictive wins**: When multiple audiences are requested, the strictest policy applies
- **Backwards compatible**: DEFAULT policy preserves existing behavior
- **Database schema v24**: Adds `entry_audience_policies` table for per-audience configuration

## What's included in this PR:
- Proto definitions for new enum and entry fields
- Server-side JTI generation based on effective policy
- Agent-side cache bypass for UNIQUE policy
- Database migration and CRUD operations
- Comprehensive tests

## What's NOT included (requires spire-api-sdk update):
- Entry API conversion logic
- CLI commands for configuring policies

These will follow in a separate PR after the spire-api-sdk is updated with the new types.

## Test plan
- [x] Unit tests for JTI generation with all policy modes
- [x] Unit tests for "most restrictive wins" multi-audience behavior
- [x] Unit tests for per-audience policy overrides
- [x] Database migration tests
- [x] Agent cache bypass tests for UNIQUE policy

Fixes #6043


sorindumitru:
Thanks @srikalyan for opening this! We discussed this today and we think it might be good to start with something simple at first by having just the entry level policy (maybe called something like `jwt_svid_jti_policy`). We think that should cover most uses. If it turns out there's still a need for the per audience version, we can add something like that later on.

We also think we should combine this flag with the changes in https://github.com/spiffe/spire/pull/6360. This avoid the need to have these seldomly used flags as individual columns in the database. I don't know what's the current state of the work from @ValFadeev, but maybe you can continue from where he left off?

srikalyan:
> Thanks @srikalyan for opening this! We discussed this today and we think it might be good to start with something simple at first by having just the entry level policy (maybe called something like `jwt_svid_jti_policy`). We think that should cover most uses. If it turns out there's still a need for the per audience version, we can add something like that later on.
> 
> We also think we should combine this flag with the changes in #6360. This avoid the need to have these seldomly used flags as individual columns in the database. I don't know what's the current state of the work from @ValFadeev, but maybe you can continue from where he left off?

@ValFadeev let me know if you are ok in me making the necessary change?

ValFadeev:
@srikalyan yes, I am happy for you to do so and to discuss if necessary. I was still aiming to allocate the time to finish that work, but would not want to hold the progress

sorindumitru:
Hi, @srikalyan We'll be merging github.com/spiffe/spire/pull/6360 in SPIRE 1.15.0, which will be the next release after 1.14.3. If you want, you can rebase this on top of that PR or wait until it gets merged into main.

srikalyan:
Thank you @sorindumitru Will do it soon.

srikalyan:
Rebased the branch and will wait for your merge to finish.

srikalyan:
@sorindumitru now that the other feature is merged, can we get this PR rolling?

sorindumitru:
Yes, I hope I can have a run over this over the weekend and provide some feedback.

sorindumitru:
> Thanks @srikalyan for opening this! We discussed this today and we think it might be good to start with something simple at first by having just the entry level policy (maybe called something like `jwt_svid_jti_policy`). We think that should cover most uses. If it turns out there's still a need for the per audience version, we can add something like that later on.

Had a look through the code for a bit, I'll look some more, but I notice we still have both the flag + the audience map. I don't think it's that likely to configure things per audience so in the interest of keeping the configuration simple we should start with a simple flag, maybe just `jws_svid_include_jti`?

You'll also have to add the flag to the new [AdditionalAttributes](https://github.com/spiffe/spire-api-sdk/blob/99d433165a01add2a207122dab97d3d1fa1f77ad/proto/spire/api/types/entry.proto#L79) structure instead of directly to the entry (this will also fix some issues around parsing the field in this PR, i think listing entries doesn't currently parse the fields).

sorindumitru:
Hi @srikalyan, just checking in to see if you're still planning on updating this to use the new field added in the SDK.

srikalyan:
@sorindumitru sorry about it. I've made most changes and forgot to push them :). It should be up and ready for review now.

srikalyan:
@sorindumitru need your help on this

sorindumitru:
> @sorindumitru need your help on this

I'll try to get to this in the next couple of days, I just didn't have enough time to look at some things a bit more carefully.

pkg/agent/manager/manager.go
```diff
@@ -300,10 +300,18 @@ func (m *manager) FetchJWTSVID(ctx context.Context, entry *common.RegistrationEn
...
-	if ok && !m.c.RotationStrategy.JWTSVIDExpiresSoon(cachedSVID, now) {
-		return cachedSVID, nil
+	var cachedSVID *client.JWTSVID
+	var ok bool
+	if !bypassCache {
+		cachedSVID, ok = m.cache.GetJWTSVID(spiffeID, audience)
+		if ok && !m.c.RotationStrategy.JWTSVIDExpiresSoon(cachedSVID, now) {
+			return cachedSVID, nil
+		}
 	}
```
sorindumitru, line 315: Would be good to have a test verify that the cache is not updated in the case we bypass the cache.

pkg/server/api/entry/v1/service.go
```diff
@@ -651,18 +651,19 @@ func (s *Service) updateEntry(ctx context.Context, e *types.Entry, inputMask *ty
...
+			Admin:                inputMask.Admin,
+			Downstream:           inputMask.Downstream,
+			EntryExpiry:          inputMask.ExpiresAt,
+			DnsNames:             inputMask.DnsNames,
+			Selectors:            inputMask.Selectors,
+			StoreSvid:            inputMask.StoreSvid,
+			X509SvidTtl:          inputMask.X509SvidTtl,
+			JwtSvidTtl:           inputMask.JwtSvidTtl,
+			Hint:                 inputMask.Hint,
+			AdditionalAttributes: inputMask.AdditionalAttributes,
```
sorindumitru, line 666: Thanks for fixing this!

pkg/server/api/svid/v1/service.go
```diff
@@ -402,7 +403,7 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svidv1.NewJWTSVIDRequest)
 		return nil, api.MakeErr(log, codes.NotFound, "entry not found or not authorized", nil)
 	}
 
-	jwtsvid, err := s.mintJWTSVID(ctx, entry.GetSpiffeId(), req.Audience, entry.GetJwtSvidTtl())
+	jwtsvid, err := s.mintJWTSVID(ctx, entry.GetSpiffeId(), req.Audience, entry.GetJwtSvidTtl(), entry.GetAdditionalAttributes().GetJwtSvidIncludeJti())
```
sorindumitru: I think `GetAdditionalAttributes()` may return `nil`, so we probably have to check if it's nil or not. I don't know if we guarantee in another place that it ends being non-nil, but I think it's best to make sure it's not nil here.

Would  be good to have tests for this here, nil and non-nil additional attributes and the values of the JTI field.

pkg/server/api/entry_test.go
```diff
@@ -750,7 +750,8 @@ func TestReadOnlyEntryClone(t *testing.T) {
 		CreatedAt:      1678731397,
 		StoreSvid:      true,
 		AdditionalAttributes: &types.Entry_AdditionalAttributes{
-			DisableX509SvidPrefetch: false,
+			DisableX509SvidPrefetch: true,
+			JwtSvidIncludeJti:       true,
```
sorindumitru, line 754: Did something change here that made us have to change the value for the prefetch bool?

pkg/server/credtemplate/builder.go
```diff
@@ -347,6 +351,10 @@ func (b *Builder) BuildWorkloadJWTSVIDClaims(ctx context.Context, params Workloa
 			"iat": jwt.NewNumericDate(now),
 		},
 	}
+
+	if params.IncludeJTI {
+		attributes.Claims["jti"] = uuid.Must(uuid.NewV4()).String()
```
sorindumitru: I think this ends up with the `xxxx-xxx..-xxx` representation which is somewhat longer, mostly for human benefits. Maybe we can go with a `base64url` of the binary, which I think is smaller. If that's not smaller, what you have here also works.

cmd/spire-server/cli/entry/update.go
```diff
@@ -107,10 +110,16 @@ func (c *updateCommand) AppendFlags(f *flag.FlagSet) {
 	f.StringVar(&c.hint, "hint", "", "The entry hint, used to disambiguate entries with the same SPIFFE ID")
 	f.BoolFunc("disableX509SVIDPrefetch", "A boolean value that, when set, disables prefetching X509 SVID for this entry",
 		func(_ string) error {
-			c.additionalAttributesSet = true
+			c.disableX509SVIDPrefetchSet = true
 			c.disableX509SVIDPrefetch = true
 			return nil
 		})
```
sorindumitru, line 113: I think it might be easier to reason about this if we also have `additionalAttributesSet` and always fetch the entry when it is set. Otherwise the more fields we introduce the more checks we have to do.

Maybe that's a sign that we should have a mask for it, but maybe that's for the future.

sorindumitru (commented):
Sorry for the delay here, @srikalyan. I'll try to prioritise looking at the updates here.

pkg/agent/manager/manager.go
srikalyan: Addressed it.

pkg/server/api/svid/v1/service.go
srikalyan: Fixed it.

pkg/server/api/entry_test.go
srikalyan: Sorry for the unintentional changes, reverted in the latest push. The only addition is JwtSvidIncludeJti: true; DisableX509SvidPrefetch is back to its prior value (false).

cmd/spire-server/cli/entry/update.go
srikalyan: Done. additionalAttributesSet now gates the fetch in Run(); per-flag ...set bits are kept only for the merge step (no per-field mask within AdditionalAttributes today). A future sub-mask would let us drop them.

sorindumitru (dismissed):
LGTM, thanks @srikalyan! Would you mind fixing up the DCO?

cmd/spire-server/cli/entry/util_posix_test.go
```diff
@@ -24,7 +24,8 @@ const (
     	The entry hint, used to disambiguate entries with the same SPIFFE ID
   -instance string
     	Instance name to substitute into socket templates (env SPIRE_SERVER_PRIVATE_SOCKET_TEMPLATE).
-  -jwtSVIDTTL int
+  -jwtSVIDIncludeJTI
+` + "    \tA boolean value that, when set, includes a unique 'jti' claim in JWT-SVIDs issued for this entry and bypasses the agent JWT-SVID cache\n" + `  -jwtSVIDTTL int
```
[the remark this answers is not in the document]
srikalyan: Won't fix, cosmetic.

srikalyan:
@sorindumitru can you give one more ship it, had to address copilot review

sorindumitru (approved):
re-LGTM

PR #6517, opened 2025-12-29
Add support for minting WIT-SVIDs

### LINKED ISSUE #6374: WIT-SVID: Implement the MintWITSVID rpc in the SVID API
Implement the MintWITSVID rpc and in the SVID API and CLI commands to access it, similarly to the x509 and jwt commands.

sorindumitru:
- [ ] Documentation updated?

**Description of change**
This adds support in spire-server for minting WIT-SVIDs using the MintWITSVID API. The BatchNewWITSVID APIs are also implemented, but are currently unused.

**Which issue this PR fixes**
fixes #6374

### DIFF
--- a/pkg/server/api/svid/v1/service.go
+++ b/pkg/server/api/svid/v1/service.go
@@ -737,6 +737,10 @@ func ValidatePublicKeyAndSigningAlgorithm(publicKey crypto.PublicKey, signingAlg
 			if signingAlgorithm != "ES384" {
 				return fmt.Errorf("unsupported signing algorithm for ec-p384 key: %s", signingAlgorithm)
 			}
+		case elliptic.P521():
+			if signingAlgorithm != "ES512" {
+				return fmt.Errorf("unsupported signing algorithm for ec-p521 key: %s", signingAlgorithm)
+			}
 		}
 	default:
 		return fmt.Errorf("unsupported key type '%T'", publicKey)


pkg/server/api/svid/v1/service.go
```diff
@@ -393,7 +407,169 @@ func (s *Service) NewJWTSVID(ctx context.Context, req *svidv1.NewJWTSVIDRequest)
...
+	}
+
+	log = log.WithField(telemetry.SPIFFEID, spiffeID.String())
+
+	witSvid, err := s.ca.SignWorkloadWITSVID(ctx, ca.WorkloadWITSVIDParams{
+		SPIFFEID: spiffeID,
+		PublicKey: jose.JSONWebKey{
+			Key: publicKey,
+		},
+		// TODO: add its own TTL
```
amartinezfayo: It would be good to mention https://github.com/spiffe/go-spiffe/issues/373 here.

test/integration/suites/fetch-wit-svids/03-mint-svid
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+for format in pretty json; do
+  for keyType in ec-p256 ec-p384 rsa-2048 rsa-4096; do
+    docker compose exec -T spire-server /opt//spire/bin/spire-server \
```
amartinezfayo:
```suggestion
    docker compose exec -T spire-server /opt/spire/bin/spire-server \
```

test/integration/suites/fetch-wit-svids/03-mint-svid
```diff
@@ -0,0 +1,18 @@
...
+      wit mint -spiffeID "spiffe://domain.test/workload" \
+      -keyType "${keyType}" \
+      -output ${format} || fail-now "could not mint WIT-SVID using key type: ${keyType}"
+  done
+done
+
+# Check that we can specify a custom TTL
+docker compose exec -T spire-server /opt//spire/bin/spire-server \
```
amartinezfayo:
```suggestion
docker compose exec -T spire-server /opt/spire/bin/spire-server \
```

test/integration/suites/fetch-wit-svids/03-mint-svid
```diff
@@ -0,0 +1,18 @@
...
+  wit mint -spiffeID "spiffe://domain.test/workload" -ttl 60s || fail-now "could not mint WIT-SVID with custom TTL"
+
+# Check that WIT-SVID can be written to a directory
+docker compose exec -T spire-server /opt//spire/bin/spire-server \
```
amartinezfayo:
```suggestion
docker compose exec -T spire-server /opt/spire/bin/spire-server \
```

cmd/spire-server/cli/cli.go
```diff
@@ -124,6 +125,9 @@ func (cc *CLI) Run(ctx context.Context, args []string) int {
 		"jwt mint": func() (cli.Command, error) {
 			return jwt.NewMintCommand(), nil
 		},
+		"wit mint": func() (cli.Command, error) {
```
amartinezfayo: I think that we should only expose `wit` commands if the feature flag is enabled.
It's unfortunate that we can't use the fflag package at this point, because fflags are not loaded until we load the config in Run.

You can refer to https://github.com/spiffe/spire/pull/5586/changes#diff-f06b238782ee9be14f22b6719b32075f5a2ec8c96fe6d2fd4885e74326d98d05L146 for an example of how we implemented this when we had the force rotation work in progress. We should probably have a better solution for this.

amartinezfayo (approved):

pkg/server/credtemplate/builder.go
```diff
@@ -357,6 +373,30 @@ func (b *Builder) BuildWorkloadJWTSVIDClaims(ctx context.Context, params Workloa
...
+	}
+
+	now := b.config.Clock.Now()
+	_, expiresAt := computeCappedLifetime(b.config.Clock, ttl, params.ExpirationCap)
+
+	return map[string]any{
+		"sub": params.SPIFFEID.String(),
+		"exp": jwt.NewNumericDate(expiresAt),
+		"iat": jwt.NewNumericDate(now),
+		"cnf": params.PublicKey,
```
amartinezfayo: According with the [WIT-SVID spec](https://github.com/spiffe/spiffe/blob/draft-wit-svid/standards/WIT-SVID.md#32-confirmation---cnf), "The structure of this claim MUST be as described by RFC7800" which I interpret that requires the nested {"cnf": {"jwk": {...}}} structure.

Shouldn't this be something like this?
```
    "cnf": map[string]any{
        "jwk": params.PublicKey,
    },
```

pkg/server/ca/ca.go
```diff
@@ -29,15 +30,25 @@ const (
 	backdate = 10 * time.Second
 )
 
+var (
+	witWorkloadKeyAllowedAlgorithms = []jose.SignatureAlgorithm{
```
amartinezfayo: It seems like this CA layer and service layer have different allowed algorithm lists.
I think that we should have all the missing algorithms added to the CA allowed list.

Also, the WIT-SVID spec mandates ES512 support, which seems to be missing.

pkg/server/credtemplate/builder.go
sorindumitru: Thanks! never noticed that extra object

amartinezfayo (approved):
Thank you @sorindumitru!

Issue #6530, opened 2026-01-03
Replace fakeaggregator.NewSimpleClientset with NewClientset when available


amartinezfayo:
The `fakeaggregator.NewSimpleClientset()` function used in `pkg/server/plugin/notifier/k8sbundle/k8sbundle_test.go` is deprecated in favor of `NewClientset()`, which provides better field management support for server-side apply testing.

However, the aggregator package (`k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/fake`) currently only provides `NewSimpleClientset` and doesn't have a `NewClientset` alternative yet. According to the deprecation notice, `NewClientset` is only available when apply configurations are generated.
We should monitor upstream kubernetes releases and replace this deprecated function once the aggregator package provides the recommended alternative.

amartinezfayo:
Waiting for kubernetes/kubernetes#135980

PR #6531, opened 2026-01-03
Update k8s.io dependencies to v0.35.0

amartinezfayo:
Update k8s.io dependencies to v0.35.0.

Replace deprecated `fake.NewSimpleClientset()` with `fake.NewClientset()` in k8sbundle tests, which adds field management support for improved server-side apply testing.

Updated test assertions to use `cmp.Diff` with `cmpopts.IgnoreFields` to ignore the TypeMeta and ManagedFields metadata populated by the new clientset.

The aggregator fake client still uses `NewSimpleClientset()` with a nolint directive since the aggregator package doesn't provide `NewClientset` until apply configurations are generated. Created https://github.com/spiffe/spire/issues/6530 to track this.

This PR supersedes https://github.com/spiffe/spire/pull/6498.

### DIFF
--- a/pkg/common/diskcertmanager/cert_manager_test.go
+++ b/pkg/common/diskcertmanager/cert_manager_test.go
@@ -220,19 +220,22 @@ func TestTLSConfig(t *testing.T) {
 	})
 
 	t.Run("update cert file with an invalid cert start error log loop", func(t *testing.T) {
+		logHook.Reset()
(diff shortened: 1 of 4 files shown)


sorindumitru (approved):
LGTM. I opened https://github.com/kubernetes/kubernetes/issues/135980 last week for this. Looks like there's an open PR for it now so maybe we'll get the non-deprecated version soon.

PR #6542, opened 2026-01-09
Adds 'agent_version' to PostStatus request

### LINKED ISSUE #6356: Can spire agent send its build version to spire server for telemetry?
**Background**
In our setup, we are managing spire servers and our customers are managing spire agents. It will be useful to know which agent versions are currently in use e.g. last 30 days. We don't have access to customer runtimes apart from what is needed for challenges.

Agent versions are needed for risk assessment for server upgrades or as a reminder to upgrade agents for the customers.

Our specific example:
we have spire agent plugin - svid store - version X. It was having a bug and we fixed it in agent plugin version Y. We announced that everyone should move to new agent image but overall we have no clue how many customers did that. (i guess we can see image pull statistics?)
3 months later we performed data migration that broke some customers who were still running on older version due to mentioned bug. In here of course svid store plugin is purely in agent side so we have no idea what versions people run, but we can weakly link to spire agent version e.g. 1.13.1

Other example:
We run spire server 1.12.* and have agents connected with versions 1.11.* and 1.12.* (+-1 minor version compatibility). We bump spire server to 1.13.* and any agents running on 1.11.* might get into compatibility issues.

**Proposal**
Agent sends its build version as part of:
a) node attestation only?
b) with every request as user-agent gRPC header? e.g. (User-Agent: spire-agent/1.13.1)

Spire server can log this information or emit in grpc metric? By default this information is untrusted but it wont be used by attestation or authorizations, only for telemetry

nweisenauer-sap:
Requires https://github.com/spiffe/spire-api-sdk/pull/86

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

- When the agent starts up it will use the [PostStatus](https://github.com/spiffe/spire-api-sdk/blob/08d43797e4291c8b7671dcc7cfff574cb1b2cb14/proto/spire/api/server/agent/v1/agent.proto#L67) API to notify the server about it's current status. That API was initially going to be used for tracking bundle propagation, but it's unused and unimplemented now. This PR adds the version to the request and implements the API in spire-server.
- The server will update the status of the agent in the datastore in the same table that is currently tracking the rest of the agent information.
- ListAgents and GetAgent should also return the version
- The CLI will report the version of the agents.

**Which issue this PR fixes**

resolves #6356


pkg/agent/manager/manager.go
```diff
@@ -195,6 +196,14 @@ func (m *manager) Initialize(ctx context.Context) error {
 	m.syncedEntries = make(map[string]*common.RegistrationEntry)
 	m.syncedBundles = make(map[string]*common.Bundle)
 
+	// Post agent status with version information to the server
+	if err := m.client.PostStatus(ctx, version.Version()); err != nil {
+		// Log the error but don't fail initialization - the server may not support this yet
+		m.c.Log.WithField(telemetry.AgentVersion, version.Version()).WithError(err).Debug("Failed to post agent status")
```
sorindumitru: We should log this at the error level so users are more likely to notice it.
```suggestion
		m.c.Log.WithField(telemetry.AgentVersion, version.Version()).WithError(err).Error("Failed to post agent status")
```

pkg/agent/manager/manager.go
```diff
@@ -195,6 +196,14 @@ func (m *manager) Initialize(ctx context.Context) error {
...
+	} else {
+		m.c.Log.WithField(telemetry.AgentVersion, version.Version()).Info("Agent status posted")
```
sorindumitru: We can probably do without logging this every time. We have the audit logs for the case where users care about this.

pkg/agent/manager/manager_test.go
```diff
@@ -1905,6 +1908,11 @@ func (h *mockAPI) RenewAgent(ctx context.Context, req *agentv1.RenewAgentRequest
 	}, nil
 }
 
+func (h *mockAPI) PostStatus(_ context.Context, req *agentv1.PostStatusRequest) (*agentv1.PostStatusResponse, error) {
+	h.lastAgentVersion = req.AgentVersion
```
sorindumitru, line 1916: This isn't used for anything yet. Maybe we can verify the version in one of the tests that use `Initialize`?

pkg/server/endpoints/middleware.go
```diff
@@ -183,6 +183,7 @@ func RateLimits(config RateLimitConfig) map[string]api.RateLimiter {
 		"/spire.api.server.agent.v1.Agent/BanAgent":                                      noLimit,
 		"/spire.api.server.agent.v1.Agent/AttestAgent":                                   attestLimit,
 		"/spire.api.server.agent.v1.Agent/RenewAgent":                                    csrLimit,
+		"/spire.api.server.agent.v1.Agent/PostStatus":                                    noLimit,
```
sorindumitru: Maybe I'm being paranoid here, but could we have limit specific for the status? It comes from the agent which is somewhat untrusted so a node could try to use it for denial of service. We could have a low limit per IP, e.g. 5. This would allow an agent to restart a few times before having to back-off.

pkg/server/api/agent/v1/service.go
```diff
@@ -476,9 +476,37 @@ func (s *Service) RenewAgent(ctx context.Context, req *agentv1.RenewAgentRequest
...
+		return nil, api.MakeErr(log, codes.Internal, "caller ID missing from request context", nil)
+	}
+
+	log = log.WithField(telemetry.SPIFFEID, callerID.String())
+	rpccontext.AddRPCAuditFields(ctx, logrus.Fields{
+		telemetry.SPIFFEID:     callerID.String(),
+		telemetry.AgentVersion: req.GetAgentVersion(),
+	})
+
+	if req.GetAgentVersion() != "" {
```
sorindumitru: We should also validate do some validations on this since it isn't coming from a trusted source:
* It should be fairly small. I think gorm has a default size of 255 for strings, so maybe you can add a test to validate that it will reject strings longer than that?
* We should probably also validate that it is valid unicode,

sorindumitru (commented):
Thanks @nweisenauer-sap, this is looking great! I've left a few comments.

pkg/server/api/agent/v1/service.go
```diff
@@ -476,9 +476,43 @@ func (s *Service) RenewAgent(ctx context.Context, req *agentv1.RenewAgentRequest
...
+			SpiffeId:     callerID.String(),
+			AgentVersion: agentVersion,
+		}
+		mask := &common.AttestedNodeMask{
+			AgentVersion: true,
+		}
+		if err := s.updateAttestedNode(ctx, update, mask, log); err != nil {
+			return nil, err
+		}
+		log.WithField(telemetry.AgentVersion, agentVersion).Info("Agent status updated")
```
sorindumitru: Missed this one on the first run through. This can still happen often enough that we shouldn't log it at Info level. Could you switch it to Debug?

pkg/server/api/agent/v1/service.go
```diff
@@ -476,9 +476,43 @@ func (s *Service) RenewAgent(ctx context.Context, req *agentv1.RenewAgentRequest
...
+	if !ok {
+		return nil, api.MakeErr(log, codes.Internal, "caller ID missing from request context", nil)
+	}
+
+	agentVersion := req.GetAgentVersion()
+
+	log = log.WithField(telemetry.SPIFFEID, callerID.String())
+	rpccontext.AddRPCAuditFields(ctx, logrus.Fields{
+		telemetry.SPIFFEID:     callerID.String(),
+		telemetry.AgentVersion: agentVersion,
```
sorindumitru: Could we only add the `AgentVersion` to the rpc logs once we validate it? Looks like it supports calling `AddRPCAuditFields` multiple times.

sorindumitru (commented):
Thanks @nweisenauer-sap, looks like it's almost ready to merge, I had just 2 small comments.

nweisenauer-sap:
Thanks for the thorough review, @sorindumitru !

PR #6570, opened 2026-01-19
Fix flaky TestTLSConfig/delete_cert_files_start_error_log_loop test

amartinezfayo:
The test `TestTLSConfig/delete_cert_files_start_error_log_loop` could fail because it used `map[time.Time]struct{}` to deduplicate error log entries by timestamp when counting them. In rapid test execution with mock clocks, multiple log entries can share identical timestamps, causing the test to count fewer unique timestamps than actual log entries.

Example of failure: https://github.com/spiffe/spire/actions/runs/21139262035/job/60788895971#step:5:413

This led to intermittent failures when the assertion `len(errLogs) >= 1 && len(errLogs) <= testTickerIterations` failed.

The fix changes the log counting logic in all three affected test cases ("update cert file with an invalid cert start error log loop", "update key file with an invalid key start error log loop", and "delete cert files start error log loop") to directly count all matching log entries instead of deduplicating by timestamp.

This fix should address the root cause related with previous fixes in this test suite (#6565, #6551, #6533, #6528, #6513).

### DIFF
--- a/pkg/common/diskcertmanager/cert_manager_test.go
+++ b/pkg/common/diskcertmanager/cert_manager_test.go
@@ -231,13 +231,14 @@ func TestTLSConfig(t *testing.T) {
 		// Assert error logs were triggered (at least 1, at most testTickerIterations)
 		// Use Eventually to give the goroutine time to process the clock ticks
 		require.Eventually(t, func() bool {
-			errLogs := map[time.Time]struct{}{}
-			for _, entry := range logHook.AllEntries() {
+			entries := logHook.AllEntries()
+			errLogCount := 0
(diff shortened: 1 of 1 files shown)


amartinezfayo:
Since it has been challenging to reproduce TestTLSConfig failures locally, I ran the unit tests three times via GitHub Actions to validate this fix. All runs succeeded.

Given the multiple previous attempts to fix this test, if this is merged and flakiness persists, I would reconsider the overall test approach rather than continue applying incremental fixes. The current pattern of using mock clocks with goroutines and log assertions appears inherently difficult to make reliably deterministic.

PR #6587, opened 2026-01-19
Fix kubectl wait syntax in integration tests

amartinezfayo:
Fixed incorrect `kubectl wait` command syntax in the upstream-authority-vault and upstream-authority-ejbca integration test suites. The commands were using `--for condition=Ready` (with a space) instead of the correct `--for=condition=Ready` (with an equals sign), causing the wait commands to fail with "error: no matching resources found". 

Example of a failure: https://github.com/spiffe/spire/actions/runs/21143694183/job/60804882737?pr=6573#step:9:1531

The corrected syntax matches other SPIRE integration tests that already use the proper `--for=condition=Ready` format.

### DIFF
--- a/test/integration/suites/upstream-authority-ejbca/02-deploy-spire
+++ b/test/integration/suites/upstream-authority-ejbca/02-deploy-spire
@@ -21,4 +21,4 @@ for secret in "${secrets[@]}"; do
 done
 
 ./bin/kubectl -n spire apply -k conf/server
-./bin/kubectl wait pods -n spire -l app=spire-server --for condition=Ready --timeout=60s
+./bin/kubectl wait pods -n spire -l app=spire-server --for=condition=Ready --timeout=60s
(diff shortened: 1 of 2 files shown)


jsnctl:
Appreciate this is just a syntax fix, but any reason for a `wait` of the Pods rather than a rollout on the StatefulSet? e.g.

```
./bin/kubectl rollout status statefulset/spire-server -n spire --timeout=60s
```

amartinezfayo:
Good question! I'd keep kubectl wait pods here because the test also needs kubectl wait ... --for=delete to wait for pods to be deleted between auth method tests, and there's no rollout status equivalent for deletion. Using wait pods for both keeps it consistent. Additionally, the test deploys 4 different SPIRE Server configurations using different kustomization overlays, and waiting on -l app=spire-server works uniformly across all of them.

Does that make sense?

jsnctl:
Thanks for the explainer @amartinezfayo, makes sense!

PR #6594, opened 2026-01-21
Implement SPIFFE Broker Endpoint & API

arndt-s:
This is a Work In Progress.

Reference implementation for https://github.com/spiffe/spiffe/pull/340

api-sdk PR: https://github.com/spiffe/spire-api-sdk/pull/92
plugin-sdk PR: https://github.com/spiffe/spire-plugin-sdk/pull/71

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**


proto/brokerapi/brokerapi.proto
arndt-s, line 1: This needs to go to spire-api-sdk I suppose?

conf/agent/agent_full.conf
arndt-s, line 85: I'm unsure whether we should allow for a dedicated allow list for the broker endpoint or reuse the same?

cmd/spire-agent/cli/run/run_posix.go
```diff
@@ -53,6 +53,30 @@ func (c *agentConfig) hasAdminAddr() bool {
 	return c.AdminSocketPath != ""
 }
 
+func (c *agentConfig) getBrokerAddr() (net.Addr, error) {
```
arndt-s, line 56: For now I'm only supporting UDS, once the Broker API evolves we may allow serving it on TCP too.

pkg/agent/attestor/node/result.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/spiffe/go-spiffe/v2/bundle/spiffebundle"
+	"github.com/spiffe/go-spiffe/v2/bundle/x509bundle"
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"github.com/spiffe/go-spiffe/v2/svid/x509svid"
+	"github.com/spiffe/spire/pkg/agent/plugin/keymanager"
+)
+
+// Allow AttestationResult to be used as go-spiffe SVID and bundle sources.
+// TODO(arndt): Check whether the key of the agent is ok to be exposed to other parties.
```
arndt-s, line 15: I think this is worth a discussion.

pkg/agent/agent.go
```diff
@@ -255,6 +256,24 @@ func (a *Agent) Run(ctx context.Context) error {
...
+		brokerEndpoints, err := broker.New(&broker.Config{
+			BindAddr:            a.c.BrokerBindAddress,
+			Manager:             manager,
+			Log:                 a.c.Log,
+			Metrics:             metrics,
+			Attestor:            workloadAttestor,
+			TrustDomain:         a.c.TrustDomain,
+			AuthorizedDelegates: a.c.AuthorizedDelegates,
+			SVIDSource:          as,
+			BundleSource:        manager.GetX509Bundle(),
```
arndt-s, line 269: I wasn't sure what's better here, us the attestation result as a bundle source or the one from the manager?

pkg/agent/broker/api/service.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+func (s *Service) FetchX509Bundles(_ *brokerapi.X509BundlesRequest, stream brokerapi.SpiffeBrokerAPI_FetchX509BundlesServer) error {
+	ctx := stream.Context()
+
+	peer, err := s.getCallerContext(ctx)
+	if err != nil {
+		return err
+	}
+	_ = rpccontext.Logger(ctx).WithField("broker_peer", peer.String())
```
arndt-s, line 120: There's no log here .. I'm considering to add Debug logs for each RPC?

pkg/agent/broker/endpoints.go
```diff
@@ -0,0 +1,165 @@
...
+	SVIDSource   x509svid.Source
+	BundleSource x509bundle.Source
+}
+
+type Endpoints struct {
+	c *Config
+}
+
+func New(c *Config) (*Endpoints, error) {
+	switch {
```
arndt-s, line 49: Is this pattern ok? Delegated Identity doesn't have it and would panic if misconfigured?

pkg/agent/broker/endpoints.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func (e *Endpoints) ListenAndServe(ctx context.Context) error {
+	unaryInterceptor, streamInterceptor := middleware.Interceptors(
+		endpoints.Middleware(e.c.Log, e.c.Metrics),
+	)
+
+	// TODO(arndt): Delegated Identity API allows to be served without any authorized peer.
+	// I think it's better to fail as it's a misconfiguration and having that socket up
+	// without any authorized peer is just a potential security risk.
```
arndt-s, line 79: this is also worth a discussion

pkg/agent/broker/endpoints.go
```diff
@@ -0,0 +1,165 @@
...
+	// This is to accommodate environments where this API is served over network.
+	tlsConfig := tlsconfig.MTLSServerConfig(e.c.SVIDSource, e.c.BundleSource, tlsconfig.AuthorizeOneOf(authorizedDelegates...))
+	server := grpc.NewServer(
+		grpc.Creds(credentials.NewTLS(tlsConfig)),
+		grpc.UnaryInterceptor(unaryInterceptor),
+		grpc.StreamInterceptor(streamInterceptor),
+	)
+
+	e.registerBrokerAPI(server)
+	reflection.Register(server)
```
arndt-s, line 100: Any objections in having it here? I don't see a drawback and could help brokers that don't want to know the proto?

pkg/agent/broker/endpoints.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			return nil, err
+		}
+		// TODO(arndt): This also differs from the admin API where it's possible to define delegates
+		// from other trust domains. Here we enforce that delegates must be in the same trust domain
+		// as the agent.
+		// I suppose technically it's not possible on the admin API to encounter this, but considering that
+		// the SPIFFE Broker Endpoint may be offered over the network in the future I think it's better to
+		// enforce this here.
+		if id.TrustDomain() != trustDomain {
```
arndt-s, line 159: I added a configuration sanity check here. Any objections?

arndt-s (commented):
Self review highlighting parts that I would like to get feedback/opinions on please.

proto/brokerapi/brokerapi.proto
MikeZappa87: Yes. That would make things more uniform of how delegated api was implemented

sorindumitru: Another option would be to have this in `go-spiffe`, similar to the workload API. The existing delegated identity isn't there because it's SPIRE specific.

That may allow having some library/helper code to interact with the API.

But the SDK would also work.

pkg/agent/broker/endpoints.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (e *Endpoints) ListenAndServe(ctx context.Context) error {
+	unaryInterceptor, streamInterceptor := middleware.Interceptors(
+		endpoints.Middleware(e.c.Log, e.c.Metrics),
+	)
+
+	// TODO(arndt): Delegated Identity API allows to be served without any authorized peer.
+	// I think it's better to fail as it's a misconfiguration and having that socket up
+	// without any authorized peer is just a potential security risk.
+	if len(e.c.AuthorizedDelegates) == 0 {
```
Dimss, line 80: @arndt-s 
This means the authorized delegates SpiffeIDs must be well known and configured ahead the deployment happens. 
The problem is that when deploying SPIRE on K8s/OCP you not necessary know what they are. 
Let's take Istio as an example. 
First user deployed SPIRE, next deploys Istio. 
The SPIFFE IDs are created dynamically based on the `ClusterSPIFFEID` [CRs](https://github.com/spiffe/spire-controller-manager/blob/main/docs/clusterspiffeid-crd.md)    
With `if len(e.c.AuthorizedDelegates) == 0` the Spire agent won't become healthy until relevant Spiffe ID is created and the Spire-Agent config map being updated. 

I think we should disable this constraint, or at least make it optional.

sorindumitru: Generally you should be able to determine what SPIFFE IDs your workload is going to receive ahead of time. For example in k8s most of the time this is defined from namespace and service account names, which I imagine you have some control over.

It's possible to include more randomised information in there, e.g. pod uid in k8s, but I don't think that's particularly common. If that's the case, the policy specified here through this authorised delegates list could get some support for wildcarding.

Before doing that we'd like to see that there is an actually need for it, so if you have a use case where that's necessary let us know. Either way, that's not something that needs to be handled by this PR.

pkg/agent/attestor/workload/workload.go
```diff
@@ -72,34 +105,22 @@ func (wla *attestor) Attest(ctx context.Context, pid int) ([]*common.Selector, e
...
-// invokeAttestor invokes attestation against the supplied plugin. Should be called from a goroutine.
-func (wla *attestor) invokeAttestor(ctx context.Context, a workloadattestor.WorkloadAttestor, pid int) (_ []*common.Selector, err error) {
-	counter := telemetry_workload.StartAttestorCall(wla.c.Metrics, a.Name())
-	defer counter.Done(&err)
-
-	selectors, err := a.Attest(ctx, pid)
-	if err != nil {
-		return nil, fmt.Errorf("workload attestor %q failed: %w", a.Name(), err)
-	}
+	// // The agent health check currently exercises the Workload API. Since this
```
arndt-s, line 118: Can broker use agent PID? I'd like to block that? What if it uses the pod UID of the agent? 

TODO for myself to find a solution so that workload.go get's to know it's itself without knowing the details of the reference.

Dockerfile
arndt-s, line 1: How do others make images with custom api & plugin packages?

proto/brokerapi/brokerapi.proto
arndt-s: https://github.com/spiffe/spire-api-sdk/pull/92

arndt-s:
Things to discuss:
- Workload Attestor v2 interface
     - v1 <-> v2 compatability
     - proto Any
- Broker Endpoint Socket
     - authn (bundle & svid source from agent <-> server comms)
         @sorindumitru mentioned that there could be an API on the server that allows both. Also to check key manager and locking during rotation.
     - authz (re-using admin socket allowlist)
     -  -> do dedicated allowlist
     - gRPC reflection API
 - Windows support

kfox1111:
@arndt-s and co, if discussing Workload Attestor v2 interface, we should also consider https://github.com/spiffe/spire/issues/6640#issuecomment-4189394765 at the same time

matheuscscp:
Superseded by: #6915

sorindumitru:
Closing this in favour of https://github.com/spiffe/spire/pull/6915

PR #6616, opened 2026-01-29
feat: support supplemental bundle path for vault upstream authority

heybronson:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Vault upstream authority plugin

**Description of change**

This PR adds `supplemental_bundle_path` support to the Vault upstream authority plugin, enabling graceful root CA rotation.

Currently, the Vault plugin only includes the current root CA in the trust bundle. When the root CA rotates, workloads with certificates signed by the old root can no longer validate because the old root is immediately replaced. This makes root CA rotation disruptive.

The new `supplemental_bundle_path` configuration option allows operators to specify a PEM file containing additional root CA certificates to include in the trust bundle. This mirrors the existing implementation in the AWS PCA upstream authority plugin.

**Which issue this PR fixes**

related to: https://github.com/spiffe/spire/issues/5101


sorindumitru:
> Currently, the Vault plugin only includes the current root CA in the trust bundle. When the root CA rotates, workloads with certificates signed by the old root can no longer validate because the old root is immediately replaced. This makes root CA rotation disruptive.

Did you actually notice this behaviour in practice? The old CA shouldn't be removed immediately, it should still stay in the bundle while it is in use and for at least 24h after that.

sorindumitru:
Hi @heybronson, just checking in if you had some time to look at this [comment](https://github.com/spiffe/spire/pull/6616#issuecomment-3840383232)

sorindumitru:
@heybronson , we'll close this for now, if you're still interested in this, please reopen.

PR #6632, opened 2026-02-07
Add some notes about supported database versions

sorindumitru:
**Affected functionality**

Added some notes to the sql plugin docs about supported version. This was triggered by lib/pq only officially supporting maintained version of postgresql. As was already the case with mysql, this doesn't mean that it isn't working it's just that we won't be able to test it.

I haven't removed postgres-13 since it hasn't been unmaintained for too long, but if it adds too much to the CI time, we can remove it.

### DIFF
--- a/.github/workflows/stalebot.yaml
+++ b/.github/workflows/stalebot.yaml
@@ -11,7 +11,7 @@ jobs:
       issues: write
       pull-requests: write
     steps:
-    - uses: actions/stale@997185467fa4f803885201cee163a9f38240193d # v10.1.1
+    - uses: actions/stale@b5d41d4e1d5dceea10e7104786b73624c18a190f # v10.2.0
       with:
         days-before-issue-stale: 365 # 1 year
         days-before-issue-close: 30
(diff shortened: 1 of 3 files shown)


doc/plugin_server_datastore_sql.md
```diff
@@ -122,6 +128,11 @@ connection_string="username:password@tcp(localhost:3306)/dbname?parseTime=true"
 
 Consult the [MySQL driver repository](https://github.com/go-sql-driver/mysql#usage) for more `connection_string` options.
 
+#### Supported MySQL versions
+
+We support all currently maintainted versions of MySQL and run [integration tests](../test/integration/suites/datastore-mysql/) agains LTS releases. Older versions of
```
amartinezfayo:
```suggestion
We support all currently maintainted versions of MySQL and run [integration tests](../test/integration/suites/datastore-mysql/) against LTS releases. Older versions of
```

amartinezfayo (approved):

PR #6655, opened 2026-02-16
Stop wrapping objects in slices when printing

### LINKED ISSUE #6654: All CLI JSON output is wrapped in a slice of one item
In `pkg/common/cliprinter/cliprinter.go` the `PrintStruct` method causes its arguments to wrapped in a second slice before printing them

```go
func (p *printer) PrintStruct(msg ...any) error {
	return p.printStruct(msg)
}
```
because the internal method is variadic too
```go
func (p *printer) printStruct(msg ...any) error
```

This is in contrast to `PrintProto` which unwraps the args before passing them on (so doesn't wrap them in an extra slice).

```go
func (p *printer) PrintProto(msg ...proto.Message) error {
	return p.printProto(msg...)
}
```

Its unclear if this is intentional or not.

It caused a problem for me as I was modifying `cmd/spire-server/cli/jwt/mint.go` to support giving output conforming to https://kubernetes.io/docs/reference/config-api/client-authentication.v1/#client-authentication-k8s-io-v1-ExecCredentialStatus - but could not with the cliprinter as is because any json output was wrapped as an array of one item.

alec-w:
**Pull Request check list**

- [x] Documentation updated? (not affected)

**Affected functionality**
CLI printing when using JSON output.

**Description of change**
Stops the cliprinter `PrintStruct` from wrapping its inputs as a slice of one item before printing them. Currently it seems the behaviour is that all JSON output is wrapped as an array of one item, which makes it more awkward to consume. Also changes the tests for the cliprinter to call the exported methods rather than the internal ones and adds an assertion on the output to stop this regressing (unclear if tests where meant to be on the exported or internal methods - if they were originally on the exported methods this would have shown up there).

**Which issue this PR fixes**
Fixes #6654

### DIFF
--- a/cmd/spire-server/cli/agent/agent_test.go
+++ b/cmd/spire-server/cli/agent/agent_test.go
@@ -561,10 +561,10 @@ SPIFFE ID         : spiffe://example.org/spire/agent/agent3
 Error             : rpc error: code = Internal desc = some error when deleting agent
 `,
 			expectedStdoutJSON: fmt.Sprintf(
-				`[{"expired_agents":[
+				`{"expired_agents":[
 {"agent_id":"%s","deleted":false,"error":"rpc error: code = Internal desc = some error when deleting agent"},
 {"agent_id":"%s","deleted":false,"error":"rpc error: code = Internal desc = some error when deleting agent"}
-]}]`,
+]}`,
 				spiffeid.RequireFromPath(td, expiredAgents[1].Id.Path).String(),
 				spiffeid.RequireFromPath(td, expiredAgents[2].Id.Path).String(),
 			),
@@ -586,7 +586,7 @@ Agents purged:
 SPIFFE ID         : spiffe://example.org/spire/agent/agent3
 `,
 			expectedStdoutJSON: fmt.Sprintf(
-				`[{"expired_agents":[{"agent_id":"%s","deleted":true}]}]`,
+				`{"expired_agents":[{"agent_id":"%s","deleted":true}]}`,
 				spiffeid.RequireFromPath(td, expiredAgents[2].Id.Path).String(),
 			),
 		},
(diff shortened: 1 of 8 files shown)


amartinezfayo (commented):
Thank you @alec-w for this contribution.
I think it would be good to update `cmd/spire-server/cli/wit/mint.go` (that's under a feature flag) similarly to `x509/mint.go`. Its `prettyPrintMint` function still uses the old double-unwrapping pattern, so I think that after this fix it will always fail to type-assert and return `ErrInternalCustomPrettyFunc`.

alec-w:
> Thank you @alec-w for this contribution. I think it would be good to update `cmd/spire-server/cli/wit/mint.go` (that's under a feature flag) similarly to `x509/mint.go`. Its `prettyPrintMint` function still uses the old double-unwrapping pattern, so I think that after this fix it will always fail to type-assert and return `ErrInternalCustomPrettyFunc`.

Thanks @amartinezfayo . I'd missed that one. Will get this updated with latest main again and apply the same change there too. Should have that today/tomorrow

amartinezfayo:
Hi @alec-w, friendly ping. Are you planning to pick this back up, or would you prefer that we take over?

alec-w:
> Hi @alec-w, friendly ping. Are you planning to pick this back up, or would you prefer that we take over?

Hi @amartinezfayo thanks for the ping, I needed the reminder. I'll get this sorted tomorrow if that's alright (in-laws visiting right now)?

alec-w:
> Hi @alec-w, friendly ping. Are you planning to pick this back up, or would you prefer that we take over?

Hi @amartinezfayo 

Sorry for the delay here. Rebased on main and applied the same changes to the wit CLI command too now (amended commit). Will make an effort to be more responsive in future :see_no_evil:

amartinezfayo (approved):
Thank you @alec-w for the contribution!

PR #6661, opened 2026-02-18
Make the periodic cache rebuild happen periodically

alec-w:
**Pull Request check list**

- [x] Documentation updated? (n/a)

**Affected functionality**
Node cache introduced in https://github.com/spiffe/spire/pull/6176.

**Description of change**
Previously the periodic cache rebuild would only be executed once, now it should happen periodically based on the rebuild interval. This appears to have been the original intention looking at the code, but currently doesn't seem to be the case.

**Which issue this PR fixes**
No open issue for this yet

### DIFF
--- a/pkg/server/cache/nodecache/cache.go
+++ b/pkg/server/cache/nodecache/cache.go
@@ -133,12 +133,13 @@ func (c *Cache) RemoveAttestedNode(spiffeId string) {
 }
 
 func (c *Cache) PeriodicRebuild(ctx context.Context) error {
-	ticker := c.clk.Tick(rebuildInterval)
+	ticker := c.clk.Ticker(rebuildInterval)
+	defer ticker.Stop()
 	for {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
-		case <-ticker:
+		case <-ticker.C:
 			if err := c.Rebuild(ctx); err != nil {
 				c.log.WithError(err).Error("Fail to rebuild the attested node cache")
 			}
--- a/pkg/server/cache/nodecache/cache_test.go
+++ b/pkg/server/cache/nodecache/cache_test.go
@@ -129,6 +129,7 @@ func TestCachePeriodicRebuild(t *testing.T) {
 		err := cache.PeriodicRebuild(rebuildCtx)
 		require.ErrorIs(t, err, context.Canceled)
 	}()
+	clk.WaitForTicker(time.Minute, "timed out waiting for rebuild task to start")
 
 	cachedFirstAgent, _ := cache.LookupAttestedNode(firstAgent.SpiffeId)
 	require.NotNil(t, cachedFirstAgent)


pkg/server/cache/nodecache/cache.go
```diff
@@ -133,14 +133,14 @@ func (c *Cache) RemoveAttestedNode(spiffeId string) {
 }
 
 func (c *Cache) PeriodicRebuild(ctx context.Context) error {
-	select {
-	case <-ctx.Done():
-		return ctx.Err()
-	case <-c.clk.Tick(rebuildInterval):
-		if err := c.Rebuild(ctx); err != nil {
-			c.log.WithError(err).Error("Fail to rebuild the attested node cache")
+	for {
```
sorindumitru, line 138: nice catch!

alec-w: Thanks!

Have just pushed a second commit that should resolve the test failure in CI (it passed locally both ways...) :crossed_fingers:

alec-w: Tests failed on macos, I think possibly a flaky failure related to creating the ticker on every loop. Have switched that so the ticker is only created once which should hopefully fix it. Tests pass locally :crossed_fingers:

pkg/server/cache/nodecache/cache.go
```diff
@@ -133,14 +133,15 @@ func (c *Cache) RemoveAttestedNode(spiffeId string) {
 }
 
 func (c *Cache) PeriodicRebuild(ctx context.Context) error {
-	select {
-	case <-ctx.Done():
-		return ctx.Err()
-	case <-c.clk.Tick(rebuildInterval):
-		if err := c.Rebuild(ctx); err != nil {
-			c.log.WithError(err).Error("Fail to rebuild the attested node cache")
+	ticker := c.clk.Tick(rebuildInterval)
```
sorindumitru: We should also call Stop on the ticker at the end (probably not that important here since it only gets called once, but better to do it):
```suggestion
	ticker := c.clk.Tick(rebuildInterval)
	defer ticker.Stop()
```

pkg/server/cache/nodecache/cache_test.go
```diff
@@ -115,32 +122,47 @@ func TestCachePeriodicRebuild(t *testing.T) {
...
 
+	done := make(chan struct{})
+	rebuildCtx, cncl := context.WithCancel(t.Context())
 	go func() {
-		err := cache.PeriodicRebuild(t.Context())
-		require.NoError(t, err)
+		defer close(done)
+		err := cache.PeriodicRebuild(rebuildCtx)
+		require.ErrorIs(t, err, context.Canceled)
 	}()
```
sorindumitru, line 131: The mock clock can be a source of flakyness here. You need to make sure that the `clk.Ticker()` has been called before doing the rest of the checks. You can do something like:

```suggestion
	}()
	clk.WaitForTicker(time.Minute, "timed out waiting for rebuild task to start")
```

sorindumitru:
Hey @alec-w, do you think you'll have a chance to address those two small comments?

alec-w:
> Hey @alec-w, do you think you'll have a chance to address those two small comments?

Hey @sorindumitru, yes, will get this done by lunchtime today (UK time). Sorry for the delay coming back to it. Appreciate your help with it.

pkg/server/cache/nodecache/cache.go
alec-w: I hadn't originally as was using `Tick()` which just returns the channel, rather than `Ticker()` so didn't think it was needed from the docs https://pkg.go.dev/time#Tick

> Before Go 1.23, this documentation warned that the underlying [Ticker](https://pkg.go.dev/time#Ticker) would never be recovered by the garbage collector, and that if efficiency was a concern, code should use NewTicker instead and call [Ticker.Stop](https://pkg.go.dev/time#Ticker.Stop) when the ticker is no longer needed. As of Go 1.23, the garbage collector can recover unreferenced tickers, even if they haven't been stopped. The Stop method is no longer necessary to help the garbage collector. There is no longer any reason to prefer NewTicker when Tick will do.

Although, since for the tests to run reliably we need to wait for it to start listening for the tick we have to use the ticker, so have switched to that with the deferred `Stop()` :+1:

pkg/server/cache/nodecache/cache_test.go
alec-w: Thanks, I've added that in.

I think there's still a race condition in the tests - if the periodic cache rebuild starts the ticker before the the wait for ticker starts waiting for the ticker we'll timeout waiting for the next ticker. Although I don't see a way around that beyond modifying `WaitForTickerMulti` so hopefully the overhead of the go routine starting should mean the wait for the ticker starts first.

alec-w:
Thanks for your help with this one @sorindumitru  I believe it is ready for re-review / CI again now :crossed_fingers:

PR #6677, opened 2026-02-24
Add a simple docker verification before builing window image

MarcosDY:
Add a new step to verify than docker is available when building windows image


.github/workflows/scripts/wait-for-docker-windows.ps1
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env pwsh
+
+Write-Host "Waiting for Docker to be ready..."
+
+for ($i = 1; $i -le 60; $i++) {
+    # Check if Docker is ready
+    $output = docker info 2>&1
+    if ($LASTEXITCODE -eq 0) {
```
sorindumitru, line 8: Is this the same as `$LastExitCode`?

PR #6724, opened 2026-03-09
Add spire-agent workload API rate limiting by pod UID with OS UID fal…

### LINKED ISSUE #2010: Hardening the Agent - Rate Limiting UDS

terahertz5k:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
SPIRE agent Workload API and Envoy SDS — adds optional per-caller rate limiting for Workload API methods and SDS resource methods.

**Description of change**
Adds an optional `ratelimit` configuration block to the SPIRE agent that enforces per-caller token-bucket rate limits on Workload API methods (`FetchX509SVID`, `FetchJWTSVID`, `FetchX509Bundles`, `FetchJWTBundles`) and Envoy SDS methods (`StreamSecrets`, `FetchSecrets`). When a caller exhausts its burst, the agent returns `Unavailable` immediately. All limits default to 0 (disabled), so existing deployments are unaffected.

Key design decisions:

- **Selector-set key after attestation** — The rate limit key is the caller's attested selector set, so workloads with the same SPIFFE identity share one bucket regardless of pod/UID. Enforcement runs inside the handler after attestation rather than as gRPC middleware, since the selector set isn't known until then. Callers with no selectors share an `<unattested>` bucket.
- **Shared `pkg/common/ratelimit` package** — The token-bucket map + two-generation GC was factored out of the server's `perIPLimiter` into a reusable `PerKeyLimiter`, and the server middleware was rewritten on top of it. O(active-callers) memory, no background goroutine, automatic eviction after ~2 minutes of inactivity.
- **Non-blocking rejection (`AllowN`)** — Unlike the server-side `perIPLimiter` which uses blocking `WaitN`, the Workload API fails fast with `Unavailable` to avoid silently queuing goroutines during reconnection storms.
- **Metric-only observability** — Rejections emit a `workload_api.rate_limit_exceeded` counter with a `method` label. Per-rejection logging was intentionally omitted to avoid `logrus.Entry` allocation pressure under heavy load.

Configuration example:

```hcl
agent {
  ratelimit {
    fetch_x509_svid    = 20
    fetch_jwt_svid     = 20
    fetch_x509_bundles = 20
    fetch_jwt_bundles  = 20
    stream_secrets     = 20
    fetch_secrets      = 20
  }
}
```

Testing:

- Unit tests: token bucket mechanics, GC, selector-set keying including unattested callers, metrics labels, concurrency with race detector
- Handler-level tests for both Workload API and SDS verifying enforcement happens post-attestation
- Load tested in a local Kubernetes cluster against a high-RPS hammer workload — peak agent memory dropped ~31% with rate limiting enabled versus the same build with limits disabled

**Which issue this PR fixes**
fixes https://github.com/spiffe/spire/issues/2010


sorindumitru:
Hi @terahertz5k, thanks for opening a PR for this! It's something missing from the agent that we'd definitely want improved.

We were wondering if you can give some details about the particular scenario that you ran into that this PR helped you.

pkg/agent/endpoints/ratelimit.go
```diff
@@ -0,0 +1,196 @@
...
+	// previous holds all the limiters that were current at the GC.
+	previous map[string]*rate.Limiter
+
+	// current holds all the limiters that have been created or moved
+	// from the previous limiters since the last GC.
+	current map[string]*rate.Limiter
+
+	// lastGC is the time of the last GC.
+	lastGC time.Time
+}
```
sorindumitru: There's a lot in common between this structure and the `perIPLimiter` in the server. Is is possible to factor out some of the common code into `pkg/common/ratelimit` so that the agent and the server can share the code?

pkg/agent/endpoints/ratelimit.go
```diff
@@ -0,0 +1,196 @@
...
+	metrics  telemetry.Metrics
+}
+
+// resolveRateLimitKey returns the rate limit key for the given caller. When a
+// pod UID resolver is available and returns a pod UID, the key is prefixed with
+// "pod:" to avoid collisions with OS UID keys. Otherwise it falls back to the
+// OS UID prefixed with "uid:".
+func (m workloadRateLimitMiddleware) resolveRateLimitKey(caller peertracker.CallerInfo) string {
+	if m.resolver != nil {
+		if podUID := m.resolver.GetPodUID(caller.PID); podUID != "" {
```
sorindumitru: This operation can be expensive and will add some RPC time even in cases where spire is not running on Kubernetes but has rate limits configured. We would need to find a way to address this where we don't do unnecessary work on nodes that don't need it.

It would be good to understand the scenario you ran into that this change helped with. We can see how can change this to handle that and also not affect non-k8s users.

terahertz5k:
hi @sorindumitru ,

We were seeing badly behaving applications hammer the socket and cause memory usage on the agent to increase high enough for k8s to kill the spire-agent pod due to OOM.

We will be using this for x509 and jwt fetch.

I'll have a look at your other two comments and see what I can do there.

pkg/agent/endpoints/ratelimit.go
terahertz5k: Agreed, the procfs reads on every RPC are unnecessary on non-k8s nodes.

Thinking that I can add two things:
1. Check KUBERNETES_SERVICE_HOST env var at resolver construction time. If unset, return a nil resolver so there's zero overhead on non-k8s nodes.
2. On k8s nodes, we could wrap the resolver with a per-PID cache (short TTL ~10s) so we only hit procfs once per PID per interval, instead of on every RPC. Not sure if this is needed or not though.

I replied separately, but we were seeing agent memory increase due to badly behaving applications hammering the socket for certain operations. This would cause k8s to kill the agent due to OOM for the entire k8s node.

sorindumitru:
> We were seeing badly behaving applications hammer the socket and cause memory usage on the agent to increase high enough for k8s to kill the spire-agent pod due to OOM.

How were these applications misbehaving?

terahertz5k:
> > We were seeing badly behaving applications hammer the socket and cause memory usage on the agent to increase high enough for k8s to kill the spire-agent pod due to OOM.
> 
> How were these applications misbehaving?

We have an application that spins up multiple short-lived pods simultaneously, with processes forking within containers and creating a new connection to the agent socket on every retry rather than reusing existing ones. We worked with the team to fix their retry logic and raised agent memory limits, which stabilized things.

We didn't know it was this application at first, but we were able to see on our graphs that something was making lots of jwtfetch requests at certain times of the day as well as the number of agent pod restarts.

That said, we want to add this rate limiting as a safety net. It's valid for applications to have multiple identities and make fetch requests for each, but a single caller shouldn't be able to overwhelm the agent to the point of OOM. We'd rather shed excess load gracefully than let one misbehaving workload take down the agent for everyone on the node.

pkg/agent/endpoints/ratelimit.go
terahertz5k: Good call.

I can extract a shared PerKeyLimiter into pkg/common/ratelimit with a Limiter interface that covers both AllowN and WaitN. The two-generation GC map logic moves there, and both the agent's perCallerRateLimiter and server's perIPLimiter become thin wrappers that just handle key extraction.

Today, the agent stores *rate.Limiter directly while the server stores its own rawRateLimiter interface (so tests can inject fakes). One shared Limiter interface replaces rawRateLimiter and works for both cases: the agent calls AllowN on it, the server calls WaitN, and the server's test fakes just need a no-op AllowN stub added.

sorindumitru:
Thanks @terahertz5k for the answers above. One issue we have with the current approach is that it's very specific to k8s (and a bit to unix). We're wondering if applying the ratelimit per workload SPIFFE ID might be a better approach. This has the benefit of being useful in non-k8s cases but also comes with some disadvantages:
* The agent is spending a bit more time attesting the workloads, so you might not such a big benefit out of it.
* The ratelimit could potentially apply to multiple workload replicas, which may be a good or a bad thing, but can also be tuned through the limits.

Do you think you might be able to apply the ratelimit after the attestation and see if it still works ok for the cases you ran into?

Pittu-Sharma:
Hello, I would like to work on this issue.

terahertz5k:
@sorindumitru , thanks for the feedback.

The issue we hit was specifically that FetchJWTSVID is unary and every call triggers a fresh attestation. Moving the rate limit to after attestation would most likely undermine the primary goal of this change, which is to protect the agent from memory exhaustion. The concurrent attestation work is what drives the OOM, I think.

The pod UID resolution is the only k8s specific piece and it's only active on Linux nodes running k8s. On all other platforms, it falls back to OS UID. It's really an enhancement for cases (that are more likely in K8s) where container images may use the same OS UID.

To validate this, I'll move the rate limit after attestation and run my load tests. I'll report back with the results.

terahertz5k:
Based on my tests at 128MB and a 10 RPS rate limit, pre-attestation code survived 5 minutes of 4 pods with 100 forked workers each hammering FetchJWTSVID on a tight loop, but post-attestation code OOMKilled instantly.

Doing more tests at 256MB but with 4x300 workers, pre-attestation code peaked at 168 MB, but post-attestation code was OOMKilled. This shows there is a large delta as load increases between pre- and post- attestation rate limiting. Granted this load is pretty extreme for a single agent.

I increased memory to 512MB just to see what the actual memory peak was at 4x300 workers with 10 RPS rate limit.
Pre-Attestation: 138.4 MiB
Post-Attestation: 307.5 MiB
No rate limiting: 341.0 MiB

sorindumitru:
Sorry for the long delay here @terahertz5k, I was out for a few weeks. Would it be possible to get some heap profiling snapshots of these? I'm curious where this extra time is spent. I've done this locally using the unix workload attestor and the main memory increase seems to come from gRPC message buffers/parsing:

image

You can configure the agent with something like the following:
```
agent {
    ...
    profiling_enabled = true
    profiling_freq = 10
    profiling_port = 9999
    profiling_names = [ "heap" ]
}
```

This will tell the agent to periodically dump heap profiling dumps to disk, in a `.profiles` folder. You can also point `go tool pprof` to the 999 port and it will read the profiling information from there. I'm curious if there's anything else taking  up the memory that we could improve.

sorindumitru:
Or you could try the following patch, which reduces the read buffer size from 32KB to 1KB:
```
diff --git a/pkg/agent/endpoints/endpoints.go b/pkg/agent/endpoints/endpoints.go
index 2cd3cedb9..5537d0497 100644
--- a/pkg/agent/endpoints/endpoints.go
+++ b/pkg/agent/endpoints/endpoints.go
@@ -20,6 +20,10 @@ import (
 )

+const (
+       readBufferSize = 1024
+)
+
 type Server interface {
        ListenAndServe(ctx context.Context) error
        WaitForListening(listening chan struct{})
@@ -107,6 +111,7 @@ func (e *Endpoints) ListenAndServe(ctx context.Context) error {
                grpc.Creds(peertracker.NewCredentials()),
+               grpc.ReadBufferSize(readBufferSize),
        )

        workload_pb.RegisterSpiffeWorkloadAPIServer(server, e.workloadAPIServer)
```

I don't know if that's the right thing to do or if this is the right size to pick, but we do usually have smaller incoming payloads, so this may be a good improvement to have.

terahertz5k:
After investigating further, I found that my refactor in 4f8c01abd introduced a memory regression.

This time around, I switched from unlimited RPS to a 50 RPS/pod cap this time around, since pre-attestation rejection is nearly instantaneous while post-attestation has to complete a full attestation first so unlimited RPS means the two modes see completely different request volumes and aren't really comparable.

The issue is the sync.Map cache that 4f8c01abd added, keyed by PID, to try and avoid repeated procfs reads. That works fine if callers are long-lived processes, but spire-agent api fetch jwt spawns a fresh subprocess per call. Every request gets a new PID and the cache never hits. The map grows without bound.

The fix is just removing the cache. The rest of the refactor, extracting PerKeyLimiter into the shared package, the cleaner agent/server code, is fine and worth keeping.

Numbers (120s sustained, 50 RPS/pod across 4 pods, 10 RPS rate limit, 512 MiB):

No RL: 74 MiB (baseline)
Pre-RL 01b606f9d (original): 47 MiB (-37% vs no RL)
Pre-RL 4f8c01abd (as committed): 264 MiB (+257% vs no RL)
Pre-RL 4f8c01abd (cache removed): 46 MiB (-38% vs no RL)
Post-RL: 51 MiB (-31% vs no RL)

Pre-attestation RL with the cache removed and post-attestation RL land at nearly the same place. Happy to go either direction.

terahertz5k:
After the cache removal fix for pre-attestation RL, I ran comparison tests on your 1k read buffer patch. Numbers are averaged across the 120s window rather than peak:

No RL: 58 MiB without, 55 MiB with (-5%)
Pre-RL (cache removed): 47 MiB without, 43 MiB with (-8%)
Post-RL: 56 MiB without, 50 MiB with (-10%)

There's a small but consistent reduction in sustained memory across all configs, most visible on post-RL at -10%.

sorindumitru:
Thanks @terahertz5k. Doing the rate limit post attestation with the key being the SPIFFE ID sounds like the best way forward. We get most of the benefits and we also end up not being k8s specific.

Out of curiosity, I see that the memory numbers you mentioned now are smaller. Was this a run with fewer workloads?

terahertz5k:
The number of workloads was the same, but the old tests were at unlimited RPS. Pre-attestation rejection would be nearly instantaneous compared to doing full attestation. That means pre-RL ends up seeing far more total requests than the other configs, making it an unequal comparison. Capping at 50 RPS/pod keeps the request volume equal across all three configs. It also takes out some of the variance of my cpu being totally maxed out and probably thermal throttling.

I'll work on the post-attestation code and commit that soon.

Did you want me to include the read buffer patch? Or is that something you're working on elsewhere?

cmd/spire-agent/cli/run/run.go
```diff
@@ -601,6 +610,24 @@ func NewAgentConfig(c *Config, logOptions []log.Option, allowUnknownConfig bool)
...
+			return 0
+		}
+		return *p
+	}
+	if v := intVal(rl.FetchX509SVID); v < 0 {
+		return nil, errors.New("ratelimit.fetch_x509_svid must not be negative")
+	}
+	if v := intVal(rl.FetchJWTSVID); v < 0 {
+		return nil, errors.New("ratelimit.fetch_jwt_svid must not be negative")
+	}
```
sorindumitru: nit: we could use the values from `ac.WorkloadAPIRateLimit`, to avoid checking for `nil` twice.

cmd/spire-agent/cli/run/run.go
```diff
@@ -114,6 +116,13 @@ type sdsConfig struct {
 	DisableSPIFFECertValidation bool   `hcl:"disable_spiffe_cert_validation"`
 }
 
+type workloadAPIRateLimitConfig struct {
+	FetchX509SVID *int `hcl:"fetch_x509_svid"`
+	FetchJWTSVID  *int `hcl:"fetch_jwt_svid"`
+
+	UnusedKeyPositions map[string][]token.Pos `hcl:",unusedKeyPositions"`
+}
```
sorindumitru, line 127: We'll need to add configuration for the rate limiting bundles too.

conf/agent/agent_full.conf
```diff
@@ -108,6 +108,18 @@ agent {
     # allowed_foreign_jwt_claims: set a list of trusted claims to be returned when validating foreign JWTSVIDs
     # allowed_foreign_jwt_claims = []
 
+    # ratelimit: Optional per-SPIFFE-ID rate limiting for Workload API methods.
+    # Each value specifies the maximum number of calls per second allowed per SPIFFE ID.
+    # Limits are enforced after workload attestation against each registration entry
+    # matched by the caller. A value of 0 (the default) disables rate limiting for that method.
```
sorindumitru: I think we should also apply these to at least the SDSv3 API. The Delegated/BrokerAPI we can probably exclude for now, since they require authorization.

doc/spire_agent.md
```diff
@@ -85,6 +86,33 @@ This may be useful for templating configuration files, for example across differ
...
+agent {
+    # ...
+    ratelimit {
+        fetch_x509_svid = 100
+        fetch_jwt_svid  = 500
+    }
+}
+```
+
+Calls exceeding the rate limit receive a `ResourceExhausted` gRPC status code.
```
sorindumitru: I think based on the [spec](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Workload_Endpoint.md#appendix-a-list-of-error-codes) we might have to use `Unavailable` here for now. `ResourceExhausted` does seem like a better fit, but we have to follow the spec.

pkg/agent/endpoints/workload/handler.go
```diff
@@ -87,6 +109,10 @@ func (h *Handler) FetchJWTSVID(ctx context.Context, req *workload.JWTSVIDRequest
 	entries := h.c.Manager.MatchingRegistrationEntries(selectors)
 	entries = filterRegistrations(entries, log)
 
+	if err := h.rateLimitByEntries(MethodFetchJWTSVID, entries); err != nil {
```
sorindumitru: I guess I didn't think this through when I proposed this. Feels a bit weird to do this for all entries, but not unreasonable either. The entry list is sorted by entryID so we could just ratelimit on the first entry since it should always be the same.

Another alternative might be to do the rate limiting based on the selectors. This would mean different pods get different rate limits which is maybe a bit better. What do you think about this? I think it would still end up being useful in your scenario.

sorindumitru (commented):
Sorry for the late reply here @terahertz5k, I'll try to review things so that this ends up in the next release.

pkg/agent/endpoints/workload/handler.go
terahertz5k: The current implementation complicates adding rate limiting for bundles and SDSv3. They don't have a spiffe id available at the rate limit point to key on.

Switching to selector-set keying is a better fit and achieves per-workload rate limiting. I'll rework it and update the PR after some testing.

pkg/common/ratelimit/perkeylimiter_test.go
```diff
@@ -0,0 +1,125 @@
...
+	wg.Add(goroutines)
+	for i := range goroutines {
+		go func(id int) {
+			defer wg.Done()
+			key := fmt.Sprintf("key:%d", id%5)
+			for range callsPerGoroutine {
+				lim := l.GetLimiter(key)
+				lim.AllowN(time.Now(), 1)
+			}
+		}(i)
```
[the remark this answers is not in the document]
terahertz5k: range over an integer has been valid Go since 1.22 (Feb 2024). works fine.

pkg/agent/endpoints/ratelimit_test.go
```diff
@@ -0,0 +1,358 @@
...
+	lim := newPerCallerRateLimiter(10)
+
+	var wg sync.WaitGroup
+	wg.Add(goroutines)
+	for i := range goroutines {
+		go func(id int) {
+			defer wg.Done()
+			// Use 5 shared keys so goroutines contend on the same buckets.
+			key := fmt.Sprintf("spiffe://example.org/workload-%d", id%5)
+			for range callsPerGoroutine {
```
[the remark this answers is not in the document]
terahertz5k: range over an integer has been valid Go since 1.22 (Feb 2024). works fine.

pkg/server/api/middleware/ratelimit.go
```diff
@@ -17,36 +15,22 @@ import (
...
 	// from the standard rate package by default production.
-	newRawRateLimiter = func(limit rate.Limit, burst int) rawRateLimiter {
+	newRawRateLimiter = func(limit rate.Limit, burst int) ratelimit.Limiter {
 		return rate.NewLimiter(limit, burst)
 	}
 )
 
+// perKeyLimiterOpts holds options for creating per-key limiters.
+// Exposed for testing (e.g., clock injection).
+var perKeyLimiterOpts []ratelimit.Option
```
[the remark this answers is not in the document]
terahertz5k: Intentional. Mirrors the existing perKeyLimiterOpts pattern in the server. Tests don't run in parallel, so it's safe today. Happy to clean both up in a follow-up.

pkg/agent/endpoints/ratelimit.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/spiffe/spire/pkg/common/util"
+	"github.com/spiffe/spire/proto/spire/common"
+	"golang.org/x/time/rate"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// perCallerRateLimiterOpts holds options for creating per-caller rate limiters.
+// Exposed for testing (e.g., clock injection).
+var perCallerRateLimiterOpts []ratelimit.Option
```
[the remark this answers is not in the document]
terahertz5k: Intentional. Mirrors the existing perKeyLimiterOpts pattern in the server. Tests don't run in parallel, so it's safe today. Happy to clean both up in a follow-up.

pkg/common/ratelimit/perkeylimiter_test.go
```diff
@@ -0,0 +1,125 @@
...
+func TestPerKeyLimiterConcurrency(t *testing.T) {
+	const goroutines = 50
+	const callsPerGoroutine = 200
+
+	l := NewPerKeyLimiter(newTestLimiterFactory(10))
+
+	var wg sync.WaitGroup
+	wg.Add(goroutines)
+	for i := range goroutines {
+		go func(id int) {
```
sorindumitru: nit: we can do `wg.Go` so we don't have to do the `Add` and `Done`.

cmd/spire-agent/cli/run/run.go
```diff
@@ -114,6 +116,17 @@ type sdsConfig struct {
 	DisableSPIFFECertValidation bool   `hcl:"disable_spiffe_cert_validation"`
 }
 
+type workloadAPIRateLimitConfig struct {
```
sorindumitru, line 118: Could we have this configuration initially in the experimental block? That would allow us to experiment a bit with this and get feedback from users while still allowing us to make changes to the configuration if needed.

We can schedule the removal of it in 1.16 or 1.17.

pkg/agent/endpoints/sdsv3/handler.go
```diff
@@ -73,6 +89,10 @@ func (h *Handler) StreamSecrets(stream secret_v3.SecretDiscoveryService_StreamSe
 		return err
 	}
 
+	if err := h.rateLimit(MethodStreamSecrets, selectors); err != nil {
+		return err
+	}
```
sorindumitru: I think this wouldn't be the case here, but it might be worth adding a check to not do the ratelimiting for the agent, similar to the workload api.

pkg/agent/endpoints/workload/handler_test.go
```diff
@@ -924,6 +925,46 @@ func TestFetchJWTSVID(t *testing.T) {
...
+	rl := &rejectAllRateLimiter{}
+	params := testParams{
+		CA:          ca,
+		Identities:  identities,
+		AsPID:       os.Getpid(),
+		RateLimiter: rl,
+	}
+	runTest(t, params, func(ctx context.Context, c workloadPB.SpiffeWorkloadAPIClient) {
+		for i := 0; i < 3; i++ {
+			resp, err := c.FetchJWTSVID(ctx, &workloadPB.JWTSVIDRequest{
```
sorindumitru: Might be good to also test some of the other methods that we rate limit

pkg/agent/endpoints/ratelimit.go
```diff
@@ -0,0 +1,125 @@
...
+	var b strings.Builder
+	for i, s := range sorted {
+		if i > 0 {
+			b.WriteByte('\x01')
+		}
+		b.WriteString(s.Type)
+		b.WriteByte('\x00')
+		b.WriteString(s.Value)
+	}
+	return b.String()
```
sorindumitru: I think some of the keys will end up being quite big. Do you think it might be worth hashing it to make sure we have smaller keys?

terahertz5k: Good idea. I was considering hashing it as well but didn't end up doing it. Let me look into this.

terahertz5k: I was originally thinking SHA256, but now I'm looking at maphash (https://pkg.go.dev/hash/maphash). It would output 8 bytes instead of 32, but has a higher collision probability of 10^-12. If there was a collision, the two workloads would share a rate limiting bucket.

Is a 10^-12 chance of collision acceptable? I want to say yes, but SHA256 should also have almost no performance impact with hardware acceleration these days. What are your thoughts on this?

sorindumitru: We use SHA256 in another [place](https://github.com/spiffe/spire/blob/10cad40dae3b001af24740f562888d2d76ecd865/pkg/agent/manager/cache/jwt_cache.go#L163), so lets go with that. It's going to be potentially called more often here, but if it ends up being a bottleneck we can always change it later.

sorindumitru (commented):
This is looking good, @terahertz5k. Would you mind addressing the CI failures?

terahertz5k:
@sorindumitru , can you trigger the the CI pipelines again? I don't think I'm able to do it.

terahertz5k:
@sorindumitru I think I fixed the lint one. The other integration test failure seems to be a docker registry timeout, so maybe we just run it again.

sorindumitru (approved):
LGTM, thanks @terahertz5k !

PR #6732, opened 2026-03-11
Remove selectors from agent logs

sorindumitru:
The selectors can contain all sorts of information, it's probably best to never log the full set of discovered selectors from plugins. There is still one more place which logs it at debug level, which should be ok since they are useful for debugging.

### DIFF
--- a/pkg/agent/attestor/workload/workload.go
+++ b/pkg/agent/attestor/workload/workload.go
@@ -74,9 +74,7 @@ func (wla *attestor) Attest(ctx context.Context, pid int) ([]*common.Selector, e
 		case err := <-errChan:
 			log.WithError(err).Error("Failed to collect all selectors for PID")
 		case <-ctx.Done():
-			// If the client times out before all workload attestation plugins have reported selectors or an error,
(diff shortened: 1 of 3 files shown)


pkg/agent/attestor/workload/workload.go
```diff
@@ -76,7 +76,7 @@ func (wla *attestor) Attest(ctx context.Context, pid int) ([]*common.Selector, e
 		case <-ctx.Done():
 			// If the client times out before all workload attestation plugins have reported selectors or an error,
 			// it can be helpful to see the partial set of selectors discovered for debugging purposes.
-			log.WithField(telemetry.PartialSelectors, selectors).Error("Timed out collecting selectors for PID")
+			log.Error("Timed out collecting selectors for PID")
```
amartinezfayo, line 77: The comment above still says "it can be helpful to see the partial set of selectors discovered for debugging purposes," but this PR removes that. I think it's worth updating or removing that comment.

amartinezfayo (approved):
🎉

PR #6746, opened 2026-03-14
Code modernizations via `go fix` + manual follow-up

c4rlo:
Modernize code using `go fix`.

On top of that, manually complete a couple of migrations that `go fix` couldn't fully do on its own:

- Switching to `(sync.WaitGroup).Go()` wherever possible (and fixing some minor race conditions in the process).
- Switching to `new(value)` expressions (as introduced in Go 1.26) instead of using helper functions for this.

This change is structured into individual commits that make clear which parts were done automatically via `go fix` (and hence should need less scrutiny).


c4rlo:
> This looks great!, Can you rebase?

Done.

PR #6753, opened 2026-03-17
Update CHANGELOG for 1.14.3 release

amartinezfayo:
Update CHANGELOG for 1.14.3 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,19 +5,16 @@
 ### Added
 
 - `spire-agent` version is now reported to `spire-server` via the PostStatus API and visible in `GetAgent`/`ListAgents` CLI output (#6542)
-- Support for minting WIT-SVIDs via the `MintWITSVID` API (#6517)
 
 ### Changed
 
 - The `RequirePQKEM` TLS policy now uses the standardized `X25519MLKEM768` instead of the draft `x25519Kyber768Draft00` (#6703)
-- OPA policy evaluation performance improved by ~2x through use of partial evaluation (#6633)
+- OPA policy evaluation performance improved by ~2x, based on benchmarking, through use of partial evaluation (#6633)
 
 ### Fixed
 
-- `ReadOnlyEntry.Clone()` was incorrectly copying the `Admin` boolean into the `Downstream` field when applying an output mask, causing clients of `GetAuthorizedEntries` and `SyncAuthorizedEntries` to receive corrupted authorization metadata (#6636)
+- `ReadOnlyEntry.Clone()` was incorrectly copying the `Admin` boolean into the `Downstream` field when applying an output mask, causing clients of `GetAuthorizedEntries` and `SyncAuthorizedEntries` to receive corrupted authorization metadata. The `Admin` and `Downstream` booleana were not used in `spire-agent` so there was no impact from this (#6636)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
 # Changelog
 
+## [1.14.3] - 2026-03-18
+
+### Added
+
+- `spire-agent` version is now reported to `spire-server` via the PostStatus API and visible in `GetAgent`/`ListAgents` CLI output (#6542)
+- Support for minting WIT-SVIDs via the `MintWITSVID` API (#6517)
```
sorindumitru: Should we mention that this is still under a feature flag?

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+
+### Changed
+
+- The `RequirePQKEM` TLS policy now uses the standardized `X25519MLKEM768` instead of the draft `x25519Kyber768Draft00` (#6703)
+- OPA policy evaluation performance improved by ~2x through use of partial evaluation (#6633)
```
sorindumitru: At least based on our benchmark, I'm hoping this also manifest in actual use. Maybe we can soften this a bit:
```suggestion
- OPA policy evaluation performance improved by ~2x, based on benchmarking, through use of partial evaluation (#6633)
```

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+
+### Fixed
+
+- `ReadOnlyEntry.Clone()` was incorrectly copying the `Admin` boolean into the `Downstream` field when applying an output mask, causing clients of `GetAuthorizedEntries` and `SyncAuthorizedEntries` to receive corrupted authorization metadata (#6636)
```
sorindumitru:
```suggestion
- `ReadOnlyEntry.Clone()` was incorrectly copying the `Admin` boolean into the `Downstream` field when applying an output mask, causing clients of `GetAuthorizedEntries` and `SyncAuthorizedEntries` to receive corrupted authorization metadata. The `Admin` and `Downstream` booleana were not used in `spire-agent` so there was no impact from this. (#6636)
```

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+- The periodic node cache rebuild was only executing once instead of running continuously at the configured interval (#6661)
+- Missing rate limiter check in the PostStatus API was causing spurious "Rate limiter went unused" error log entries (#6635)
```
sorindumitru: I think we could exclude this, since the rate limit was added in the current release (1.14.3)
```suggestion
```

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+- SIGINT/SIGTERM errors were being logged during normal graceful shutdown (#6717)
```
sorindumitru: Same here, this was only an issue with Go 1.26, which we only upgraded to in this release:
```suggestion
```

CHANGELOG.md
amartinezfayo: Actually, I think we should not include this. We traditionally don't include functionality in CHANGELOG that's under feature flags, which is meant more for developers (as it's not even in experimental mode).

sorindumitru (approved):

PR #6763, opened 2026-03-18
Don't overwrite agent version on re-attestation

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Reported agent version after re-attestation.

**Description of change**
Once the agent re-attests we update the node in the datastore using a nil mask. This overwrites the reported version with an empty string.

### DIFF
--- a/pkg/server/api/agent/v1/service.go
+++ b/pkg/server/api/agent/v1/service.go
@@ -387,14 +387,7 @@ func (s *Service) AttestAgent(stream agentv1.Agent_AttestAgentServer) error {
 			CertSerialNumber: svid[0].SerialNumber.String(),
 			CanReattest:      attestResult.CanReattest,
 		}
-		mask := &common.AttestedNodeMask{
-			CertNotAfter:        true,
-			CertSerialNumber:    true,
-			NewCertNotAfter:     true,
-			NewCertSerialNumber: true,
-			CanReattest:         true,
-		}
-		if _, err := s.ds.UpdateAttestedNode(ctx, node, mask); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/server/api/agent/v1/service.go
```diff
@@ -387,7 +387,14 @@ func (s *Service) AttestAgent(stream agentv1.Agent_AttestAgentServer) error {
 			CertSerialNumber: svid[0].SerialNumber.String(),
 			CanReattest:      attestResult.CanReattest,
 		}
-		if _, err := s.ds.UpdateAttestedNode(ctx, node, nil); err != nil {
+		mask := &common.AttestedNodeMask{
```
amartinezfayo: I'm wondering if it would be worth extracting the mask into a shared package-level variable or a small helper to avoid the two definitions of mask (here and in endpoints package) diverging in the future. Both call sites set the same fields, so if a new field were added to AttestedNode that should also be preserved during re-attestation, we would need to remember to update both places.

pkg/server/api/agent/v1/service_test.go
```diff
@@ -3496,6 +3500,10 @@ func (s *serviceTest) assertAgentWasStored(t *testing.T, expectedID string, expe
 	agentSelectors, err := s.ds.GetNodeSelectors(ctx, expectedID, datastore.RequireCurrent)
 	require.NoError(t, err)
 	require.EqualValues(t, expectedSelectors, agentSelectors)
+
+	if expectedVersion != "" {
```
amartinezfayo: Should we consider always asserting the version rather than only when `expectedVersion != ""`?

PR #6777, opened 2026-03-22
agent: better request error logging

c4rlo:
This is a follow-up to #6744, where we added logging of the request context error in agent request handling failure paths.

Here we additionally add the request duration, and (if present) the context cancellation cause, if the request context is done (cancelled or timed out). We also apply it to the remaining cases where a Workload API request failure is being logged, excluding cases where a timeout is less likely/relevant (e.g. bad request).

It looks like this if the client is timing out when trying to obtain a JWT-SVID, for example:

    ERRO[0004] No identity issued   duration_ms=2000 error="context canceled" method=FetchJWTSVID pid=224164 registered=false service=WorkloadAPI subsystem_name=endpoints

Notice the new `duration_ms` field, which reflects the timeout applied by the client.

The added context cancellation cause is mainly future-proofing, as it currently doesn't seem to be populated in the cases I've tested. In particular, when the agent is shutting down due to receiving a signal, the cause on the root context (which gets set by [signal.NotifyContext](https://pkg.go.dev/os/signal#NotifyContext)) is not currently propagated into the request context. This could be done with more plumbing, but it did not seem worth the extra complexity since it's generally fairly obvious from the logs when the agent is shutting down.


amartinezfayo (approved):
Thank you @c4rlo!

PR #6789, opened 2026-03-25
Add instance flag support

### LINKED ISSUE #6247: Instance flag and pathing
**Background:**

One of the nice things about SPIRE is the ability to run multiple instances of both the `spire-agent` and `spire-server`.

SystemD is being used more and more on linux systems for deploying services. It suipports having multiple instances of a service. You create a unit file like spire-agent@.service, and then can start a particular instance by providing a name after the `@`.

Ex, start two instances of `spire-agent` named `a` and `b`:
```
systemctl start spire-agent@a spire-agent@b
```

The SystemD unit uses a template to path out variables such as /var/run/spire/agent/sockets/%i/public/spiffe.sock

This all works pretty well except for the cli tools.

**Problem:**
Say you want to run an agent command against the agent.

Ex:
```
spire-agent api watch -socketPath  /var/run/spire/agent/sockets/a/public/spiffe.sock
```

This makes for a pretty hard to use cli though, as you have to remember where the system stores its sockets.

**Proposed solution:**

We make some new variables (exact names TBD):
SPIFFE_PUBLIC SOCKET_TEMPLATE
SPIRE_AGENT_PRIVATE_SOCKET_TEMPLATE
SPIRE_SERVER_PRIVATE_SOCKET_TEMPLATE

ex:
```
export SPIFFE_PUBLIC_SOCKET_TEMPLATE=/var/run/spire/agent/sockets/%i/public/spiffe.sock
```

and add a new `-i` flag to the cli's.
```
spire-agent -i b api watch
```

kfox1111:
**Pull Request check list**

- [x] Documentation updated?

**Description of change**
Adds easy multiinstance support

**Which issue this PR fixes**
fixes #6247


cmd/spire-agent/cli/run/run_posix.go
```diff
@@ -16,21 +16,32 @@ import (
...
+	flags.StringVar(&c.Instance, "i", "", "Instance name to substitute into socket templates (env SPIFFE_PUBLIC_SOCKET_TEMPLATE and SPIRE_AGENT_PRIVATE_SOCKET_TEMPLATE). If omitted and the env var(s) are set, defaults to 'main'.")
 }
 
 func (c *agentConfig) setPlatformDefaults() {
 	c.SocketPath = common.DefaultSocketPath
 }
 
 func (c *agentConfig) getAddr() (net.Addr, error) {
-	return util.GetUnixAddrWithAbsPath(c.SocketPath)
+	resolved := common.ResolveSocketPath(c.SocketPath, common.DefaultSocketPath, "SPIFFE_PUBLIC_SOCKET_TEMPLATE", c.Instance)
```
sorindumitru: I think this variable should be `SPIRE_PUBLIC_SOCKET_TEMPLATE`. We can't use anything from the `SPIFFE_` namespace since the spec might want to use that later on.

kfox1111: Updated

cmd/spire-agent/cli/common/config_posix.go
```diff
@@ -11,18 +11,22 @@ import (
 
 type ConfigOS struct {
 	socketPath string
+	instance   string
 }
 
 func (c *ConfigOS) AddOSFlags(flags *flag.FlagSet) {
 	flags.StringVar(&c.socketPath, "socketPath", DefaultSocketPath, "Path to the SPIRE Agent API Unix domain socket")
+	flags.StringVar(&c.instance, "i", "", "Instance name to substitute into socket templates (env SPIRE_AGENT_PUBLIC_SOCKET_TEMPLATE). If omitted and the env var is set, defaults to 'main'.")
```
sorindumitru, line 19: I wonder if it's not a better UX to error out instead of defaulting to main, since main is not particularly meaningful in any way and it might not be an instance name.

I'm thinking that people might forget the `-i` in the case where those env vars are available and they accidentally target the wrong instance.

cmd/spire-agent/cli/common/template.go
```diff
@@ -0,0 +1,22 @@
+package common
+
+import (
+	"os"
+	"strings"
+)
+
+func ResolveSocketPath(socketPath, defaultPath, templateEnv, instance string) string {
+	tpl := os.Getenv(templateEnv)
+	if tpl != "" && strings.Contains(tpl, "%i") {
```
sorindumitru, line 10: What do you think about erroring out if the env vars are set but they don't include a template.

cmd/spire-agent/cli/api/api_posix_test.go
```diff
@@ -8,6 +8,8 @@ const (
     	comma separated list of audience values
   -format value
     	deprecated; use -output
+  -i string
```
sorindumitru, line 11: I think we should have it as `-instance` instead of `-i`, or we could have both maybe?

cmd/spire-agent/cli/common/config_posix.go
kfox1111: without instance flag support, it defaults to /tmp/xxxxxxx, so there is precedent where people expect not having a flag to "just work"

I'm kind of trying to reproduce that feeling too when configured with system packages and not just use /tmp on a systemd system which is pretty insecure. The system packages would set the path by default and then if you only have one instance, things "just work" like they always have.

But, if its a show stopper, requiring -i if the env vars set is still probably better then not having support at all

cmd/spire-agent/cli/common/template.go
kfox1111: interesting...

Depends on the answer above I think.

If we allow an assumed -i main when the var is set, then I think the algorithm should be:
 if env set, and doesnt have %i, it is usable only if -i is not specified or it only can ever be -i main or fail

Otherwise, yeah, we can have it fail if the env var is set and doesnt contain %i

cmd/spire-agent/cli/api/api_posix_test.go
kfox1111: its a lot of typing. I'd be good with both. (yeah, I know what sounds a bit odd. but when your doing it a lot in a day, it adds up quick)

cmd/spire-agent/cli/common/config_posix.go
kfox1111: I kind of envisioned the env vars here along with a default -i, fulfilling the role of:
https://github.com/spiffe/spire/issues/5770

cmd/spire-agent/cli/api/api_posix_test.go
sorindumitru: We discussed this today and we agree this can can be confusing when all there flags are "long names". So we'd prefer to go with `-instance`. We'll leave the `-i` for when we have support for short flags in the CLI.

cmd/spire-agent/cli/common/config_posix.go
sorindumitru: Discussed this today and we would prefer erroring out if the instance is not present. Any default here is likely to be very opinionated and likely not useful for a lot of people.

For #5770 I think we can still use `SPIRE_SERVER_ADMIN_SOCKET`, `SPIRE_AGENT_ADMIN_SOCKET` and `SPIFFE_ENDPOINT_SOCKET`. That would cover the non-instanced use case and would not require any instance specified.

kfox1111: ok. so, if no instance is specified, SPIRE_SERVER_ADMIN_SOCKET, SPIRE_AGENT_ADMIN_SOCKET and SPIFFE_ENDPOINT_SOCKET will be used if set and the regular defaults if not set.

If -instance is set and the template env vars are set, then those are used. otherwise, it will complain

kfox1111: actually, I ran into the issue with SPIFFE_ENDPOINT_SOCKET being incompatible with socket path before since one expects unix:// in front and the other does not. For consistency, I coded up everything to have the _TEMPLATE equivalent variable name without the _TEMPLATE

cmd/spire-agent/cli/common/template.go
kfox1111: done

kfox1111:
Ok, this should be ready

cmd/spire-agent/cli/common/config_posix.go
```diff
@@ -11,18 +11,28 @@ import (
 
 type ConfigOS struct {
 	socketPath string
+	instance   string
 }
 
 func (c *ConfigOS) AddOSFlags(flags *flag.FlagSet) {
 	flags.StringVar(&c.socketPath, "socketPath", DefaultSocketPath, "Path to the SPIRE Agent API Unix domain socket")
+	flags.StringVar(&c.instance, "instance", "", "Instance name to substitute into socket templates (env SPIRE_AGENT_PUBLIC_SOCKET_TEMPLATE). If omitted and the env var is set, defaults to 'main'.")
```
sorindumitru: This still mentions a default of 'main', similar for the other help locations.

cmd/spire-server/util/util_windows.go
```diff
@@ -27,7 +27,7 @@ func dialer(ctx context.Context, addr string) (net.Conn, error) {
 	return winio.DialPipeContext(ctx, npipeAddr)
 }
 
-func (a *Adapter) getGRPCAddr() string {
+func (a *Adapter) getGRPCAddr() (string, error) {
```
sorindumitru, line 30: We should also make sure we apply the template variables for windows too.

cmd/spire-agent/cli/common/template.go
```diff
@@ -0,0 +1,33 @@
...
+	"strings"
+)
+
+func ResolveSocketPath(socketPath, defaultPath, templateEnv, instance string) (string, error) {
+	baseEnv := strings.TrimSuffix(templateEnv, "_TEMPLATE")
+	tpl := os.Getenv(templateEnv)
+	sock := os.Getenv(baseEnv)
+	if instance != "" {
+		if tpl == "" {
+			return "", fmt.Errorf("you must define %s to use the instance flag", templateEnv)
```
sorindumitru:
```suggestion
			return "", fmt.Errorf(" %s environment variable must be defined when using the instance flag", templateEnv)
```

cmd/spire-agent/cli/common/config_posix.go
```diff
@@ -11,18 +11,28 @@ import (
...
+	return util.GetUnixAddrWithAbsPath(resolved)
 }
 
 func (c *ConfigOS) GetTargetName() (string, error) {
-	addr, err := util.GetUnixAddrWithAbsPath(c.socketPath)
+	resolved, err := ResolveSocketPath(c.socketPath, DefaultSocketPath, "SPIRE_AGENT_PUBLIC_SOCKET_TEMPLATE", c.instance)
+	if err != nil {
+		return "", err
+	}
+	addr, err := util.GetUnixAddrWithAbsPath(resolved)
```
sorindumitru: Seems like you could call  `GetAddr` to avoid some code duplication.

doc/multiple_instances.md
```diff
@@ -0,0 +1,27 @@
+# Running Multiple SPIRE Instances
+
+Systemd (and other init systems) can run multiple instances of the same service using a template unit. To make it easier to target a particular instance of `spire-agent` or `spire-server`, the CLI supports environment-variable socket templates and an -instance` flag.
+
+Environment variables
+
+- `SPIRE_AGENT_PUBLIC_SOCKET_TEMPLATE` — template for agent public/workload socket, use `%i` for the instance name.
+- `SPIRE_AGENT_PRIVATE_SOCKET_TEMPLATE` — template for agent private/admin socket, use `%i` for the instance name.
```
sorindumitru: Is this actually supported? I don't see any support for it in the code.

cmd/spire-server/util/util_windows.go
kfox1111: I think that would require a significant additional changes to this already big pr. Can that be done as a different pr?

cmd/spire-agent/cli/common/config_posix.go
kfox1111: Good catch. Fixed.

doc/multiple_instances.md
kfox1111: Your right.. Fixed.

cmd/spire-agent/cli/common/config_posix.go
kfox1111: true. fixed.

sorindumitru (commented):
This is mostly looking ok, but it would be good to have some tests for `ResolveSocketPath` and `getGRPCAddr` since that's where the main logic for this is and we do not have any tests for it.

kfox1111:
Added requested tests.

cmd/spire-server/util/util_windows.go
sorindumitru: Ok, I think that's fine, but would you mind opening an issue for this so we remember to do it at some point?

sorindumitru (approved):
LGTM, thanks @kfox1111 !

PR #6805, opened 2026-03-30
fix: prevent refreshAliasesTask from reverting rotated aliases in aws_kms plugin

angabini:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

`aws_kms` KeyManager plugin — specifically the `refreshAliasesTask` background task that periodically refreshes alias timestamps.

**Description of change**

In multi-replica SPIRE server deployments sharing a `key_identifier_value`, `refreshAliasesTask` on non-leader replicas can revert a KMS alias to a stale cached key after the leader rotates it. If the old key is subsequently scheduled for deletion, the alias ends up pointing to a disabled key, causing all future server startups to crash with `FailedPrecondition`.

This PR adds a `DescribeKey` check in `refreshAliases()` before calling `UpdateAlias`. If the alias's current target differs from the local cache (indicating a rotation by another replica), the refresh is skipped to avoid reverting the rotation.

**Which issue this PR fixes**

fixes https://github.com/spiffe/spire/issues/6804


pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -551,7 +551,29 @@ func (p *Plugin) refreshAliases(ctx context.Context) error {
...
+			KeyId: aws.String(entry.AliasName),
+		})
+		if err != nil {
+			p.log.Error("Failed to describe key for alias refresh", aliasNameTag, entry.AliasName, reasonTag, err)
+			errs = append(errs, err.Error())
+			continue
+		}
+		if describeResp == nil || describeResp.KeyMetadata == nil || describeResp.KeyMetadata.Arn == nil {
+			p.log.Error("Malformed describe key response during alias refresh", aliasNameTag, entry.AliasName)
+			continue
```
amartinezfayo, line 567: I think this path should also append to errs, similar to the DescribeKey error path.

pkg/server/plugin/keymanager/awskms/awskms_test.go
```diff
@@ -1396,6 +1396,100 @@ func TestRefreshAliases(t *testing.T) {
...
+	ts.clockHook.Add(6 * time.Hour)
+	err = waitForSignal(t, refreshAliasesSignal)
+	require.NoError(t, err)
+
+	// Assert: the alias must still point to the rotated key, not reverted
+	aliasAfter := ts.fakeKMSClient.store.aliases["alias/SPIRE_SERVER/test_example_org/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee/id_01"]
+	require.Equal(t, "key_id_rotated", *aliasAfter.KeyEntry.KeyID, "alias should not have been reverted to the stale cached key")
+}
+
+func TestRefreshAliasesDescribeKeyError(t *testing.T) {
```
amartinezfayo, line 1457: Maybe we could also add a test for the malformed response branch (where DescribeKey succeeds but returns nil KeyMetadata or nil Arn)?

amartinezfayo (commented):
Thank you @angabini for this contribution!

pkg/server/plugin/keymanager/awskms/awskms.go
angabini: done. does it look ok?

pkg/server/plugin/keymanager/awskms/awskms_test.go
angabini: done.

pkg/server/plugin/keymanager/awskms/awskms.go
amartinezfayo: Yes, thanks!

amartinezfayo (approved):
Thank you @angabini for this contribution!

PR #6807, opened 2026-03-31
Fix azure_imds node attestation for standalone VMs

ravishen:
**Affected functionality**

`azure_imds` node attestation for standalone Azure VMs.

**Description of change**

The standalone VM NIC query converts `securityGroup` to a string for grouping (`summarize ... by tostring(securityGroup)`), but never converts it back to an object before returning. The Go unmarshaler expects an object (`azureimds.SecurityGroup`), not a string, causing attestation to fail with:

      json: cannot unmarshal string into Go struct field
      NetworkInterface.securityGroup of type azureimds.SecurityGroup

**Note:** VMSS nodes are unaffected (different ARG query path).

fixes https://github.com/spiffe/spire/issues/6806

### DIFF
--- a/pkg/server/plugin/nodeattestor/azureimds/client.go
+++ b/pkg/server/plugin/nodeattestor/azureimds/client.go
@@ -168,7 +168,7 @@ func (c *azureClient) getNetworkInterfaces(ctx context.Context, vmId string, sub
 	query := fmt.Sprintf(`
 	Resources
 	| where type == "microsoft.network/networkinterfaces"
-	| where properties.virtualMachine.id == "%s"
+	| where tolower(properties.virtualMachine.id) == tolower("%s")
 	| mv-expand ipConfig = properties.ipConfigurations
 	| extend subnetId = tostring(ipConfig.properties.subnet.id)
 	| extend vnetName = extract(@"virtualNetworks/([^/]+)", 1, subnetId)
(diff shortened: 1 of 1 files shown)


pkg/server/plugin/nodeattestor/azureimds/client.go
```diff
@@ -168,7 +168,7 @@ func (c *azureClient) getNetworkInterfaces(ctx context.Context, vmId string, sub
 	query := fmt.Sprintf(`
 	Resources
 	| where type == "microsoft.network/networkinterfaces"
-	| where properties.virtualMachine.id == "%s"
+	| where tolower(properties.virtualMachine.id) == tolower("%s")
```
amartinezfayo, line 171: I have a question on the `tolower` addition: is this fixing an observed casing mismatch you've seen, or is it speculative hardening? Asking because looks like a user-visible behavior change separate from the unmarshal fix.

pkg/server/plugin/nodeattestor/azureimds/client.go
```diff
@@ -178,10 +178,10 @@ func (c *azureClient) getNetworkInterfaces(ctx context.Context, vmId string, sub
 	| extend nsgRg = extract(@"resourceGroups/([^/]+)",1,nsgId)
 	| extend nsgName = extract(@"networkSecurityGroups/([^/]+)",1,nsgId)
 	| extend securityGroup = bag_pack("resourceGroup", nsgRg, "name",nsgName)
-	| summarize 
+	| summarize
 		subnets = make_list(subnetObj)
-		by id, name, resourceGroup, tostring(securityGroup)
-	| project name, resourceGroup, subnets, securityGroup`, vmId)
+		by id, name, resourceGroup, securityGroup_str = tostring(securityGroup)
+	| project name, resourceGroup, subnets, securityGroup = todynamic(securityGroup_str)`, vmId)
```
amartinezfayo, line 184: It may be worth adding a regression test here. The existing fakeAPIClient mocks at the apiClient interface and returns pre-built *VirtualMachine structs, so nothing currently exercises the ARG-response-to-struct unmarshal path where this bug lives. A small test feeding `extractArmResourceGraphItems[NetworkInterface]` a JSON fixture shaped like a real ARG response would lock in the output contract.

amartinezfayo (commented):
Thank you for this contribution, @ravishen!

amartinezfayo:
@ravishen just checking-in here. Did you have a chance to look at the comments?

amartinezfayo:
@ravishen do you think you may have some time to come back to this? I can also take it over from here. Thank you for the contribution!

amartinezfayo (approved):
Thanks again for this fix, @ravishen.
I went back to the two comments I left earlier and I'm comfortable merging as-is. On the tolower question, I checked it against the Azure Resource Graph documentation and the `resourceGroup` segment of the `id` column on the `resources` table is lowercased, while nested property values like `properties.virtualMachine.id` keep their original casing, so an exact match can miss on standalone VMs. Case-insensitive matching is the pattern Azure recommends for exactly this situation, so the change is correct rather than speculative hardening.
On the regression test, I'd still like to lock in the ARG-response-to-struct contract, and since nothing about it blocks the fix I'll bring it in a separate follow-up PR along with a short comment recording the casing rationale.
Thank you for the contribution!

PR #6812, opened 2026-04-01
Add SPIRE-backed Prometheus TLS identity and SPIFFE allowlist

### LINKED ISSUE #6780: [spire] prometheus spiffe cert support
tls support was added to spire's prometheus endpoint recently, but to use it with spire issued certs requires a lot of effort, including the spire-agent and spiffe-helper, along with a bunch of glue scripting, and runs into ordering issues. There is then still no way to restrict the spiffeid of the attaching client.

Can we please update the feature so that:
* the user can specify via a flag that they just want to use spire itself for the web servers identity
* a list of spiffeid's that are allowed to connect to it.

This would allow easy securing of the prometheus telementry endpoint using spire itself.

aviralgarg05:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

The Prometheus telemetry endpoint TLS support for both SPIRE Server and SPIRE Agent.

This change makes it possible to:
- serve the Prometheus endpoint with SPIRE's own SVID instead of requiring a separately managed web certificate
- restrict access to that endpoint to an explicit allowlist of client SPIFFE IDs

**Description of change**

This updates the Prometheus telemetry TLS configuration so operators can secure the metrics endpoint using SPIRE-native identity rather than external certificate management and helper processes.

Specifically, the change:
- adds `use_spire_svid` to let the Prometheus listener present the current rotating SPIRE SVID
- adds `authorized_spiffe_ids` to allow only specific SPIFFE IDs to connect
- reuses existing SPIRE SVID and trust bundle state already maintained by the agent/server, instead of introducing any sidecar or glue logic
- preserves the existing file-based `cert_file` / `key_file` path for users who still want to serve the endpoint with a regular web certificate
- validates incompatible combinations such as `authorized_spiffe_ids` with `client_ca_file`
- adds regression coverage for:
  - SPIRE SVID-backed TLS
  - SPIFFE ID client authorization
  - config validation failures
  - existing file-based TLS behavior

The configuration examples and telemetry documentation were also updated to describe the new options.

**Which issue this PR fixes**

Fixes #6780


MarcosDY (dismissed):
Looks great!!
thanks!

MarcosDY (dismissed):

sorindumitru (approved):

PR #6864, opened 2026-04-11
fix: Adds gRPC conn release() method for common Agent-side client

jsnctl:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

Agent management of gRPC connections is now `defer` closed via new `release()` method

**Description of change**

Out of interest, I was perusing the SPIRE codebase with Claude to identify any simple, uncontroversial improvements we could make in terms of quality etc. Noticed this wasn't being done by the Agent, so here's the change to ensure it closes the gRPC connections in a manner similar to the way the SPIRE Server does (albeit the Server does have different connection responsibilities).

There's no unit test for this, as we don't do anything equivalent on the Server - happy to add this if we think it'd be useful

**Which issue this PR fixes**

N/A


amartinezfayo (approved):
Thank you @jsnctl!

PR #6869, opened 2026-04-13
feat(node-attestor/gcp_iit): use service account email from identity token

jeengbe:
According to https://docs.cloud.google.com/compute/docs/reference/rest/v1/instances/get#body.Instance.FIELDS.service_account, a Compute instance can only have one service account. Unsure why it's an array, then, but I understand it like there's only max one service account in that list.

Since a Compute identity token already contains an `email` claim (https://docs.cloud.google.com/compute/docs/instances/verifying-instance-identity#payload), we can use that one instead and can provide its selector without `use_instance_metadata`.

Additionally, it's useful to create Agent SPIFFE IDs from the (single) service account email too, so provide that claim in the agent path template data. I am not sure about that change, since https://github.com/spiffe/spire-plugin-sdk/blob/48d8694e60b270075c369ebef9fa0428e31d0d87/proto/spire/plugin/server/nodeattestor/v1/nodeattestor.proto#L49 suggests Agent IDs must be globally unique; i.e. two agents may not share the same ID?


pkg/server/plugin/nodeattestor/gcpiit/iit.go
```diff
@@ -175,16 +177,17 @@ func (p *IITAttestorPlugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServ
...
 		}
 	}
 
 	selectorValues := []string{
-		makeSelectorValue("project-id", identityMetadata.ProjectID),
-		makeSelectorValue("zone", identityMetadata.Zone),
-		makeSelectorValue("instance-name", identityMetadata.InstanceName),
+		makeSelectorValue("project-id", computeEngineMetadata.ProjectID),
+		makeSelectorValue("zone", computeEngineMetadata.Zone),
+		makeSelectorValue("instance-name", computeEngineMetadata.InstanceName),
```
amartinezfayo, line 189: I want to flag the compatibility consequences of this change. Moving `gcp_iit:sa` from metadata-gated to always-emitted is additive (existing registrations won't stop matching), but a registration that today only matches agents running with `use_instance_metadata = true` would start matching agents without it, which is a subtle behavior change. It's probably ok to include this with a proper entry in the CHANGELOG in the upcoming 1.15.0, which is a minor version bump.

doc/plugin_server_nodeattestor_gcp_iit.md
```diff
@@ -79,13 +79,14 @@ Details about the template engine are available [here](template_engine.md).
...
 | Value                      | Description                                                      |
-|----------------------------|------------------------------------------------------------------|
+| -------------------------- | ---------------------------------------------------------------- |
 | .PluginName                | The name of the plugin                                           |
 | .ProjectID                 | The ID for the project where the instance was created            |
 | .InstanceID                | The unique ID for the instance to which this token belongs.      |
 | .ProjectNumber             | The unique number for the project where you created the instance |
 | .Zone                      | The zone where the instance is located                           |
 | .InstanceCreationTimestamp | A Unix timestamp indicating when you created the instance.       |
+| .ServiceAccount            | The service account email ("@" is replaced with "_")             |
```
amartinezfayo, line 89: I suggest we add a note that the service account email is not per-instance unique: multiple VMs can share the same SA, so any custom `agent_path_template` that relies on `.ServiceAccount` without also including a per-instance value like `.InstanceID` would produce colliding agent IDs. Combined with TOFU, the first agent would attest successfully and any subsequent agent with the same resolved template would fail.

To answer your question in the PR description about agent ID uniqueness: the default template still uses `.InstanceID`, so it is unaffected, but exposing `.ServiceAccount` without guidance seems to be easy to misuse.

doc/plugin_server_nodeattestor_gcp_iit.md
```diff
@@ -9,7 +9,7 @@ This plugin requires an allow list of ProjectID from which nodes can be attested
 ## Configuration
 
 | Configuration             | Description                                                                                                                             | Default                                                   |
-|---------------------------|-----------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------|
+| ------------------------- | --------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------- |
```
amartinezfayo: nit: I wonder if the non-content changes to the existing tables (widening column separators from `|---|` to `| --- |` and re-spacing cells) could be reverted so the diff stays focused on the two new rows. Both formats are valid GitHub markdown, and keeping the stylistic delta out makes the behavioral change easier to review.

amartinezfayo (commented):
Thank you @jeengbe for this contribution@

pkg/server/plugin/nodeattestor/gcpiit/iit.go
jeengbe: Yes, thank you for noting this. Could you take care of fixing that in the changelog, please? Not sure how to go about the last pieces to get this merged (changelog, version bump, etc.)

doc/plugin_server_nodeattestor_gcp_iit.md
jeengbe: Are you saying that an agent's identity must _always_ be unique? That is, that two agents can't attest with the same ID? If so, I would suggest we make that clearer in the SPIRE docs somewhere. I did not know about that.

doc/plugin_server_nodeattestor_gcp_iit.md
jeengbe: Autoformatting did that, I thought nothing of it. Absolutely 🙂

pkg/server/plugin/nodeattestor/gcpiit/iit.go
amartinezfayo: No action needed on your side for the CHANGELOG or version bump, those are curated by the maintainers at release time in a dedicated PR. When 1.15.0 is cut, an entry for this change will be added under the `Changed` section referencing this PR. Thanks again for the contribution.

jeengbe: Thank you for the review and maintaining the project 🙇🏻‍♂️

doc/plugin_server_nodeattestor_gcp_iit.md
amartinezfayo: Yes, agent SPIFFE IDs must be unique within a trust domain. In this plugin that's enforced by the TOFU check: the first agent to attest with a given ID wins, and any subsequent attempt is rejected with `attestation data has already been used to attest an agent`. The datastore reinforces this at the schema level with a unique index on SPIFFE ID, so duplicates can't coexist.
This could be clearer in the docs. The current `## Security Considerations` section here mentions TOFU but doesn't frame it in terms of the uniqueness invariant the operator has to preserve when customizing `agent_path_template`. I think a short note in the Agent Path Template section of this file is a good scope for this PR if you are up for that. Thank you!

jeengbe: Added a note in the readme. But given that this rule applies to all node attestors (k8s, etc.), I think it would be wise to include this in the general spiffe.io SPIRE docs.

amartinezfayo (approved):
Thank you @jeengbe for this contribution!

PR #6875, opened 2026-04-15
Specify a read buffer size for agent WorkloadAPI server

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Hopefully none, but it may lead to multiple read() calls when reading large requests for WorkloadAPI, SDSv3 and gRPC Healthcheck.

**Description of change**
gRPC allocates a 32KB buffer per active connection for reading data from callers. This is excessive since we don't usually have large incoming requests.

In a testing setup this dropped the memory usage of spire-agent from ~340MB to ~200MB

In the case where we have larger incoming payloads, we would end up making multiple read() calls to read the incoming data, so no functionality is affected but if there are larger requests (maybe for the SDSv3 API?) incoming we could have some performance degradation.

### DIFF
--- a/pkg/agent/endpoints/endpoints.go
+++ b/pkg/agent/endpoints/endpoints.go
@@ -20,6 +20,10 @@ import (
 	"github.com/spiffe/spire/pkg/common/telemetry"
 )
 
+const (
+	readBufferSize = 4096
+)
+
 type Server interface {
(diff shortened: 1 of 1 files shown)


amartinezfayo (approved):
🎉

Issue #6867, opened 2026-04-13
x509pop workload api support


kfox1111:
Currently you can use the spiffe-helper along with the spire-agent with x509pop node attestor to exchange a workload svid for a full node svid. This works very well for nested environments where you may want to have a spire-agent on the host and one in k8s, or you want to exchange the pod identity for the vm running inside it.

But, to do this, you need to write out the keys to disk. Instead I'd like to modify the x509pop agent to optionally be configured to just get the cert/key/trustbundle via the spiffe workload api directly, simplifying the process.

tjons:
We have a custom attester plugin that does something very similar. We have a trust domain from which we issue machine identities, where each machine identity is an alias referencing every node in the trust domain. We then have a number of trust domains that federate with this root trust domain, which we call "trunk" trust domains. Each agent in the trunk trust domain is registered as a workload in the root trust domain, giving it a unique SPIFFE ID. Our plugin provides the ability for the trunk agents to retrieve an X.509 SVID from the root trust domain via the root agent's workload API socket, and then perform x509pop attestation in the `trunk` trust domain _via_ the agent SVID from the root trust domain. 

It's been a fairly nifty solution to allow us to decouple infrastructure trust (the root trust domain) from the workload trust domains (the trunks).

kfox1111:
@tjons thanks fro sharing. :)

I was thinking of something very similar. The remaining missing feature I think after this pr to make it very generic, would be to enable the mode=spiffe setting on the server to also specify a federated trust domain name to use.

This piece is useful all by itself though.

kfox1111:
This merged.

PR #6884, opened 2026-04-17
x509pop agent workload api support

kfox1111:
- [x] Documentation updated?

Enable using a spiffe workload api socket to get the certificate for a x509pop node attested agent.

**Which issue this PR fixes**
https://github.com/spiffe/spire/issues/6867


sorindumitru (commented):
This looks ok, @kfox1111, thanks for the changes! Would it be possible to test the changes using something like the [fake workload api](https://github.com/spiffe/spire/blob/main/test/fakes/fakeworkloadapi/workloadapi.go) implementation?

kfox1111:
> This looks ok, @kfox1111, thanks for the changes! Would it be possible to test the changes using something like the [fake workload api](https://github.com/spiffe/spire/blob/main/test/fakes/fakeworkloadapi/workloadapi.go) implementation?

Done

pkg/agent/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -118,14 +122,14 @@ func (p *Plugin) AidAttestation(stream nodeattestorv1.NodeAttestor_AidAttestatio
...
-func (p *Plugin) Configure(_ context.Context, req *configv1.ConfigureRequest) (*configv1.ConfigureResponse, error) {
+func (p *Plugin) Configure(ctx context.Context, req *configv1.ConfigureRequest) (*configv1.ConfigureResponse, error) {
 	newConfig, _, err := pluginconf.Build(req, buildConfig)
 	if err != nil {
 		return nil, err
 	}
 
 	// make sure the configuration produces valid data
-	if _, err := loadConfigData(newConfig); err != nil {
+	if _, err := loadConfigData(ctx, newConfig); err != nil {
```
sorindumitru: We were discussing this between the maintainers and we usually try to keep the `Configure` call offline. Could we only load the cert from the worload api during attestation? We do the same in other plugins, for example [spire UpstreamAuthority](https://github.com/spiffe/spire/blob/7e6f00b327af7b853a717ecdb89478c36b054a33/pkg/server/plugin/upstreamauthority/spire/spire.go#L105)

pkg/agent/plugin/nodeattestor/x509pop/x509pop_test.go
```diff
@@ -68,6 +75,69 @@ func (s *Suite) TestAttestSuccessWithIntermediates() {
...
 }
 
+func getTestAddress(path string) string {
+	if runtime.GOOS == "windows" {
+		return "npipe:" + strings.TrimPrefix(path, `\\.\pipe\`)
+	}
+	return "unix://" + path
+}
+
+func (s *Suite) TestAttestSuccessWithWorkloadAPI() {
```
sorindumitru, line 86: Could we also have a negative test for the attestation?

pkg/agent/plugin/nodeattestor/x509pop/x509pop.go
kfox1111: Wouldnt that cause a lot more round trips at a critical time to the socket when a bunch of spire-agents try and attest to the server in a batch?

sorindumitru: I don't think that's an issue. You could say that the same happens when the system starts up, so you're likely to run into that issue either way.

kfox1111: Doh, your right. I was mistakenly thinking of the server side, not the agent side. Its been a long week....

Yeah, pulling it in during attestation makes sense to me too.

kfox1111: hmm....

It looks like loadConfigData happens in two places. Configure and AidAttestation. during configure, it has the comment 
```
// make sure the configuration produces valid data
```

Which makes sense to do I think?

So other then configure check time, its only loading it during attestation.

Do we remove the config time check?

sorindumitru: I think it's good to still load the certs from the filesystem, it's just about operations that may take a long time.

pkg/agent/plugin/nodeattestor/x509pop/x509pop_test.go
kfox1111: done

pkg/agent/plugin/nodeattestor/x509pop/x509pop.go
kfox1111: ok. done.

test/fixture/nodeattestor/x509pop/generate.go
```diff
@@ -38,6 +38,7 @@ func main() {
 		SerialNumber:          big.NewInt(0x4d5e6f),
 		BasicConstraintsValid: true,
 		IsCA:                  true,
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
```
sorindumitru, line 42: Was this a new requirement that you encountered somewhere? Should we also add it to the root CA in that case?

kfox1111: sort of. the certs in the test never ran through spiffe workload api before. the go spiffe code does some sanity checks that break now when running through the new spiffe socket test code

sorindumitru: That makes sense, could you also add then to the root? I imagine it should have them too

kfox1111: done

sorindumitru (approved):
LGTM, thanks @kfox1111!

PR #6889, opened 2026-04-18
Add HashiCorp Vault Key Manager plugin

MarcosDY:
**Summary**
Add a new built-in KeyManager "hashicorp_vault" server plugin that stores and rotates SPIRE server keys using HashiCorp Vault's [Transit secret engine](https://developer.hashicorp.com/vault/docs/secrets/transit). It also refactors the existing upstreamauthority/vault plugin to share a common Vault client package, eliminating significant code duplication between the two plugins.

**What's new**
pkg/server/plugin/keymanager/hashicorpvault — New key manager plugin. Keys are created in Vault's Transit engine under a _<UUID>-<SPIRE-key-id>_ naming scheme to support HA deployments where multiple server instances must keep their keys distinct. 

Authentication methods: Token, Client Certificate, AppRole, and Kubernetes auth are all supported (matching the upstream authority plugin's feature set). Credentials can be supplied via config or environment variables.

Integration tests: A new key-manager-vault suite in test/integration/suites/ exercises all four auth methods against a real Vault instance in a kind cluster.

Thanks to @InverseIntegral for writing this plugin!!!


pkg/server/common/vault/client.go
```diff
@@ -19,17 +22,19 @@ import (
...
-	envVaultNamespace       = "VAULT_NAMESPACE"
+	EnvVaultAddr              = "VAULT_ADDR"
+	EnvVaultToken             = "VAULT_TOKEN"
+	EnvVaultClientCert        = "VAULT_CLIENT_CERT"
+	EnvVaultClientKey         = "VAULT_CLIENT_KEY"
+	EnvVaultCACert            = "VAULT_CACERT"
+	EnvVaultAppRoleID         = "VAULT_APPROLE_ID"
+	EnvVaultAppRoleSecretID   = "VAULT_APPROLE_SECRET_ID" // #nosec G101
+	EnvVaultNamespace         = "VAULT_NAMESPACE"
+	EnvVaultTransitEnginePath = "VAULT_TRANSIT_ENGINE_PATH"
```
sorindumitru, line 34: Is this a standard environment variable for vault? I didn't find anything about it. If it isn't, it may be worth having a comment to specify that. Also needs mentioning in the docs.

pkg/server/common/vault/client.go
```diff
@@ -414,3 +466,173 @@ func (c *Client) SignIntermediate(ttl string, csr *x509.CertificateRequest) (*Si
...
+// See: https://developer.hashicorp.com/vault/api-docs/secret/transit#sign-data
+func (c *Client) SignData(ctx context.Context, keyName string, data []byte, hashAlgo TransitHashAlgorithm, signatureAlgo TransitSignatureAlgorithm) ([]byte, error) {
+	encodedData := base64.StdEncoding.EncodeToString(data)
+
+	body := map[string]any{
+		"input":                 encodedData,
+		"marshalling_algorithm": "asn1",
+		"prehashed":             true,
+	}
+	if signatureAlgo != TransitSignatureAlgorithmNone {
```
sorindumitru, line 519: Should we support the none algorithm? Might be safer to just not allow it.

pkg/server/plugin/upstreamauthority/vault/vault.go
```diff
@@ -209,6 +153,7 @@ func (p *Plugin) MintX509CAAndSubscribe(req *upstreamauthorityv1.MintX509CAReque
 	p.mtx.Lock()
 	defer p.mtx.Unlock()
 
+	// TODO: this is flaky, we should have a better way to manage the Vault client lifecycle
```
sorindumitru: Is this still flaky?

sorindumitru (commented):
Did not get a chance to go through most of it yet, I'll continue later this week

pkg/server/common/vault/client.go
```diff
@@ -44,11 +49,44 @@ const (
...
+	KeyType string
+	KeyData map[string]any
+}
+
 // ClientConfig represents configuration parameters for vault client
 type ClientConfig struct {
 	Logger hclog.Logger
 	// vault client parameters
-	clientParams *ClientParams
+	ClientParams *ClientParams
```
sorindumitru, line 90: Why are we exporting this member of the structure?

pkg/server/common/vault/client.go
```diff
@@ -414,3 +466,173 @@ func (c *Client) SignIntermediate(ttl string, csr *x509.CertificateRequest) (*Si
...
 	return resp, nil
 }
+
+// CreateKey creates a new key in the specified transit secret engine
+// See: https://developer.hashicorp.com/vault/api-docs/secret/transit#create-key
+func (c *Client) CreateKey(ctx context.Context, keyName string, keyType TransitKeyType) error {
+	arguments := map[string]any{
+		"type":       keyType,
+		"exportable": false, // SPIRE keys are never exportable
+	}
```
sorindumitru, line 478: We should also set `auto_rotate_period` to 0, just in case. It's currently the default, but just in case it will change.

pkg/server/common/vault/client.go
```diff
@@ -414,3 +466,173 @@ func (c *Client) SignIntermediate(ttl string, csr *x509.CertificateRequest) (*Si
...
+	if listResp == nil {
+		return []*KeyEntry{}, nil
+	}
+
+	keys, ok := listResp.Data["keys"]
+	if !ok {
+		return nil, status.Errorf(codes.Internal, "transit engine list keys call was successful but keys are missing")
+	}
+
+	keyNames, ok := keys.([]any)
```
sorindumitru, line 571: Would it make sense to cast to `[]string` directly?

pkg/server/plugin/keymanager/hashicorpvault/hashicorp_vault.go
```diff
@@ -0,0 +1,599 @@
...
+
+	spireKeyID := req.KeyId
+	newKeyEntry, err := p.createKey(ctx, spireKeyID, req.KeyType)
+	if err != nil {
+		return nil, err
+	}
+
+	if keyEntry, ok := p.getKeyEntry(spireKeyID); ok {
+		select {
+		case p.scheduleDelete <- keyEntry.KeyName:
```
sorindumitru: We fetch previous key from our in-memory cache and schedule it for deletion because it must not be in use anymore, right? Wouldn't there be some window where a key fails to be deleted if the server restarts at the wrong time?

pkg/server/common/vault/client.go
MarcosDY: TransitSignatureAlgorithmNone is safe,
it doesn't skip or weaken signing. It's used only for ECDSA keys (P-256/P-384) where Vault's signature_algorithm field (pss/pkcs1v15) is RSA-specific and must be omitted from the request; sending it would cause a Vault API error.

pkg/server/plugin/upstreamauthority/vault/vault.go
MarcosDY: maybe not :D

pkg/server/common/vault/client.go
MarcosDY: I use them on test (key manager and upstream authority, an alternative is to return a getter

pkg/server/common/vault/client.go
MarcosDY: No=, Vault SDK deserializes all JSON into map[string]any using encoding/json, which always produces []interface{}  for arrays

pkg/common/telemetry/names.go
```diff
@@ -210,9 +210,18 @@ const (
 	// CertFilePath tags a certificate file path used for TLS connections.
 	CertFilePath = "cert_file_path"
 
+	// KeyID tags a key ID
+	KeyID = "key_id"
+
 	// KeyFilePath tags a key file path used for TLS connections.
 	KeyFilePath = "key_file_path"
 
+	// KeyID tags the key nane
```
amartinezfayo:
```suggestion
	// KeyName tags the key name
```

doc/plugin_server_keymanager_hashicorp_vault.md
```diff
@@ -0,0 +1,187 @@
...
+| `VAULT_CACERT`              | Yes      | Path to a CA certificate file                                 |
+| `VAULT_CLIENT_CERT`         | Yes      | Path to a client certificate file (cert auth)                 |
+| `VAULT_CLIENT_KEY`          | Yes      | Path to a client private key file (cert auth)                 |
+| `VAULT_NAMESPACE`           | Yes      | Vault namespace (Enterprise only)                             |
+| `VAULT_TOKEN`               | Yes      | Vault token (token auth)                                      |
+| `VAULT_APPROLE_ID`          | **No**   | AppRole role ID — SPIRE-specific, not a Vault SDK variable    |
+| `VAULT_APPROLE_SECRET_ID`   | **No**   | AppRole secret ID — SPIRE-specific, not a Vault SDK variable  |
+| `VAULT_TRANSIT_ENGINE_PATH` | **No**   | Transit engine path — SPIRE-specific, falls back to `transit` |
+
+The plugin supports **Client Certificate**, **Token** and **AppRole** authentication methods.
```
amartinezfayo: I wonder if this sentence should also list the Kubernetes auth method. Leaving Kubernetes out here might lead a reader to assume only three methods are supported. The same line exists in `doc/plugin_server_upstreamauthority_vault.md`, so we may want to update both.

test/integration/suites/key-manager-vault/conf/spire.hcl
```diff
@@ -0,0 +1,3 @@
+path "transit/*" {
+  capabilities = [ "create", "read", "update", "delete", "list" ]
```
amartinezfayo: Should we consider scoping this policy to match the minimum policy described in the doc (paths under `transit/keys/*` and `transit/sign/*` rather than `transit/*`)? With the broader `transit/*` here, a future change that ends up needing a capability outside the documented set would still pass the integration suite, and the doc could silently drift from what actually works.

pkg/server/plugin/keymanager/hashicorpvault/hashicorp_vault.go
```diff
@@ -0,0 +1,637 @@
...
+	}
+
+	serverID := config.KeyIdentifierValue
+	p.logger.Debug("Loaded server id", "server_id", serverID)
+
+	if config.InsecureSkipVerify {
+		p.logger.Warn("TLS verification of Vault certificates is skipped. This is only recommended for test environments.")
+	}
+
+	p.mu.Lock()
```
amartinezfayo, line 120: I'm wondering if it would be worth releasing `p.mu` while `genVaultClient` and `vc.GetKeys` do their Vault round-trips. The way I read this, the lock is held from here through the network calls until `setCache` returns, so any concurrent `GenerateKey`, `SignData`, or `GetPublicKey` will stall for the duration of authentication and the keys-list call. In other plugins we build the client and fetch key entries before taking `p.mu` and only locking for the cache mutation, avoiding that stall. Would it make sense to mirror that ordering here?

amartinezfayo: It turned out that implementing this suggestion would require a lot of code changes, which doesn’t seem worth the benefit.

amartinezfayo (approved):
LGTM!

PR #6897, opened 2026-04-20
cli: add shorthand aliases to server and agent commands

### LINKED ISSUE #6790: short vs long args
spire uses the less conventional single dash with long name for all arguments. In a future build it would be nice to support both long and short names like:
```
--parentID|-p
```

Pittu-Sharma:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
CLI flag parsing and help documentation for both SPIRE Server and SPIRE Agent.

**Description of change**
This PR introduces short-form flag aliases (shorthands) across the primary SPIRE command suites to improve usability and align with standard POSIX CLI conventions. 

Key improvements:
- **Flag Infrastructure**: Added reusable shorthand helper functions in `pkg/common/cli/flags.go` that wrap the standard Go `flag` package to support aliases without introducing external dependencies.
- **Applied Aliases**: Integrated consistent shorthands for frequently used flags, including:
  - `-c` for `-config`
  - `-d` for `-dataDir`
  - `-v` for `-logLevel` / `-verbose`
  - `-s` for `-socketPath` / `-shallow`
  - `-i` for `-id` / `-spiffeID`
  - `-p` for `-parentID` / `-path` / `-namedPipeName`
- **Documentation**: Updated the corresponding reference tables in `doc/spire_server.md` and `doc/spire_agent.md`.
- **Validation**: Updated existing CLI test suites to verify both long and short-form flag variants and confirmed that all help-output assertions pass.

**Which issue this PR fixes**
fixes #6790


sorindumitru:
Thanks @Pittu-Sharma for opening this PR. We do want to have support for short flags, but we'd want to first switch to a flag library that has native support for this. This avoids the need to have both flags appear as separate options in the help output.

If we do make this change, they would have to happen in a backwards compatible way so that no existing users encounter breaking changes.

sorindumitru:
@Pittu-Sharma, another option we could explore here is to modify the [Usage](https://pkg.go.dev/flag#pkg-variables) variable to de-duplicate the output in the help line, so it could say something like:

```
-socketPath, -s string
    Path to bind the SPIRE Server API socket to
```

I don't know if it's going to work very well with our subcommands, but if it does and the help out still looks reasonable that may be a good way forward.

sorindumitru:
Closing this for now, since there's no progress

PR #6901, opened 2026-04-21
sigstore: promote out of experimental

### LINKED ISSUE #6900: sigstore: promote out of experimental
sigstore has been available for a while now and it's now also tested as part of the integration test. It's to move the sigstore support out of the experimental block.

The support will still remain optional, this is just about making it officially supported.

mateenali66:
moves the sigstore block to the top level of the k8s and docker workload attestor configs. the `experimentalK8SConfig` and `experimentalConfig` wrapper structs are gone and `buildConfig` reads from `newConfig.Sigstore` directly.

per discussion on the issue, this is a hard cutover with no deprecation window. existing configs using `experimental { sigstore {...} }` need to move the `sigstore` block out of the wrapper. sample configs, plugin docs, and the k8s-sigstore integration test suite are updated to match.

tested locally with `go test ./pkg/agent/...` and `make lint-code`.

Fixes #6900


amartinezfayo (approved):
Thank you for this contribution!

PR #6906, opened 2026-04-23
Reject stale experimental sigstore config on upgrade

amartinezfayo:
The k8s workload attestor previously decoded its HCL config without an `unusedKeyPositions` map, so unknown top-level keys were silently accepted. This adds `UnusedKeyPositions` to `HCLConfig` and reports an `unknown configurations detected: ...` error when any are found, matching the behavior of the docker workload attestor.

In particular, a stale `experimental { sigstore { ... } }` block left over from the sigstore promotion (#6901) is now rejected, giving operators a clear signal that the `sigstore` block needs to move to the top level of `plugin_data`.

This PR also extracts the duplicated unused-key reporting from the docker and k8s workload attestors and the `gcp_secretmanager` SVID store into a new shared `pluginconf.ReportUnusedKeys` helper and adds the missing `sigstore` row to the docker workload attestor configuration table, so its docs match the k8s workload attestor.

### DIFF
--- a/pkg/agent/plugin/workloadattestor/k8s/k8s.go
+++ b/pkg/agent/plugin/workloadattestor/k8s/k8s.go
@@ -20,6 +20,7 @@ import (
 	"github.com/andres-erbsen/clock"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/hcl"
+	hcltoken "github.com/hashicorp/hcl/hcl/token"
 	workloadattestorv1 "github.com/spiffe/spire-plugin-sdk/proto/spire/plugin/agent/workloadattestor/v1"
 	configv1 "github.com/spiffe/spire-plugin-sdk/proto/spire/service/common/config/v1"
 	"github.com/spiffe/spire/pkg/agent/common/sigstore"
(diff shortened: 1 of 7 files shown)


pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -135,6 +136,8 @@ type HCLConfig struct {
 
 	// Sigstore contains sigstore specific configs.
 	Sigstore *sigstore.HCLConfig `hcl:"sigstore,omitempty"`
+
+	UnusedKeyPositions map[string][]hcltoken.Pos `hcl:",unusedKeyPositions"`
```
sorindumitru, line 140: Would this also cause issues for users that have extra keys in the non-experimental configuration? Not an issue since this is going in a minor release, but might be good to remember to point out in the changelog.

sorindumitru (approved):

pkg/agent/plugin/workloadattestor/k8s/k8s.go
amartinezfayo: Yes, I thought about that. I'll make sure we call out this in the CHANGELOG

Issue #6769, opened 2026-03-20
Optionally verify client certificate IP address


sorindumitru:
In x509pop and sshpop plugins we could additionally verify the IP address of the client. This would allow administrators which have stable IP addresses to ensure that a client certificate would only be used from the intended location.

This would need to be made optional since current certs likely don't have any IP addresses specified. It would also likely have to stay optional going forward since it wouldn't work with load balancers in front of the server.

To make this work we'd need to:
* Pass the incoming client IP address to the node attestor plugins
* Verify the ip address is present on the certificates the agent presents.

tjons:
big +1, this would be really helpful. it would be great to add something similar for the httpchallenge plugin too.

jsnctl:
> Pass the incoming client IP address to the node attestor plugins

Was having a look at what changes this might require - would we take a similar approach to passing via a header as with the Host per:

https://github.com/spiffe/spire/blob/51beb341b5640f0bb40292d0ea85ec20d507d340/pkg/server/plugin/nodeattestor/v1.go#L17-L23

Is there an alternative whereby we extend the protos passed in the request (or a similarly extensible approach) to include the IP?

jsnctl:
x509pop implementation + plugin changes in this PR https://github.com/spiffe/spire/pull/6911

If we're happy with this approach, I can roll out across the other relevant attestors and update the documentation accordingly

sorindumitru:
Thanks @jsnctl for taking this on. Yes, the current implementation you have seems like a good option and something that can be generalised to the other 2 plugins.

samuraixwandering:
I'd like to take the remaining work on this if it's still up for grabs.

I had a look at what landed in #6911. The facade already forwards the observed peer IP via `X-Forwarded-Client-IP`, and `x509pop` has optional `verify_client_ip` against the leaf IP SANs. Docs for x509pop are updated. What is left looks like rolling the same option out to the other two plugins called out in the thread (sshpop and httpchallenge), plus docs.

For x509pop the check is straightforward because the leaf has IP SANs. sshpop is SSH certificates, so there is no IP SAN in the same sense. Before I write code I want to confirm the intended rule there.

My working assumption:
1. httpchallenge: optional `verify_client_ip`, same metadata key, fail closed when enabled and the IP is missing or does not match whatever host/address material the plugin already trusts.
[...]

PR #6911, opened 2026-04-28
feat: Support for optional verification of client cert IPs (x509pop)

jsnctl:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

* v1 implementation of server plugins examine new forwarded header
* x509pop server plugin

**Description of change**

Adds optional client cert IP verification, specifical for x509pop. Other attestors will benefit from this, but I wanted to get buy-in on these changes before rolling out across the board.

* Switchable with verify_client_ip feature flag
* When enabled, connecting peer's IP is extracted from the gRPC context
  header, forwarded a X-Forwarded-Client-IP, and validated against the
  leaf cert IP SANs

**Which issue this PR fixes**

Partial (x509pop) implementation of https://github.com/spiffe/spire/issues/6769


pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -253,6 +259,27 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
...
+		for _, certIP := range leaf.IPAddresses {
+			if certIP.Equal(clientIP) {
+				matched = true
+				break
+			}
+		}
+		if !matched {
+			return status.Errorf(codes.PermissionDenied, "client IP %s does not match any certificate IP SAN", clientIP)
+		}
+	}
```
sorindumitru, line 284: We should do this after we verify the presented certificate is valid, to avoid parsing and using potentially invalid input. Avoids issues with processing large number of IP SANs.

jsnctl: Is this later in the `Attest` method, or elsewhere in the codebase? I note there's some validation (or perhaps just parsing) earlier in this method, or did you have something else in mind? 

https://github.com/jsnctl/spire/blob/97dbf524e646ea4bf86493ad932bfb9bcf1c1864/pkg/server/plugin/nodeattestor/x509pop/x509pop.go#L219-L222

sorindumitru: Yes, I realised after I wrote it that parsing is not the best word there since we still parse. We just don't trust it for us to iterate through the list of ip address before we verify that it is signed by a CA that we trust.

This avoid DOS issues where someone sends us arbitrary certificates with long lists of IP address and we just spend time going through them for no reason. Not that big of an issue but we should only do as much as is actually needed with the certificate before we verify it.

jsnctl: Ah I see what you mean. I think moving this new logic to after the `VerifyChallengeResponse()` step will address this - we're happy at this stage the X.509 is legit, and we can perform the extra scrutiny on the IP SANs if this is enabled.

I've pushed an adjustment here https://github.com/spiffe/spire/pull/6911/changes/571d77934770017c80e77e0c1e894a32b47a5e97

jsnctl: The above was failing the unit tests due to the adjustments, fixed now

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -288,6 +294,27 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
 		return status.Errorf(codes.PermissionDenied, "challenge response verification failed: %v", err)
 	}
 
+	if config.verifyClientIP {
+		ips := metadata.ValueFromIncomingContext(stream.Context(), nodeattestor.XForwardedClientIPKey)
+		if len(ips) == 0 || ips[0] == "" {
+			return status.Error(codes.Internal, "client IP not available for verification")
+		}
```
sorindumitru: We may want to eventually put this somewhere else, but I think this is ok for now.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -288,6 +294,27 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
...
+		clientIP := net.ParseIP(ips[0])
```
sorindumitru: This should always contain a single field in our case, right? Could we log something if that's not the case?

pkg/server/plugin/nodeattestor/x509pop/x509pop_test.go
```diff
@@ -553,6 +556,107 @@ func expectNoChallenge(context.Context, []byte) ([]byte, error) {
 	return nil, errors.New("challenge is not expected")
 }
 
+func (s *Suite) TestVerifyClientIP() {
+	clientIP := net.ParseIP("192.0.2.1")
```
sorindumitru, line 560: Could we also add a test to verify that IPv6 is accepted? I think that should already be the case.

sorindumitru (commented):
Sorry for the delay @jsnctl, just a few small comments

sorindumitru:
We should also document the new flag in the documentation for the plugin.

sorindumitru:
Hi @jsnctl, checking in to see if you think you'll have time to address the review comments. If not, we can push the changes on top of the branch.

jsnctl:
Hi @sorindumitru - apologies I've been busy with new job. I'll have some time in the next week or so to push changes, but in the meantime if you'd prefer to get this merged then feel free to push

sorindumitru (approved):
LGTM, thanks @jsnctl !

PR #6917, opened 2026-04-30
Add a nightly job to update AWS EC2 CAs

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Description of change**

We maintain a list of certificates that are truested when validating AWS EC2 IIDs. We do not regularly update it with new CAs, although new ones are created from time to time. The rsa1024 ones are also somewhat shorter lived, so we should have a way to automate this.

This introduces another job in the nightly run that will parse the list of CAs from AWS, update our go files with the CA and open a PR for us to merge in.

https://github.com/sorindumitru/spire/pull/4 is an example of what it does (although it messed up by including another commit).


kfox1111:
well... if it is needing to be updated this frequently, it is probably better to add support to spire to do it so we're not having to build new releases every time they release a ca?

sorindumitru:
> well... if it is needing to be updated this frequently, it is probably better to add support to spire to do it so we're not having to build new releases every time they release a ca?

My mistake, the ones that have earlier timestamps are rsa1024, maybe that just points to a phasing out of that? I think in general the new CAs appear when new zones are added so a PR to update them is fine.

If these are available through some API, we can consider fetching them at runtime, but it didn't seem like there is one. At least [this doc](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-iid.html) doesn't specify a different method.

pkg/server/plugin/nodeattestor/awsiid/aws-ca-parser/main.go
```diff
@@ -0,0 +1,265 @@
...
+		vn := regionToVar[e.region]
+		if !seen[vn] {
+			seen[vn] = true
+			vars = append(vars, varEntry{vn, varToPEM[vn], varToRegions[vn]})
+		}
+	}
+
+	var buf bytes.Buffer
+	fmt.Fprintf(&buf, "package %s\n\n", packageName)
+	fmt.Fprintln(&buf, "const (")
```
amartinezfayo, line 216: I think running this generator as is would break the build of the package it overwrites, if I follow the logic correctly.
The current awsrsa1024/cacerts.go exports an AWSCACert constant that getAWSCACertificate in awsca.go falls back to whenever a region is missing from `CACerts`, but as I read generateFile, the generator emits only per-region constants like usEast1Cert and never produces AWSCACert. Once the bot replaces the file, awsrsa1024.AWSCACert would be undefined and the generated PR's CI would fail to compile.
Am I missing something? Should we either teach the generator to emit AWSCACert (mapping it to whatever AWS lists as the default all-regions cert), or drop that fallback in `awsca.go` as part of this change?

pkg/server/plugin/nodeattestor/awsiid/aws-ca-parser/main.go
```diff
@@ -0,0 +1,265 @@
...
+//
+//	## US East (N. Virginia) — us-east-1
+//	#### [ RSA ]
+//	```
+//	-----BEGIN CERTIFICATE-----
+//	...
+//	-----END CERTIFICATE-----
+//	```
+func fetchCerts() (map[string]regionEntry, error) {
+	resp, err := http.Get(certsURL)
```
amartinezfayo: I wonder if we should check `resp.StatusCode` before scanning, and swap the bare `http.Get` for an `http.Client` with a `Timeout`. If AWS rotates the URL or returns a 4xx or 5xx with an HTML error body, the scanner would just walk it and the only signal the workflow surfaces is the generic "no certificates found; page structure may have changed". Including the status code and the URL in that error path would make a future failing nightly easier to triage, and a real timeout would protect us from a stuck connection consuming the workflow's full duration.

pkg/server/plugin/nodeattestor/awsiid/aws-ca-parser/main.go
```diff
@@ -0,0 +1,265 @@
...
+// The markdown structure is:
+//
+//	## US East (N. Virginia) — us-east-1
+//	#### [ RSA ]
+//	```
+//	-----BEGIN CERTIFICATE-----
+//	...
+//	-----END CERTIFICATE-----
+//	```
+func fetchCerts() (map[string]regionEntry, error) {
```
amartinezfayo, line 32: Maybe we could add a small unit test for `fetchCerts` against a fixture markdown file checked into testdata? That would allow us to catch regressions in the regexes during PR review rather than on the next nightly run.

.github/workflows/nightly_build.yaml
```diff
@@ -38,3 +38,49 @@ jobs:
...
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          BRANCH="bot/update-aws-ca-certs"
+          git checkout -b "$BRANCH"
+          git add \
+            pkg/server/plugin/nodeattestor/awsiid/awsrsa1024/cacerts.go \
+            pkg/server/plugin/nodeattestor/awsiid/awsrsa2048/cacerts.go
+          git commit -s -m "chore: update AWS EC2 CA certificates"
+          git push --force origin "$BRANCH"
```
amartinezfayo: Would it be worth using `--force-with-lease` instead of plain `--force` here? Force pushing is fine while only the workflow writes to this branch, but if we ever push a commit on top of `bot/update-aws-ca-certs` the next nightly run silently overwrites it.
I think `--force-with-lease` would refuse the push in that case and surface the conflict instead.

amartinezfayo (commented):
Thank you @sorindumitru for this!

pkg/server/plugin/nodeattestor/awsiid/aws-ca-parser/main.go
```diff
@@ -0,0 +1,274 @@
...
+	for _, e := range entries {
+		vn := regionToVar[e.region]
+		if !seen[vn] {
+			seen[vn] = true
+			vars = append(vars, varEntry{vn, varToPEM[vn], varToRegions[vn]})
+		}
+	}
+
+	var buf bytes.Buffer
+	fmt.Fprintf(&buf, "package %s\n\n", packageName)
```
amartinezfayo, line 215: I'm wondering if the generated files should carry a "// Code generated by ... DO NOT EDIT." header, since a nightly bot overwrites them and a contributor could otherwise hand-edit cacerts.go and lose it on the next run. The RSA-2048 file's doc comment above its const block would also be dropped, since the template emits only the package clause and the const and map blocks. Would it be worth emitting the header, and keeping a short descriptive comment?

pkg/server/plugin/nodeattestor/awsiid/aws-ca-parser/main.go
```diff
@@ -0,0 +1,274 @@
...
+	return sb.String() + "Cert"
+}
+
+// certExpiry parses the NotAfter date from a PEM certificate.
+func certExpiry(pemStr string) string {
+	block, _ := pem.Decode([]byte(pemStr))
+	if block == nil {
+		return ""
+	}
+	cert, err := x509.ParseCertificate(block.Bytes)
```
amartinezfayo, line 151: I think it could be worth having the generator hard-fail if a cert block does not decode or parse as an X.509 certificate. As I read it, only certExpiry parses, and it returns an empty string on error, so a malformed entry from the AWS page would still be written into cacerts.go and only surface as a runtime attestation failure rather than a failed nightly. Since crypto/x509 is already imported, validating each PEM during generation would catch a bad page during the run.

pkg/server/plugin/nodeattestor/awsiid/aws-ca-parser/main.go
```diff
@@ -0,0 +1,276 @@
...
+		vn := regionToVar[e.region]
+		if !seen[vn] {
+			seen[vn] = true
+			vars = append(vars, varEntry{vn, varToPEM[vn], varToRegions[vn]})
+		}
+	}
+
+	var buf bytes.Buffer
+	fmt.Fprintf(&buf, "package %s\n\n", packageName)
+	fmt.Fprintf(&buf, "\n// Code generated by aws-ca-parser. DO NOT EDIT\n")
```
amartinezfayo: I checked this against go/ast.IsGenerated, and the marker as emitted here is not recognized (returns false) for two independent reasons: it needs to sit before the package clause rather than after it, and per go.dev/s/generatedcode it has to match "^// Code generated .* DO NOT EDIT\.$", so it needs a trailing period. With both fixes (marker line before the package clause, ending in a period) IsGenerated returns true.
Can we change this to emit it that way?

amartinezfayo (approved):
Thank you @sorindumitru!

PR #6924, opened 2026-05-01
[gcp_kms] add small sleep to get public key

### LINKED ISSUE #6923: gcp_kms keymanager plugin gives up too quickly when getting the public key for an rsa-4096 size key
It appears it finishes its 10 attempts to get the key in under a second, which is not long enough for my deployment it seems.

I will make a PR to double the number of max attempts if using rsa-4096.

daescha:
10 attempts is too short for some deployments with rsa-4096, adding small sleep increases the chance it will work.

**Pull Request check list**

- [x] Documentation updated?
  - not needed

**Affected functionality**

**Description of change**

add sleep with exponential backoff between attempts to get the public key when using gcp_kms

**Which issue this PR fixes**

fixes #6923


pkg/server/plugin/keymanager/gcpkms/gcpkms.go
```diff
@@ -1114,6 +1113,9 @@ func getPublicKeyFromCryptoKeyVersion(ctx context.Context, log hclog.Logger, kms
 		log.Warn("Could not get the public key because the CryptoKeyVersion is still being generated. Trying again.")
 		attempts++
 		kmsPublicKey, errGetPublicKey = kmsClient.GetPublicKey(ctx, &kmspb.GetPublicKeyRequest{Name: cryptoKeyVersionName})
+		if errGetPublicKey != nil && algorithm == kmspb.CryptoKeyVersion_RSA_SIGN_PKCS1_4096_SHA256 {
+			time.Sleep(300 * time.Millisecond)
```
amartinezfayo: I think that we should use the plugin's clock abstraction rather than calling `time.Sleep` directly, which lets tests drive retries with a mock clock instead of accruing real wall-time delay. I also think that we should align with the existing backoffMin/backoffMax exponential backoff pattern that we already have, rather than a fixed 300ms.

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
```diff
@@ -1078,10 +1078,9 @@ func getOrCreateServerID(idPath string) (string, error) {
 
 // getPublicKeyFromCryptoKeyVersion requests Cloud KMS to get the public key
 // of the specified CryptoKeyVersion.
-func getPublicKeyFromCryptoKeyVersion(ctx context.Context, log hclog.Logger, kmsClient cloudKeyManagementService, cryptoKeyVersionName string) ([]byte, error) {
+func getPublicKeyFromCryptoKeyVersion(ctx context.Context, log hclog.Logger, kmsClient cloudKeyManagementService, cryptoKeyVersionName string, algorithm kmspb.CryptoKeyVersion_CryptoKeyVersionAlgorithm) ([]byte, error) {
 	kmsPublicKey, errGetPublicKey := kmsClient.GetPublicKey(ctx, &kmspb.GetPublicKeyRequest{Name: cryptoKeyVersionName})
```
amartinezfayo, line 1086: I think the gate on a specific algorithm constant could be dropped: a small unconditional delay would be simpler and would also help in the (rare) cases where smaller key sizes hit the same issue. If we wanted to keep the gate, the algorithm is already available on `cryptoKeyVersion.Algorithm` from the `kmsClient.GetCryptoKeyVersion` response inside the loop, so the new parameter and the call-site threading could be avoided, though I'd lean toward dropping the gate entirely.

amartinezfayo (commented):
Thank you @daescha for this contribution!

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
daescha: Makes sense to me :+1:

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
daescha: this makes sense to me, though it requires knowledge I dont have on how this codebase works, so it will take me a little extra time to figure out how to implement what you are asking for.

daescha: @amartinezfayo how would you like me to pass the clk in for use by the function? was thinking making it a method on plugin but use of the function by the keyfetcher struct would prevent that from being as simple of a chbage as I would like. Just pass in as parameter?

amartinezfayo: Passing it as a parameter sounds good. Since `keyFetcher` already holds `kmsClient`, `log`, etc. as fields, I'd suggest adding a `clk clock.Clock` field there and setting it from `p.hooks.clk` at the construction site in `Configure`. Then `getPublicKeyFromCryptoKeyVersion` can take a `clock.Clock` argument that the `*Plugin` callers pass as `p.hooks.clk` and the keyFetcher caller passes as `kf.clk`.
It would be great if we could also have the wait follow the exponential backoff pattern already used in `scheduleDestroyTask` (`backoffMin`/`backoffMax` with `clk.Sleep(backoff)`) rather than a flat 300ms. That could need some changes across the testing, so if that gets too complex, we can start with a fixed 300ms duration.

daescha: @amartinezfayo done.

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
```diff
@@ -1114,6 +1118,10 @@ func getPublicKeyFromCryptoKeyVersion(ctx context.Context, log hclog.Logger, kms
 		log.Warn("Could not get the public key because the CryptoKeyVersion is still being generated. Trying again.")
 		attempts++
 		kmsPublicKey, errGetPublicKey = kmsClient.GetPublicKey(ctx, &kmspb.GetPublicKeyRequest{Name: cryptoKeyVersionName})
+		if errGetPublicKey != nil {
+			backoff = min(backoff*2, backoffMax)
+			clk.Sleep(backoff)
```
amartinezfayo, line 1123: @daescha I ran the unit-test jobs locally and I think I know what's happening.
SPIRE's mock clock blocks on `Sleep` until the clock is advanced, and nothing advances it inside this retry loop, so the cases that retry more than once deadlock until the test timeout. The two that hit this are `TestConfigure`'s "get public key error max attempts" and `TestGenerateKey`'s `getPublicKeyErrCount: 5` (the `getPublicKeyErrCount: 1` case is fine because it succeeds before it ever sleeps).

The backoff logic itself looks right, so I think this is fixable purely on the test side: run the call that drives the retries in a goroutine, then release each backoff with `WaitForSleep` + `Add`. The loop sleeps once per failed retry, which is `getPublicKeyErrCount - 1` (one sleep for every error after the initial `GetPublicKey` call), so the count can come straight from the test case rather than a literal:

```go
errCh := make(chan error, 1)
go func() {
	_, err := ts.plugin.GenerateKey(ctx, tt.generateKeyReq)
	errCh <- err
}()

for range tt.getPublicKeyErrCount - 1 {
	ts.clockHook.WaitForSleep(testTimeout, "timed out waiting for backoff sleep")
	ts.clockHook.Add(time.Second) // >= backoffMax, releases the pending sleep
}

require.NoError(t, <-errCh)
```

The same shape works for `TestConfigure`'s max-attempts case: wrap `ts.plugin.Configure(...)`, and since it sets `getPublicKeyErrCount` to `getPublicKeyMaxAttempts + 1` the same expression yields `getPublicKeyMaxAttempts` sleeps, with the error asserted from `errCh` rather than `NoError`.

pkg/server/plugin/keymanager/gcpkms/gcpkms_test.go
```diff
@@ -401,6 +401,12 @@ func TestConfigure(t *testing.T) {
 				require.Nil(t, tt.config, "The test case must define a configuration or a configuration request, not both.")
 				configureRequest = tt.configureRequest
 			}
+			go func() {
+				for range tt.getPublicKeyErrCount - 1 {
+					ts.clockHook.WaitForSleep(testTimeout, "timed out waiting for backoff sleep")
+					ts.clockHook.Add(time.Second) // >= backoffMax, releases the pending sleep
+				}
+			}()
```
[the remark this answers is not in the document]
amartinezfayo: This one is subtler than it looks. The timeout path in WaitForSleep does call t.Fatalf, which wouldn't be reported correctly from a spawned goroutine. In practice it's never reached here because the pump loops exactly as many times as the retry loop sleeps (getPublicKeyErrCount - 1), so every WaitForSleep receives a sleep and returns normally. Driving the clock fully synchronously isn't an option, since the call under test blocks inside clk.Sleep and a separate goroutine has to advance the mock clock. I also tried the obvious restructuring (run the call under test in the goroutine and keep WaitForSleep on the test goroutine), and it actually deadlocks: advancing the mock clock by a full second coalesces several of the small backoff timers into one Add, and the mock's sleep channel is cap-1 with a drop-on-full send, so the surplus sleep signals get dropped and the pump waits forever. So the current arrangement is intentional.

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
```diff
@@ -1078,7 +1079,10 @@ func getOrCreateServerID(idPath string) (string, error) {
 
 // getPublicKeyFromCryptoKeyVersion requests Cloud KMS to get the public key
 // of the specified CryptoKeyVersion.
-func getPublicKeyFromCryptoKeyVersion(ctx context.Context, log hclog.Logger, kmsClient cloudKeyManagementService, cryptoKeyVersionName string) ([]byte, error) {
+func getPublicKeyFromCryptoKeyVersion(ctx context.Context, log hclog.Logger, kmsClient cloudKeyManagementService, cryptoKeyVersionName string, clk clock.Clock) ([]byte, error) {
+	backoffMin := 10 * time.Millisecond
+	backoffMax := time.Second
+	backoff := backoffMin
```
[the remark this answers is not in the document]
amartinezfayo: This mirrors the existing scheduleDestroyTask backoff in this same file, which also doubles the backoff before the first Sleep, so the behavior is intentional and consistent with the established pattern. The exact value of the first wait (10ms versus 20ms) has no functional impact, so we're keeping it as-is for consistency.

pkg/server/plugin/keymanager/gcpkms/gcpkms.go
```diff
@@ -1114,6 +1118,10 @@ func getPublicKeyFromCryptoKeyVersion(ctx context.Context, log hclog.Logger, kms
 		log.Warn("Could not get the public key because the CryptoKeyVersion is still being generated. Trying again.")
 		attempts++
 		kmsPublicKey, errGetPublicKey = kmsClient.GetPublicKey(ctx, &kmspb.GetPublicKeyRequest{Name: cryptoKeyVersionName})
+		if errGetPublicKey != nil {
+			backoff = min(backoff*2, backoffMax)
+			clk.Sleep(backoff)
+		}
```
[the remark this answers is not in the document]
amartinezfayo: This mirrors the existing scheduleDestroyTask backoff in this same file, which also doubles the backoff before the first Sleep, so the behavior is intentional and consistent with the established pattern. The exact value of the first wait (10ms versus 20ms) has no functional impact, so we're keeping it as-is for consistency.

pkg/server/plugin/keymanager/gcpkms/gcpkms_test.go
```diff
@@ -401,6 +401,12 @@ func TestConfigure(t *testing.T) {
 				require.Nil(t, tt.config, "The test case must define a configuration or a configuration request, not both.")
 				configureRequest = tt.configureRequest
 			}
+			go func() {
+				for range tt.getPublicKeyErrCount - 1 {
```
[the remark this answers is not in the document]
amartinezfayo: This is fine on our supported Go version: go.mod targets Go 1.26, where range-over-integer is fully supported. Ranging over a non-positive value runs zero iterations, so no explicit clamp is needed when getPublicKeyErrCount is 1 or 0.

amartinezfayo (approved):
Thank you @daescha!

PR #6929, opened 2026-05-03
Suppress RPC metrics for agent health check loopback calls

### LINKED ISSUE #5335: spire-agent health check report spire_agent_rpc_workload_api_fetch_x509_bundles{status="PermissionDenied"} metrics
pire-agent health check invoke this method: sendX509BundlesResponse,  because spire-agent do not have a spire entry that select spire-agent pod, it will return codes.PermissionDenied. I will report this metrics: spire_agent_rpc_workload_api_fetch_x509_bundles{status="PermissionDenied"}, so when this exception occurs in a normal request for non health checks, I cannot distinguish between them.

```
func sendX509BundlesResponse(update *cache.WorkloadUpdate, stream workload.SpiffeWorkloadAPI_FetchX509BundlesServer, log logrus.FieldLogger, allowUnauthenticatedVerifiers bool, previousResponse *workload.X509BundlesResponse, quietLogging bool) (*workload.X509BundlesResponse, error) {
    if !allowUnauthenticatedVerifiers && !update.HasIdentity() {
       if !quietLogging {
          log.WithField(telemetry.Registered, false).Error("No identity issued")
       }
       return nil, status.Error(codes.PermissionDenied, "no identity issued")
    }
    resp, err := composeX509BundlesResponse(update)
    if err != nil {
       log.WithError(err).Error("Could not serialize X509 bundle response")
       return nil, status.Errorf(codes.Unavailable, "could not serialize response: %v", err)
    }
    if proto.Equal(resp, previousResponse) {
       return previousResponse, nil
    }
```

* **Version**: v1.9.6
* **Platform**: linux-amd64
* **Subsystem**: spire-agent

rausingh-rh:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Agent RPC metrics for Workload API calls made by the agent's own health check (both the gRPC Health/Check service and the HTTP `/ready` `/live` endpoints). These internal loopback calls no longer emit per-call RPC metrics (counter and elapsed_time). Connection metrics are unaffected. Metrics for calls from real workloads are unaffected.

**Description of change**

- Added `Discard()` method to `telemetry.CallCounter` that marks the counter as done without emitting metrics
- Added `discardAgentCallMetrics` postprocess middleware step in the agent endpoint middleware chain that checks the caller PID via peertracker; when the caller is the agent itself, the RPC call counter is discarded
- Updated existing integration tests to reflect that agent-PID calls no longer emit RPC metrics
- Added regression test `TestAgentCallMetricsDiscarded` with three cases: agent PID (suppressed), non-agent PID (emitted), and no peertracker (emitted)

**Which issue this PR fixes**

Fixes #5335

### DIFF
--- a/pkg/agent/endpoints/endpoints_test.go
+++ b/pkg/agent/endpoints/endpoints_test.go
@@ -89,13 +89,9 @@ func TestEndpoints(t *testing.T) {
 				{Type: fakemetrics.IncrCounterType, Key: []string{"workload_api", "connection"}, Val: 1},
 				{Type: fakemetrics.SetGaugeType, Key: []string{"workload_api", "connections"}, Val: 1},
 				{Type: fakemetrics.SetGaugeType, Key: []string{"workload_api", "connections"}, Val: 0},
-				// Call counter
-				{Type: fakemetrics.IncrCounterWithLabelsType, Key: []string{"rpc", "workload_api", "fetch_jwtsvid"}, Val: 1, Labels: []metrics.Label{
-					{Name: "status", Value: "OK"},
-				}},
-				{Type: fakemetrics.MeasureSinceWithLabelsType, Key: []string{"rpc", "workload_api", "fetch_jwtsvid", "elapsed_time"}, Val: 0, Labels: []metrics.Label{
-					{Name: "status", Value: "OK"},
-				}},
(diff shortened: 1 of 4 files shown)


sorindumitru (approved):
LGTM, thanks @rausingh-rh!

PR #6943, opened 2026-05-08
datastore: add DataConsistency param to FetchRegistrationEntries and FetchAttestedNode

heybronson:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

Events-driven entry cache reload (`AuthorizedEntryFetcherEvents`): the step that fetches changed entry and node data after detecting events via `ListRegistrationEntryEvents` / `ListAttestedNodeEvents`.

**Description of change**

`FetchRegistrationEntries` and `FetchAttestedNode` had no `DataConsistency` parameter and always queried the primary database. This is inconsistent with the rest of the events-based cache reload path, where `ListRegistrationEntryEvents`, `ListAttestedNodeEvents`, and `GetNodeSelectors` already route to the read-only replica via `DataConsistency: TolerateStale`.

This PR adds a `dataConsistency DataConsistency` trailing parameter to both methods (matching the existing `GetNodeSelectors` convention) and routes to the replica when `TolerateStale` is passed and `roDb` is configured. The two cache-updater call sites are updated to pass `TolerateStale`; all other callers pass `RequireCurrent` to preserve existing behavior.

`fetchAttestedNode` (inner function) is also refactored to accept `(ctx context.Context, db *sqlDB, spiffeID string)` so the request context propagates correctly to the SQL call, consistent with `fetchRegistrationEntries` and `getNodeSelectors`.

**Which issue this PR fixes**

N/A

### DIFF
--- a/pkg/common/telemetry/server/datastore/wrapper.go
+++ b/pkg/common/telemetry/server/datastore/wrapper.go
@@ -125,10 +125,10 @@ func (w metricsWrapper) DeleteRegistrationEntryEventForTesting(ctx context.Conte
 	return w.ds.DeleteRegistrationEntryEventForTesting(ctx, eventID)
 }
 
-func (w metricsWrapper) FetchAttestedNode(ctx context.Context, spiffeID string) (_ *common.AttestedNode, err error) {
+func (w metricsWrapper) FetchAttestedNode(ctx context.Context, spiffeID string, dataConsistency datastore.DataConsistency) (_ *common.AttestedNode, err error) {
 	callCounter := StartFetchNodeCall(w.m)
 	defer callCounter.Done(&err)
-	return w.ds.FetchAttestedNode(ctx, spiffeID)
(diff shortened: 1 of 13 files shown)


sorindumitru:
Thanks for opening this @heybronson. I don't think we can use `TolerateStale` here because we only update the registration entries/attested nodes when we see an event. So you could have the following sequence:
* Entry gets updated, master node starts propagating it to replicas (lets assume async replicas)
* spire-server reads the event about the registration entry from DB instance X
* spire-server reads the registration entry from DB instance Y
* db instance Y learns about the changes to the registration entry

In this case spire-server would read the old version of the entry and the changes to the entry will be lost until the next time the entry changes or when a full sync is done.

sorindumitru:
@heybronson, I'll close this for now since this doesn't seem like the right approach. Could you open an issue describing the problem you're trying to fix?

PR #6946, opened 2026-05-08
fix: cascade-delete registered_entries on attested node deletion

### LINKED ISSUE #6944: registered_entries with parent = pruned attested node are not cascade-deleted
### Background

When `CreateJoinToken` is called **with `AgentId` set** (`pkg/server/api/agent/v1/service.go:536-591`), the server writes a `registered_entries` alias row via `createJoinTokenRegistrationEntry` (service.go:594) in addition to the normal `join_tokens` row. The alias row has:

- `parent_id = <join_token SVID>` (e.g. `spiffe://<td>/spire/agent/join_token/<uuid>`)
- `spiffe_id = <AgentId>`
- `selectors = [{type: "spiffe_id", value: parent_id}]`

This alias is created only when `AgentId` is provided — without `AgentId`, the API just mints a token and no registration entry is written
(service.go:583). When the agent later attests with the token, the `attested_node_entries` row is created/updated for the join-token SVID, and the alias is the mechanism by which workloads parented under `<AgentId>` can receive SVIDs through the agent at runtime.

### The problem

When the `attested_node_entries` row is later pruned (via `prune_attested_nodes_expired_for`, added in #6152) or deleted by an admin via `DeleteAgent`, **the corresponding `registered_entries` alias row is not deleted**. Its `parent_id` now references a non-existent attested node, so it can never produce an SVID again — it is functionally dead, but continues to occupy a `registered_entries` row and flows through every server replica's in-memory entry cache and the events-cache.

In deployments that rely on `CreateJoinToken(AgentId=...)` for node onboarding — where agents re-attest over their lifetime and older agents are eventually retired — this causes unbounded growth of `registered_entries` that does not self-correct. At scale (fleets on the order of thousands to millions of agents), the orphan fraction grows to dominate the table: on a fleet I've measured, orphans reached a clear majority of all rows before any cleanup was attempted, with measurable DB, entry-cache-memory, and events-cache-reload cost on every replica.

### Verification in current code

- `pruneAttestedExpiredNodes` (`sqlstore.go:1828`) → `deleteAttestedNodeAndSelectors` (`sqlstore.go:2397`) deletes `AttestedNode` + `NodeSelector` rows only.
- `DeleteAttestedNode` uses the same helper. Same gap.
- `PruneRegistrationEntries` (`sqlstore.go:590`) operates on `expiry != 0`; `createJoinTokenRegistrationEntry` creates alias rows without setting `expiry`, so this existing prune never catches them.

### Related work already in-tree

- #6152 added `PruneAttestedExpiredNodes`. Its scope was explicitly attested nodes + selectors; entries were not in scope.
- #3873 extended `DeleteAttestedNode` to cascade-delete `node_resolver_map_entries` — same shape this fix would extend further.

### Proposed fixes (preference order)

**Option A (revised) — cascade in `deleteAttestedNodeAndSelectors`, guarded by `node.DataType == "join_token"`.** Same shape as #3873, same code path, but only fires for join-token-attested nodes. Cleans up the auto-alias unconditionally on prune (since prune is meant to be silent background maintenance and can't ask for a flag), without changing behavior for any other attestor.

**Option B — periodic sweep.** Add a new `PruneOrphanedEntries` analogous to `PruneAttestedExpiredNodes` that scans `registered_entries` for rows whose parent is a missing attested node. Independent schedule; catches historical orphans; requires new config.

I'd like to implement Option A and am opening this issue first.

**Option A vs B** — preference?

Happy to refactor if maintainers prefer Option B or a different approach. A PR implementing Option A will follow shortly.

angabini:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Server datastore — the attested-node delete path (`DeleteAttestedNode` and `PruneAttestedExpiredNodes`), both of which flow through `deleteAttestedNodeAndSelectors` in `pkg/server/datastore/sqlstore/sqlstore.go`. Also the events-based cache (one `RegistrationEntryEvent` per cascaded entry) and, transitively, the agent API delete path (`DeleteAgent`, i.e. `spire-server agent evict`/`agent purge`).

**Description of change**

`deleteAttestedNodeAndSelectors` now cascades to `registered_entries` for join-token-attested nodes. Before deleting the attested-node row, when `DataType == "join_token"` it finds child entries (`parent_id == <node spiffe_id>`) that match the auto-alias shape `createJoinTokenRegistrationEntry` writes — a single `spiffe_id` selector whose value equals the parent SVID — deletes each via the existing `deleteRegistrationEntrySupport` helper, and emits a `RegistrationEntryEvent` per deletion so the events-based cache invalidates. Each cascaded delete logs at `Info` with the standard `SPIFFEID`/`ParentID`/`RegistrationID` fields, matching `pruneRegistrationEntries`.

This closes the gap from #6152 (node prune scheduler scoped to `attested_node_entries` + `node_selectors`) by also cleaning up the `registered_entries` alias row that `CreateJoinToken` (with `AgentId`) creates. It mirrors #3873, which added the `node_resolver_map_entries` cascade in the same function.

**Scope is deliberately narrow.** The cascade fires only when (a) the node's `DataType == "join_token"` and (b) the child entry matches the auto-alias shape. That auto-alias is the only `registered_entries` row SPIRE itself creates with `parent_id == <node SVID>`, so user-registered workload entries parented on an agent SVID are never touched. A broader cascade for other attestors would need to be opt-in (e.g. a `--cascade` flag), per discussion on the issue.

**Behavior**
- Prune path: `PruneAttestedExpiredNodes` deletes the matching alias entry atomically with each expired join-token node.
- Manual delete path: `DeleteAttestedNode` (`spire-server agent evict`/`agent purge`) deletes the matching alias entry atomically with the node.
- Both paths emit entry-delete events so servers using the events-based cache invalidate correctly.

**Forward-only.** The cascade runs only when a node is actually deleted or pruned, so it cleans up aliases going forward but does not revisit nodes removed before this change. Pre-existing orphaned alias entries must be cleaned up separately (e.g. a one-off batch job); this is called out in `doc/spire_server.md`.

**Avoiding shape drift.** The alias shape lives in two places — the writer (`createJoinTokenRegistrationEntry` in `pkg/server/api/agent/v1/service.go`) and the cascade matcher (`deleteAttestedNodeAndSelectors`). Both sites now cross-reference each other, and `TestCascadeDeleteJoinTokenAliasEntry` exercises the real `CreateJoinToken` → `DeleteAgent` path (`fakedatastore` is backed by the sqlstore), so a change to the writer that the matcher doesn't track fails the test.

**Tests**
- `TestCascadeDeleteJoinTokenAliasEntry` (`pkg/server/api/agent/v1/service_test.go`) — real create→delete path; asserts the alias entry is cascaded and an event emitted.
- `TestDeleteAttestedNodeCascadesEntries` — single-child cascade + event on the manual delete path.
- `TestDeleteAttestedNodeJoinTokenPreservesNonAliasChildEntries` — alias child cascaded; a user-managed (non-alias) child on the same node preserved; node selectors cleared; exactly one event emitted.
- `TestPruneAttestedExpiredNodesCascadesEntries` — cascade fires on the prune path; a non-expired sibling node's child entry is preserved.
- `TestDeleteAttestedNodeNonJoinTokenDoesNotCascade` — alias-shaped child under a non-join-token node is preserved (DataType filter).

**Which issue this PR fixes**

fixes #6944


amartinezfayo:
Thank you @angabini for this contribution!
Did you have a chance to look at this [comment](https://github.com/spiffe/spire/issues/6944#issuecomment-4453891653) in the issue?

angabini:
> Thank you @angabini for this contribution! Did you have a chance to look at this [comment](https://github.com/spiffe/spire/issues/6944#issuecomment-4453891653) in the issue?

Yes! and the new commit is to address that. see [this](https://github.com/spiffe/spire/issues/6944#issuecomment-4547689273).

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -2394,18 +2394,43 @@ func updateAttestedNode(tx *gorm.DB, n *common.AttestedNode, mask *common.Attest
...
+			if err := createRegistrationEntryEvent(tx, &datastore.RegistrationEntryEvent{
+				EntryID: entry.EntryID,
+			}); err != nil {
+				return nil, err
+			}
+			logger.WithFields(logrus.Fields{
+				telemetry.SPIFFEID:       entry.SpiffeID,
+				telemetry.ParentID:       entry.ParentID,
+				telemetry.RegistrationID: entry.EntryID,
+			}).Info("Cascade-deleted registration entry on attested node deletion")
```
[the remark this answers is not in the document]
angabini: This matches the existing pattern in this file. pruneRegistrationEntries (sqlstore.go:4197–4220) loops over expired entries and emits exactly the same per-entry Info log — "Pruned an expired registration" — with the same field set (spiffe_id / parent_id / registration_id). 

The cascade I added uses "Cascade-deleted registration entry on attested node deletion" and the same fields, deliberately, so an operator auditing why an entry disappeared sees one consistently shaped log line per deletion regardless of which code path removed it.

Happy to move to Debug if the maintainers prefer the quieter default, but I'd argue we should change pruneRegistrationEntries at the same time so the two related lifecycle events log at the same level.

angabini:
@amartinezfayo any action item on my side?

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -2394,18 +2394,51 @@ func updateAttestedNode(tx *gorm.DB, n *common.AttestedNode, mask *common.Attest
...
+		var candidates []RegisteredEntry
+		if err := tx.Where("parent_id = ?", spiffeID).Find(&candidates).Error; err != nil {
+			return nil, newWrappedSQLError(err)
+		}
+		for _, entry := range candidates {
+			var selectors []Selector
+			if err := tx.Where("registered_entry_id = ?", entry.ID).Find(&selectors).Error; err != nil {
+				return nil, newWrappedSQLError(err)
+			}
+			if len(selectors) != 1 || selectors[0].Type != "spiffe_id" || selectors[0].Value != entry.ParentID {
```
amartinezfayo, line 2460: I think the narrow alias-shape match is the right call over cascading every join-token child. My concern is that the shape lives both here and in createJoinTokenRegistrationEntry with nothing linking them, so a change there would silently stop this cleanup, and the hand-built test would drift with it rather than catch the mismatch. Could we cross-reference both sites and add a test over the real CreateJoinToken to delete path?

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -2394,18 +2394,51 @@ func updateAttestedNode(tx *gorm.DB, n *common.AttestedNode, mask *common.Attest
...
+	if err := tx.Find(&nodeModel, "spiffe_id = ?", spiffeID).Error; err != nil {
 		return nil, newWrappedSQLError(err)
 	}
 
-	if err := tx.Find(&nodeModel, "spiffe_id = ?", spiffeID).Error; err != nil {
+	// Cascade only for join-token-attested nodes, and only for entries that match
+	// the auto-alias shape that createJoinTokenRegistrationEntry writes (single
+	// "spiffe_id" selector whose value is the parent SVID). Other parent_id-keyed
+	// entries are user-managed workload entries and must be preserved.
+	if nodeModel.DataType == "join_token" {
```
amartinezfayo, line 2450: As I read it the cascade only runs when a node is actually deleted or pruned, so it cleans up the alias going forward but never revisits nodes that were already pruned before this lands. Those existing orphan entries would stay in the table. Since the issue reports that backlog is already most of the table, should we call out that this is a forward-only fix and track draining the existing orphans separately, for example via the periodic sweep floated in the issue?

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -2394,18 +2394,51 @@ func updateAttestedNode(tx *gorm.DB, n *common.AttestedNode, mask *common.Attest
...
+			if err := createRegistrationEntryEvent(tx, &datastore.RegistrationEntryEvent{
+				EntryID: entry.EntryID,
+			}); err != nil {
+				return nil, err
+			}
+			logger.WithFields(logrus.Fields{
+				telemetry.SPIFFEID:       entry.SpiffeID,
+				telemetry.ParentID:       entry.ParentID,
+				telemetry.RegistrationID: entry.EntryID,
+			}).Info("Cascade-deleted registration entry on attested node deletion")
```
amartinezfayo, line 2475: It seems reasonable to keep this at Info, since it mirrors the per-entry line pruneRegistrationEntries already emits. If we want it quieter, I'd move both together so the two lifecycle events stay at the same level.

doc/spire_server.md
```diff
@@ -56,39 +56,39 @@ SPIRE configuration files may be represented in either HCL or JSON. Please see t
...
+| `profiling_port`                   | Port number of the [net/http/pprof](https://pkg.go.dev/net/http/pprof) endpoint. Only used when `profiling_enabled` is `true`.                                                                                                                                                                                                                                                                                                                                                              |                                                                |
+| `prune_attested_nodes_expired_for` | Enables periodic purging of attested node records with expired SVIDs where the expiry time further in the past than the specified duration. Non-reattestable nodes are not pruned unless `prune_tofu_nodes` is set to `true`. Banned nodes are not pruned. When a join-token-attested node is purged, the alias registration entry that `CreateJoinToken` creates with `AgentId` is cascade-deleted (with a corresponding registration entry event). Non-join-token nodes are not cascaded. |                                                                |
```
amartinezfayo: Should we also note that a manual agent delete picks up the same behavior, since the cascade fires in deleteAttestedNodeAndSelectors on both the prune path and DeleteAttestedNode? An operator reading only this row might not realize it.

pkg/server/datastore/sqlstore/sqlstore.go
angabini: yes, that's correct and my intention here is only forward-only fix. People can run batch jobs or so to cleanup up the orphan entires prior to this change. 
Where would be a good place to call it out? `spire_server.md`?

angabini: Added a note in the doc.

pkg/server/datastore/sqlstore/sqlstore.go
angabini: added a test.

amartinezfayo (approved):
Thank you @angabini!

PR #6915, opened 2026-04-30
Implement SPIFFE Broker Endpoint & API

matheuscscp:
Reference implementation for https://github.com/spiffe/spiffe/pull/340 https://github.com/arndt-s/spiffe/pull/1

go-spiffe PR: https://github.com/spiffe/go-spiffe/pull/388 plugin-sdk PR: https://github.com/spiffe/spire-plugin-sdk/pull/74

Follow-up issues: #7074 #7075
Follow-up PRs: https://github.com/matheuscscp/spire/pull/1 (waiting for this one to merge)

**Pull Request check list**

* [x]  Commit conforms to CONTRIBUTING.md?
* [x]  Proper tests/regressions included? Added [here](https://github.com/spiffe/spire/compare/b5dc6bf8f8a735127fbd312e55d3dd989fb7aa1d..3548750f3c727d8a0ddfe9b9de942cb1a7dfbcd8)
* [x]  Documentation updated? Added [here](https://github.com/spiffe/spire/compare/bc98f2eff6cf883e109c5a37506637c721ab1ace..ec0888b4777d5aad6febc7bf25df1c97f3358b02)

**Affected functionality**

**Description of change**

**Which issue this PR fixes**

Supersedes: #6594

PR #6957, opened 2026-05-13
Add disable_group_name_selectors option to Windows workload attestor

### LINKED ISSUE #6902: Add a mechanism to skip group SID to name lookup in Windows workload attestor
Proposing improvements to: [Agent plugin: WorkloadAttestor "windows"](https://github.com/spiffe/spire/blob/main/doc/plugin_agent_workloadattestor_windows.md)

We have a use case where Windows user accounts/service accounts are added to 100s of groups and the group SID to group name lookup for each group is taking around ~88 ms - [this line of code](https://github.com/spiffe/spire/blob/81cb7d970c55444e984d8069aa0e213d43895808/pkg/agent/plugin/workloadattestor/windows/windows_windows.go#L170). 

In one case, the user was added to 501 groups and in total the group name lookup took 44 seconds to complete timing out the first x509 fetch call for that user - it is being cached somewhere as the subsequent calls right after were quick but unclear what the TTL of the cache as the calls timeout every now and then. 

Proposing to add an additional configuration flag to skip this lookup. 

| Configuration               | Description                              | Default |
|----------------------------|------------------------------------------|---------|
| skip_group_name_selectors  | Skips collecting the group name selectors | false   |

I can make the required code changes as well.

jananiarunachalam:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Windows workload attestor group SID resolution.

**Description of change**
Adds a `disable_group_name_selectors` boolean configuration option (default: `false`) to the Windows workload attestor. When enabled, the attestor skips resolving group SIDs to human-readable names via `LookupAccountSid`, avoiding timeouts for workloads whose principals belong to many groups (~88ms per group, 44+ seconds for 500+ groups). Group SID selectors (`group_sid`) are always collected regardless of this setting, only `group_name` selectors are skipped.

Also adds debug-level timing instrumentation to the group name lookup path so operators can observe lookup latency without enabling the flag.

**Which issue this PR fixes**
fixes #6902

### DIFF
--- a/doc/plugin_agent_workloadattestor_windows.md
+++ b/doc/plugin_agent_workloadattestor_windows.md
@@ -3,10 +3,11 @@
 The `windows` plugin generates Windows-based selectors for workloads calling the agent.
 It does so by opening an access token associated with the workload process. The system is then interrogated to retrieve user and group account information from that access token.
 
-| Configuration            | Description                                                                                                                                                | Default |
-|--------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
-| `discover_workload_path` | If true, the workload path will be discovered by the plugin and used to provide additional selectors                                                       | false   |
-| `workload_size_limit`    | The limit of workload binary sizes when calculating certain selectors (e.g. sha256). If zero, no limit is enforced. If negative, never calculate the hash. | 0       |
+| Configuration                  | Description                                                                                                                                                                                                            | Default |
+|--------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
+| `discover_workload_path`       | If true, the workload path will be discovered by the plugin and used to provide additional selectors                                                                                                                   | false   |
+| `workload_size_limit`          | The limit of workload binary sizes when calculating certain selectors (e.g. sha256). If zero, no limit is enforced. If negative, never calculate the hash.                                                             | 0       |
+| `disable_group_name_selectors` | If true, skips resolving group SIDs to human-readable names, avoiding expensive Domain Controller lookups. Group SID selectors (`group_sid`) are always collected regardless. Only `group_name` selectors are affected. | false   |
 
 ## Workload Selectors
 
--- a/doc/plugin_agent_workloadattestor_windows.md
+++ b/doc/plugin_agent_workloadattestor_windows.md
@@ -3,10 +3,10 @@
 The `windows` plugin generates Windows-based selectors for workloads calling the agent.
 It does so by opening an access token associated with the workload process. The system is then interrogated to retrieve user and group account information from that access token.
 
-| Configuration                  | Description                                                                                                                                                                                                            | Default |
-|--------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
-| `discover_workload_path`       | If true, the workload path will be discovered by the plugin and used to provide additional selectors                                                                                                                   | false   |
-| `workload_size_limit`          | The limit of workload binary sizes when calculating certain selectors (e.g. sha256). If zero, no limit is enforced. If negative, never calculate the hash.                                                             | 0       |
+| Configuration                  | Description                                                                                                                                                                                                                                                          | Default |
+|--------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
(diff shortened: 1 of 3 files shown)


doc/plugin_agent_workloadattestor_windows.md
```diff
@@ -3,10 +3,11 @@
...
+|--------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
+| `discover_workload_path`       | If true, the workload path will be discovered by the plugin and used to provide additional selectors                                                                                                                   | false   |
+| `workload_size_limit`          | The limit of workload binary sizes when calculating certain selectors (e.g. sha256). If zero, no limit is enforced. If negative, never calculate the hash.                                                             | 0       |
+| `disable_group_name_selectors` | If true, skips resolving group SIDs to human-readable names, avoiding expensive Domain Controller lookups. Group SID selectors (`group_sid`) are always collected regardless. Only `group_name` selectors are affected. | false   |
```
amartinezfayo: I wonder if it would help to surface in the Notes section that enabling this flag will cause existing workload registration entries using `group_name` selectors to stop matching, so operators know to audit those entries (or switch them to `group_sid` selectors) before flipping it on. As a smaller wording detail, I think that `LookupAccount` is not strictly a Domain Controller call in every environment (local SAM, AD DS, or other providers can answer); something like "potentially expensive account name resolution (e.g. against a Domain Controller)" might be more accurate.

pkg/agent/plugin/workloadattestor/windows/windows_windows.go
```diff
@@ -160,20 +162,26 @@ func (p *Plugin) newProcessInfo(pid int32, queryPath bool) (*processInfo, error)
...
+				continue
+			}
+			// If the LookupAccount call succeeded, we know that groupAccount is not empty
+			processInfo.groups = append(processInfo.groups, enabledSelector+":"+parseAccount(groupAccount, groupDomain))
 		}
-		// If the LookupAccount call succeeded, we know that groupAccount is not empty
-		processInfo.groups = append(processInfo.groups, enabledSelector+":"+parseAccount(groupAccount, groupDomain))
+	}
+	if !disableGroupNames {
+		p.log.Debug("lookupAccount (groups) completed", "count", len(groups), "elapsed_ms", time.Since(start).Milliseconds())
```
amartinezfayo: A couple of small things on this Debug log. I don't think we have any other `elapsed_ms` keys already logged. I think `telemetry.ElapsedTime` and the `duration_ms` field are the closest that we have, and maybe we could use one of those?
In terms of the wording, maybe "Group account name lookups completed" might feel more familiar.

It could also be worth considering whether this is redundant with the host-level metric from `telemetry_workload.StartAttestorCall`, which already captures the full per-attestor Attest time and is dominated by this loop for the Windows attestor.

amartinezfayo (commented):
Thank you @jananiarunachalam for this contribution!

pkg/agent/plugin/workloadattestor/windows/windows_windows.go
jananiarunachalam: Addressed all comments. 

This debug log specifically isolates the group name resolution portion with a count, which is still useful for diagnosing whether the lookups are the bottleneck.

pkg/agent/plugin/workloadattestor/windows/windows_windows.go
```diff
@@ -195,6 +203,10 @@ func (p *Plugin) Configure(_ context.Context, req *configv1.ConfigureRequest) (*
 	defer p.mu.Unlock()
 	p.config = newConfig
 
+	if newConfig.DisableGroupNameSelectors {
+		p.log.Info("Group name selectors disabled; skipping LookupAccount for groups, only group_sid selectors will be produced")
```
amartinezfayo: I'm wondering if this operator-facing message could drop the LookupAccount reference, since that's an internal call name rather than something an operator reading the logs would recognize. Something like "Group name selectors disabled; only group_sid selectors will be produced for groups" might read a little more clearly while still conveying the behavior change.

amartinezfayo (commented):
Thank you @jananiarunachalam!

jananiarunachalam:
Rebased on latest main to resolve a merge conflict in `windows_windows.go`. No code changes - only incorporated the new `AttestReference` method added by #6915. Only thing to review is: https://github.com/spiffe/spire/pull/6957/commits/31a4ac0efdb08bef391c81c2918b25286572495f which has the updated log line changes requested.

amartinezfayo (approved):
🚀

Issue #6952, opened 2026-05-12
Azure IMDS Plugin: Attestation Failures Due to Incomplete CA Pattern Coverage


ChanghengGu:
The Azure IMDS node attestor fails for VMs in certain Azure regions because it only validates one of three CA naming patterns Microsoft uses:

  - Pattern 1: Microsoft Azure RSA TLS Issuing CA XX (validated)
  - Pattern 2: Microsoft TLS G1 RSA CA OCSP XX (fails)
  - Pattern 3: Microsoft TLS G2 RSA CA OCSP XX (fails)

  VMs with Pattern 2/3 certificates fail validation despite having legitimate Microsoft-issued certificates. This causes unpredictable attestation failures across Azure regions.

  Root cause: The plugin validates intermediate CA names via regex instead of following Azure's official guidance (https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#signature-validation-guidance) to validate the certificate Subject Alternative Name (SAN).

Additional issue: Missing Microsoft TLS RSA Root G2 root certificates (both self-signed and cross-signed versions), causing failures for 3-level certificate chains.

ravishen:
Following up here per @amartinezfayo's suggestion, so other maintainers can weigh in. ( @evan2645  @sorindumitru @MarcosDY @rturner3 )

This builds on a discussion with him (on #6312 and offline).

The immediate failures in this issue, SAN validation and the missing *Microsoft TLS RSA Root G2* roots were fixed in #6960. 

What I'd like to raise is the fragility that remains: `azure_imds` chains the signing certificate to a hardcoded root list in `roots.go`, and since Azure rotates its PKI without reliable advance notice, a future rotation that doesn't terminate at a pinned root would again break attestation until `roots.go` is updated and a new SPIRE release ships.

In #6312 it was suggested that fetching the CA from a secure, well-known location would be the best approach. I looked into that, but Azure doesn't seem to offer any programmatic or machine-readable source, only a human-readable [CA details article](https://learn.microsoft.com/en-us/azure/security/fundamentals/azure-certificate-authority-details), and IMDS explicitly won't notify on CA changes, so a dynamic fetch isn't practical. Azure's [pinning guidance](https://learn.microsoft.com/en-us/azure/security/fundamentals/certificate-pinning) is neutral on pinning, but does say pinned certs should be "updated on short notice," which is the hard part today given the release coupling.

[...]

PR #6960, opened 2026-05-14
Fix Azure IMDS Plugin Signature Validation

ChanghengGu:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Azure IMDS plugin

**Description of change**
Previous implementation checks intermediate CA's against a pinned list, this requires a frequent update to the issuer regex as Azure frequently rotates without announcement. Updated the plugin validate signature by checking cert's SAN instead. As recommended by Azure https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#signature-validation-guidance

Added a configuration to the plugin allowedMetadataDomains, this defaults to metadata.azure.com if no specified. User can override by specifying this through config

Also adds 2 root that was missing

MicrosoftTLSRSARootG2SelfSignedPem
MicrosoftTLSRSARootG2CrossSignedPem

**Which issue this PR fixes**
https://github.com/spiffe/spire/issues/6952
SPIFFE Slack thread https://spiffe.slack.com/archives/CBNCC2V17/p1778103005230459


pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
```diff
@@ -146,58 +137,31 @@ func getIntermediateCertificate(ctx context.Context, signingCert *x509.Certifica
...
+// by checking the Subject Alternative Name (SAN) extension. Per RFC 6125 and modern PKI standards,
+// SAN is the authoritative source for certificate identity validation (Subject CN is deprecated).
+func validateAzureCertificate(cert *x509.Certificate, allowedDomains []string) error {
+	// Check SAN DNS names
+	for _, dnsName := range cert.DNSNames {
+		for _, allowedDomain := range allowedDomains {
+			if dnsName == allowedDomain {
+				return nil
+			}
+		}
```
c4rlo:
```suggestion
		if slices.Contains(allowedDomains, dnsName) {
			return nil
		}
```

pkg/server/plugin/nodeattestor/azureimds/roots.go
c4rlo, line 4: The comment doesn't apply to the new roots you've added. Please add a similar comment for your new roots. Also, please include the URLs that you obtained them from, like we have for the DigiCert roots.

pkg/server/plugin/nodeattestor/azureimds/roots.go
```diff
@@ -54,7 +54,88 @@ hMAtudXH/vTBH1jLuG2cenTnmCmrEbXjcKChzUyImZOMkXDiqw8cvpOp/2PV5Adg
 PnlUkiaY4IBIqDfv8NZ5YBberOgOzW6sRBc4L0na4UU+Krk2U886UAb3LujEV0ls
 YSEY1QSteDwsOoBrp+uvFRTp2InBuThs4pFsiv9kuXclVzDAGySj4dzp30d8tbQk
 CAUw7C29C79Fv1C5qfPrmAESrciIxpg0X40KPMbp1ZWVbd4=
+-----END CERTIFICATE-----`
+
+	// Microsoft TLS RSA Root G2 (self-signed)
+	// Valid until: 24/Jun/2045
+	// Issuer: Microsoft TLS RSA Root G2
+	// Subject: Microsoft TLS RSA Root G2
+	MicrosoftTLSRSARootG2SelfSignedPem = `-----BEGIN CERTIFICATE-----
```
c4rlo, line 62: This is not a valid cert. `openssl x509` rejects it, and `openssl asn1parse` print some of it before erroring out. What asn1parse does show is a "Baltimore CyberTrust Root" that expired in Dec 2024.

Where did you get this from??

Btw this is exposing a lack of error checking: these roots are parsed in `validateCertificateChain`, which does not check the return value of `AppendCertsFromPEM`. That seems worth fixing.

pkg/server/plugin/nodeattestor/azureimds/roots.go
```diff
@@ -54,7 +54,88 @@ hMAtudXH/vTBH1jLuG2cenTnmCmrEbXjcKChzUyImZOMkXDiqw8cvpOp/2PV5Adg
...
+/chqtaQlNYPP8lxGK0jK1RRMEOvzJJNPBLvpMkGEfNRvhPvUBHNnFRWCHXJ1qOCG
+VE0y8ZzXJTXFhT5uKQxDZ0h3DT6/fXCHJWjKALJYGPQh6oaQCJ1P8L4VY3BePQPa
+z8E=
+-----END CERTIFICATE-----`
+
+	// Microsoft TLS RSA Root G2 (cross-signed by DigiCert Global Root G2)
+	// Valid until: 19/Jun/2029
+	// Issuer: DigiCert Global Root G2
+	// Subject: Microsoft TLS RSA Root G2
+	MicrosoftTLSRSARootG2CrossSignedPem = `-----BEGIN CERTIFICATE-----
```
c4rlo, line 100: This cert too does not parse.

c4rlo (changes requested):
Please also update `doc/plugin_server_nodeattestor_azure_imds.md`.

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
ChanghengGu: Updated in the latest commit.

pkg/server/plugin/nodeattestor/azureimds/roots.go
ChanghengGu: Added to the new certs.

pkg/server/plugin/nodeattestor/azureimds/roots.go
ChanghengGu: I have the correct certs in our forked spire branch where I made and test the changes, I asked AI agent to copy the changes to spire repo, it apparently changed the certs instead of copying over... 

I have updated the PR with the correct certs and add a check in the validateCertificateChain.

pkg/server/plugin/nodeattestor/azureimds/roots.go
ChanghengGu: Updated to the correct cert in the new commit.

pkg/server/plugin/nodeattestor/azureimds/roots.go
c4rlo: Got it. Ok, I can confirm this looks ok now: the cert here matches the stated download URL, which is listed on https://www.microsoft.com/pkiops/docs/repository.htm, which is the first Google result for "microsoft pki roots" for me.

pkg/server/plugin/nodeattestor/azureimds/roots.go
c4rlo: Likewise, this cert appears on https://www.microsoft.com/pkiops/docs/repository.htm and matches the stated download URL.

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
```diff
@@ -146,58 +138,29 @@ func getIntermediateCertificate(ctx context.Context, signingCert *x509.Certifica
 }
 
 // validateAzureCertificates performs Azure-specific certificate validation
-func validateAzureCertificates(signingCert, intermediateCert *x509.Certificate) error {
-	// Validate signing certificate subject
-	if err := validateCertificateSubject(signingCert, AzureMetadataSubject); err != nil {
-		return fmt.Errorf("signing certificate subject validation failed: %w", err)
+// Following Azure's recommendation to validate the certificate Subject Alternative Name (SAN)
+// to confirm it's from Azure, rather than pinning specific intermediate CA names.
+// See: https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#signature-validation-guidance
```
c4rlo, line 149: That page says:

> Please note that the Azure Instance Metadata Service will NOT offer notifications for future Certificate Authority changes. Instead, you must follow the centralized [Azure Certificate Authority details](https://learn.microsoft.com/en-us/azure/security/fundamentals/azure-ca-details?tabs=root-and-subordinate-cas-list) article for all future updates.

The linked page lists many more CA roots – so presumably those should all be trusted also. The page seems to be sourced from https://github.com/MicrosoftDocs/azure-docs/blob/main/articles/security/fundamentals/azure-certificate-authority-details.md, based on the _Edit_ link in the ⋮ menu; we could parse that Markdown. In fact, that is the approach taken in #6917 for the AWS IID node attestor root CAs.

Anyway, that could be a longer-term improvement, I don't think it's a reason not to add the additional CAs here as a short-term fix.

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
```diff
@@ -146,58 +138,29 @@ func getIntermediateCertificate(ctx context.Context, signingCert *x509.Certifica
...
+func validateAzureCertificates(signingCert *x509.Certificate, allowedMetadataDomains []string) error {
```
c4rlo: This is a pointless wrapper over `validateAzureCertificate`.

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
```diff
@@ -11,31 +11,23 @@ import (
...
-	// Expected issuer patterns
-	MicrosoftAzureRSATLSIssuer = regexp.MustCompile(`^Microsoft Azure RSA TLS Issuing CA \d{2}$`)
-	// The azure Docs state that it should be DigiCert Global Root CA, but it is actually DigiCert Global Root G2 which is the newer version
-	DigiCertGlobalRootCA = regexp.MustCompile(`^DigiCert Global Root G2$`)
-)
-
 const (
-	// expected microsoft issuer host
+	// Default metadata domain for commercial Azure
+	DefaultMetadataDomain = "metadata.azure.com"
```
c4rlo, line 23: [Microsoft docs](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#signature-validation-guidance) list four different metadata domains; I guess it's reasonable to default to this one, which seems to be the main one, and require users to configure something else if needed.

However, the docs also state:

> The certificates might not have an exact match for the domain. For this reason, the certification validation should accept any subdomain (for example, in public cloud general availability regions accept `*.metadata.azure.com`).

That's not how `validateAzureCertificate()` works currently.

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
```diff
@@ -11,31 +11,23 @@ import (
...
+
+	// Expected microsoft issuer host for intermediate certificate fetching
 	MicrosoftIntermediateIssuerHost = "www.microsoft.com"
```
c4rlo, line 26: I know you didn't change the logic related to this, but I'm not sure if it's correct, couldn't find it explicitly documented.

[This doc](https://learn.microsoft.com/en-us/azure/security/fundamentals/azure-certificate-authority-details?tabs=root-and-subordinate-cas-list#certificate-downloads-and-revocation-lists) has a list of AIA hosts that includes `www.microsoft.com` but also a few others, so perhaps that would be a reasonable set.

Could leave unchanged for this PR, but may be worth a follow-up.

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
ChanghengGu: Base on testing on our production nodes I noticed cert details always has 2 SAN, ```metadata.azure.com``` and a region specific SAN, for example ```centralus.metadata.azure.com```, the implementation looks for the ```metadata.azure.com```, or ones that user configured.

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
ChanghengGu: Yea that seems a reasonable set to allowlist, I can do it as a follow up PR.

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
ChanghengGu: Updated in the latest commit

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
ChanghengGu: I checked each root CA in [Azure's doc](https://learn.microsoft.com/en-us/azure/security/fundamentals/azure-certificate-authority-details?tabs=root-and-subordinate-cas-list#root-certificate-authorities), we don't need all of them. 
```DigiCert Global Root G3```,  ```DigiCert TLS ECC P384 Root G5```, and ```Microsoft ECC Root CA 2017``` are ECC based and will not be used for IMDS (RSA).

I'm a bit hesitant on whether adding ```DigiCert TLS RSA 4096 Root G5``` and ```Microsoft RSA Root Certificate Authority 2017```, the RSA intermediate CAs listed on the documentation does not chain to either of these roots.

File: doc/plugin_server_nodeattestor_azure_imds.md [hunk 8]
```diff
@@ -15,10 +15,11 @@ attestation or to resolve selectors.
...
+| `tenants`                   | Required | A map of tenants, keyed by tenant domain, that are authorized for attestation. Tokens for unspecified tenants are rejected.                                                                                      |                                                                          |
+| `agent_path_template`       | Optional | A URL path portion format of Agent's SPIFFE ID. Describe in text/template format.                                                                                                                                | `"/{{ .PluginName }}/{{ .TenantID }}/{{ .SubscriptionID }}/{{ .VMID }}"` |
+| `allowed_metadata_domains`  | Optional | A list of allowed Azure metadata domains for certificate validation. Used to validate the Subject Alternative Name (SAN) in certificates from Azure IMDS. Useful for testing with custom Azure environments. | `["metadata.azure.com"]`                                                 |
```
sorindumitru: The table needs to be aligned

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
sorindumitru: I think it might be best to follow what the Azure validation docs say and do the wildcard match. This also means that users won't have to put in all possible values for their deployment, which I'm not sure how easy it would be for them to procure.

It maybe that now they all have the `metadata.azure.com` value, but they might not in the future.

ChanghengGu: Yeah, that makes sense. I think an alternative would be to let the user configure the base domain name, for example ```metadata.azure.com```, and have validation check either:

1. an exact match on the base domain, or
2. a match against the pattern ```*.configured_base_domain```.
That raises the question of whether multi-level subdomains should also be allowed. 

Azure’s documentation says that certificate validation should accept any subdomain, but not mentioned anything about depth. The example it gives is only a single-level subdomain of the form region.base_domain, ``` eastus.metadata.azure.com```
I think we should match multi level subdomain. If the user trusts the base domain, then any subdomain under it, regardless of depth, should also be trusted.

sorindumitru: Yes, I think configuring the base domain makes sense. The docs don't really specify this very well, but I agree with you, if trust `foo.metadata.azure.com` you probably also trust `bar.foo.metadata.azure.com`. So it may be enough to check if any of the DNS SANs end with any of the configured base domain names?

I think it's probably still ok to allow multiple base domain names, similar to what's in the PR now.

ChanghengGu: Updated to allow multiple base domains and validation would attempt for each configured base domain 1. exact match on the base domain 2. try to match the pattern *.configured_base_domain.

doc/plugin_server_nodeattestor_azure_imds.md
ChanghengGu: Aligned the tabled in the latest commit.

sorindumitru:
Thanks @ChanghengGu, looks like some commits are missing DCO, could you fix that up?

ChanghengGu:
> Thanks @ChanghengGu, looks like some commits are missing DCO, could you fix that up?

Fixed. Added sign-offs to commits that were missing them.

sorindumitru (approved):
thanks @ChanghengGu !

PR #6963, opened 2026-05-17
Update golangci-lint version + add `modernize` linter

### LINKED ISSUE #7076: spire-server 1.15.x: server TCP endpoint TLS handshake aborts with "session ticket keys unavailable" for clients offering psk_dhe_ke (e.g. Envoy)
## Summary
spire-server 1.15.x: server TCP endpoint TLS handshake aborts with "session ticket keys unavailable" for clients offering `psk_dhe_ke` (e.g. Envoy)

## Affected
- spire-server **1.15.0** and **1.15.1** (Go 1.26). **1.14.x is unaffected.**
- Server TCP endpoint (`bind_address`/`bind_port`). Any client whose ClientHello advertises `psk_key_exchange_modes: psk_dhe_ke` — Envoy/BoringSSL always does; a Go `crypto/tls` client only does with a `ClientSessionCache`.

## Symptom
mTLS to the server TCP endpoint fails. The server sends a ServerHello then resets the connection with no TLS alert. With `GRPC_GO_LOG_SEVERITY_LEVEL=info`:

```
grpc: Server.Serve failed to create ServerTransport: connection error:
  desc = "ServerHandshake("<peer>") failed: tls: internal error: session ticket keys unavailable"
```

Clients see a connection reset / `EOF` reading the HTTP/2 server preface.

## Root cause
`createTCPServer` sets `SessionTicketsDisabled: true` on the **top-level** `tls.Config` (`endpoints.go:311`, added in #6715) but the per-client config returned by `getTLSConfig` / `GetConfigForClient` does **not** set it.

Go 1.26 `crypto/tls` splits these (`handshake_server.go:173-176`):
- `c.ticketKeys = originalConfig.ticketKeys(configForClient)` → keys come from the **original** config → `nil` (disabled there).
- `c.config = configForClient`, and `shouldSendSessionTickets()` gates on **`c.config.SessionTicketsDisabled`** → the **per-client** value → **false** (not set).

So when the client offers `psk_dhe_ke`, the server attempts a TLS 1.3 session ticket, has no keys, and `encryptTicket` returns `tls: internal error: session ticket keys unavailable` (`ticket.go`), aborting the handshake right after ServerHello. (Go 1.25 tolerated the asymmetric config; 1.26 does not.)

## Fix
Also disable session tickets on the per-client config:
```go
spiffeTLSConfig.SessionTicketsDisabled = true
```
This line is **already on `main`** ([here](https://github.com/spiffe/spire/blob/main/pkg/server/endpoints/endpoints.go#L471)), it was added seemingly incidentally by the golangci-lint PR #6963, not as a recognized fix, and is in **no released tag** (1.15.0 and 1.15.1 both lack it).

## Verified end-to-end
Built spire-server from the v1.15.1 source twice and ran it behind an Envoy sidecar terminating client mTLS to the TCP endpoint:
- **unpatched** → handshake fails with the error above; the gRPC client sees connection reset / `EOF`.
- **+ `spiffeTLSConfig.SessionTicketsDisabled = true`** → handshake completes, mTLS calls succeed.

## Request
Backport to `release/1.15` for a 1.15.2, so deployments fronting spire-server with Envoy can upgrade off 1.14.x.

This is blocking us from upgrading to 1.15.x.

c4rlo:
- Bump `golangci-lint` to latest version (2.12.2) and address findings (or in some cases, suppress them when they occur in test code).
- Add [`modernize`](https://golangci-lint.run/docs/linters/configuration/#modernize) linter. This uses the official https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize so hopefully is uncontroversial to enable.

Fixes: #7076

### DIFF
--- a/pkg/agent/attestor/node/node.go
+++ b/pkg/agent/attestor/node/node.go
@@ -273,7 +273,7 @@ func (a *attestor) serverConn(bundle *spiffebundle.Bundle) (*grpc.ClientConn, er
 	// reached what appears to be the right trust domain server.
 	tlsConfig := &tls.Config{
 		InsecureSkipVerify: true, //nolint: gosec // this is required in order to do non-hostname based verification
-		VerifyPeerCertificate: func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+		VerifyPeerCertificate: func(rawCerts [][]byte, _ [][]*x509.Certificate) error { //nolint:gosec // we don't need SessionTicketsDisabled
 			a.c.Log.Warn("Insecure bootstrap enabled; skipping server certificate verification")
 			if len(rawCerts) == 0 {
 				// This is not really possible without a catastrophic bug
--- a/pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
+++ b/pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
@@ -209,7 +209,7 @@ func (p *Plugin) serveNonce(ctx context.Context, l net.Listener, agentName strin
 		fmt.Fprintln(w, nonce)
 	})
 
-	go func() {
+	go func() { //nolint:gosec // G118: complaint about context.Background()
 		<-ctx.Done()
 		_ = s.Shutdown(context.Background())
 	}()
--- a/pkg/agent/plugin/nodeattestor/v1_test.go
+++ b/pkg/agent/plugin/nodeattestor/v1_test.go
@@ -186,7 +186,7 @@ func (plugin *fakeV1Plugin) AidAttestation(stream nodeattestorv1.NodeAttestor_Ai
 func challengeResponses(ss ...[]byte) map[string]string {
 	set := make(map[string]string)
 	for i := 0; i < len(ss); i += 2 {
-		set[string(ss[i])] = string(ss[i+1])
+		set[string(ss[i])] = string(ss[i+1]) //nolint:gosec // incorrect "G602: slice index out of range" finding
 	}
 	return set
 }
--- a/pkg/agent/plugin/nodeattestor/v1_test.go
+++ b/pkg/agent/plugin/nodeattestor/v1_test.go
@@ -184,9 +184,12 @@ func (plugin *fakeV1Plugin) AidAttestation(stream nodeattestorv1.NodeAttestor_Ai
 }
 
 func challengeResponses(ss ...[]byte) map[string]string {
-	set := make(map[string]string)
-	for i := 0; i < len(ss); i += 2 {
-		set[string(ss[i])] = string(ss[i+1]) //nolint:gosec // incorrect "G602: slice index out of range" finding
+	if len(ss)%2 != 0 {
+		panic(fmt.Sprintf("challengeResponse: arg len (%d) must be even", len(ss)))
 	}
-	return set
+	result := make(map[string]string)
+	for i := 0; i < len(ss); i += 2 {
+		result[string(ss[i])] = string(ss[i+1]) //nolint:gosec // silence "G602: slice index out of range" finding
+	}
+	return result
 }
--- a/pkg/agent/plugin/workloadattestor/k8s/k8s.go
+++ b/pkg/agent/plugin/workloadattestor/k8s/k8s.go
@@ -516,6 +516,7 @@ func (p *Plugin) reloadKubeletClient(config *k8sConfig) (err error) {
 	// with the raw kubelet certs that we can verify directly.
 	case config.NodeName == "":
 		tlsConfig.InsecureSkipVerify = true
+		tlsConfig.SessionTicketsDisabled = true
 		tlsConfig.VerifyPeerCertificate = func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
 			var certs []*x509.Certificate
 			for _, rawCert := range rawCerts {
--- a/pkg/common/catalog/external.go
+++ b/pkg/common/catalog/external.go
@@ -166,7 +166,7 @@ func (p *hcClientPlugin) GRPCClient(ctx context.Context, b *goplugin.GRPCBroker,
 		}
 	})
 
-	ctx, cancel := context.WithCancel(ctx)
+	ctx, cancel := context.WithCancel(ctx) //nolint:gosec // G118: false complaint about cancel not being called
 	wg.Go(func() {
 		<-ctx.Done()
 		if !gracefulStopWithTimeout(server) {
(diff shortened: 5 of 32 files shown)


pkg/agent/attestor/node/node.go
```diff
@@ -273,7 +273,7 @@ func (a *attestor) serverConn(bundle *spiffebundle.Bundle) (*grpc.ClientConn, er
 	// reached what appears to be the right trust domain server.
 	tlsConfig := &tls.Config{
 		InsecureSkipVerify: true, //nolint: gosec // this is required in order to do non-hostname based verification
-		VerifyPeerCertificate: func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+		VerifyPeerCertificate: func(rawCerts [][]byte, _ [][]*x509.Certificate) error { //nolint:gosec // we don't need SessionTicketsDisabled
```
c4rlo, line 276: [Docs](https://pkg.go.dev/crypto/tls#Config.SessionTicketsDisabled):

> Note that on clients, session ticket support is also disabled if ClientSessionCache is nil.

We're not setting `ClientSessionCache` on the config, hence session tickets are disabled anyway.

pkg/agent/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -209,7 +209,7 @@ func (p *Plugin) serveNonce(ctx context.Context, l net.Listener, agentName strin
 		fmt.Fprintln(w, nonce)
 	})
 
-	go func() {
+	go func() { //nolint:gosec // G118: complaint about context.Background()
```
c4rlo, line 212: Using `ctx` instead of `context.Backgroud()` below wouldn't make sense, since `ctx` is known to be cancelled at that point.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -516,6 +516,7 @@ func (p *Plugin) reloadKubeletClient(config *k8sConfig) (err error) {
 	// with the raw kubelet certs that we can verify directly.
 	case config.NodeName == "":
 		tlsConfig.InsecureSkipVerify = true
+		tlsConfig.SessionTicketsDisabled = true
```
c4rlo, line 519: Again session tickets are disabled anyway since it's a client config without `ClientSessionCache` set, but it also doesn't hurt to set this flag here.

pkg/common/catalog/external.go
```diff
@@ -166,7 +166,7 @@ func (p *hcClientPlugin) GRPCClient(ctx context.Context, b *goplugin.GRPCBroker,
 		}
 	})
 
-	ctx, cancel := context.WithCancel(ctx)
+	ctx, cancel := context.WithCancel(ctx) //nolint:gosec // G118: false complaint about cancel not being called
```
c4rlo, line 169: `cancel` is being assigned to `hcPlugin.closers` below (via `closerFuncs`).

pkg/server/endpoints/endpoints.go
```diff
@@ -468,6 +468,7 @@ func (e *Endpoints) getTLSConfig(ctx context.Context) func(*tls.ClientHelloInfo)
 		spiffeTLSConfig.MinVersion = tls.VersionTLS12
 		spiffeTLSConfig.NextProtos = []string{http2.NextProtoTLS}
 		spiffeTLSConfig.VerifyPeerCertificate = e.serverSpiffeVerificationFunc(bundleSrc)
+		spiffeTLSConfig.SessionTicketsDisabled = true
```
c4rlo, line 471: Note that `createTCPServer()` already sets `SessionTicketsDisabled`, therefore it probably does make sense to set it here too. I haven't looked into whether it actually makes a difference, I guess it depends where session tickets come in relative to receipt of ClientHello (which is when `GetConfigForClient` is documented to get called).

pkg/server/plugin/keymanager/azurekeyvault/client_fake.go
```diff
@@ -398,14 +400,30 @@ func toRSAKey(publicKey crypto.PublicKey, kmsKeyID string, keyOperations []*azke
...
-		Crv: new(curveName),
-		//D:      ecdsaKey.D.Bytes(),
+		Crv:    new(curveName),
 		KID:    new(azkeys.ID(keyName)),
 		KeyOps: keyOperations,
 		Kty:    new(azkeys.KeyTypeEC),
-		X:      ecdsaKey.X.Bytes(),
-		Y:      ecdsaKey.Y.Bytes(),
+		X:      slices.Clone(encodedPoint[1 : 1+coordinateLength]),
+		Y:      slices.Clone(encodedPoint[1+coordinateLength:]),
```
c4rlo, line 426: A reasonable alternative might be to keep the old code and suppress the linter complaint. However, there is a risk that one day `ecdsaKey.X` will no longer be available as a public field, hence I opted for this more cumbersome way.

pkg/agent/plugin/nodeattestor/v1_test.go
```diff
@@ -186,7 +186,7 @@ func (plugin *fakeV1Plugin) AidAttestation(stream nodeattestorv1.NodeAttestor_Ai
 func challengeResponses(ss ...[]byte) map[string]string {
 	set := make(map[string]string)
 	for i := 0; i < len(ss); i += 2 {
-		set[string(ss[i])] = string(ss[i+1])
+		set[string(ss[i])] = string(ss[i+1]) //nolint:gosec // incorrect "G602: slice index out of range" finding
```
sorindumitru: This doesn't seem like it's incorrect, it's just that the map always has an even number of items. Maybe we can add a `t.Fail()` if that's the case?

pkg/server/datastore/sqlstore/mysql.go
```diff
@@ -125,7 +125,7 @@ func configureConnection(cfg *configuration, isReadOnly bool) (*mysql.Config, er
 
 	// MySQL still allows, and in some places requires, older TLS versions. For example, when built with yaSSL, it is limited to TLSv1 and TLSv1.1.
 	// TODO: consider making this more secure by default
-	tlsConf := tls.Config{} //nolint: gosec // see above
+	tlsConf := tls.Config{}
```
sorindumitru: Would probably should also remove the comment above since `tls.Config{}` has been defaulting to TLS 1.2 for a while now. We also dropped official support for MySQL 5.X, hopefully newer versions do not require older TLS versions.

pkg/agent/plugin/nodeattestor/v1_test.go
c4rlo: Addressed in the latest commit. I've opted for a `panic` to avoid having to pass a `testing.T` into this simple helper.

pkg/server/datastore/sqlstore/mysql.go
c4rlo: Addressed in the latest commit.

sorindumitru (approved):
LGTM, thanks @c4rlo!

PR #6964, opened 2026-05-17
Keep CA journal in-memory state after datastore save failures

c4rlo:
Callers of journal mutation methods (i.e. CA managers) treat journal save failures as non-fatal: they log the error and continue using the newly prepared or activated authority/key. Rolling back only the in-memory journal makes it diverge from the manager’s live slot state, and it does not fix the already-stale persisted journal.

Therefore, stop rolling back in-memory journal mutations when datastore saves fail. Keeping the in-memory journal advanced lets a later successful save persist the latest known manager state.

The previous rollback logic was actually flawed even on its own terms: the `backup` value used was only a reference, not its own independent copy. We also weren't restoring the `AuthorityId` field in `UpdateJWTKeyStatus` etc.

Also:
- simplify journal size trimming with `slices.Clone`
- update outdated journal comments and add missing journal test coverage for WITs

### DIFF
--- a/pkg/server/ca/manager/journal.go
+++ b/pkg/server/ca/manager/journal.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/x509"
 	"fmt"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


amartinezfayo (approved):
Thank you so much @c4rlo for this fix! Great cleanup also.

PR #6972, opened 2026-05-20
delegatedAPI: do not server JWT-SVIDs for dmin or downstream entries

sorindumitru:
We already do this for X509-SVID and it would be good to have this done for JWT-SVIDs too.


amartinezfayo (approved):
Thank you @sorindumitru!

PR #6981, opened 2026-05-22
agent: add `log_selectors` config item

### LINKED ISSUE #6936: agent: log non-sensitive selectors
#6179 added logging of selector values as part of "No identity issued" log entries (i.e. failures to issue SVIDs). This was useful for troubleshooting; e.g. if the cause was a missing registration entry, it would make it possible to see what the entry should have been.

#6732 removed these selectors again from all log entries out of concern that selector values can contain sensitive data and therefore should not be logged.

Since the original feature was useful, I would like to propose finding a middle-ground and logging only selector values that are non-sensitive. For sensitive ones, we could either leave them out completely, or log them as (e.g.) `docker:env:<redacted>`.

There's a few different ways this could work that I can think of, in increasing order of sophistication:

1. spire-agent could have a hardcoded list of selector types that it logs. This would imply that selectors from custom plugins never get logged.
2. It could be a new agent config item, e.g.: `log_selectors = ["k8s:ns", "k8s:sa", "unix:user"]`.
3. Plugins could advertise which selectors are safe to log through a new request type on the [WorkloadAttestor](https://github.com/spiffe/spire-plugin-sdk/blob/main/proto/spire/plugin/agent/workloadattestor/v1/workloadattestor.proto) gRPC service. Existing plugins that don't implement this are assumed to have no loggable selectors. It could be something like (pseudo-protobuf-code):
    ```protobuf
    rpc GetNonSensitiveSelectorTypes() returns (repeated string);
    ```

Personally I'd probably lean towards option 1 at least initially, to keep things simple.

The next question would be which specific selector types across all the built-in workload attestation plugins should be considered non-sensitive / loggable. I think the answer is the majority of them. Exceptions should include `docker:env`, and perhaps also `docker:label` and `k8s:pod-label` and maybe a few others.

I'd be happy to help implement this, if the maintainers agree to this enhancement in some form.

c4rlo:
- Add `log_selectors` config item to log given selector values on relevant errors (which are all "No identity issued").
- Minor refactor to compensate for added method parameters (methods already had uncomfortably many params, I didn't want to make it worse).

Fixes #6936.


pkg/agent/endpoints/workload/handler.go
```diff
@@ -502,8 +507,48 @@ func loggerWithContextInfo(ctx context.Context, log logrus.FieldLogger, start ti
...
 }
 
-func logNoIdentityIssued(ctx context.Context, log logrus.FieldLogger, start time.Time) {
-	loggerWithContextInfo(ctx, log.WithField(telemetry.Registered, false), start, nil).Error("No identity issued")
+func (h *Handler) logNoIdentityIssued(ctx context.Context, log logrus.FieldLogger, selectors []*common.Selector, start time.Time) {
+	fields := logrus.Fields{
+		telemetry.Registered: false,
+	}
+	if loggableSelectors := filterLoggableSelectors(selectors, h.c.LogSelectors); len(loggableSelectors) > 0 {
+		fields[telemetry.Selectors] = loggableSelectors
```
amartinezfayo: I'm wondering if it would be worth rendering the loggable selectors as a comma-joined `type:value` string here rather than storing the raw `[]*common.Selector`. I think the existing convention for the `telemetry.Selectors` field elsewhere is a string. `api.SelectorFieldFromProto` emits something like `k8s:ns:default,unix:user:root`, whereas a slice of proto pointers renders as proto text under the text formatter and as `[{"Type":...,"Value":...}]` under the JSON formatter.
Matching the existing string form would keep the `selectors` field shape consistent and read a little more cleanly.

amartinezfayo (commented):
Thank you @c4rlo for this contribution!

pkg/agent/endpoints/workload/handler.go
c4rlo: I see your point. What I did here is reflect how selector logging in the agent was originally implemented in #6179, but I see that spire-server tends to log selectors as a single string as you say. I've now changed it to do it the same way – I see the benefit of consistency as well as less verbosity.

I also considered a middle-ground of logging as `["k8s:ns:default", "unix:user:root"]`, which might be slightly more readable (in JSON format at least) and avoid ambiguity in case a selector value contains a comma. But I don't have a strong opinion on it so I went with your suggestion.

c4rlo: Actually this brings me to another point: it might be good to sort the selectors before logging them, as they would appear to be unordered. I think having them in some canonical order could be quite helpful, e.g. when searching logs etc.

I've now done this too. If this seems reasonable, I might submit a follow-up PR to sort selectors in a few other code paths too where we log selectors (and also convert them to log them as a single string where that's not the case).

amartinezfayo: Thank you for the updates! Sorting before logging makes sense to me as well, deterministic output should make searching and correlating logs easier.
A follow-up PR to apply the same treatment (sorting and single-string rendering) to the other code paths that log selectors sounds good to me. One thing to keep in mind for that follow-up is that it would change the shape of log output that users may already be parsing or alerting on, so I think we should treat it as a change that goes into a minor release with a changelog note. I would avoid having that change in a patch release.

amartinezfayo (approved):
🚀

PR #6982, opened 2026-05-24
Makefile: go-check: allow suffixed Go versions

c4rlo:
Some Go toolchains report a version suffix, for example `go1.26.3-X:nodwarf5` on my Arch Linux machine. Strip any dash suffix before comparing the installed Go version against the required version from `.go-version`.

This avoid unnecessary Go toolchain downloads.


sorindumitru (approved):
LGTM, thanks @c4rlo!

PR #6983, opened 2026-05-24
Upgrade protoc and markdownlint versions

c4rlo:
Upgrade to the latest versions of protoc and markdownlint.

protoc did not introduce any changes aside from the updated version number.

markdownlint has two new rules:
- [MD059 - Link text should be descriptive](https://github.com/DavidAnson/markdownlint/blob/main/doc/md059.md)
- [MD060 - Table column style](https://github.com/DavidAnson/markdownlint/blob/main/doc/md060.md)

Fortunately AI makes it quite easy to comply with these (though I had to prod it a bit to keep the MD060 changes minimal).

Of course we also have the option of choosing to ignore one of both of these in [.markdownlint.yaml](https://github.com/spiffe/spire/blob/main/.markdownlint.yaml) if we prefer.


sorindumitru (approved):
LGTM, thanks @c4rlo!

PR #6994, opened 2026-05-28
events: keep previous first/lastEvent information when reloading

### LINKED ISSUE #6876: Skipped event ID spikes after full cache rebuild delay SVID issuance
## Summary

We observe periodic spikes in `spire_server_node_skipped_node_event_ids_count` and `spire_server_entry_skipped_entry_event_ids_count` that correlate with the `full_cache_reload_interval` cycle. During these spikes, newly attested agents do not receive authorized entries for 70–90+ seconds, causing workloads to fail with "no identity issued."

## Environment

- SPIRE Server v1.14.4
- `events_based_cache = true`
- HA deployment (4 server pods)
- Single database instance (no read replicas)

## What We Observed

1. Every ~24 hours (matching the default `full_cache_reload_interval`), `spire_server_node_skipped_node_event_ids_count` spikes to 10,000–25,000 across server pods.

2. The spike duration varies between 2 minutes and 2 hours depending on the pod, with the higher spikes taking proportionally longer to resolve.

3. The spike amplitude varies too, commonly they peak at ~7,000–10,000, but sometimes they reach ~25,000.

4. Different server pods in the HA deployment spike to different levels and resolve at different rates.

5. During these spikes, agents that complete node attestation successfully do not receive any authorized entries from `SyncAuthorizedEntries` for 70–90+ seconds. Server-side logs confirm:
   - Node attestation completes (e.g., at 09:37:23)
   - The agent repeatedly calls for authorized entries and receives 0
   - X.509-SVID is finally signed ~91 seconds later (09:38:54)

6. Outside of these spike windows, newly attested agents receive their entries and SVIDs within a few seconds.

## Impact

Applications depending on SPIRE for mTLS credentials fail to start or crash when they exhaust their retry budget (60–90 seconds) waiting for X.509 credentials from the SPIRE agent workload API. This has caused incidents on multiple occasions.

## Possibly Related Issues

- #5624
- #5021

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**
fixes #6876

### DIFF
--- a/pkg/server/endpoints/authorized_entryfetcher.go
+++ b/pkg/server/endpoints/authorized_entryfetcher.go
@@ -44,7 +44,7 @@ type AuthorizedEntryFetcherEvents struct {
 type eventsBasedCache interface {
 	updateCache(ctx context.Context) error
 	loadCache(ctx context.Context, cache *authorizedentries.Cache) error
-	commitReload(cache *authorizedentries.Cache)
+	swapCache(cache *authorizedentries.Cache)
 }
 
 func NewAuthorizedEntryFetcherEvents(ctx context.Context, trustDomain string, c AuthorizedEntryFetcherEventsConfig) (*AuthorizedEntryFetcherEvents, error) {
@@ -153,8 +153,8 @@ func (a *AuthorizedEntryFetcherEvents) reloadCache(ctx context.Context) error {
 		return err
 	}
 
-	a.registrationEntries.commitReload(cache)
-	a.attestedNodes.commitReload(cache)
+	a.registrationEntries.swapCache(cache)
+	a.attestedNodes.swapCache(cache)
 
 	a.mu.Lock()
 	a.cache = cache
--- a/pkg/server/endpoints/authorized_entryfetcher.go
+++ b/pkg/server/endpoints/authorized_entryfetcher.go
@@ -43,6 +43,8 @@ type AuthorizedEntryFetcherEvents struct {
 
 type eventsBasedCache interface {
 	updateCache(ctx context.Context) error
+	loadCache(ctx context.Context, cache *authorizedentries.Cache) error
+	commitReload(cache *authorizedentries.Cache)
 }
 
 func NewAuthorizedEntryFetcherEvents(ctx context.Context, trustDomain string, c AuthorizedEntryFetcherEventsConfig) (*AuthorizedEntryFetcherEvents, error) {
@@ -91,7 +93,7 @@ func (a *AuthorizedEntryFetcherEvents) RunUpdateCacheTask(ctx context.Context) e
 			return ctx.Err()
 		case <-cacheReloadTicker.C:
 			if fullCacheReload {
-				if err := a.buildCache(ctx); err != nil {
+				if err := a.reloadCache(ctx); err != nil {
 					a.c.log.WithError(err).Error("Failed to full refresh entry cache")
 					continue
 				}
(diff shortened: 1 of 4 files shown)


pkg/server/endpoints/authorized_entryfetcher_attested_nodes.go
```diff
@@ -239,6 +239,11 @@ func (a *attestedNodes) updateCachedNodes(ctx context.Context) error {
 	return nil
 }
 
+func (a *attestedNodes) commitReload(cache *authorizedentries.Cache) {
```
MarcosDY: NIT: may we call it swapCache?

pkg/server/endpoints/authorized_entryfetcher.go
```diff
@@ -140,6 +141,24 @@ func (a *AuthorizedEntryFetcherEvents) updateCache(ctx context.Context) error {
...
+func (a *AuthorizedEntryFetcherEvents) reloadCache(ctx context.Context) error {
+	cache := authorizedentries.NewCache(a.c.clk, a.trustDomain)
+
+	if err := a.registrationEntries.reloadCache(ctx, cache); err != nil {
+		return err
+	}
+
+	if err := a.attestedNodes.reloadCache(ctx, cache); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
MarcosDY: this sounds like a real issue?

pkg/server/endpoints/authorized_entryfetcher.go
```diff
@@ -140,6 +142,27 @@ func (a *AuthorizedEntryFetcherEvents) updateCache(ctx context.Context) error {
...
+	if err := a.registrationEntries.loadCache(ctx, cache); err != nil {
+		return err
+	}
+
+	if err := a.attestedNodes.loadCache(ctx, cache); err != nil {
+		return err
+	}
+
+	a.registrationEntries.commitReload(cache)
+	a.attestedNodes.commitReload(cache)
```
MarcosDY: mmm this is not in a lock right? can cause some issues?

pkg/server/endpoints/authorized_entryfetcher_test.go
```diff
@@ -1073,6 +1073,359 @@ func TestFullCacheReloadRecoversFromSkippedAttestedNodeEvents(t *testing.T) {
 	compareEntries(t, entries, entry)
 }
 
+func TestReloadCacheRecoversFromSkippedRegistrationEntryEvents(t *testing.T) {
```
MarcosDY, line 1077: NIT: a unit test to validate concurrent writes (it may be hard to write... so avoid if it complicate test too much)

pkg/server/endpoints/authorized_entryfetcher.go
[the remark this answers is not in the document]
sorindumitru: I think I addressed that with the `commitReload`

pkg/server/endpoints/authorized_entryfetcher.go
sorindumitru: Had to check this one, but I think it's safe. We only call into the cache when the cache is being updated/refreshed and that happens from a single goroutine.

pkg/server/endpoints/authorized_entryfetcher_test.go
sorindumitru: There's only one writer with these caches so I've added one with concurrent reads, just in case it can expose some issues.

MarcosDY (approved):

Issue #7001, opened 2026-05-29
x509pop grouping selector


kfox1111:
## Problem Statement
When deploying a the spire-iidentity-exchange alongside a spire-agent inside a standard Kubernetes Deployment (rather than a DaemonSet), each individual pod's spire-agent must obtain a unique SPIFFE ID via a node attestation method like x509pop to avoid kicking each other out.

However, because these agents act on behalf of a singular, logical "identity exchange" service, a SPIRE administrator needs a clean way to register downstream workload entries using a single, static parentID (e.g., spiffe://example.org/spire-identity-exchange). Currently, there is a missing link to bridge the gap between unique per-agent SPIFFE IDs and a common identity/selector that a Node Alias can target.

Without this, managing individual registration entries for every scaled replica of an identity-exchange pod becomes unmaintainable.

## Proposed Solution Architecture
The goal is to deploy spire-identity-exchange as a standard Kubernetes Deployment.

The host-level agents run standard node attestation (e.g., k8s_psat).

The specialized Deployment pods use the SPIFFE CSI driver to receive an SVID, then use x509pop to attest to the SPIRE server.

The server needs to generate a non-unique, common selector based on the attested agent's path metadata so a Node Alias can target all instances of this deployment.

### Option 1: Introduce Path/Parent Selector Logic to x509pop (Preferred)
Modify the x509pop server node attestor to generate a selector based on a "parent" or "trimmed" path derived from the configured agent_path_template.

Example Structure:

SVID provided to agent: spiffe://spire-exchange/k8s/clusterfoo/spire-identity-exchange/abcdefg

Generated Agent SVID: spiffe://spire/agent/x509pop/k8s/clusterfoo/spire-identity-exchange/abcdefg

New Auto-Generated Selector:
x509pop:agent_path_parent:spiffe://spire/agent/x509pop/k8s/clusterfoo/spire-identity-exchange (analogous to a dirname operation on the path template).

A single Node Alias entry can then target x509pop:agent_path_parent:spiffe://spire/agent/x509pop/k8s/clusterfoo/spire-identity-exchange and project the target spiffe://example.org/spire-identity-exchange parent ID to all instances seamlessly.

### Option 2: Static Group Config for x509pop
Introduce an option such as extra_groups or static selectors into the x509pop configuration that allows adding generic classification selectors during attestation.

Example Config: config.x509pop.extraGroups = ["spire-identity-exchange"]

Resulting Selector: x509pop:static_group:spire-identity-exchange

[...]

PR #7002, opened 2026-05-29
Add a group selector to x509pop

kfox1111:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
x509pop mode gets a new selector that lets you easily setup node aliases.

**Description of change**
When in x509pop, a new selector is added that enables grouping for node aliases.
 
**Which issue this PR fixes**
https://github.com/spiffe/spire/issues/7001


doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -48,11 +48,12 @@ A sample configuration:
...
+| Common Name      | `x509pop:subject:cn:example.org`                                  | The Subject's Common Name (see X.500 Distinguished Names)                                                                                                                                                     |
+| SHA1 Fingerprint | `x509pop:ca:fingerprint:0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33` | The SHA1 fingerprint as a hex string for each cert in the PoP chain, excluding the leaf.                                                                                                                      |
+| SerialNumber     | `x509pop:serialnumber:0a1b2c3d4e5f`                               | The leaf certificate serial number as a lowercase hexadecimal string                                                                                                                                          |
+| San              | `x509pop:san:<key>:<value>`                                       | The san selectors on the leaf certificate. The expected format of the uri san is `x509pop://<trust_domain>/<key>/<value>`. One selector is exposed per uri san corresponding to x509pop uri scheme. string    |
```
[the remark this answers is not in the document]
kfox1111: Unrelated to the change at hand. the line was changed just for spacing.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -313,11 +313,23 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
...
+
+	if config.mode == "spiffe" {
+		agentIDStr := spiffeid.String()
+
+		if lastSlash := strings.LastIndex(agentIDStr, "/"); lastSlash != -1 {
+			parentID := agentIDStr[:lastSlash]
+			parentID = strings.TrimSuffix(parentID, "/")
+			selectors = append(selectors, "agent_id_parent:" + parentID)
+		}
+	}
```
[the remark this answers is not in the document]
kfox1111: why did you change the order of the removal of the last slash? that ends up changing the behavior?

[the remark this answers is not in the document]
kfox1111: since you cant seem to answer and produced bad code, lets skip this suggestion.

pkg/server/plugin/nodeattestor/x509pop/x509pop_test.go
```diff
@@ -182,6 +182,14 @@ func (s *Suite) TestAttestSuccess() {
 				{Type: "x509pop", Value: "san:environment:production"},
 				{Type: "x509pop", Value: "san:key:path/to/value"},
 			}
+
+			if tt.desc == "success with spiffe exchange" {
+				expectedSelectors = append(expectedSelectors, &common.Selector{
+					Type:  "x509pop",
+					Value: "agent_id_parent:spiffe://example.org/spire/agent/x509pop",
+				})
+			}
```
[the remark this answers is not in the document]
kfox1111: Suggestion accepted.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -313,11 +313,23 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
...
+	selectors := buildSelectorValues(leaf, chains, sanSelectors)
+
+	if config.mode == "spiffe" {
+		agentIDStr := spiffeid.String()
+
+		if lastSlash := strings.LastIndex(agentIDStr, "/"); lastSlash != -1 {
+			parentID := agentIDStr[:lastSlash]
+			parentID = strings.TrimSuffix(parentID, "/")
+			selectors = append(selectors, "agent_id_parent:"+parentID)
+		}
```
sorindumitru: This is a very single-use selector, for the case that you mentioned. We can publish the SPIFFE ID as the selector. For your particular case you could make the agent SPIFFE ID template be derived from the SHA1 fingerprint, to get unique SPIFFE IDs, and have the full SPIFFE ID be `spiffe://trust domain/x509pop/identity-exchange`, which you can then use as a node alias. You might have to make sure that no 2 agents get deployed to the same node so that they do not use the same exact SVID to attest themselves.

kfox1111: Can you please give an example of what the selector for the node alias would actually be in this configuration? I dont understand how it is possible with the approach your suggesting.

the docs mention selectors:
```
Common Name	x509pop:subject:cn:example.org	The Subject's Common Name (see X.500 Distinguished Names)
SHA1 Fingerprint	x509pop:ca:fingerprint:0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33	The SHA1 fingerprint as a hex string for each cert in the PoP chain, excluding the leaf.
SerialNumber	x509pop:serialnumber:0a1b2c3d4e5f	The leaf certificate serial number as a lowercase hexadecimal string
San	x509pop:san:<key>:<value>	The san selectors on the leaf certificate. The expected format of the uri san is x509pop://<trust_domain>/<key>/<value>. One selector is exposed per uri san corresponding to x509pop uri scheme. string
```

the san selector wont work as a spiffe svid will never have a url with x509pop.

the serial number/sha1 fingerprint selectors are not useful as you will never know what it is to alias with. common name wont be defined with spiffe certs.
the experimental spiffe_id selector will always be the full spiffeid that is genreated from the template. if it has the fingerprint in it, its not useable as a generic point to alias over either.

So, there is a lack of a selector that fits the use case I think.

It is ok if it is narrowly targeted a selector. it only is needed for x509 pop with spiffe mode. There are lots of selectors that only are useful for certain use cases.

sorindumitru: The SPIFFE ID of the SVID the agent presents to spire-server to attest would be:
* `spiffe://trust domain/spire-identity-exchange` (or something like that this). The same one for all agents.
* The SPIFFE of the SVID resulting from the attestation with the above SVID would be `spiffe://trust domain/spire/agent/x509pop/<cert fingerprint>`, instead of the usual format for spiffe mode. You'd need to make sure that the agents get unique "spire-identity-exchange" SVIDs so that they get unique fingerprints.

We'd still need to add the id as a selector so you can use in the node aliase, but that feels like a reasonable thing to do. 

> It is ok if it is narrowly targeted a selector. it only is needed for x509 pop with spiffe mode.

It's not that it's only needed for spiffe mode, it's only needed for this particular deployment. Most deployments wouldn't care about the "parent id".

kfox1111: Ah. I see.

I think that has two drawbacks.
1. I would be worried about an agent who's identifier can change over time. When reattesting, it will get a new cert which will have a new fingerprint.
2. You'd never know the fingerprint if you still did want to hang off svds off of the individual agents up front. Forcing it to be the fingerprint id would make it harder to predict. I was planning on using the {{ pod.Name }} in it for stable identifiers in case of a statefulset in k8s.

I agree though, either way, we need a new selector. We're just debating the name of it I guess. So we could pick a name potentially independent of the rest.

This does raise a discussion I wanted to have at some point about dynamic entries. This idea does kind of match an idea I've been rolling around in my head. It would be useful to simplify my own entry management, as well as potentially helping dynamic entries for spire-identity-exchange too.

In most of my usage, I've run across needing to tack on the node name over and over to identifiers to guarantee uniqueness of identity.
```
example.com/spire/agent/node123
example.com/sshd/node123
example.com/fluent-bit/node123
example.com/node-agent/node123
example.com/kubelet/node123
```
Times the number of nodes. It gets a bit painful after a while.

But, what if we could set some labels on the node:
example.com/spire/agent/node123 (label nodeName=node123)

we could put generic workload entries on it:
{example.com/node123, example.com/node234}-> example.com/k8s-node-group -> {example.com/sshd, example.com/fluent-bit, example.com/node-agent, example.com/kubelet}

Each one would use those spiffeid's for lookups and resolving workloads. but, during issuence of the svid, it would have an extra template that would resolve before issuing the final spiffeid. so it could say, append "/{{ nodeLabels.nodeName }}"

sorindumitru: It's not clear to me if the solution above is ok for you, which would mean that we'd have just an `agent_spiffe_id` selector, based on the full SPIFFE ID of the presented SVID.

We can continue the other discussion independently of this.

kfox1111: If your asking if the proposed solution of just making the underlying spiffe id be the common group part, and then making them unique by adding the SHA1 fingerprint automatically to it for this special case, then no, I dont think that is a good way to solve it. It makes it too hard to:
* hang off individual spiffeids off the individual agents when needed
* It can make an agent get a completely different agent id when it gets a new svid from the underlying system and its sha changes. I dont think spire-agent will handle this right, along with the registration entries now pointing at the wrong agent when individual ones are registered.

sorindumitru: If I understand correctly, your proposed solution here also has a random, changing, identifier in the agent identity, the pod UUID. So that's something you have to deal with anyway. I also don't think it's that big of a deal since you're planning on using the node alias.

I'm not sure it's worth adding this selector which is very specific to your deployment and is likely hard to reason about in a large deployment. The full spiffe id would be ok to have.

kfox1111: You keep falling back to very specific cases, designing a feature that is so specific, when the next case comes out, we're back to square one. IMO we should design features that are useful for more then just one use case so we're not constantly adding features.

The way I'm using it in this particular case, yes a pod uid is somewhat unknowable up front. But, it will never change during the life of the pod. So adding it later is stable. Reboot the node unexpectedly, the pod uid stays the same. The spire-agent gets reattested, it still gets the same spiffeid while the fingerprint changes. If you really want to put a stable identity on it so you can hang off per agent entries, you use the {.Metadata.Name}} from the pod and a statefulset to give it a persistent name. Persistent identity. done. I'm not 100% sure I will never need to do that. So, why preoptimize the ability to do that completely away? Besides, we still havent addressed how your proposal will work when the fingerprint changes. That is deal breaking on its own IMO. So, I'm still at the point of, that isnt a good solution.

I"m not asking for a ton of code here. We're talking for like 12 lines of code and a unit test. This really shoudlnt be burdensome on the spire team. That saves 57 lines of code https://github.com/spiffe/spire-identity-exchange/blob/main/cmd/spire-credentialcomposer-identity-exchange/main.go
and a whole separate plugin and infrastructure, and is much easier for the end user to deal with. This is a fair optimization I think. Over all the spire projects, spire is maintains less code.

kfox1111: I had to make the credential composer configurable for its prefix. so its up to 87 lines now.

sorindumitru:
In general we have one main issue with this proposal, that we start interpreting the SPIFFE ID provided and assuming some grouping. This is not something that we can do, since users are allowed to specify SPIFFE-IDs in ways that do not provide any grouping or may not provide a grouping in the way that the selector is constructed.

The selector also can be confusing and hard to use since you end up needing to have all nodes in a trust domain follow a particular pattern, otherwise you end up with confusing selectors/aliases. If we were to reconsider this in the future, we'd like to find out if this is a problem or not.

We'd be open to reconsidering something like this if we see more users which have a similar requirement. In that case we'd still need to see how to make it work for all users so it may or may not end up being different than this proposal.

In the mean time, I think you have a few options for dealing with this, including the one you're currently using with external code.

kfox1111:
The workaround I came up with was for the spire-identity-exchange with x509pop attestors. we're going to hit the same issue with all pod agents using x509pop in the chart next. We'll suffer through with our existing solution until we can convince you guys of the need.

The only thing saving you from this in the k8s psat pod attestor just added is the existing selectors. if all you had was a spiffeid to work with like this one does, youd be in the same boat as us. You just havint had to use this particular configuration yet to suffer like I am, and for the record, that sucks. :(

I did leave other options open, such as having a static set of selectors configurable. That would still be preferable to a whole credential composer mutating things IMO.

I hope you can agree at least that the need for the credential composer is a bit hackish. But there has been no other viable alternative presented yet that supports chained trust. Just a k8s psat attestor that doesnt meet the need.

What other, viable, options are there that would be acceptable?

sorindumitru:
Hi @kfox1111,

Would something like the following config work for you?
```
group_path_template = "{{ dir .SVIDPathTrimmed }}"
groups = [
    "/cluster/foo/identity-exchange",
    "/cluster/foo/bar",
]
```

The `group_path_template` would be applied to the SPIFFE-ID/any selector and it must result in a group from `group`. If the template is empty, no group selector is produced. I think this address my concerns with this being a single format for groups and make it hard to tell which prefixes are used for node aliases to avoid using those prefixes in other SPIFFE IDs.

kfox1111:
Hmm... Yeah, I think that would work well. Using the template that way is good. It does resolve any ambiguity on how the groups are generated.

But, I'm just trying to understand what you are trying to protect against with the group list. Maybe there is something I'm missing.

The x509pop spiffe mode already has a gate for if a spiffe_id is allowed in or not. svid_prefix. if the given x509 cert is not under svid_prefix, it's not allowed to connect at all. So we're already filtering out things that shouldn't connect.

The current code is just for an additional selector. The additional selector is not useful without a registration entry using it. So the sysadmin is directly involved in specifying if a group is valid or not by sticking it in a registration entry.

The proposal puts in a second gate saying a selector cant be generated unless it is in the allowed list. Is there a use case I'm missing where that gate provides additional protection that svid_prefix + registration entries are not enough?

sorindumitru:
> The proposal puts in a second gate saying a selector cant be generated unless it is in the allowed list. Is there a use case I'm missing where that gate provides additional protection that svid_prefix + registration entries are not enough?

We can't assume that there's inherent grouping here. As I mentioned before, that's true in your deployment but it's not going to be true for everyone. So defining which groups can exist allows to know when the group selector should be emitted or not.

It also provides a place where the admin states the intent of which group selectors they want to exist. They can be visible in the entries, yes, but in large deployments it's a lot easier to view them here. Care needs to be taken with this feature to make sure that agents do not accidentally end up with a group selector  that they shouldn't have, so visibility into what groups are being used is good to have.

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -48,12 +50,13 @@ A sample configuration:
...
+| SerialNumber     | `x509pop:serialnumber:0a1b2c3d4e5f`                               | The leaf certificate serial number as a lowercase hexadecimal string                                                                                                                                                 |
+| San              | `x509pop:san:<key>:<value>`                                       | The san selectors on the leaf certificate. The expected format of the uri san is `x509pop://<trust_domain>/<key>/<value>`. One selector is exposed per uri san corresponding to x509pop uri scheme. string           |
+| group            | `x509pop:group:/cluster/foo/identity-exchange`                    | Group path derived from `group_path_template` in SPIFFE mode. Only emitted when the derived path is in `groups`.                                                                                       |
```
sorindumitru: To match the other rows:
```suggestion
| Group            | `x509pop:group:/cluster/foo/identity-exchange`                    | Group path derived from `group_path_template` in SPIFFE mode. Only emitted when the derived path is in `groups`.                                                                                       |
```

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -343,11 +361,43 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
 		return status.Errorf(codes.Internal, "failed to make spiffe id: %v", err)
 	}
 
+	selectors := buildSelectorValues(leaf, chains, sanSelectors)
+
+	if config.mode == "spiffe" && config.groupPathTemplate != nil {
```
sorindumitru: There's no need to guard this on `spiffe` mode anymore.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -343,11 +361,43 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
...
+			Certificate:     leaf,
+			SerialNumberHex: x509pop.SerialNumberHex(leaf.SerialNumber),
+			Fingerprint:     x509pop.Fingerprint(leaf),
+			PluginName:      x509pop.PluginName,
+			TrustDomain:     config.trustDomain.Name(),
+			SVIDPathTrimmed: svidPath,
+			URISanSelectors: sanSelectors,
+		}
+		groupPath, err := config.groupPathTemplate.Execute(groupData)
+		if err == nil {
```
sorindumitru: Is it possible to log something on `err != nil`? Or would that be too noisy?

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
```diff
@@ -343,11 +361,43 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
...
+			for _, allowed := range config.groups {
```
sorindumitru: As the littered mentioned, please use `slices.Contains`.

pkg/server/plugin/nodeattestor/x509pop/x509pop_test.go
```diff
@@ -149,6 +150,21 @@ func (s *Suite) TestAttestSuccess() {
 			certs:         s.leafBundle, // has 1 leaf and 1 intermediate
 			serialnumber:  "serialnumber:0a1b2c3d4e5f",
 		},
+		{
+			desc:          "success with group selector",
+			expectAgentID: "spiffe://example.org/spire/agent/x509pop/testhost",
+			giveConfig:    s.createConfigurationModeSPIFFE(`group_path_template = "{{ .SVIDPathTrimmed }}"` + "\n" + `groups = ["testhost"]`),
```
sorindumitru: The test looks good, but it would be good if there was some grouping involved here.

pkg/server/plugin/nodeattestor/x509pop/x509pop_test.go
```diff
@@ -149,6 +150,21 @@ func (s *Suite) TestAttestSuccess() {
...
+			certs:         s.svidExchange,
+			serialnumber:  "serialnumber:0a1b2c3d4e7f",
+			expectGroup:   "group:testhost",
```
sorindumitru: To void some issues when writing these tests we could have just the group name in here:
```suggestion
			expectGroup:   "testhost",
```

and then add the `group:` prefix In the expectation. That way nobody can accidentally expect a non-group selector using this field.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
kfox1111: I dont follow. SVIDPath / SVIDPathTrimmed is only valid in mode spiffe.

Maybe this is some of the confusion from before about guarding with the group list... Are you assuming this mechanism will be used with external_pki as well?

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
kfox1111: We could do it as a debug level log message maybe?

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
sorindumitru: It could be used with external PKI too, you're passing in the certificate to the template which can then be used to look at a bunch of fields. Similarly for the URI san selectors.

Will it be useful in that mode? I don't know. But there doesn't seem to be a reason to restrict it to just the spiffe mode.

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
sorindumitru: Yes, we could

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
kfox1111: Hm... yeah, I see. Ok. I'll update it so that it works either way.

kfox1111: Actually, I think that means we can emit just groups, and not paths. so maybe we should rename it to just group_template

pkg/server/plugin/nodeattestor/x509pop/x509pop.go
kfox1111: reworked

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -26,6 +26,8 @@ spiffe://<trust_domain>/spire/agent/x509pop/<fingerprint>
...
 | `verify_client_ip`    | If `true`, validates the connecting peer's IP against the leaf certificate's IP SANs. Attestation fails if no SAN matches. Reflects the immediate peer - may not represent true client origin behind load balancers.                           | false                                                          |
+| `group_template`      | Go text/template used to derive one or more group values from the certificate. If empty, no group selector is produced. See [Group Template](#group-template).                                                                                 |                                                                |
+| `groups`              | Allowlist of group values that may be emitted as selectors. A group produced by `group_template` must be in this list to be emitted. Required when `group_template` is set.                                                                    |                                                                |
```
sorindumitru: I think it might be good to stick with just one group for now. Supporting multiple groups adds a lot of extra complexity to this PR and doesn't seem like something that would be used too much.

kfox1111: ok. backed it off to be single group only again.

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -84,3 +87,53 @@ Some useful values are:
...
+    -selector x509pop:group:/cluster/foo/identity-exchange
+```
+
+The selector value comes from a template, and selector values are not otherwise
+validated, so the `groups` allowlist rather than the template is what bounds
+which groups a node can claim. It is required for that reason.
+
+```hcl
+    group_template = "{{ .SVIDPathTrimmed }}"
+    groups = ["/cluster/foo/identity-exchange"]
```
sorindumitru: Looking at what `SVIDPathTrimmed` is, it may not include the `/`. So a more accurate example here maybe without the `/`.

doc/plugin_server_nodeattestor_x509pop.md
```diff
@@ -26,6 +26,8 @@ spiffe://<trust_domain>/spire/agent/x509pop/<fingerprint>
...
 | `verify_client_ip`    | If `true`, validates the connecting peer's IP against the leaf certificate's IP SANs. Attestation fails if no SAN matches. Reflects the immediate peer - may not represent true client origin behind load balancers.                           | false                                                          |
+| `group_template`      | Go text/template used to derive a group value from the certificate. If empty, no group selector is produced. See [Group Template](#group-template).                                                                                            |                                                                |
+| `groups`              | Allowlist of group values that may be emitted as selectors. A group produced by `group_template` must be in this list to be emitted. Required when `group_template` is set.                                                                    |                                                                |
```
sorindumitru: `groups` feels a bit too generic, let's call this `allowed_groups`.

sorindumitru (commented):
Looking good, just two small changes might improve the documentation a bit.

sorindumitru (approved):
LGTM, thanks @kfox1111 !

PR #7006, opened 2026-05-30
Add tag-based key discovery support in the `aws_kms` KeyManager plugin

### LINKED ISSUE #6269: AWS KMS Plugin - Code scan all the KMS keys in the account after every 48 hours

amartinezfayo:
The `aws_kms` KeyManager plugin currently discovers its keys by listing every KMS alias and calling `DescribeKey` across the account's keys. This is not the most efficient way to discover SPIRE-managed keys, and it becomes problematic in deployments with a large number of keys in AWS KMS, where it is slow and requires broad `kms:ListKeys` and `kms:DescribeKey` permissions.

This PR adds an opt-in tag-based key discovery mode that uses the AWS Resource Groups Tagging API (`tag:GetResources`) to find only the keys managed by the plugin, identified by SPIRE-specific tags. This mirrors the label-based approach already used by the `gcp_kms` plugin.

Changes:
- New `enable_tag_based_key_discovery` configuration option (boolean, default `false`).
- When enabled, keys are discovered via the Resource Groups Tagging API filtered by SPIRE tags (`spire-server-td`, `spire-server-id`, `spire-active`) and looked up by `spire-key-id`.
- Keys are stamped with `spire-last-update` at creation and during migration, and the value is refreshed periodically (every 6h). Stale keys (no refresh for two weeks) are marked inactive and scheduled for deletion, mirroring the alias-based liveness and reclamation contract.
- Transparent migration: on first startup with the flag enabled, the plugin detects existing alias-managed keys and applies the SPIRE tags to them automatically. No manual steps are required.
- Aliases are still created and refreshed in tag-based mode so that a downgrade to alias-based discovery remains safe.
- When tag-based discovery is enabled, a missing `tag:GetResources` permission now produces an actionable error explaining that the permission is identity-based and cannot be granted through the KMS key policy, instead of a raw AWS error.

Permissions:
- `kms:TagResource` is required when tag-based discovery is enabled (or when using `key_tags`).
- `tag:GetResources` is required when tag-based discovery is enabled. It belongs to the Resource Groups Tagging API and must be granted in an identity-based IAM policy.
- `kms:ListKeys` is only required for the legacy alias-based discovery path.

Deprecation:
Alias-based key discovery will be deprecated in a future version and removed in a later one. The default remains alias-based for now, and a warning is logged when it is in use.

HA and operational note:
Within a trust domain, `enable_tag_based_key_discovery` should be configured consistently across all servers. The two discovery modes use independent liveness signals (alias `LastUpdatedDate` vs the `spire-last-update` tag), so a partial rollback sustained beyond the two-week window can cause tag-based peers to reclaim a rolled-back server's still-in-use keys. This is documented in the plugin docs.

Documentation:
Updated `doc/plugin_server_keymanager_aws_kms.md` with the new option, the tag schema, the migration behavior, the permission split, the deprecation note, and the HA configuration-consistency guidance.

Fixes #6269


pkg/server/plugin/keymanager/awskms/fetcher.go
```diff
@@ -142,3 +149,238 @@ func (kf *keyFetcher) spireKeyIDFromAlias(aliasName string) (string, bool) {
...
+			spireKeyID, ok := kf.spireKeyIDFromTags(resource.Tags)
+			if !ok {
+				kf.log.Warn("Could not get SPIRE key ID from tags", "key_arn", keyArn)
+				continue
+			}
+
+			// Trigger a goroutine to get the details of the key
+			g.Go(func() error {
+				entry, err := kf.fetchKeyEntryDetailsFromArn(ctx, keyArn, spireKeyID)
+				if err != nil {
```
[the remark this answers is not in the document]
amartinezfayo: `keyArn` and `spireKeyID` are declared inside the loop body, so each goroutine captures its own copy (and the range var is per-iteration on Go 1.22+ anyway). No race here.

pkg/server/plugin/keymanager/awskms/awskms_test.go
```diff
@@ -2563,3 +2572,606 @@ func waitForSignal(t *testing.T, ch chan error) error {
...
+			if tt.err != "" {
+				require.EqualError(t, err, tt.err)
+				return
+			}
+			require.NoError(t, err)
+
+			// Wait for expected deletions to be processed
+			for range tt.expectDeleteCount {
+				_ = waitForSignal(t, deleteSignal)
+			}
```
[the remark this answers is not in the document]
amartinezfayo: Range-over-int is valid since Go 1.22 (we are on 1.26 in SPIRE).

pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -1078,6 +1202,29 @@ func buildKeyTags(tags map[string]string) []types.Tag {
 	return keyTags
 }
 
+// buildSPIRETags creates the SPIRE-specific tags that are added to all KMS keys
+// at creation time. These tags enable efficient key discovery via the AWS
+// Resource Groups Tagging API.
+//
+// Note: spire-last-update is intentionally omitted here. It is set exclusively
+// by keepActiveKeys, which runs on a regular schedule.
+func (p *Plugin) buildSPIRETags(serverID, trustDomain string) []types.Tag {
```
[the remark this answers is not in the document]
amartinezfayo: Fixed.

pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -107,6 +129,20 @@ type Config struct {
...
+	// spire-server-id, spire-active).
+	// This eliminates the need for broad ListKeys + DescribeKey permissions and reduces API costs.
+	//
+	// Default: false (uses legacy alias-based discovery)
+	// In a future SPIRE version, this will default to true. The alias-based
+	// approach will be deprecated in a future version and removed in a later one.
+	//
+	// Note: When enabled, the plugin requires permission to use the
+	// resourcegroupstaggingapi:GetResources API action.
+	EnableTagBasedKeyDiscovery bool `hcl:"enable_tag_based_key_discovery" json:"enable_tag_based_key_discovery"`
```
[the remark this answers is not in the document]
amartinezfayo: Fixed.

pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -34,16 +37,29 @@ const (
...
 	aliasPrefix = "alias/SPIRE_SERVER/"
 
+	// Logging tags
 	keyArnTag    = "key_arn"
 	aliasNameTag = "alias_name"
 	reasonTag    = "reason"
 
+	// KMS resource tags for key discovery
+	tagKeyServerTD   = "spire-server-td"   // Trust domain (no hashing needed - AWS allows dots and long values)
+	tagKeyServerID   = "spire-server-id"   // Server identifier
```
sorindumitru, line 47: Double checking my assumptions: I see that AWS KMS has a limit on the tag key to be 128 chars and the value to be 256 chars. This should be ok for both of these since the trust domain should be a maximum of 256 chars and the server ID we require to be a UUIDv4, right?

pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -235,10 +314,23 @@ func (p *Plugin) Configure(ctx context.Context, req *configv1.ConfigureRequest)
...
+	// Build the tag list applied to every new key. SPIRE-specific tags are
+	// only included when tag-based discovery is enabled, so that the legacy
+	// alias-based path does not require the kms:TagResource permission.
+	switch {
+	case useTagBasedDiscovery && len(newConfig.KeyTags) > 0:
+		userTags := buildKeyTags(newConfig.KeyTags)
+		// Build a fresh slice to avoid mutating the spireTags backing array.
+		p.keyTags = make([]types.Tag, 0, len(spireTags)+len(userTags))
+		p.keyTags = append(p.keyTags, spireTags...)
+		p.keyTags = append(p.keyTags, userTags...)
```
sorindumitru, line 328: Should we check that there are no duplicate tags between `spireTags` and `userTags`? I don't know if it matters to the API, but the users might want to know.

pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -1094,3 +1242,218 @@ func decodeKeyID(keyID string) string {
...
+				continue
+			}
+
+			// Mark the key as inactive by updating the spire-active tag.
+			_, err := p.kmsClient.TagResource(ctx, &kms.TagResourceInput{
+				KeyId: &keyArn,
+				Tags: []types.Tag{
+					{
+						TagKey:   aws.String(tagKeyActive),
+						TagValue: aws.String("false"),
```
sorindumitru, line 1443: I think it's possible for the key to be enqueued for deletion and for us to mark the key as inactive, but for the key to not actually be deleted, for example if the key was deleting some other key and then shuts down.

Do we actually need to do the `ScheduleKeyDeletion` call asynchronously?

pkg/server/plugin/keymanager/awskms/awskms.go
amartinezfayo: Thanks for checking. That's right: the tag keys are fixed constants under 128 chars, and the values fit within 256. The trust domain is length-bounded by the SPIFFE spec, and the server ID is either an auto-generated UUID or a `key_identifier_value` that we already cap at 256 chars in `buildConfig`. The remaining values (`spire-active`, `spire-last-update`, `spire-key-id`) are short.

pkg/server/plugin/keymanager/awskms/awskms.go
amartinezfayo: A collision can't happen: `validateTags` rejects any user tag key beginning with the reserved `spire-` prefix (and `aws:`) at config time, and every plugin-managed tag uses `spire-`, so user tags and SPIRE tags never overlap. I also updated the docs, which previously called `spire-` "reserved for future use", to say "reserved for SPIRE-managed tags" since it's in use now.

pkg/server/plugin/keymanager/awskms/awskms.go
amartinezfayo: Thanks for raising this. You're describing a real gap: the alias path tolerates it because `disposeKeys` sweeps alias-less keys by creation date as a fallback, but tag mode doesn't run that sweeper, so a crash between marking the key inactive and the async deletion would orphan it. I changed disposal to call `ScheduleKeyDeletion` synchronously and mark the key inactive only after deletion is scheduled, so the key can't be dropped from the active filter without a deletion pending. An already-pending-deletion result is treated as success so a re-run still clears the tag.

pkg/server/plugin/keymanager/awskms/awskms.go
```diff
@@ -1411,22 +1411,31 @@ func (p *Plugin) disposeKeysViaTags(ctx context.Context) error {
...
+			_, err := p.kmsClient.ScheduleKeyDeletion(ctx, &kms.ScheduleKeyDeletionInput{
+				KeyId:               &keyArn,
+				PendingWindowInDays: aws.Int32(7),
+			})
+			// A key that is already pending deletion (for example, scheduled on
+			// a previous cycle that stopped before marking it inactive) is
+			// treated as success so the spire-active tag can still be cleared.
+			// Any other error leaves the key active so it is retried next cycle.
+			var invalidState *types.KMSInvalidStateException
+			if err != nil && !errors.As(err, &invalidState) {
```
sorindumitru, line 1430: Would `errors.Is(err, type.KMSInvalidStateException)` also work here? It's a bit shorter

amartinezfayo: It would be nice, but errors.Is doesn't work for this one. Passing the type directly doesn't compile (types.KMSInvalidStateException (type) is not an expression), and the SDK error type has no Is(error) bool method, so even constructing an instance would only match by == equality against a fresh value and never match the real error.

sorindumitru (approved):
LGTM, thanks @amartinezfayo !

PR #7017, opened 2026-06-03
spire-agent: implement logger service

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**


cmd/spire-agent/util/util.go
```diff
@@ -0,0 +1,110 @@
...
+		fmt.Fprintln(a.env.Stderr, "Error: "+err.Error())
+		return 1
+	}
+
+	return 0
+}
+
+func (a *Adapter) Help() string {
+	return a.flags.Parse([]string{"-h"}).Error()
+}
```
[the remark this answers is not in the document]
sorindumitru: This is consistent with what the server does. So I think this should be fine.

cmd/spire-agent/util/util.go
```diff
@@ -0,0 +1,110 @@
...
+
+	return a
+}
+
+func (a *Adapter) Run(args []string) int {
+	ctx := context.Background()
+
+	if err := a.flags.Parse(args); err != nil {
+		return 1
+	}
```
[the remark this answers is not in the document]
sorindumitru: This is consistent with what the server does. So I think this should be fine.

pkg/common/api/logger/printers_test.go
```diff
@@ -2,15 +2,34 @@ package logger_test
...
-	"github.com/spiffe/spire/cmd/spire-server/cli/logger"
+	"github.com/spiffe/spire/pkg/common/api/logger"
 	commoncli "github.com/spiffe/spire/pkg/common/cli"
+	"github.com/stretchr/testify/require"
 )
 
+type errorWriter struct {
+	ReturnError error
+	buf         strings.Builder
+}
```
c4rlo, line 18: I see you added this here because this file uses that type. But you've still got another copy of it in cmd/spire-server/cli/logger/mocks_test.go and another one in cmd/spire-agent/cli/logger/mocks_test.go, both of which are now unused.

pkg/agent/api/logger/v1/service.go
```diff
@@ -0,0 +1,96 @@
...
+	return s.createAPILogger(), nil
+}
+
+func (s *Service) SetLogLevel(ctx context.Context, req *loggerv1.SetLogLevelRequest) (*apitype.Logger, error) {
+	rpccontext.AddRPCAuditFields(ctx, logrus.Fields{telemetry.NewLogLevel: req.NewLevel.String()})
+	log := rpccontext.Logger(ctx)
+
+	if req.NewLevel == apitype.LogLevel_UNSPECIFIED {
+		return nil, api.MakeErr(log, codes.InvalidArgument, "newLevel value cannot be LogLevel_UNSPECIFIED", nil)
+	}
```
c4rlo, line 63: Seems like this would be caught by the `LogrusLevel` lookup below so can remove this check.

pkg/agent/api/logger/v1/service_test.go
c4rlo, line 1: There are a lot of very similar test cases in here – not sure they're all worth their keep...

pkg/agent/api/logger/v1/service.go
sorindumitru: Yeah, maybe. This mostly follows the server side where I think it wants to give a better error message for the case where you forgot to put in a value in the request, as opposed to putting in some invalid value.

I agree the next check catches this, but I'll leave this in to be consistent with the server side and to get marginally better error messages.

pkg/agent/api/logger/v1/service_test.go
sorindumitru: Yes, again this mostly follows the server side of things which has a similar array of tests. I agree it's a bit much for the functionality here, but it follows what we do for the server.

pkg/common/api/logger/printers_test.go
sorindumitru: Thanks for catching this

rausingh-rh:
This PR will close issue https://github.com/spiffe/spire/issues/4986

pkg/agent/api/logger/v1/service.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+	loggerv1 "github.com/spiffe/spire-api-sdk/proto/spire/api/agent/logger/v1"
+	apitype "github.com/spiffe/spire-api-sdk/proto/spire/api/types"
+	commonlogger "github.com/spiffe/spire/pkg/common/api/logger"
+	"github.com/spiffe/spire/pkg/common/telemetry"
+	"github.com/spiffe/spire/pkg/server/api"
+	"github.com/spiffe/spire/pkg/server/api/rpccontext"
```
rausingh-rh: importing `pkg/server/api/rpccontext` and `pkg/server/api` creates a cross-boundary dependency where the agent package depends on server internals. The agent has its own `pkg/agent/api/rpccontext package` can we use that- so that we are consistent with how it's done in `DelegatedIdentity`?

rausingh-rh:
There's no `AgentLoggerServiceName` in `pkg/common/api/middleware/names.go` and no Logger switch case in `pkg/agent/endpoints/metrics.go`. Due to this, every Logger API call will hit the default branch switch case block and emit an `unrecognized service for connection metrics` error log. Similar issue was fixed in #6878 for the Debug API. The existing [TestAllAgentServicesHandledByConnectionMetrics](https://github.com/spiffe/spire/blob/f4070d88a87fdc1d1be1f3b52a424c81a8f94a21/pkg/agent/endpoints/metrics_test.go#L83) in  test should catch this. worth running it with the Logger service registered to verify.

pkg/agent/api/logger/v1/service.go
```diff
@@ -0,0 +1,96 @@
...
+	if !ok {
+		return nil, api.MakeErr(log, codes.InvalidArgument, "unsupported log level", nil)
+	}
+
+	log.WithFields(logrus.Fields{
+		telemetry.NewLogLevel: newLogLevel.String(),
+	}).Info("SetLogLevel Called")
+	s.log.SetLevel(newLogLevel)
+
+	rpccontext.AuditRPC(ctx)
```
rausingh-rh: `AuditRPC` calls in this file will silently no-op on the agent since the agent middleware doesn't set up audit tracking on the context

doc/spire_agent.md
```diff
@@ -354,6 +354,31 @@ Checks SPIRE agent's health.
 | `-socketPath` | Path to the SPIRE Agent API socket    | /tmp/spire-agent/public/api.sock |
 | `-verbose`    | Print verbose information             |                                  |
 
+### `spire-agent logger get`
+
+Gets the current logging level of the SPIRE Agent.
+
+| Command       | Action                             | Default                          |
+|:--------------|:-----------------------------------|:---------------------------------|
+| `-socketPath` | Path to the SPIRE Agent API socket | /tmp/spire-agent/public/api.sock |
```
rausingh-rh: CLI defaults to `/tmp/spire-agent/private/admin.sock` for `-socketPath`. Also the description should say `admin API socket` instead of `API socket` since the Logger service is only registered on the admin server, not the workload API server.

pkg/agent/api/logger/v1/service.go
sorindumitru: This ended up being a bigger change because it was a pre-existing issue, but it makes sense to fix it. Kept it in a separate commit

pkg/common/api/middleware/names.go
```diff
@@ -19,7 +19,7 @@ const (
 	EnvoySDSv3ServiceShortName         = "SDS.v3"
 	HealthServiceName                  = "grpc.health.v1.Health"
 	HealthServiceShortName             = "Health"
-	LoggerServiceName                  = "logger.v1.Logger"
+	LoggerServiceName                  = "spire.api.agent.logger.v1.Logger"
```
MarcosDY: I think this LoggerServiceName change (`logger.v1.Logger` to `spire.api.agent.logger.v1.Logger`) might regress the server side, it is shared for both.

`spire.api.server.logger.v1.Logger` from `Logger` to `logger.v1.Logger` while `spire.api.agent.logger.v1.Logger` from `spire.api.agent.Logger` to `Logger`,.

So the agent resolves cleanly now but the server logger's metric key regresses from `logger.<method>` to `logger.v1.logger.<method>`, 
Looks like there is no test for logger naming in pkg/common/api/middleware/*_test.go to catch this.

Can you add a full name for agent and keep the old one at it was?
and maybe add a test to catch it?
something like:

```
for _, tt := range []struct {
		name       string
		fullMethod string
	}{
		{name: "server", fullMethod: "/spire.api.server.logger.v1.Logger/SetLogLevel"},
		{name: "agent", fullMethod: "/spire.api.agent.logger.v1.Logger/SetLogLevel"},
	} {
			names := makeNames(tt.fullMethod)
			assert.Equal(t, "Logger", names.Service)
			assert.Equal(t, []string{"logger", "set_log_level"}, names.MetricKey)
		})
	}
```

cmd/spire-agent/cli/logger/set_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+func TestSet(t *testing.T) {
+	for _, tt := range []struct {
+		name string
+		// service state
+		service *mockLoggerService
+		// input
+		args []string
+		// expected items
+		expectedSetValue types.LogLevel
```
MarcosDY, line 33: Looks like expected is never used?
can you add `require.Equal(t, tt.expectedSetValue, *test.server.receivedSetValue)`?

doc/spire_agent.md
```diff
@@ -394,6 +394,31 @@ Checks SPIRE agent's health.
 | `-socketPath` | Path to the SPIRE Agent API socket    | /tmp/spire-agent/public/api.sock |
 | `-verbose`    | Print verbose information             |                                  |
 
+### `spire-agent logger get`
+
+Gets the current logging level of the SPIRE Agent.
+
+| Command       | Action                             | Default                             |
+|:--------------|:-----------------------------------|:------------------------------------|
+| `-socketPath` | Path to the SPIRE Agent API socket | /tmp/spire-agent/private/admin.sock |
```
MarcosDY: NIT: can you update `SPIRE Agent API socket` to `SPIRE Agent admin API socket`
Just to avoid confusion with another apis

cmd/spire-agent/cli/logger/set_test.go
sorindumitru: Same problem in the server code, fixed it there too

pkg/common/api/middleware/names.go
sorindumitru: Fixed this now. It's good to have the test, it shows the logger service is a bit different, it doesn't include the `v1` in the service name. Same for the delegated API one.

We can see about fixing them in 1.16.0, since it's a breaking change.

MarcosDY (approved):

PR #7019, opened 2026-06-03
Fix Dependabot auto-merge by exempting dependency manifests from code owner review

amartinezfayo:
The Dependabot auto-merge workflow has never successfully merged a PR. After CI passes, the workflow assigns the milestone and approves, but the merge step fails with `GraphQL: Auto merge is not allowed for this repository (enablePullRequestAutoMerge)`. An example of a failure is here: https://github.com/spiffe/spire/actions/runs/26658532935

The root cause is that `go.mod` and `go.sum` fall under the `*` rule in `CODEOWNERS`, so the "require review from code owners" branch protection requirement is never satisfied by the workflow's `github-actions[bot]` approval. Because the PR can never become merge-ready, `gh pr merge` falls back to enabling auto-merge, which is a feature that is intentionally disabled on this repository.

This PR exempts `/go.mod` and `/go.sum` from code owner review using the documented CODEOWNERS pattern of leaving the owners empty. With that, a verified Dependabot PR meets all branch protection requirements as soon as the workflow approves it, and `gh pr merge` adds it directly to the merge queue, the same as clicking "Merge when ready". The "Allow auto-merge" repository setting remains disabled and the review process for every other PR is unchanged. The merge step now also retries briefly, since GitHub recomputes the mergeability state asynchronously after the approval is submitted.

The workflow now only approves and merges PRs whose changed files are limited to `go.mod` and `go.sum`. For other Dependabot PRs, such as `github-actions` ecosystem updates that modify workflow files, it assigns the milestone and leaves review and merge to the maintainers instead of failing at the merge step.

Security considerations:

- No significant new attack vector is created for human-authored PRs. The workflow only approves and merges PRs that were opened by `dependabot[bot]` and contain exactly one commit cryptographically signed by Dependabot. A regular PR touching only `go.mod` and `go.sum` still requires an approving review from a collaborator with write access, and only a collaborator can add it to the merge queue. A PR touching `go.mod` together with any other file still requires code owner review, since ownership applies per file.
- A new `Module directive check` step in the lint job rejects `replace` and `exclude` directives in `go.mod`, so a `go.mod`-only PR cannot redirect a module to a different source repository. This closes the one meaningful capability that removing the code owner requirement would otherwise add.
- A 7-day Dependabot `cooldown` lets new upstream releases age before an update PR is opened, preserving a window for the ecosystem to detect and report a compromised release before it can land unattended. This mitigates the main residual supply chain exposure (a malicious version published upstream), which exists today as well, since routine version bumps are typically merged within a day or two of being opened. The cooldown applies to both the `gomod` and `github-actions` ecosystems.
- The merge queue still revalidates CI before any PR lands, and the workflow re-verifies that the PR head SHA has not changed between commit verification and merge.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,12 @@ updates:
     interval: "daily"
     time: "09:00"
     timezone: "America/Los_Angeles"
+  # Wait for new releases to age before opening update PRs. Since
+  # Dependabot patch/minor updates are auto-merged once CI passes,
+  # the cooldown preserves a window for the ecosystem to detect and
+  # report compromised releases before they can land unattended.
+  cooldown:
+    default-days: 7
   groups:
     actions:
       patterns:
@@ -50,4 +56,10 @@ updates:
   directory: "/"
   schedule:
    interval: daily
+  # Let new action releases age before opening update PRs, so that a
+  # compromised release has a window to be detected and reported by the
+  # ecosystem before it is proposed for review (e.g. the March 2025
+  # tj-actions/changed-files compromise was reverted within days).
+  cooldown:
+    default-days: 7
   open-pull-requests-limit: 20
--- a/.github/workflows/dependabot_automerge.yml
+++ b/.github/workflows/dependabot_automerge.yml
@@ -97,9 +97,23 @@ jobs:
 
           HEAD_SHA=$(echo "$COMMIT" | jq -r '.sha')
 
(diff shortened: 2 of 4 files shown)


amartinezfayo:
Closing in favor of #7070, which takes a different direction. We have decided to step back from automated Dependabot merges and switch to grouped weekly Dependabot PRs that maintainers review and merge by hand.

PR #7020, opened 2026-06-03
Experimental External CA mode

### LINKED ISSUE #7018: FR: Sign X.509-SVIDs with a static, externally-anchored CA (no SPIRE-minted downstream intermediate)
### Summary

SPIRE Server always signs X.509-SVIDs with an intermediate CA that SPIRE generates and rotates itself — either self-signed, or minted as a downstream of a configured UpstreamAuthority. There is no way to sign directly with a pre-existing, long-lived intermediate that the server doesn't own.

This requests an "external CA" mode where SPIRE Server loads an operator-supplied intermediate CA (+ root), signs SVIDs directly with its key via crypto.Signer (PKCS#11 as the first backend), treats the X.509 CA as static (no rotation), and keeps JWT key rotation unchanged.

###   The gap

  1. Intermediates are commonly issued leaf-constrained (CA:TRUE, pathlen:0), forbidding any CA beneath them. SPIRE's flow requires minting a downstream sub-CA, which such a cert cannot legally sign — chains fail path validation at conforming verifiers.
  2. Mandatory rotation breaks non-SPIFFE consumers. SPIRE rotates its X.509 CA frequently by design. Heterogeneous systems consuming SPIRE-issued certs — with no SPIFFE-aware edge to abstract trust away — can't continuously re-fetch the bundle and refresh chains. A stable, externally-anchored CA removes that friction.
  3. The workaround adds CA-management burden. Provisioning a pathlen≥1 intermediate just so SPIRE can mint short-lived sub-CAs relaxes existing PKI policy and introduces a new problem: securely authorizing each spire-server to request and hold a signing-capable sub-CA. Signing directly with a fixed CA sidesteps this.

###   Proposed behavior

An experimental server { experimental { external_ca { … } } } block that:
  - takes a root cert, an intermediate (signing) cert, and a crypto.Signer reference for the key (PKCS#11 first; design is signer-agnostic);
  - validates at startup that the intermediate is signed by the root and is a CA cert (fail fast otherwise);
  - uses that intermediate as the active X.509 signing authority and publishes root + intermediate in the trust bundle;
  - disables X.509 CA rotation; leaves JWT rotation untouched.

Config sketch
```
  server {
    experimental {
      external_ca {
        enabled              = true
        root_cert_file_path  = "/etc/spire/certs/root.pem"
        cert_file_path       = "/etc/spire/certs/intermediate.pem"
        pkcs11 {
          pkcs11_uri    = "pkcs11:module-path=/usr/lib/.../libXXX.so;token=SPIRE;pin-value=…"
          pkcs11_object = "pkcs11:object=intermediate-ca-key"
        }
      }
    }
  }
```

###   Scope / non-goals

  - Not changing JWT-SVID behavior — JWT keys still rotate.
  - Not rotating/re-issuing the external CA — its lifecycle is owned by the external PKI; SPIRE consumes it read-only.
  - Not replacing UpstreamAuthority — this is "sign directly with a supplied static CA," not "mint a downstream from an upstream."
  - SPIRE already supports HSM/KMS-backed signing via several UpstreamAuthority/KeyManager plugins. The novelty here is direct signing with a static, externally-owned CA.

###   Alternatives considered

  - UpstreamAuthority (disk/vault/PKCS#11-backed): still mints and rotates a SPIRE-owned downstream; incompatible with `pathlen:0`, still rotates the signing CA and doesn't scale well in heterogenous environments.
  - Provision a `pathlen≥1` intermediate: relaxes PKI policy and adds sub-CA bootstrapping/management on every server.

###   Contribution
I have this experimental feature implemented and I will shortly link a PR to support the above which can help with the discussion.

calderonth:
**Pull Request check list**

- [x] Documentation updated

**Affected functionality**
Adding an experimental feature which allows to configure `spire-server` to use an existing intermediate CA with signing happening via pkcs11 in a HSM.

**Description of change**
Allow SPIRE Server to sign X.509-SVIDs directly with a pre-existing,
long-lived intermediate CA instead of generating and rotating its own.
The intermediate certificate (and its root) are supplied by the operator
and the signing key is accessed through a crypto.Signer, with PKCS#11 as
the first supported backend (covers HSM and SoftHSM).

When external CA mode is enabled:
  * the supplied intermediate is used as the active X.509 signing authority
  * the root and intermediate are published in the trust bundle
  * X.509 CA rotation is disabled (the external CA is static)
  * JWT key rotation continues normally
  * startup validates that the intermediate is signed by the root and is a
    CA certificate, failing fast otherwise
 
**Which issue this PR fixes**
Fixes #7018

### DIFF
--- a/cmd/spire-server/cli/run/external_ca_test.go
+++ b/cmd/spire-server/cli/run/external_ca_test.go
@@ -1,0 +1,98 @@
+package run
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/spiffe/spire/pkg/common/log"
+	"github.com/stretchr/testify/require"
+)
+
+// TestParseConfigExternalCA verifies the experimental external_ca / pkcs11 HCL
+// block is decoded into the config struct.
+func TestParseConfigExternalCA(t *testing.T) {
+	hclConfig := `
+server {
+	bind_address = "127.0.0.1"
+	bind_port = "8081"
+	trust_domain = "example.org"
+	data_dir = "."
+
+	experimental {
+		external_ca {
+			enabled = true
+			root_cert_file_path = "/opt/spire/certs/root-cert.pem"
+			cert_file_path = "/opt/spire/certs/intermediate-cert.pem"
+			pkcs11 {
+				pkcs11_uri = "pkcs11:module-path=/usr/lib/softhsm/libsofthsm2.so;token=SPIRE;pin-value=1234"
+				pkcs11_object = "pkcs11:object=intermediate-ca-key"
+			}
+		}
+	}
+}
+plugins {}
+`
+	dir := t.TempDir()
+	confPath := filepath.Join(dir, "server.conf")
+	require.NoError(t, os.WriteFile(confPath, []byte(hclConfig), 0o600))
+
+	c, err := ParseFile(confPath, false)
+	require.NoError(t, err)
+
+	extCA := c.Server.Experimental.ExternalCA
+	require.NotNil(t, extCA)
+	require.True(t, extCA.Enabled)
+	require.Equal(t, "/opt/spire/certs/root-cert.pem", extCA.RootCertFilePath)
+	require.Equal(t, "/opt/spire/certs/intermediate-cert.pem", extCA.CertFilePath)
+	require.NotNil(t, extCA.PKCS11)
+	require.Equal(t, "pkcs11:module-path=/usr/lib/softhsm/libsofthsm2.so;token=SPIRE;pin-value=1234", extCA.PKCS11.Pkcs11URI)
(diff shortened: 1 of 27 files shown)


amartinezfayo:
Thank you for this contribution. I posted some feedback and a few clarifying questions on #7018: I have concerns about introducing a static, non-rotating X.509 signing authority, and I believe there may be paths that satisfy the underlying requirements while keeping rotation. I'd suggest we settle the design discussion on the issue before investing more time in this PR, so the implementation effort goes toward whichever direction we land on.

sorindumitru:
@calderonth We'll close this PR for now, we can discuss more in the issue, but the solution for this is likely to involve a new plugin type.

PR #7030, opened 2026-06-04
Added Azure bundle publisher

sabre1041:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Extend existing `BundlePublisher` capability by adding support for publishing the bundle to Azure Blob

**Description of change**
New BundlePublisher for integrating with Azure Blob. Similar capability to existing AWS and GCP bundle publisher

**Which issue this PR fixes**
N/A

### DIFF
--- a/doc/plugin_server_bundlepublisher_azure_blob.md
+++ b/doc/plugin_server_bundlepublisher_azure_blob.md
@@ -1,0 +1,104 @@
+# Server plugin: BundlePublisher "azure_blob"
+
+The `azure_blob` plugin puts the current trust bundle of the server in a designated
+Azure Blob Storage container, keeping it updated.
+
+The plugin accepts the following configuration options:
+
+| Configuration        | Description                                                                                                                                                                          | Required                                             | Default               |
+|----------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------|-----------------------|
+| storage_account_name | The name of the Azure Storage account.                                                                                                                                               | Yes.                                                 |                       |
+| storage_account_key  | The base64-encoded access key for the Azure Storage account. Used for shared key authentication (found in the Azure Portal, **Access keys**).                                        | Required only when using shared key authentication.  |                       |
+| container_name       | The name of the blob container to which the trust bundle is uploaded.                                                                                                                | Yes.                                                 |                       |
+| blob_name            | The blob name inside the container.                                                                                                                                                  | Yes.                                                 |                       |
+| format               | Format in which the trust bundle is stored, &lt;spiffe &vert; jwks &vert; pem&gt;. See [Supported bundle formats](#supported-bundle-formats) for more details.                       | Yes.                                                 |                       |
+| service_endpoint     | The Azure Blob Storage service endpoint.                                                                                                                                             | No.                                                  | blob.core.windows.net |
+| tenant_id            | The Azure tenant ID for client secret credential authentication.                                                                                                                     | Required only when using client secret credentials.  |                       |
+| app_id               | The Azure application (client) ID for client secret credential authentication.                                                                                                       | Required only when using client secret credentials.  |                       |
+| app_secret           | The Azure application client secret for client secret credential authentication.                                                                                                     | Required only when using client secret credentials.  |                       |
+| refresh_hint         | Sets the refresh hint for the bundle when using the spiffe format. Specified as string e.g. '10m', '1h'. See [time.ParseDuration](https://pkg.go.dev/time#ParseDuration) for details | No.                                                  |                       |
+
+## Supported bundle formats
+
+The following bundle formats are supported:
+
+### SPIFFE format
+
+The trust bundle is represented as an RFC 7517 compliant JWK Set, with the specific parameters defined in the [SPIFFE Trust Domain and Bundle specification](https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Trust_Domain_and_Bundle.md#4-spiffe-bundle-format). Both the JWT authorities and the X.509 authorities are included.
+
+### JWKS format
+
+The trust bundle is encoded as an RFC 7517 compliant JWK Set, omitting SPIFFE-specific parameters. Both the JWT authorities and the X.509 authorities are included.
+
+### PEM format
+
+The trust bundle is formatted using PEM encoding. Only the X.509 authorities are included.
+
+## Authentication
+
+The plugin supports three authentication methods. Only one method may be used at a time; shared key authentication and client secret credentials are mutually exclusive.
+
+### Shared key authentication
+
+When `storage_account_key` is provided, the plugin authenticates using the storage account's access key. This method does not require Azure AD and is useful in environments where Azure AD is not available.
+
+### Client secret credentials
+
+When `tenant_id`, `app_id`, and `app_secret` are all provided, the plugin authenticates using Azure client secret credentials. All three fields must be specified together.
+
+### Default Azure credentials
+
+When neither shared key nor client secret credentials are configured, the plugin uses the [Azure Default Credential](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication) chain. This supports Managed Identity, environment variables, Azure CLI credentials, and other methods provided by the Azure SDK.
+
+## Required permissions
+
+The authenticated identity must have the `Storage Blob Data Contributor` role (or equivalent permissions to write blobs) on the configured storage account or container.
+
+## Sample configuration using Default Azure Credentials
+
+The following configuration uploads the local trust bundle contents to the `example.org` blob in the `spire-bundle` container within the `mystorageaccount` storage account. Since client secret credentials are not configured, [Default Azure Credentials](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication) are used.
+
+```hcl
+    BundlePublisher "azure_blob" {
+        plugin_data {
+            storage_account_name = "mystorageaccount"
+            container_name = "spire-bundle"
+            blob_name = "example.org"
+            format = "spiffe"
+        }
+    }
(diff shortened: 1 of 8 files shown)


pkg/server/plugin/bundlepublisher/azureblob/azureblob.go
```diff
@@ -0,0 +1,303 @@
...
+func (p *Plugin) Configure(_ context.Context, req *configv1.ConfigureRequest) (*configv1.ConfigureResponse, error) {
+	newConfig, notes, err := pluginconf.Build(req, buildConfig)
+	if err != nil {
+		return nil, err
+	}
+	for _, note := range notes {
+		p.log.Warn(note)
+	}
+
+	serviceEndpoint := newConfig.ServiceEndpoint
```
rausingh-rh: can we put some validation on `serviceEndpoint` here as done [here](https://github.com/spiffe/spire/blob/7122d4865f41687106abd452b766ec8ff2ddb97f/pkg/server/plugin/bundlepublisher/awss3/awss3.go#L81) ?

pkg/server/plugin/bundlepublisher/azureblob/azureblob.go
```diff
@@ -0,0 +1,303 @@
...
+		bundleToPublish.RefreshHint = config.parsedRefreshHint
+	}
+
+	formatter := bundleformat.NewFormatter(bundleToPublish)
+	bundleBytes, err := formatter.Format(config.bundleFormat)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "could not format bundle: %v", err.Error())
+	}
+
+	_, err = p.blobClient.UploadBuffer(ctx, config.ContainerName, config.BlobName, bundleBytes, nil)
```
rausingh-rh: we should pass appropriate `UploadBufferOptions` with `HTTPHeaders` containing the correct `content-type` based on `config.bundleFormat`. Right now `UploadBufferOptions` is passed as `nil`, which means the blob is uploaded with Azure's default content type (`application/octet-stream`).

pkg/server/plugin/bundlepublisher/azureblob/azureblob.go
```diff
@@ -0,0 +1,305 @@
...
+		blobClient, err = p.hooks.newBlobClientFunc(cred, accountURL)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to create client: %v", err)
+		}
+	}
+	p.blobClient = blobClient
+
+	p.setConfig(newConfig)
+	p.setBundle(nil)
+	return &configv1.ConfigureResponse{}, nil
```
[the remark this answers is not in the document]
amartinezfayo: This is a pre-existing pattern in the aws_s3 and gcp_cloudstorage publishers, where the client field is assigned in Configure and read in PublishBundle without a lock. It only became concurrently reachable after dynamic reconfiguration was added, since Configure previously ran once at initialization. I'll open a separate PR to fix it in the other two plugins, so addressing it here is optional and either way is fine with me.

pkg/server/plugin/bundlepublisher/azureblob/azureblob.go
```diff
@@ -0,0 +1,305 @@
...
+
+	currentBundle := p.getBundle()
+	if proto.Equal(req.Bundle, currentBundle) {
+		return &bundlepublisherv1.PublishBundleResponse{}, nil
+	}
+
+	bundleToPublish := proto.Clone(req.Bundle).(*types.Bundle)
+	if config.parsedRefreshHint != 0 {
+		bundleToPublish.RefreshHint = config.parsedRefreshHint
+	}
```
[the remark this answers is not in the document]
amartinezfayo: No change needed. The dedup compares the raw req.Bundle, so the only effect is re-uploading byte-identical content when an incoming bundle changes solely in RefreshHint while refresh_hint is configured, which is harmless and matches the existing aws_s3 and gcp_cloudstorage behavior.

pkg/server/plugin/bundlepublisher/azureblob/azureblob.go
```diff
@@ -0,0 +1,305 @@
...
+
+	formatter := bundleformat.NewFormatter(bundleToPublish)
+	bundleBytes, err := formatter.Format(config.bundleFormat)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "could not format bundle: %v", err.Error())
```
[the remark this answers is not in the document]
amartinezfayo: No change needed. This matches the identical format error in aws_s3 and gcp_cloudstorage, and %v on the resulting string is equivalent, so changing only this plugin would introduce an inconsistency for no behavioral gain.

doc/plugin_server_bundlepublisher_azure_blob.md
```diff
@@ -0,0 +1,104 @@
...
+
+| Configuration        | Description                                                                                                                                                                          | Required                                             | Default              |
+|----------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------|----------------------|
+| storage_account_name | The name of the Azure Storage account.                                                                                                                                               | Yes.                                                 |                      |
+| storage_account_key  | The Azure Storage account access key for shared key authentication.                                                                                                                  | Required only when using shared key authentication.  |                      |
```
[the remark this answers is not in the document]
amartinezfayo: This one is worth addressing. Clarifying in the description and sample that storage_account_key expects the storage account access key, and not a connection string or SAS token, would help avoid a real misconfiguration.

pkg/server/plugin/bundlepublisher/azureblob/azureblob.go
amartinezfayo: Thanks for flagging this. I agree some validation here would be good. One thing to keep in mind is that the aws_s3 check uses url.ParseRequestURI because its endpoint is a full URL, whereas service_endpoint here is a bare host suffix (default blob.core.windows.net), so that exact check would reject the valid default. It would be better to validate the host form, or validate the constructed account URL, and to do it in buildConfig so that spire-server validate catches it too.

pkg/server/plugin/bundlepublisher/azureblob/azureblob.go
amartinezfayo: Thanks for raising this. Passing nil does leave the blob stored as application/octet-stream. Worth noting that neither aws_s3 nor gcp_cloudstorage set a content type today, so I would treat this as an optional improvement rather than something needed for parity with the existing publishers. If we do want correct content types (json for spiffe and jwks, a pem type for pem), I think it would be best to apply it consistently across all three publishers in a separate change.

pkg/server/plugin/bundlepublisher/azureblob/azureblob.go
```diff
@@ -0,0 +1,303 @@
...
+
+	serviceEndpoint := newConfig.ServiceEndpoint
+	if serviceEndpoint == "" {
+		serviceEndpoint = defaultEndpoint
+	}
+	accountURL := fmt.Sprintf("https://%s.%s", newConfig.StorageAccountName, serviceEndpoint)
+
+	var blobClient blobStorage
+
+	switch {
```
amartinezfayo, line 168: It seems that the auth-field validation here only runs in Configure, so the Validate RPC that backs spire-server validate won't catch those misconfigurations. So a config that sets both storage_account_key and the client-secret fields, or only a partial service principal, would validate as OK and then fail at startup.
I think we should move the pure validation into buildConfig via status.ReportError (the way aws_s3 validates its endpoint there) and keep only the credential and client creation in Configure.

doc/plugin_server_bundlepublisher_azure_blob.md
```diff
@@ -0,0 +1,104 @@
+# Server plugin: BundlePublisher "azure_blob"
```
amartinezfayo, line 1: This new plugin needs an entry in the BundlePublisher table in doc/spire_server.mdto to keep the built-in plugin index updated.

amartinezfayo (commented):
Thank you @sabre1041 for this contribution!

pkg/server/plugin/bundlepublisher/azureblob/azureblob.go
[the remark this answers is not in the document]
amartinezfayo: Opened #7082 to fix this in the aws_s3 and gcp_cloudstorage publishers.

sabre1041:
Hi @amartinezfayo 

Thanks for the review. Based on the reviews thus far, please share any actions that I would be able to take for this contribution. Happy to address any area that may need enhancement in order to resolve any identified deficiencies.

amartinezfayo (commented):
Hi @sabre1041, thanks for your patience and for offering to keep iterating on this.

The open review threads above capture everything that needs action before merge. The key ones are moving the authentication field validation and the service_endpoint validation into buildConfig so the Validate RPC catches them, adding the azure_blob entry to the BundlePublisher table in doc/spire_server.md, and clarifying in the docs that storage_account_key expects the storage account access key rather than a connection string or SAS token. The content type and the client locking can wait for a separate cross-publisher change, so they are not blockers here.

Thanks again for the contribution.

sabre1041:
> Hi @sabre1041, thanks for your patience and for offering to keep iterating on this.
> 
> The open review threads above capture everything that needs action before merge. The key ones are moving the authentication field validation and the service_endpoint validation into buildConfig so the Validate RPC catches them, adding the azure_blob entry to the BundlePublisher table in doc/spire_server.md, and clarifying in the docs that storage_account_key expects the storage account access key rather than a connection string or SAS token. The content type and the client locking can wait for a separate cross-publisher change, so they are not blockers here.
> 
> Thanks again for the contribution.

Think I was able to account for the requested enhancements. Please take a look whenever you have some time

amartinezfayo (approved):
Thank you @sabre1041!

PR #7037, opened 2026-06-04
Remove the unused kind configuration from the cert-manager integration test

amartinezfayo:
The cert-manager integration test suite has a kind configuration whose only effect was pinning the node image to `kindest/node:v1.21.1`. That pin never took effect: `start-kind-cluster` always creates the cluster with `--image "${K8SIMAGE}"`, and kind's `--image` flag overrides the per-node image in the config, so the suite has been running on the version selected by `K8SIMAGE` (currently `v1.31.12`) all along. With the pin gone the configuration only declared a single `control-plane` node, which is kind's default.

This removes the configuration file entirely and drops the `--config` argument from the cluster creation. The suite now provisions its cluster the same way as the `k8s`, `key-manager-vault`, `upstream-authority-vault`, and `upstream-authority-ejbca` suites, which already create kind clusters without a config file.

I noticed the stale pin while adding image pull retries to `start-kind-cluster` in #7031.

### DIFF
--- a/test/integration/suites/upstream-authority-cert-manager/00-setup-kind
+++ b/test/integration/suites/upstream-authority-cert-manager/00-setup-kind
@@ -14,7 +14,7 @@ download-kind "${KIND_PATH}"
 download-kubectl "${KUBECTL_PATH}"
 
 # Start the kind cluster.
-start-kind-cluster "${KIND_PATH}" cert-manager-test ./conf/kind-config.yaml
+start-kind-cluster "${KIND_PATH}" cert-manager-test
 
 # Load the given images in the cluster.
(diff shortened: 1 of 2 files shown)


sorindumitru (dismissed):
LGTM. We might not need the `kind` config file at anymore now?

amartinezfayo:
Thanks for pointing this out. The only thing the config did was pin that stale node image, and with that gone it just declared the default single control-plane node. I removed the file entirely and dropped the --config argument, so the suite now provisions its cluster the same way as the k8s, vault, and ejbca suites that already run without a kind config.

sorindumitru (approved):

PR #7044, opened 2026-06-05
Tolerate mountinfo lines with an empty mount source

### LINKED ISSUE #7036: docker/k8s workload attestor fails on tmpfs mount with empty source
The docker workload attestor (and k8s, which shares the code) fails attestation outright when a workload process has a tmpfs mount with an empty source. No container ID is extracted, no selectors are produced, and no identity is issued. This is with the default container locator (`use_new_container_locator=true`).

The error looks like this:

```
level=error msg="Failed to collect all selectors for PID" error="workload attestor \"docker\" failed: rpc error: code = Internal desc = workloadattestor(docker): failed to parse mount info at \"/proc/943/mountinfo\": wrong number of fields in (expected at least 10, got 9): 119 206 0:68 / /local rw,relatime - tmpfs  rw,size=8192k" pid=943 subsystem_name=workload_attestor
```

The attestor parses `/proc/<pid>/mountinfo` via `mount.ParseMountInfo` from `k8s.io/mount-utils` (`pkg/common/containerinfo/extract.go:80`), which splits each line with `strings.Fields` and requires at least 10 fields. Per `proc(5)`, the mount source (field 10) can be empty, and the kernel escapes real whitespace in paths as octal (`\040`), so the double space between `tmpfs` and `rw,size=8192k` unambiguously means an empty source. `strings.Fields` collapses that empty field, so a valid line parses as 9 fields and the whole file is rejected.

SPIRE failing rather than emitting selectors from partially parsed data is intentional and correct, so the bug is not the fail-closed behavior. The problem is that this is a valid mountinfo line the upstream parser mishandles. The fix is to parse the trailing `fstype` / `source` / `super-options` triple position-aware so an empty source is retained, ideally upstream in `k8s.io/mount-utils`, or with a SPIRE-side parser otherwise.

Workaround: the legacy locator (`use_new_container_locator=false` with `container_id_cgroup_matchers`) reads `/proc/<pid>/cgroup` instead of mountinfo and sidesteps this, though that path is deprecated.

arpitjain099:
**Pull Request check list**

- [x] Commit conforms to CONTRIBUTING.md? (DCO sign-off included)
- [ ] Documentation updated? (internal parser only, no user-facing docs affected)

**Affected functionality**

The docker and k8s workload attestors, which share `pkg/common/containerinfo` to read `/proc/<pid>/mountinfo` and recover the container ID.

**Description of change**

Implements the SPIRE-side approach confirmed on #7036. The attestor parses mountinfo via `k8s.io/mount-utils` `ParseMountInfo`, which splits each line with `strings.Fields` and requires at least 10 fields. Per proc(5) the mount source may be empty (a tmpfs mount has no source), rendering as an empty field between the filesystem type and the super options. The kernel escapes any real whitespace in a path as octal (`\040`), so the double space in a line like:

    119 206 0:68 / /local rw,relatime - tmpfs  rw,size=8192k

unambiguously means an empty source. `strings.Fields` collapses that empty field, so the line counts as 9 fields and `ParseMountInfo` rejects the whole file with "wrong number of fields"; the attestor then extracts no container ID and no SVID is issued.

This replaces the upstream call with a small SPIRE-side mountinfo parser that is position-aware around the `-` separator, so the trailing fstype/source/super-options triple is read by position and an empty source is tolerated. Only the mount root and filesystem type are consumed by the extractor. Since this was the only user of `k8s.io/mount-utils`, the change also drops that dependency (and its transitive `moby/sys/mountinfo`).

A regression fixture with the empty-source tmpfs line and unit tests for the parser are included.

**Which issue this PR fixes**

fixes #7036


cbartleroblox (approved):

cbartleroblox:
Thank you!

sorindumitru:
Thanks for opening this PR @arpitjain099 

> Per proc(5) the mount source may be empty (a tmpfs mount has no source), rendering as an empty field between the filesystem type and the super options.

Could you share a source for this? My `man 5 proc_pid_mountinfo` says the following:
```
mount source: filesystem-specific information or "none".
```

Could you share some details about the system where you encountered this or a way to reproduce (e.g. can I just start a docker container with some options)?

arpitjain099:
Thanks @sorindumitru! This is specifically the tmpfs case. You're right that `proc(5)` documents the source as "filesystem-specific information or 'none'", but tmpfs mounts in practice have no source at all - Docker's own docs state ["There is no source for tmpfs mounts"](https://docs.docker.com/engine/storage/tmpfs/#use-a-tmpfs-mount-in-a-container), and the kernel emits a literal empty field rather than `none`.

The concrete line is in #7036 (filed by @amartinezfayo):
```
119 206 0:68 / /local rw,relatime - tmpfs  rw,size=8192k
```
Note the double space between `tmpfs` and `rw,size=8192k` - that is the empty source field (fstype `tmpfs`, empty source, super-options `rw,size=8192k`). Because the kernel octal-escapes real whitespace in paths as `\040`, a literal double space unambiguously means an empty source rather than a space in a name. `strings.Fields` in the upstream `k8s.io/mount-utils` parser collapses that empty field, so the line reads as 9 fields and the whole file is rejected - which is why a position-aware parse of the trailing `fstype`/source/super-options triple is needed.

To reproduce with Docker:
```
docker run --rm --tmpfs /local:rw,size=8192k alpine sh -c 'grep " tmpfs " /proc/self/mountinfo'
```
You should see the tmpfs line with the empty source. A Kubernetes pod with an `emptyDir: {medium: Memory}` volume produces the same kind of mount.

sorindumitru:
@arpitjain099 this is what I get from that:
```
735 726 0:131 / /local rw,nosuid,nodev,noexec,relatime - tmpfs tmpfs rw,size=8192k,inode64,huge=within_size
```

This is with a recent kernel version, though 7.0. Are you running this on linux, and which distro/kernel version, or some other operating system?

sorindumitru:
> @arpitjain099 this is what I get from that:
> 
> ```
> 735 726 0:131 / /local rw,nosuid,nodev,noexec,relatime - tmpfs tmpfs rw,size=8192k,inode64,huge=within_size
> ```
> 
> This is with a recent kernel version, though 7.0. Are you running this on linux, and which distro/kernel version, or some other operating system?

Was able to reproduce by calling `mount` syscall with an empty source.

arpitjain099:
Thanks for digging in @sorindumitru, and good that the `mount` syscall with an empty source reproduces it. That is the validation that matters here: an empty source is valid mountinfo the parser has to accept.

To be upfront about the environment question: I didn't hit this on my own machine. The concrete line came from @amartinezfayo's report in #7036 (a real workload, `/proc/943/mountinfo`), and my part was diagnosing the root cause - `strings.Fields` collapsing the empty source field so the line reads as 9 columns and `k8s.io/mount-utils` rejects the whole file - plus the position-aware fix. @amartinezfayo would have the exact distro/kernel for the original.

As you found, the trigger is the mount source being literally empty (a `mount` call with an empty source, as some runtimes do for in-memory mounts) rather than the kernel writing `tmpfs` or `none`. Recent kernels populating `tmpfs` for `docker --tmpfs` just means that one path doesn't hit it on 7.0; the empty-source form is still valid and still reaches this parser, which is what the fix handles. Happy to add a test fixture using the exact 9-column line from #7036 if that helps lock it in.

cbartleroblox:
Validated that this change doesn't trigger the parser issue.

pkg/common/containerinfo/mountinfo.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, line := range strings.Split(string(content), "\n") {
+		if line == "" {
+			continue
+		}
+		info, err := parseMountInfoLine(line)
+		if err != nil {
+			return nil, err
+		}
+		infos = append(infos, info)
+	}
```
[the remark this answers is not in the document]
sorindumitru: This seems useful to do and could simplify things a bit.

pkg/common/containerinfo/mountinfo.go
```diff
@@ -0,0 +1,102 @@
...
+// Fields up to the "-" separator are whitespace-separated and never empty, so
+// strings.Fields is safe there. The three fields after the separator are taken
+// by position so an empty source (field 10) is preserved rather than collapsed.
+func parseMountInfoLine(line string) (mountInfo, error) {
+	// Split the line into the part before the "-" separator and the part after
+	// it. The separator is a standalone "-" token, so it is surrounded by
+	// spaces. Splitting on " - " keeps it unambiguous: optional-field tags
+	// before the separator never equal a bare "-", and the post-separator
+	// fields are matched by position below.
+	sep := strings.Index(line, " - ")
```
sorindumitru: Looking at what `strings` offers, I think `strings.Split(line, " ")` should do what we want here and maybe it would simplify things a bit. What do you think?

arpitjain099: Good catch. Switched to strings.Split(line, " ") so empty fields survive as empty strings in the slice. This lets us locate the separator and all post-separator fields by index directly, no more separate Index + Fields split. Pushed in 4af7f2f.

pkg/common/containerinfo/mountinfo.go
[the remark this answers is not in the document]
arpitjain099: Done in the same commit (4af7f2f). Swapped ReadFile + Split for bufio.Scanner, so it streams line by line now.

pkg/common/containerinfo/mountinfo.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// Locate the "-" separator. Everything before it is the fixed +
+	// optional-tag fields; everything after is fstype, source, super options.
+	sepIdx := -1
+	for i, f := range fields {
+		if f == "-" {
+			sepIdx = i
+			break
+		}
+	}
```
sorindumitru: I think this should work here, but might need an extra import:
```suggestion
	sepIdx := slices.Index(fields, "-")
```

sorindumitru:
Hi @arpitjain099, do you think you'll have time to address that small comment I left and resolve the conflicts? If you won't I can try to push the changes over the branch, but we'd want to include this in the next release.

arpitjain099:
Thanks for the patience @sorindumitru. Rebased onto main (the conflict was just go.mod/go.sum from the mount-utils removal) and applied your suggestion: separator lookup is now `slices.Index(fields, "-")` with the `slices` import added. Also ran go mod tidy so the now-unused k8s.io/mount-utils is dropped cleanly. Build, vet, and the containerinfo package tests pass locally.

sorindumitru (approved):
LGTM, thanks @arpitjain099

PR #7060, opened 2026-06-10
Prefer determining pod and container ids from cgroup file

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Determining container and pod id.

**Description of change**
Currently we determine the information about the pod and containers first from /proc/pid/mountinfo and then from /proc/pid/cgroup. The cgroup file is the authoritative source here so we should prefer it over mountinfo. mountinfo is only required for the cases where the caller is in the same pod or container as the agent. A privileged workload can change information from mountinfo, e.g. via bind mounts, so this avoids issues where we wrongly identify a task.


pkg/common/containerinfo/extract.go
```diff
@@ -50,22 +50,28 @@ func (e *Extractor) GetPodUIDAndContainerID(pid int32, log hclog.Logger) (types.
...
 	if err != nil {
 		return "", "", err
 	}
 
 	if containerID == "" {
-		podUID, containerID, err = e.extractPodUIDAndContainerIDFromCGroups(pid, log, extractPodUID)
+		podUID, containerID, err = e.extractPodUIDAndContainerIDFromMountInfo(pid, log, extractPodUID)
 		if err != nil {
 			return "", "", err
 		}
```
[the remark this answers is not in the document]
sorindumitru: This was also true before, when preferring fetching information from `mountinfo` so I think it's safe to ignore this comment.

pkg/common/containerinfo/extract_test.go
```diff
@@ -78,6 +78,14 @@ func TestExtractPodUIDAndContainerID(t *testing.T) {
 	t.Run("falls back to cgroup file", func(t *testing.T) {
```
MarcosDY: is this really testing the falls back?  looks like it is now exercising the primary path?
and if that is the case how to test the fallback no?

MarcosDY: something like:

```
	t.Run("cgroup file has container ID (mountinfo mounts at root)", func(t *testing.T) {
		assertFound(t, "testdata/other/fallback", "", testContainerID)
	})

	t.Run("falls back to mountinfo when cgroup has no container ID", func(t *testing.T) {
		assertFound(t, "testdata/other/cgroup-no-id", testPodUID, testContainerID)
	})
```

with cgroup:
```
0::/
```

MarcosDY (commented):
looks great just a quick question about test

pkg/common/containerinfo/extract_test.go
sorindumitru: yeah, those don't really make sense that much. Changed them to match the current expectations

MarcosDY (approved):
LGTM!

PR #7085, opened 2026-06-20
agent: k8s: coordinate kubelet pod list fetching

c4rlo:
Move kubelet pod-list fetching, caching and backoff timing into a long-lived fetcher that versions pod-list responses, instead of using the `singleflight` package.

This now enables us to cache successful kubelet responses until another request is eligible to start, which makes more sense. Previously, they were cached for _half_ the polling interval, which could lead to the following situation (illustrated with the default polling interval of 500 ms):

- T = 0 ms: Attestation request 1 makes pod-list request, finds no result, sleeps for 500 ms.
- T = 300 ms: Attestation request 2 finds cache expired, makes pod-list request that contains an entry that request 1 was looking for.
- T = 500 ms: Only now does request 1 wake up and find the desired entry in the cached pod list.

Fixing this with the `singleflight` approach would not have been easily possible, as it could cause retries to re-use the same cached pod list response as in the previous iteration.

The new approach also enforces a minimum interval between kubelet request _start times_ across concurrent attestations. Unlike the previous fixed delay after each request, this allows the next request to start immediately when the previous one took longer than the configured interval.

In addition, several other improvements (that could also have been implemented with the old `singleflight` mechanism):

- kubelet requests: do not cancel an inflight request just because its _original_ caller is no longer interested, as other callers may be waiting for it. (With `singleflight`, this could have been fixed via the `Group.DoChan` method.)
- kubelet requests: retry failed fetches via the same backoff mechanism, rather than immediately returning a failure response. Previously, it was only used when the pod list did not contain the workload.
- kubelet requests: when reloading the client, whether through an explicit `Configure()` call or periodically via `reload_interval`, only create a fresh `http.Transport` object if necessary based on config or certificate changes, ensuring we do not trash its connection pool unnecessarily.
- kubelet requests: treat an invalid `items` JSON type in the response as an error instead of empty.
- kubelet requests: replace the `attempt` log value rather than appending a new one with each attempt.
- kubelet requests: report kubelet and malformed-response failures as `Unavailable` instead of `Internal`. (Failure to find the workload in the pod list after attempts are exhausted continues to be `DeadlineExceeded`.)
- kubelet requests: use `net.JoinHostPort` for endpoints, enabling IPv6 support.
- kube API server requests: correctly reflect runtime changes to the `api_server.cache.enabled` setting via `Configure()`.
- `Validate()`: perform more thorough validation, including checking that kubelet certificates can be read and parsed.
- gRPC return codes: preserve whether caller termination was cancellation or deadline expiry.
- gRPC return codes: report failed Sigstore policy verification as `PermissionDenied` instead of `Internal`.
- tests: fix a flaky, order-dependent Sigstore selector test failure caused by mutating shared expected selectors.
- misc: small code cleanups.
- docs: add missing descriptions of `max_poll_attempts`, `poll_retry_interval`, and `reload_interval`.
- docs: assorted fixes.

---

Open questions:

1. I appreciate that this is quite a lot of new code and complexity. Is it worth the gain, maybe counting only the improvements we could not get with the existing `singleflight` approach? I think so, as it is a more principled approach with more logical semantics, but it is probably debatable.
2. If we go with this new approach, do we want to change the default value for the `poll_retry_interval` config setting? Currently it is 500 ms. But its meaning changes with this PR: it used to be the sleep time between retry attempts, with the cache TTL being set to half that value (250 ms); now it is the minimum interval between fetch start times for both new requests and retries, with the cache TTL also being the same. I would be tempted to reduce the default value now, perhaps to 250 ms, but I do not have a strong opinion either.

### DIFF
--- a/doc/plugin_agent_workloadattestor_k8s.md
+++ b/doc/plugin_agent_workloadattestor_k8s.md
@@ -48,25 +48,28 @@ need PID-based workload attestation or agent-node-scoped pod references.
 **Note** To run on Windows containers, Kubernetes v1.24+ and containerd v1.6+ are required,
 since [hostprocess](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/) container is required on the agent container.
 
-| Configuration                           | Description                                                                                                                                                                                                                             |
-|-----------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `disable_container_selectors`           | If true, container selectors are not produced. This can be used to produce pod selectors when the workload pod is known but the workload container is not ready at the time of attestation.                                             |
-| `disable_kubelet_client`                | If true, disables kubelet client setup and kubelet pod-list calls. PID-based workload attestation and `pod_reference_scope = "agent_node"` pod references are unavailable. Mutually exclusive with kubelet client settings.             |
-| `enable_namespace_labels`               | If true, namespace labels are fetched from the Kubernetes API server and produced as selectors. Requires RBAC `get`, `list` and `watch` on `namespaces`. Disabled by default.                                                           |
-| `kubelet_read_only_port`                | The kubelet read-only port. This is mutually exclusive with `kubelet_secure_port`.                                                                                                                                                      |
-| `kubelet_secure_port`                   | The kubelet secure port. It defaults to `10250` unless `kubelet_read_only_port` is set.                                                                                                                                                 |
-| `kubelet_ca_path`                       | The path on disk to a file containing CA certificates used to verify the kubelet certificate. Required unless `skip_kubelet_verification` is set. Defaults to the cluster CA bundle `/run/secrets/kubernetes.io/serviceaccount/ca.crt`. |
-| `skip_kubelet_verification`             | If true, kubelet certificate verification is skipped                                                                                                                                                                                    |
-| `token_path`                            | The path on disk to the bearer token used for kubelet authentication. Defaults to the service account token `/run/secrets/kubernetes.io/serviceaccount/token`                                                                           |
-| `certificate_path`                      | The path on disk to client certificate used for kubelet authentication                                                                                                                                                                  |
-| `private_key_path`                      | The path on disk to client key used for kubelet authentication                                                                                                                                                                          |
-| `use_anonymous_authentication`          | If true, use anonymous authentication for kubelet communication                                                                                                                                                                         |
-| `node_name_env`                         | The environment variable used to obtain the node name. Defaults to `MY_NODE_NAME`.                                                                                                                                                      |
-| `node_name`                             | The name of the node. Overrides the value obtained by the environment variable specified by `node_name_env`.                                                                                                                            |
-| `experimental`                          | The experimental options that are subject to change or removal (see below).                                                                                                                                                             |
-| `sigstore`                              | Sigstore options. Options described below. See [Sigstore options](#sigstore-options). When set, enables verification of container image signatures and attestations.                                                                    |
-| `use_new_container_locator`             | If true, enables the new container locator algorithm that has support for cgroups v2. Defaults to true.                                                                                                                                 |
-| `verbose_container_locator_logs`        | If true, enables verbose logging of mountinfo and cgroup information used to locate containers. Defaults to false.                                                                                                                      |
+| Configuration                           | Description                                                                                                                                                                                                                                 |
+|-----------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `disable_container_selectors`           | If true, container selectors are not produced. This can be used to produce pod selectors when the workload pod is known but the workload container is not ready at the time of attestation.                                                 |
+| `disable_kubelet_client`                | If true, disables kubelet client setup and kubelet pod-list calls. PID-based workload attestation and `pod_reference_scope = "agent_node"` pod references are unavailable. Mutually exclusive with kubelet client settings.                 |
+| `enable_namespace_labels`               | If true, namespace labels are fetched from the Kubernetes API server and produced as selectors. Requires RBAC `get`, `list` and `watch` on `namespaces`. Disabled by default.                                                               |
+| `kubelet_read_only_port`                | The kubelet read-only port. This is mutually exclusive with `kubelet_secure_port`.                                                                                                                                                          |
+| `kubelet_secure_port`                   | The kubelet secure port. It defaults to `10250` unless `kubelet_read_only_port` is set.                                                                                                                                                     |
+| `kubelet_ca_path`                       | The path on disk to a file containing CA certificates used to verify the kubelet certificate. Required unless `skip_kubelet_verification` is set. Defaults to the cluster CA bundle `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`. |
+| `skip_kubelet_verification`             | If true, kubelet certificate verification is skipped.                                                                                                                                                                                       |
+| `token_path`                            | The path on disk to the bearer token used for kubelet authentication. Defaults to the service account token `/var/run/secrets/kubernetes.io/serviceaccount/token`.                                                                          |
+| `certificate_path`                      | The path on disk to client certificate used for kubelet authentication.                                                                                                                                                                     |
+| `private_key_path`                      | The path on disk to client key used for kubelet authentication.                                                                                                                                                                             |
+| `use_anonymous_authentication`          | If true, use anonymous authentication for kubelet communication.                                                                                                                                                                            |
+| `node_name_env`                         | The environment variable used to obtain the node name for kubelet requests. Defaults to `MY_NODE_NAME`.                                                                                                                                     |
+| `node_name`                             | The node name used for kubelet requests. Overrides the value obtained from the env var specified by `node_name_env`.                                                                                                                        |
+| `max_poll_attempts`                     | The maximum number of kubelet polling attempts while locating the workload container/pod. Defaults to `60`.                                                                                                                                 |
+| `poll_retry_interval`                   | The interval between kubelet polling attempts while locating the workload container/pod. Defaults to `500ms`.                                                                                                                               |
+| `reload_interval`                       | How often kubelet TLS and token configuration is reloaded from disk. Defaults to `1m`.                                                                                                                                                      |
+| `experimental`                          | The experimental options that are subject to change or removal (see below).                                                                                                                                                                 |
+| `sigstore`                              | Sigstore options. Options described below. See [Sigstore options](#sigstore-options). When set, enables verification of container image signatures and attestations.                                                                        |
+| `use_new_container_locator`             | If true, enables the new container locator algorithm that has support for cgroups v2. Defaults to true.                                                                                                                                     |
+| `verbose_container_locator_logs`        | If true, enables verbose logging of mountinfo and cgroup information used to locate containers. Defaults to false.                                                                                                                          |
 
 These are the current experimental configurations.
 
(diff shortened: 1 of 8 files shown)


pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -313,14 +309,23 @@ func (p *Plugin) Attest(ctx context.Context, req *workloadattestorv1.AttestReque
...
+		podListResult := p.podListFetcher.fetchNext(ctx, podListVersion)
+		podList, podListErr := podListResult.pods, podListResult.err
+		if podListErr != nil {
+			if ctx.Err() != nil {
+				return nil, ctx.Err()
+			}
+			// Otherwise, we'll log podListErr below, and we may retry.
+		} else {
+			podListVersion = podListResult.version
 		}
```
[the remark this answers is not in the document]
c4rlo: Should be unlikely to happen, but I've addressed it now.

pkg/agent/plugin/workloadattestor/k8s/podlistfetcher.go
```diff
@@ -0,0 +1,513 @@
...
+			config:   config,
+			client:   client,
+			resultCh: request.resultCh,
+		}
+		sendOrStop(f, f.configResultCh, result)
+	})
+}
+
+func (f *podListFetcher) handleConfigResult(result podListConfigResult) {
+	f.config = new(result.config)
```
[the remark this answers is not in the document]
c4rlo: See #7086.

matheuscscp:
Please hold this PR. The Broker API PR is changing many things in the k8s plugin, this PR should come on top of it.

matheuscscp:
Hey @c4rlo, the Broker API is merged, pls rebase :pray:

c4rlo:
Rebased, and included a few more small improvements; commit message & PR description updated accordingly.

pkg/agent/plugin/workloadattestor/k8s/podlistfetcher.go
```diff
@@ -0,0 +1,457 @@
...
+func (f *podListFetcher) registerFetchRequest(afterVersion uint64, resultCh chan<- podListFetchResult) {
+	if f.cachedPodList.pods != nil && afterVersion < f.cachedPodList.version &&
+		f.clock.Now().Before(f.cachedFetchStart.Add(f.config.pollRetryInterval)) {
+		resultCh <- podListFetchResult{versionedPodList: f.cachedPodList}
+		return
+	}
+
+	f.fetchWaiters[resultCh] = struct{}{}
+	f.scheduleFetch()
+}
```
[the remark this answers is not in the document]
c4rlo: Shouldn't happen since `Configure()` gets called before any `Attest()` calls afaik, but still good to defend against this, hence addressed in cbf43854d.

pkg/agent/plugin/workloadattestor/k8s/podlistfetcher.go
```diff
@@ -0,0 +1,457 @@
...
+	result := make(map[string]*fastjson.Value, len(items))
+
+	for _, podValue := range items {
+		uid := string(podValue.Get("metadata", "uid").GetStringBytes())
+		if uid == "" {
+			f.log.Warn("Pod has no UID", "pod", podValue)
+			continue
+		}
+		result[uid] = podValue
+	}
```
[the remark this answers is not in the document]
c4rlo: This can't happen because `fastjson.Value.Get()` and `fastjson.Value.GetStringBytes()` both have a check for whether they're being called on a `nil` value:

https://github.com/valyala/fastjson/blob/d652a1b1909d3520389b2c287ca3cf3aa3791451/parser.go#L725-L728

https://github.com/valyala/fastjson/blob/d652a1b1909d3520389b2c287ca3cf3aa3791451/parser.go#L860-L864

(Also, this isn't new code, it's just been moved around.)

matheuscscp:
Hey @c4rlo #7094 merged and introduced some conflicts, sorry about that, can you pls rebase again? :pray: 

@sorindumitru We should try to review this one soon so @c4rlo doesn't have to keep solving conflicts :pray:

c4rlo:
Rebased again.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -833,25 +825,25 @@ func (p *Plugin) getBrokerEntryIfPresent(ctx context.Context, config *k8sConfig)
 // can match pod-specific fields like container images and service accounts
 // that aren't present on a PartialObjectMetadata.
 func (p *Plugin) attestByPodReference(ctx context.Context, brokerEntry *k8sBrokerEntry, objRef *broker.KubernetesObjectReference) (*attestReferenceResult, error) {
+	config, _, _, err := p.getConfig()
```
MarcosDY: get config can return `return nil, nil, nil, status.Error(codes.FailedPrecondition, "not configured")` thi s internal can cause backcompat since we are expecting another error in that case
can you revert return the returned error instead? (plz verify if there are more cases like this)

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -509,20 +486,22 @@ type Plugin struct {
...
 
 func New() *Plugin {
-	return &Plugin{
-		clock:  clock.New(),
+	pluginClock := clock.New()
+	p := &Plugin{
+		clock:  pluginClock,
 		getenv: os.Getenv,
 	}
+	p.podListFetcher = newPodListFetcher(pluginClock, p.rootDir)
```
MarcosDY, line 543: it is not a bug "now" but we are creating fetcher on new, no in configure, 
in a possible scenario we allow rootDIr to be configurable, fetcher is not going to get the updated value

c4rlo:
Rebased again.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -1460,7 +1453,13 @@ func (p *Plugin) Configure(ctx context.Context, req *configv1.ConfigureRequest)
 }
 
 func (p *Plugin) Validate(_ context.Context, req *configv1.ValidateRequest) (resp *configv1.ValidateResponse, err error) {
-	_, notes, err := pluginconf.Build(req, p.buildConfig)
+	newConfig, notes, err := pluginconf.Build(req, p.buildConfig)
+	if err == nil && !newConfig.DisableKubeletClient {
+		err = p.podListFetcher.validate(newConfig.podListFetcherConfig)
```
MarcosDY: It looks to me like validate no longer only validates but can performs disk I/O and can mutate fetcher state.
am i right?
so 2 topics here:
- Validate now reads CA/token/cert files from disk. For spire-agent validate run outside a pod, `/var/run/secrets/kubernetes.io/serviceaccount/token` does not exist, so a config that is perfectly valid for the runtime environment will report Valid: `false`. This is a behavior change for the CLI validate path that isn't mentioned in the PR description. Is that intended?
- Validate reads f.buildClient — a field owned by the run goroutine per the struct comment. In production nothing else writes it, but it breaks the stated ownership invariant, and the tests do reassign fetcher.buildClient from the test goroutine. May we make validate a free function over podListFetcherConfig + rootDir so it never touches fetcher state.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -1414,6 +1401,7 @@ func buildKubeClientOptions() (*rest.Config, client.Options, error) {
 }
 
 func (p *Plugin) Close() error {
+	p.podListFetcher.close()
```
MarcosDY, line 1407: this may panic if we call twice, may we run int inside a sync.Once?

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -692,23 +668,32 @@ func (p *Plugin) attestByPIDReference(ctx context.Context, pid int32) (*attestRe
...
+		if podListErr != nil {
+			if ctx.Err() != nil {
+				return nil, ctx.Err()
+			}
+			if errors.Is(podListErr, errPodListFetcherClosed) {
+				return nil, status.Error(codes.Unavailable, podListErr.Error())
+			}
+			// Otherwise, we'll log podListErr below, and we may retry.
+		} else {
+			podListVersion = podList.version
```
MarcosDY, line 690: is there a guaranty of that podList is not nil after an error? 
when error podList is the zero versionedPodList, so podList.pods is nil and ranging is a no-op. 
So it depends on fetchNext always returning the zero value on error. 
Worth an explicit continue-style guard or at least a comment, since a future change returning a partial list would silently attest against stale data.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -786,18 +774,22 @@ func (p *Plugin) attestByPIDReference(ctx context.Context, pid int32) (*attestRe
 		}
 
 		// if the container was not located after the maximum number of attempts then the search is over.
-		if attempt >= config.MaxPollAttempts {
+		switch {
+		case attempt >= config.MaxPollAttempts:
+			if podListErr != nil {
+				log.Warn("Unable to get pod list; giving up", telemetry.Error, podListErr)
+				return nil, status.Error(codes.Unavailable, podListErr.Error())
```
MarcosDY, line 782: is this catching context cancelation? (we used to return no selector found in that case)

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -743,9 +728,12 @@ func (p *Plugin) attestByPIDReference(ctx context.Context, pid int32) (*attestRe
...
 				if sigstoreVerifier != nil {
 					log.Debug("Attempting to verify sigstore image signature", "image", containerStatus.Image)
-					sigstoreSelectors, err := p.sigstoreVerifier.Verify(ctx, containerStatus.ImageID)
+					sigstoreSelectors, err := sigstoreVerifier.Verify(ctx, containerStatus.ImageID)
 					if err != nil {
-						return nil, status.Errorf(codes.Internal, "error verifying sigstore image signature for imageID %s: %v", containerStatus.ImageID, err)
+						if ctx.Err() != nil {
+							return nil, ctx.Err()
+						}
+						return nil, status.Errorf(codes.PermissionDenied, "error verifying sigstore image signature for imageID %s: %v", containerStatus.ImageID, err)
```
MarcosDY: QQ: why to update the error code? can the change cause a regresion?

doc/plugin_agent_workloadattestor_k8s.md
```diff
@@ -48,25 +48,28 @@ need PID-based workload attestation or agent-node-scoped pod references.
...
+| `use_anonymous_authentication`          | If true, use anonymous authentication for kubelet communication.                                                                                                                                                                            |
+| `node_name_env`                         | The environment variable used to obtain the node name for kubelet requests. Defaults to `MY_NODE_NAME`.                                                                                                                                     |
+| `node_name`                             | The node name used for kubelet requests. Overrides the value obtained from the env var specified by `node_name_env`.                                                                                                                        |
+| `max_poll_attempts`                     | The maximum number of kubelet polling attempts while locating the workload container/pod. Defaults to `60`.                                                                                                                                 |
```
MarcosDY, line 66: Previously max_poll_attempts counted "pod not yet found" polls, and a kubelet error returned immediately. Now kubelet errors consume attempts too.
But the interaction is worth flagging: with the default 60 × 500ms, a kubelet that's down for 30s will now hold an attestation request open for the full 30s instead of failing fast. 
That's a behavior change for the "kubelet is genuinely broken" case, and it multiplies across concurrent workloads. 
I think we need to mention that behavior or try to catch kubelet unrecoverable errors)

MarcosDY (commented):
Open questions:

> 1. I appreciate that this is quite a lot of new code and complexity. Is it worth the gain, maybe counting only the improvements we could not get with the existing singleflight approach? I think so, as it is a more principled approach with more logical semantics, but it is probably debatable.

The versioned-pod-list core is worth it; it fixes a real bug that singleflight structurally cannot. However, the PR bundles ~10 unrelated improvements (IPv6 JoinHostPort, Validate hardening, gRPC code changes, doc fixes, strconv.Itoa → %d, isValidCGroupPathMatches simplification, the flaky-test fix). 
That increment the complexity of reviewing and things we need to verify. 
However good work :D 

> 2. If we go with this new approach, do we want to change the default value for the poll_retry_interval config setting? Currently it is 500 ms. But its meaning changes with this PR: it used to be the sleep time between retry attempts, with the cache TTL being set to half that value (250 ms); now it is the minimum interval between fetch start times for both new requests and retries, with the cache TTL also being the same. I would be tempted to reduce the default value now, perhaps to 250 ms, but I do not have a strong opinion either.

Yeap, semantic changed, so we need to change the default, 500ms is going to go against your changes. 
250ms preserves the current cache freshness, another option is to decouple them: a separate cache TTL from the rate limit would be clearer than overloading one knob, but that's more config surface...

pkg/agent/plugin/workloadattestor/k8s/k8s.go
c4rlo: Apologies, this was an unintentional change. Undone.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
c4rlo: I can't really see a reason why we'd ever want to change `rootDir` at runtime. Hence I'd prefer to keep this unchanged for simplicity.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
c4rlo: To your first point: I had assumed that `Plugin.Validate()` is something that would only be called on a real running agent, i.e. where it would be expected to have access to all its key material. However, on reflection one could imagine it also getting called in other scenarios. It looks like `spire-agent validate` does not actually call it, however `spire-server validate` does and so in future it seems quite possible `spire-agent validate` may also. And in that context one would probably expect validation to not require the key material to be in the configured locations.

Moreover, it looks like other plugins also don't do this kind of thing.

I've therefore changed it to not attempt to load any key material and only validate the config in the more usual less-thorough way.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
c4rlo: Done as defence-in-depth.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
c4rlo: I've expanded the comment to hopefully clarify it. I wanted to avoid making the loop conditional because that would add another level of indentation there, or cause other code complications.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
c4rlo: Yes: see the error handling just after `fetchNext` above.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
c4rlo: I though it would be a little more correct, but it's probably debatable since the error could signify various different things. I've therefore undone this change to be conservative.

doc/plugin_agent_workloadattestor_k8s.md
c4rlo: Yes, that is an intentional change; I feel it makes more sense to keep trying _both_ if the pod isn't in the list, and if we couldn't talk to the kubelet.

In principle it would be nice if we could tell whether a kubelet error was recoverable or not. But I don't think we can:
- Connection failure? Maybe temporary network glitch.
- Authentication failure? Maybe our token expired but it's about to be renewed (via the `ReloadInterval` timer of due to a `Configure()` call).
- Internal Server Error? Maybe someone broke Kubernetes but is already close to unbreaking it.

So given we have the retry loop, doesn't it make sense to use it for any such errors, for as long as the client's context isn't cancelled?

I'll admit this is a noteworthy change in behaviour and should perhaps be called out in the release notes...

c4rlo:
> > 2. If we go with this new approach, do we want to change the default value for the poll_retry_interval config setting? Currently it is 500 ms. But its meaning changes with this PR: it used to be the sleep time between retry attempts, with the cache TTL being set to half that value (250 ms); now it is the minimum interval between fetch start times for both new requests and retries, with the cache TTL also being the same. I would be tempted to reduce the default value now, perhaps to 250 ms, but I do not have a strong opinion either.
> 
> Yeap, semantic changed, so we need to change the default, 500ms is going to go against your changes. 250ms preserves the current cache freshness, another option is to decouple them: a separate cache TTL from the rate limit would be clearer than overloading one knob, but that's more config surface...

I don't think the behaviour has changed _that_ much that we completely need new config settings for this.

On reflection, perhaps it would be best to leave the default poll interval at 500 ms in order to minimise user-observable changes:

- Keeping 500 ms means:
    - **(Unchanged-ish)** Requests keep waiting 500 ms between retries when pod isn't in the list (still a change here in that the 500 ms is now a minimum duration between request start times, which is less pessimistic). They will make up to 60 attempts, for a total of at least 30 sec.
    - **(Changed)** But requests may now be served from a cache that is allowed to be up to 500 ms stale, rather than 250 ms (though again the 500 ms is now measured from the request start time).
- Changing to 250 ms would mean:
    - **(Changed)**: Requests now wait for only 250 ms between retries when pod isn't in the list. Total attempt time is at least 15 sec.
    - **(Unchanged)** Requests are served from a cache that is allowed to be up to 250 ms stale.

I feel like the first option is the more minimal one and hence preferable.

If we want to tune behaviour further, that can wait for another PR.

c4rlo:
Btw, I've addressed all the comments via additional commits, per contribution guidelines. The original commit is unchanged except for a (conflict-free) rebase against main.

MarcosDY (approved):

PR #7093, opened 2026-06-26
fix: make rpc_timeout and max_bundle_workers configurable on the agent

### LINKED ISSUE #6490: SPIRE agent can exceed the RPC timeout when synchronising bundles with many federations
SPIRE agent periodically syncs entries, federated bundles and SVIDs from the SPIRE server. This happens every 5 seconds, with a backoff applied on failure. There is a 30 second timeout for the update process (in `pkg/agent/client/client.go`):

```go
const rpcTimeout = 30 * time.Second
```

If there are many federations for the entries allowed for the agent, this timeout can be exceeded. If the timeout is exceeded, then the update fails and no entries, bundles or SVIDs are updated. If this happens persistently then SVIDs may expire. This is particularly apparent in resource constrained environments.

Part of the problem here is that federated bundles are fetched sequentially (in `fetchBundles()`).

salimeid:
Both values were package-level constants with no way to tune them at runtime. Expose them as experimental HCL options following the same pattern as jwt_svid_cache_hit_timeout.

**Affected functionality**
Fetching of federated trust bundles.

**Description of change**
Makes spire-agent gRPC Timeout and federated bundle fetch's concurrency configurable.

**Which issue this PR fixes**
Fixes: #6490

### DIFF
--- a/pkg/agent/client/client.go
+++ b/pkg/agent/client/client.go
@@ -29,14 +29,15 @@ import (
 )
 
 const (
-	rpcTimeout = 30 * time.Second
-
-	// maxBundleWorkers is the maximum number of worker goroutines to use when fetching bundles.
+	rpcTimeout       = 30 * time.Second
 	maxBundleWorkers = 10
 )
 
 var (
 	ErrUnableToGetStream = errors.New("unable to get a stream")
+
+	RPCTimeout       = rpcTimeout
+	MaxBundleWorkers = maxBundleWorkers
 
 	entryOutputMask = &types.EntryMask{
 		SpiffeId:             true,
@@ -56,6 +57,14 @@ var (
 	// in the Experimental Config of the Agent, and can be set as low as 5 seconds
 	RPCTimeoutWithCacheHit = rpcTimeout
 )
+
+func SetRPCTimeout(d time.Duration) {
+	RPCTimeout = d
+}
+
+func SetMaxBundleWorkers(n int) {
+	MaxBundleWorkers = n
+}
 
 func SetJWTSVIDCacheHitTimeout(d time.Duration) {
 	RPCTimeoutWithCacheHit = d
@@ -146,7 +155,7 @@ func (c *client) FetchUpdates(ctx context.Context) (*Update, error) {
 	c.c.RotMtx.RLock()
 	defer c.c.RotMtx.RUnlock()
 
-	ctx, cancel := context.WithTimeout(ctx, rpcTimeout)
+	ctx, cancel := context.WithTimeout(ctx, RPCTimeout)
 	defer cancel()
 
 	protoEntries, err := c.fetchEntries(ctx)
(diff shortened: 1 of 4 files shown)


pkg/agent/client/client.go
```diff
@@ -57,6 +58,14 @@ var (
 	RPCTimeoutWithCacheHit = rpcTimeout
 )
 
+func SetRPCTimeout(d time.Duration) {
+	RPCTimeout = d
+}
```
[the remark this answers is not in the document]
salimeid: Leaving this comment unresolved, reviewers can decide of which approach to take here

pkg/agent/client/client.go
```diff
@@ -29,15 +29,16 @@ import (
 )
 
 const (
-	rpcTimeout = 30 * time.Second
-
-	// maxBundleWorkers is the maximum number of worker goroutines to use when fetching bundles.
```
sorindumitru, line 36: We should keep this comment.

pkg/agent/client/client.go
```diff
@@ -29,15 +29,16 @@ import (
...
+	rpcTimeout       = 30 * time.Second
 	maxBundleWorkers = 10
```
sorindumitru: Could you rename these to `defaultX` to differentiate them from the other ones

pkg/agent/client/client.go
```diff
@@ -29,15 +29,16 @@ import (
...
 )
 
 var (
 	ErrUnableToGetStream = errors.New("unable to get a stream")
 
+	RPCTimeout       = rpcTimeout
+	MaxBundleWorkers = maxBundleWorkers
```
sorindumitru: These do not need to be exported, maybe after renaming the consts, these can become `rpcTimeout` and `maxBundleWorkers`.

sorindumitru:
Hi @salimeid, Checking in to see if you're still interested in the changes in this PR. If so I've left some comments that need addressing.

pkg/agent/client/client.go
salimeid: restored the comment in the `var` block

pkg/agent/client/client.go
salimeid: done
`defaultRPCTimeout` and `defaultMaxBundleWorkers`

pkg/agent/client/client.go
salimeid: done

pkg/agent/client/client.go
```diff
@@ -368,6 +379,9 @@ func (c *client) NewX509SVIDs(ctx context.Context, csrs map[string][]byte) (map[
 }
 
 func (c *client) NewJWTSVID(ctx context.Context, entryID string, audience []string, hasCacheHit bool) (*JWTSVID, spiffeid.ID, error) {
+	c.c.RotMtx.RLock()
+	defer c.c.RotMtx.RUnlock()
```
sorindumitru: We do this in `fetchJWTSVID` so there's no need to do it here. Maybe this is from a merge gone wrong?

salimeid: Rebase gone wrong indeed. Removed it, sorry bout that!

sorindumitru (approved):
LGTM, thanks @salimeid

PR #7094, opened 2026-06-28
feat: add namespace labels to k8s workload attestor

### LINKED ISSUE #7010: Add namespace labels to k8s attestor
For our organization, it would be useful to determine SPIFFE IDs based on (curated) namespace labels and pod labels. At the moment both pod labels and node labels are available to the k8s attestor but this isn't the case for namespace labels.

Would it be possible to add namespace labels to the attestor as well? Or isn't this inline with SPIFFE / SPIRE philosophy?

rausingh-rh:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
k8s workload attestor selector generation

**Description of change**
Adds opt-in `enable_namespace_labels` config to the k8s workload attestor. When enabled, the plugin fetches namespace labels via the existing controller-runtime kube client and produces `ns-label:<key>:<value>` selectors.
Labels are cached for 5 minutes with expired-entry pruning. If the fetch fails, attestation proceeds without ns-label selectors. Requires RBAC `get` on `namespaces` for the agent service account.

**Which issue this PR fixes**
fixes #7010


matheuscscp:
Hi @rausingh-rh, please rebase, the k8s workloadattestor has just changed a lot :pray:

matheuscscp:
I see two big problems in this PR:

1. It's adding support only for `WorkloadPIDReference`. That's no longer okay after introducing the Broker API, which supports also the `KubernetesObjectReference`. Adding namespace labels as selectors is valuable for any namespaced Kubernetes resource type, not just Pods. This plugin is no longer just about Pods. Please fix this :pray: 
2. This custom caching pattern is exactly what Kubernetes users don't expect to see. Five minutes delay for observing a change in namespace labels that can result in a workload / broker no longer having the right to get an SVID is not okay security-wise. One of Kubernetes' major features is the ability to report changes to objects in real-time via watches, and that's why we have the cache-backed controller-runtime client that you can acquire with `getOrCreateKubeMetadataClient()`. Please get rid of this custom cache and use the `kubeMetadataClient` :pray:

sorindumitru:
> 2\. Please get rid of this custom cache and use the `kubeMetadataClient` 🙏

I agree with this, that seems to be the best way to maintain this cache.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -677,6 +685,15 @@ func (p *Plugin) attestByPIDReference(ctx context.Context, pid int32) (*attestRe
 					selectorValues = append(selectorValues, getSelectorValuesFromWorkloadContainerStatus(containerStatus)...)
 				}
 
+				if config.EnableNamespaceLabels {
+					nsLabels, err := p.getNamespaceLabels(ctx, pod.Namespace)
+					if err != nil {
+						log.Warn("Unable to get namespace labels", "namespace", pod.Namespace, "err", err)
```
matheuscscp: I'm not sure it's okay to fail softly here. Allowing the attestation to succeed in this case when namespace labels are required feels wrong to me. Kubernetes users will enable this feature specifically due to security labels in namespaces.

rausingh-rh: Good point, agreed. when `enable_namespace_labels` is set to true, since it's an explicit opt-in, so silently dropping those selectors on failure could be unsafe. Updated to hard-fail attestation on namespace label fetch errors and updated tests accordingly.

conf/agent/agent_full.conf
```diff
@@ -406,6 +406,11 @@ plugins {
             # Defaults to false. (Linux only)
             # verbose_container_locator_logs = false
 
+            # enable_namespace_labels: If true, namespace labels are fetched
+            # from the Kubernetes API server and produced as selectors.
+            # Requires RBAC permissions to get namespaces. Defaults to false.
```
matheuscscp:
```suggestion
            # Requires RBAC permissions to get, list and watch namespaces. Defaults to false.
```

doc/plugin_agent_workloadattestor_k8s.md
```diff
@@ -49,6 +49,7 @@ since [hostprocess](https://kubernetes.io/docs/tasks/configure-pod-container/cre
 | Configuration                           | Description                                                                                                                                                                                                                             |
 |-----------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `disable_container_selectors`           | If true, container selectors are not produced. This can be used to produce pod selectors when the workload pod is known but the workload container is not ready at the time of attestation.                                             |
+| `enable_namespace_labels`               | If true, namespace labels are fetched from the Kubernetes API server and produced as selectors. Requires RBAC `get` on `namespaces`. Disabled by default.                                                                               |
```
matheuscscp:
```suggestion
| `enable_namespace_labels`               | If true, namespace labels are fetched from the Kubernetes API server and produced as selectors. Requires RBAC `get`, `list` and `watch` on `namespaces`. Disabled by default.                                                                               |
```

sorindumitru (approved):
LGTM, thanks @rausingh-rh!

matheuscscp:
Thanks @rausingh-rh! LGTM! :rocket: (a bit late :sweat_smile:)

rausingh-rh:
> Thanks @rausingh-rh! LGTM! 🚀 (a bit late 😅)

thanks for the valuable feedback on the PR @matheuscscp :saluting_face:

PR #7108, opened 2026-07-01
Fix flaky fetch-jwt-svids integration test by waiting for the restarted server

amartinezfayo:
The `08-test-disabled-jwt` step of the `fetch-jwt-svids` integration suite restarts the SPIRE server with `disable_jwt_svids = true` and then checks that JWT operations fail with a "JWT functionality is disabled" error. It intermittently fails on the final check, where the agent fetches a JWT-SVID, with a transport error instead:

```
JWT fetch expected JWT disabled error, but got: rpc error: code = Unavailable desc = could not fetch JWT-SVID: failed to fetch JWT SVID: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 172.18.0.2:8081: connect: connection refused"
```

The root cause is that the readiness check after the restart is not restart-safe. `check-server-started` counts occurrences of the `Starting Server APIs` log line and returns once it sees at least two (the TCP and UDS listeners). A `docker compose restart` reuses the same container, so its log still contains the two markers from the original boot in step `01`. The count is already satisfied the instant the check runs, so it returns immediately without waiting for the restarted server to actually be listening again.

The six checks that run `spire-server` commands inside the server container reach the server over its local UDS admin socket, which comes up quickly, so they pass. The final check runs `spire-agent api fetch jwt`, which makes the agent reach the server over its TCP API on port `8081`. During the window where the restarted server (or the agent's connection to it) is not ready yet, that fetch gets a transient `connection refused` instead of the expected disabled error, and the step fails.

This PR makes the step wait for the server that was actually restarted and tolerate the brief post-restart window:

- Adds `count-server-api-starts` and `wait-for-server-restarted` helpers to `test/integration/common`. The step now records the marker count before the restart and waits for it to grow by two, so it waits for the restarted server's own listeners rather than returning on the stale markers.
- Wraps each JWT-disabled check in a retry loop that rides out transient transport errors. A command that succeeds still fails the step immediately (JWT must be disabled), a command that returns the expected disabled error passes, and any other error is retried until the server settles.

An example of a failure is here: https://github.com/spiffe/spire/actions/runs/28516261957/job/84530600505?pr=7105

### DIFF
--- a/test/integration/common
+++ b/test/integration/common
@@ -144,6 +144,37 @@ check-server-started() {
   fail-now "timed out waiting for server to start"
 }
 
+# Counts how many times the server has logged that it started its APIs. The
+# server emits the "Starting Server APIs" line twice per boot (TCP and UDS
+# listeners).
+count-server-api-starts() {
+  # grep -c exits non-zero when there are no matches, which would trip the
+  # step's errexit/pipefail, so swallow that into a zero count.
+  docker compose logs "$1" | grep -c "Starting Server APIs" || true
+}
+
+# Like check-server-started but safe to use after "docker compose restart".
+# A restart reuses the same container, so its log still contains the
+# "Starting Server APIs" markers from the pre-restart boot. Callers capture the
+# marker count before restarting and pass it here so we wait for the restarted
+# server to emit its own two markers rather than returning on the stale ones.
+wait-for-server-restarted() {
+  local container="$1"
+  local baseline="$2"
+  local want=$((baseline + 2))
+  MAXCHECKS=30
+  CHECKINTERVAL=1
+  for ((i=1;i<=MAXCHECKS;i++)); do
+      log-info "checking for restarted server APIs ($i of $MAXCHECKS max)..."
+      if [ "$(count-server-api-starts "$container")" -ge "$want" ]; then
+          return 0
+      fi
+      sleep "${CHECKINTERVAL}"
+  done
+
+  fail-now "timed out waiting for server to restart"
+}
+
 check-log-line() {
   # Check at most 30 times (with one second in between) that the agent has
   # successfully synced down the workload entry.
--- a/test/integration/common
+++ b/test/integration/common
@@ -148,9 +148,12 @@ check-server-started() {
 # server emits the "Starting Server APIs" line twice per boot (TCP and UDS
 # listeners).
 count-server-api-starts() {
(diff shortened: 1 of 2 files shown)


test/integration/suites/fetch-jwt-svids/08-test-disabled-jwt
```diff
@@ -4,23 +4,40 @@ log-info "testing disabled JWT functionality..."
...
 sed -i.bak "s#disable_jwt_svids = false#disable_jwt_svids = true#g" conf/server/server.conf
+
+# The restart reuses the container, so its log keeps the "Starting Server APIs"
+# markers from the original boot. Capture that count first and wait for the
+# restarted server to emit its own markers, otherwise the readiness check would
+# return on the stale ones before the server is actually listening again.
+baseline=$(count-server-api-starts spire-server)
 docker compose restart spire-server
-check-server-started spire-server
+wait-for-server-restarted spire-server "$baseline"
```
sorindumitru: Could we also have `docker-spire-server-restart` helper with this, in case we need to do this in other places too?

amartinezfayo: Thanks for the suggestion. I added a docker-spire-server-restart helper in test/integration/common that baselines the "Starting Server APIs" marker count, restarts the containers, then waits for each one to come back up, and step 08 now just calls docker-spire-server-restart spire-server. It accepts multiple container names so other suites can reuse it.

test/integration/common
```diff
@@ -144,6 +144,37 @@ check-server-started() {
...
 }
 
+# Counts how many times the server has logged that it started its APIs. The
+# server emits the "Starting Server APIs" line twice per boot (TCP and UDS
+# listeners).
+count-server-api-starts() {
+  # grep -c exits non-zero when there are no matches, which would trip the
+  # step's errexit/pipefail, so swallow that into a zero count.
+  docker compose logs "$1" | grep -c "Starting Server APIs" || true
+}
```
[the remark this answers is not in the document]
amartinezfayo: Reworked count-server-api-starts to capture the output into a variable first (logs=$(docker compose logs "$1")), so a docker compose logs failure still aborts under errexit/pipefail. The || true now applies only to the echo | grep -c pipeline, so it tolerates just grep's no-match exit code.

sorindumitru (approved):

PR #7114, opened 2026-07-03
GORM migration: Move common code to a new package

MarcosDY:
**Affected functionality**
In order to migrate GORM start to move code in common to a new package

**Description of change**
Create a new common package to allocate shared code between GORM v1 and GORM v2

**Which issue this PR fixes**
Part of: #1832


pkg/server/datastore/sqlcommon/sqlite_connstring.go
```diff
@@ -0,0 +1,45 @@
...
+	"path/filepath"
+	"runtime"
+)
+
+// EmbellishSQLite3ConnString adds query values supported by
+// github.com/mattn/go-sqlite3 to enable journal mode and foreign key support.
+// These query values MUST be part of the connection string in order to be
+// enabled for *each* connection opened by db/sql. If the connection string is
+// not already a file: URI, it is converted first.
+func EmbellishSQLite3ConnString(connectionString string) (string, error) {
```
sorindumitru, line 14: This might be better placed inside the `sqlite.go` file

MarcosDY: Good catch, I renamed connstring.go to sqlite_connstring.go (and its test) so it is with the other sqlite helper. 
I kept it as a separate file rather than folding it into sqlite.go because sqlite.go is //go:build cgo (for the mattn constraint check) and EmbellishSQLite3ConnString has no cgo dependency, it's used in both cgo and non-cgo builds, so moving it there would break the !cgo build.

sorindumitru (approved):
LGTM, thanks @MarcosDY

PR #7117, opened 2026-07-04
server: k8s_psat: validate pod UID & fix error message grammar

c4rlo:
TokenReview authenticates the projected service account token and returns the pod UID bound to that token. The server then fetches the live pod by namespace and name to derive node identity, selectors, and the agent SPIFFE ID.

Reject the attestation when the fetched pod UID does not match the TokenReview pod UID so a stale or raced pod lookup cannot bind authenticated token data to a different live pod. Update the unit tests to model pod UIDs explicitly and cover the mismatch case.

This closes a TOCTOU issue, though it's probably somewhat theoretical.

---

While I was in there, I also fixed some error message grammar (`fail to` → `failed to`).

### DIFF
--- a/pkg/server/plugin/nodeattestor/k8spsat/psat.go
+++ b/pkg/server/plugin/nodeattestor/k8spsat/psat.go
@@ -207,7 +207,7 @@ func (p *AttestorPlugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer)
 
 	namespace, serviceAccountName, err := k8s.GetNamesFromTokenStatus(tokenStatus)
 	if err != nil {
-		return status.Errorf(codes.Internal, "fail to parse username from token review status for cluster %q: %v", attestationData.Cluster, err)
+		return status.Errorf(codes.Internal, "failed to parse username from token review status for cluster %q: %v", attestationData.Cluster, err)
 	}
 	fullServiceAccountName := fmt.Sprintf("%v:%v", namespace, serviceAccountName)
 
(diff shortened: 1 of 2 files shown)


pkg/server/plugin/nodeattestor/k8spsat/psat.go
```diff
@@ -224,6 +224,9 @@ func (p *AttestorPlugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer)
 	if err != nil {
 		return status.Errorf(codes.Internal, "fail to get pod from k8s API server for cluster %q: %v", attestationData.Cluster, err)
 	}
+	if string(pod.UID) != podUID {
+		return status.Errorf(codes.Internal, "pod UID mismatch for cluster %q", attestationData.Cluster)
+	}
```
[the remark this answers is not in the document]
c4rlo: Good call, addressed in d4bc0fc.

pkg/server/plugin/nodeattestor/k8spsat/psat.go
```diff
@@ -212,22 +212,25 @@ func (p *AttestorPlugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer)
...
+		return status.Errorf(codes.Internal, "failed to get pod UID from token review status for cluster %q: %v", attestationData.Cluster, err)
 	}
 
 	pod, err := cluster.client.GetPod(stream.Context(), namespace, podName)
 	if err != nil {
-		return status.Errorf(codes.Internal, "fail to get pod from k8s API server for cluster %q: %v", attestationData.Cluster, err)
+		return status.Errorf(codes.Internal, "failed to get pod from k8s API server for cluster %q: %v", attestationData.Cluster, err)
+	}
+	if string(pod.UID) != podUID {
+		return status.Errorf(codes.PermissionDenied, "pod UID mismatch for cluster %q", attestationData.Cluster)
```
amartinezfayo: I'm wondering if we could include a bit more context in this error to help operators diagnose a rejection. Including the pod name and the token's UID, for example `pod UID mismatch for pod %q in cluster %q: token bound to pod UID %q`, would make it clearer which pod was involved, while keeping the error limited to values the caller already presented in its token.

amartinezfayo (commented):
Thank you @c4rlo!

pkg/server/plugin/nodeattestor/k8spsat/psat.go
c4rlo: Agreed – done.

I was thinking of also including `pod.UID`, as that could be useful for debugging too, but it might not be ideal to include that in the response as it divulges some info that could help an attacker. Instead we could only log that, but I've left that out since other code paths don't do it and it might not be that important.

amartinezfayo (approved):
Thank you @c4rlo!

PR #7120, opened 2026-07-04
GORM migration: extract sqltest suite

MarcosDY:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
The SQL datastore test suite. 
This is a test-only refactor, no production datastore behavior changes.
It's the second step of the ongoing work to allow an optional gorm v2 datastore to live alongside the existing one, following the sqlcommon extraction that already merged (#7114).

**Description of change**
The test suite (the big PluginSuite) used to live inside the sqlstore package's test file, so only the v1 datastore could run it. This PR lifts it out into a new shared pkg/server/datastore/sqltest package, parameterized over a datastore factory, so a future gorm v2 datastore can be validated against the exact same tests.

A few decisions worth calling out, in case you'd have done them differently:

- The raw-SQL escape hatch (RawScan/RawExec/DatabaseType/Rebind) is exported on the concrete *Plugin in a non-test file. A handful of tests need raw DB access that bypasses the ORM (PRAGMA checks, inserting deliberately-corrupt rows, etc.). Because the shared suite is now a separate package, these methods can't live in a _test.go file or behind a build tag. So they have to be exported. They're documented as test-support only and aren't wired to anything in production. **Open to suggestions if there's a cleaner pattern I missed**.
- RawQuerier is embedded in DataStoreUnderTest so the escape-hatch contract is checked at compile time.
- Placeholder rewriting has a single source of truth. The suite briefly had its own copy of the ? → $1 rebind logic. That's now a Rebind() method delegating to the production maybeRebind, so the tests can't drift from what production actually generates.
- The suite still uses v1's HCL config keys (database_type, connection_string, etc.). This is intentional, the planned v2 datastore reuses the same config schema on purpose (same parser), so there's nothing to abstract yet.

No docs change — this is test infrastructure only.

**Which issue this PR fixes**
Part of #1832


sorindumitru (dismissed):
LGTM, thanks @MarcosDY

MarcosDY:
@sorindumitru there was a flaky test on windows I applied a change to try to resolve lets me know what do you think

sorindumitru (approved):

PR #7121, opened 2026-07-04
GORM Migration: Add skeleton

MarcosDY:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Adds a new, optional SPIRE server datastore built on gorm v2 (gorm.io/gorm), living alongside the existing gorm v1 datastore. 
This is the skeleton step: the new datastore can connect and be configured, but has no query methods yet and is not wired into the catalog, so it is unreachable by operators. Continues the gorm v2 migration after the sqlcommon extraction (#7114).

**Description of change**
New package pkg/server/datastore/sqlstorev2 (PluginName = "sql_v2") built on gorm v2.

A few decisions worth flagging:

- sqlite mirrors v1 exactly, same cgo mattn/go-sqlite3 driver via gorm.io/driver/sqlite, behind //go:build cgo, with the same "not supported without cgo" behavior under CGO_ENABLED=0.
- AWS RDS / IAM uses sql.Open + Conn injection. gorm v2 has no RegisterDialect, so the AWS path opens a *sql.DB from the existing awsrds driver and hands it to gorm v2 via postgres.New. The awsrds driver itself is untouched.
- MySQL custom TLS is now shared and preserved.
- AWS IAM DSN assembly and the db-type predicates are de-duplicated into sqlcommon (BuildAWS{MySQL,Postgres}DSN, Is{MySQL,Postgres,SQLite}DbType) rather than copied per datastore. Two small changes that affects v1, both moving it toward correctness:
  - v1's AWS MySQL path now rejects a password on an IAM connection (it already did this for postgres).
  - The postgres IAM password check uses pgx.ParseConfig().Password, which correctly recognizes URL, ?password=, and keyword-form passwords. It does merge libpq env fallbacks (PGPASSWORD / .pgpass), so tests isolate the environment with t.Setenv("PGPASSWORD", ""), the same guard awsrds_test.go already uses.

Known limitation (pre-existing, not introduced here): if an operator has PGPASSWORD set in their environment (not the connection string), the AWS postgres IAM path is rejected — because the awsrds driver re-parses with pgx at connect time and pgx merges env vars. This affects v1 today too; it's out of scope for this skeleton and worth a separate issue against the awsrds driver.

No docs change — the new datastore isn't selectable yet; docs land with the catalog-wiring PR.

**Which issue this PR fixes**
Part of: #1832


pkg/server/datastore/sqlcommon/awsdsn.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/jackc/pgx/v5"
+	"github.com/spiffe/spire/pkg/server/datastore/sqldriver/awsrds"
+)
+
+// BuildAWSPostgresDSN builds the AWS RDS / IAM DSN for a postgres connection.
+// It rejects a connection string carrying a password, since IAM auth supplies
+// a rotating token instead. Shared by the v1 and v2 datastores so IAM DSN
+// handling has a single source of truth.
+func BuildAWSPostgresDSN(cfg *Configuration) (string, error) {
+	connString := GetConnectionString(cfg, false)
```
sorindumitru, line 17: Should this be passing in true for the read only connection?

pkg/server/datastore/sqlcommon/awsdsn.go
```diff
@@ -0,0 +1,54 @@
...
+		DbUser:          c.User,
+		DriverName:      awsrds.PostgresDriverName,
+		ConnString:      connString,
+	}
+	return awsrdsConfig.FormatDSN()
+}
+
+// BuildAWSMySQLDSN builds the AWS RDS / IAM DSN for a MySQL connection.
+func BuildAWSMySQLDSN(cfg *Configuration, mysqlConfig *mysql.Config) (string, error) {
+	if mysqlConfig.Passwd != "" {
```
sorindumitru, line 40: This check wasn't there in the previous version. It's good to have, but it may be a breaking change? Might be best to remove for and replace with a comment + an issue so we remember to do it in 1.16?

pkg/server/datastore/sqlstore/sqlstore.go
```diff
@@ -5103,58 +5040,16 @@ func deleteCAJournal(tx *gorm.DB, caJournalID uint) error {
...
-	return dbType == MySQL || dbType == AWSMySQL
+	return sqlcommon.IsMySQLDbType(dbType)
 }
 
 func isPostgresDbType(dbType string) bool {
-	return dbType == PostgreSQL || dbType == AWSPostgreSQL
+	return sqlcommon.IsPostgresDbType(dbType)
 }
 
 func isSQLiteDbType(dbType string) bool {
```
sorindumitru, line 5051: Could we remove the usage of these `is*DbType` and replace them with `sqlcommon.Is*DbType`? If there's too many of them, it's ok to have them in a separate PR

pkg/server/datastore/sqlstorev2/errors.go
```diff
@@ -0,0 +1,8 @@
+package sqlstorev2
+
+import "github.com/spiffe/spire/pkg/server/datastore/sqlcommon"
+
+// newSQLError and newWrappedSQLError are thin aliases so the rest of the
+// package reads like v1 while all error construction lives in sqlcommon.
+func newSQLError(format string, args ...any) error { return sqlcommon.NewSQLError(format, args...) }
+func newWrappedSQLError(err error) error           { return sqlcommon.NewWrappedSQLError(err) }
```
sorindumitru, line 8: Since this is new code, we could probably use `sqlcommon.New...` directly, to avoid these wrappers.

Issue #7116, opened 2026-07-04
Feature Request: Support a Kubernetes Deployment for TCP-only Broker API


matheuscscp:
PRs: #7122, #7123, #7141, #7142

Unblocks: https://github.com/spiffe/helm-charts-hardened/pull/868

My plan for Deployment agents in terms of registration entries is as follows. It supports multiple Deployment agents in the Kubernetes cluster, in any namespaces. (The below is fully written by hand, no AI, please read everything!)

### `k8s_psat` node attestor cluster entries

Cluster entries in the `k8s_psat` node attestor configuration are just named rules the agents can choose among for defining the format of the SPIFFE ID they will get, they do not define selectors or entry authorization (see next sections). The selectors emitted for agents attesting by this method are [fixed](https://github.com/spiffe/spire/blob/main/doc/plugin_server_nodeattestor_k8s_psat.md).

In a Kubernetes cluster with both DaemonSet agents and Deployment agents, two entries are needed in the server-side config of `k8s_psat`:

* The usual one, defaulting to node-UID SPIFFE IDs for agents
* An additional  entry using `use_pod_uid_for_agent_id`

If the cluster has multiple Deployments acting as SPIRE agents, all of them can reference the second (pod-UID-based) cluster entry without conflicts.

The DaemonSet agent uses the first cluster entry (node-UID-based). If there's only one DaemonSet agent in the cluster (e.g. the one deployed by the SPIRE Helm chart), this is enough. If there are multiple DaemonSets, then only one can use the first node-UID-based cluster entry, all the others have to use the pod-UID-based entry or create a new node-UID-based cluster entry.

### Alias entries

A single alias is needed for each SPIRE agent, regardless if Deployment or DaemonSet. Multiple aliases for the same group-of-pods SPIRE agent do not improve security, because at the end all of them have to select the same set of agent pods. In other words, from the perspective of alias entries, a group of pods representing the same logical SPIRE agent is the atomic unit. In Kubernetes, that is represented by the ServiceAccount being shared by all the pods in that group.

So for each Deployment or DaemonSet that is a SPIRE agent in the cluster, I'd recommend a single alias entry:

* Parent ID: SPIRE server ID
* SPIFFE ID: the alias ID for a specific group of pods (Deployment or DaemonSet)
* Selectors:
  *  `k8s_psat:cluster:<cluster>` (The server-side cluster entry the agent has chosen to join, see previous section)
  *  `k8s_psat:agent_ns:<namespace>`
[...]

PR #7122, opened 2026-07-06
Introduce flags for disabling Workload and SDS APIs

matheuscscp:
Part of: #7116

Introduces service-level switches in SPIRE Agent for disabling APIs that are normally served on the public agent endpoint:

* `agent.disable_workload_api` / `-disableWorkloadAPI`
* `agent.disable_sds_api` / `-disableSDSAPI`

The public endpoint remains enabled while at least one of those APIs is enabled:

* `disable_workload_api = false`, `disable_sds_api = false`: expose both Workload API and SDS
* `disable_workload_api = true`, `disable_sds_api = false`: expose only SDS
* `disable_workload_api = false`, `disable_sds_api = true`: expose only Workload API
* `disable_workload_api = true`, `disable_sds_api = true`: do not bind the public endpoint

This applies uniformly to the Unix socket configured by `agent.socket_path` and the Windows named pipe configured by `agent.experimental.named_pipe_name`. Existing configurations keep the current behavior because both flags default to `false`.

This is useful for deploying Broker API-only agents.


pkg/agent/agent.go
```diff
@@ -540,6 +547,14 @@ func (a *Agent) newAdminEndpoints(metrics telemetry.Metrics, mgr manager.Manager
 
 // CheckHealth is used as a top-level health check for the agent.
 func (a *Agent) CheckHealth() health.State {
+	if a.c.BindAddress == nil || a.c.DisableWorkloadAPI {
```
amartinezfayo: I'm wondering if we lose a useful readiness signal for SDS-only agents here. With the Workload API disabled but the endpoint still serving SDS, readiness becomes just the started flag, which is set once at startup and never re-evaluated. Since checkWorkloadAPI only fails on Unavailable, a server that has SDS registered but not the Workload API would return Unimplemented for the fetch and still pass, so it seems the dial-based check could be kept whenever BindAddress is non-nil and would still verify that the listener is actually accepting connections. Would that be worth preserving for the SDS-only case?

pkg/agent/agent.go
```diff
@@ -274,20 +274,25 @@ func (a *Agent) Run(ctx context.Context) error {
...
+	if a.c.BindAddress != nil {
+		agentEndpoints := a.newEndpoints(metrics, mgr, workloadAttestor)
+		go func() {
+			agentEndpoints.WaitForListening(readyForHealthChecks)
+			a.started = true
+		}()
+		tasks = append(tasks, agentEndpoints.ListenAndServe)
+	} else {
+		a.started = true
+		close(readyForHealthChecks)
```
amartinezfayo, line 294: Maybe we could log something in this branch so an operator can tell from the logs that the public endpoint was intentionally not bound. If both settings are enabled by accident, the only symptom would be workloads failing to connect to a socket that doesn't exist, and a line noting that the Workload API and SDS are disabled would make that quick to diagnose.

pkg/agent/endpoints/endpoints.go
```diff
@@ -139,15 +152,15 @@ func (e *Endpoints) ListenAndServe(ctx context.Context) error {
 	e.log.WithFields(logrus.Fields{
 		telemetry.Network: e.addr.Network(),
 		telemetry.Address: e.addr,
-	}).Info("Starting Workload and SDS APIs")
+	}).Infof("Starting %s", e.apiNames)
```
amartinezfayo: This makes the log message text dynamic. In general, we try to keep messages fixed and convey context through structured fields to keep message cardinality down.
What do you think about keeping a static message and putting the enabled API names in a field, something like log.WithField("apis", e.apiNames).Info("Starting agent APIs"), here and in the corresponding stop message?

doc/spire_agent.md
```diff
@@ -307,9 +309,11 @@ the following flags are available:
 | `-logFile`                       | File to write logs to                                                                         |                       |
 | `-logFormat`                     | Format of logs, &lt;text&vert;json&gt;                                                        |                       |
 | `-logLevel`                      | DEBUG, INFO, WARN or ERROR                                                                    |                       |
+| `-disableWorkloadAPI`            | Disable the SPIFFE Workload API                                                               | false                 |
```
amartinezfayo: The rows in this table look alphabetically sorted, so should we move -disableSDSAPI and -disableWorkloadAPI up next to -dataDir to keep the ordering?

doc/spire_agent.md
```diff
@@ -66,7 +66,9 @@ This may be useful for templating configuration files, for example across differ
...
-| `socket_path`                     | Location to bind the SPIRE Agent API socket (Unix only)                                                                                                                                                                                           | /tmp/spire-agent/public/api.sock |
+| `socket_path`                     | Location to bind the SPIRE Agent Workload API and SDS socket (Unix only). The socket is exposed unless both `disable_workload_api` and `disable_sds_api` are `true`.                                                                              | /tmp/spire-agent/public/api.sock |
+| `disable_workload_api`            | Disable the SPIFFE Workload API. The public socket or named pipe remains exposed if SDS is enabled.                                                                                                                                               | false                            |
```
amartinezfayo: It could be worth noting that the spire-agent healthcheck command targets the public endpoint, so it won't work when both APIs are disabled, and deployments using it as a liveness or readiness probe would need to switch to the health_checks HTTP listener. Perhaps a short note in the spire-agent healthcheck command section would help too, since that's where someone would look up the command's requirements.

amartinezfayo (commented):
Thank you @matheuscscp!

matheuscscp:
@amartinezfayo I addressed everything in https://github.com/spiffe/spire/pull/7122/commits/3bd8c53d00c65ad54c92df73e1ce6fd1600c89c9

As a follow-up from this PR, I realized we need to add the Broker Endpoints to the healthchecks as well, especially since we are preparing the SPIRE agent to run only with Broker API.

amartinezfayo (approved):
Thank you @matheuscscp!

PR #7125, opened 2026-07-08
Move JWTSVIDCache outside of the LRUCache

sorindumitru:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
There should be no functionality change here. Just extracting this cache out of the LRU cache to make room for caching the new WIT-SVID.

**Description of change**

There's no particular reason to keep these together and it makes it a bit weirder if we want to use the LRU cache for different SVID types (i.e. X509 and WIT)

### DIFF
--- a/pkg/agent/manager/cache/jwt_cache.go
+++ b/pkg/agent/manager/cache/jwt_cache.go
@@ -39,8 +39,8 @@ type jwtSvidElement struct {
 }
 
 func (c *JWTSVIDCache) CountJWTSVIDs() int {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+	c.mu.RLock()
+	defer c.mu.RUnlock()
 
 	return len(c.svids)
 }
--- a/pkg/agent/manager/cache/jwt_cache.go
+++ b/pkg/agent/manager/cache/jwt_cache.go
@@ -74,6 +74,8 @@ func (c *JWTSVIDCache) GetJWTSVID(spiffeID spiffeid.ID, audience []string) (*cli
 }
 
 func (c *JWTSVIDCache) SetJWTSVID(spiffeID spiffeid.ID, audience []string, svid *client.JWTSVID) {
+	defer func() { agent.SetJWTSVIDCacheSize(c.metrics, c.CountJWTSVIDs()) }()
+
 	key := jwtSVIDKey(spiffeID, audience)
 
 	c.mu.Lock()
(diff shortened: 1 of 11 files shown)


pkg/agent/manager/sync.go
```diff
@@ -87,7 +82,7 @@ func (m *manager) processTaintedAuthorities(ctx context.Context, bundle *spiffeb
 			Debug("New tainted JWT authorities found")
 
 		// Taint JWT-SVIDs in the cache
-		m.cache.TaintJWTSVIDs(ctx, jwtAuthorities)
+		m.jwtCache.TaintJWTSVIDs(ctx, jwtAuthorities)
```
amartinezfayo, line 85: Now that TaintJWTSVIDs is no longer part of the SVIDCache interface, I think the no-op TaintJWTSVIDs on the SVID store cache (storecache.Cache) is left without any callers, since it only existed to satisfy that interface.
Can we remove it?

pkg/agent/manager/cache/jwt_cache.go
```diff
@@ -74,6 +74,8 @@ func (c *JWTSVIDCache) GetJWTSVID(spiffeID spiffeid.ID, audience []string) (*cli
 }
 
 func (c *JWTSVIDCache) SetJWTSVID(spiffeID spiffeid.ID, audience []string, svid *client.JWTSVID) {
+	defer func() { agent.SetJWTSVIDCacheSize(c.metrics, c.CountJWTSVIDs()) }()
```
amartinezfayo, line 77: I'm wondering if we could simply make CountJWTSVIDs take the read lock instead of the exclusive lock, matching what CountX509SVIDs does in the LRU cache. That would keep this deferred emission cheap and I think it would also address the contention concern from the Copilot comments without restructuring the closures.

pkg/agent/manager/cache/jwt_cache_test.go
```diff
@@ -198,6 +213,11 @@ func TestJWTSVIDCache(t *testing.T) {
 					Val:    0,
 					Labels: []metrics.Label{{Name: "status", Value: "OK"}},
 				},
+				{
+					Type: fakemetrics.SetGaugeType,
+					Key:  []string{telemetry.JWTSVIDCacheSize},
+					Val:  3,
```
amartinezfayo, line 219: It could be worth also asserting the gauge in TestJWTSVIDCacheSize, for example checking that it reads the max size after an insert evicts the oldest entry. As far as I can see, the assertions here only cover the gauge as emitted through TaintJWTSVIDs, since the values emitted by SetJWTSVID are cleared by the reset before tainting.

amartinezfayo (approved):

PR #7132, opened 2026-07-09
svid.v1 client: add support for building WIT-SVIDs

sorindumitru:


pkg/agent/client/client.go
```diff
@@ -402,6 +409,49 @@ func (c *client) NewJWTSVID(ctx context.Context, entryID string, audience []stri
...
+	svids := make(map[string]*WITSVID)
+	var params []*svidv1.NewWITSVIDParams
+	for entryID, publicKey := range publicKeys {
+		pkixPublicKey, err := x509.MarshalPKIXPublicKey(publicKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal PKIX public key for entry %q: %w", entryID, err)
+		}
+
+		params = append(params, &svidv1.NewWITSVIDParams{
+			EntryId:   entryID,
```
MarcosDY: signing_algorithm is never set, is not this going to fail on every call?

https://github.com/spiffe/spire-api-sdk/blob/main/proto/spire/api/server/svid/v1/svid.proto#L208-L210

pkg/agent/client/client.go
```diff
@@ -768,6 +818,39 @@ func (c *client) fetchSVIDs(ctx context.Context, params []*svidv1.NewX509SVIDPar
...
+	resp, err := svidClient.BatchNewWITSVID(ctx, &svidv1.BatchNewWITSVIDRequest{
+		Params: params,
+	})
+	if err != nil {
+		c.release(connection)
+		c.withErrorFields(err).Error("Failed to batch new WIT-SVID(s)")
+		return nil, fmt.Errorf("failed to batch new WIT-SVID(s): %w", err)
+	}
+
+	okStatus := int32(codes.OK)
```
MarcosDY, line 837: we are not validating response the same way we do for jwt:
```
	case svid.IssuedAt == 0:
		return nil, spiffeid.ID{}, errors.New("JWTSVID missing issued at")
	case svid.ExpiresAt == 0:
		return nil, spiffeid.ID{}, errors.New("JWTSVID missing expires at")
	case svid.IssuedAt > svid.ExpiresAt:
		return nil, spiffeid.ID{}, errors.New("JWTSVID issued after it has expired")
	}
```

this can result on invalid values,  NewWITSVIDs converts both via time.Unix so a zero ExpiresAt becomes 1970.

pkg/agent/client/client.go
sorindumitru: Ups, yes, this needs to be set. Added it as parameter

pkg/agent/client/client.go
sorindumitru: Added that now

MarcosDY (approved):

PR #7134, opened 2026-07-09
Add Marshalling support for WIT-SVID keys

sorindumitru:


pkg/common/bundleutil/marshal.go
```diff
@@ -67,6 +76,25 @@ func StandardJWKS() MarshalOption {
...
+}
+
+func MarshalWITSVIDBundle(bundle *spiffebundle.Bundle, opts ...MarshalOption) (string, error) {
+	allOpts := append([]MarshalOption{NoX509SVIDKeys(), NoJWTSVIDKeys()}, opts...)
+	marshaledBundle, err := Marshal(bundle, allOpts...)
+	if err != nil {
+		return "", err
+	}
+	return string(marshaledBundle), nil
+}
```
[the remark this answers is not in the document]
sorindumitru: I think this decision was made for us by `go-spiffe` already. I don't think we can fix it there now, since it would be a breaking API change.

pkg/common/bundleutil/marshal.go
```diff
@@ -67,6 +76,25 @@ func StandardJWKS() MarshalOption {
...
+	allOpts := append([]MarshalOption{NoJWTSVIDKeys(), NoWITSVIDKeys()}, opts...)
+	return Marshal(bundle, allOpts...)
+}
+
+func MarshalJWTSVIDBundle(bundle *spiffebundle.Bundle, opts ...MarshalOption) ([]byte, error) {
+	allOpts := append([]MarshalOption{NoX509SVIDKeys(), NoWITSVIDKeys()}, opts...)
+	return Marshal(bundle, allOpts...)
+}
+
+func MarshalWITSVIDBundle(bundle *spiffebundle.Bundle, opts ...MarshalOption) (string, error) {
```
MarcosDY, line 89: why it to be a string but all another ones are []byte?

sorindumitru: This matches the shape of the API. See [JWTBundles](https://github.com/spiffe/spiffe/blob/dc4e9d9b4eff8aa181a54cd330ff9f877186060e/standards/workloadapi.proto#L160) vs [WITBundles](https://github.com/spiffe/spiffe/blob/dc4e9d9b4eff8aa181a54cd330ff9f877186060e/standards/workloadapi.proto#L237).

We can probably do `[]byte` here but it would mean ugly conversions where we use it.

MarcosDY (approved):

PR #7143, opened 2026-07-13
azure_imds: add configurable trust_bundle_path for additional root CAs

ravishen:
**Affected functionality**

`azure_imds` server node attestor: certificate chain validation / configuration.

**Description of change**

Adds an optional `trust_bundle_path` configuration option to the `azure_imds` node attestor. When set, the PEM file at that path is loaded at configure time and its certificates are added to the set of root CAs used to validate the attested document's signing certificate chain.

The configured roots are a **union** with the roots embedded in SPIRE, not a replacement, embedded roots remain the default and are always trusted. This lets operators react to an Azure root CA change (e.g. a newly cross-signed root)
without waiting for a new SPIRE release, addressing the root-pinning maintenance concern raised in #6952.

Details:
- Added `trust_bundle_path` HCL field, following the `ca_bundle_path`
  precedent in the `x509pop` attestor.
- Updated docs

**Which issue this PR fixes**

Relates to #6952


pkg/server/plugin/nodeattestor/azureimds/imds.go
```diff
@@ -202,11 +206,21 @@ func (p *IMDSAttestorPlugin) buildConfig(coreConfig catalog.CoreConfig, hclText
 		allowedMetadataDomains = []string{DefaultMetadataDomain}
 	}
 
+	var additionalRoots []*x509.Certificate
+	if newConfig.TrustBundlePath != "" {
+		certs, err := util.LoadCertificates(newConfig.TrustBundlePath)
```
MarcosDY, line 211: can you plz verify that the loaded certificates are bundles before adding them?

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
```diff
@@ -80,8 +80,8 @@ func validateAttestedDocument(ctx context.Context, doc *azure.AttestedDocument,
 	}
 
 	// Step 8: Validate certificate chain
-	if err := validateCertificateChain(signingCert, intermediateCert); err != nil {
-		return nil, fmt.Errorf("certificate chain validation failed: %w", err)
+	if err := validateCertificateChain(signingCert, intermediateCert, additionalRoots); err != nil {
+		return nil, err
```
MarcosDY: qq: why to remove the error with context about why it failed?

doc/plugin_server_nodeattestor_azure_imds.md
```diff
@@ -20,6 +20,7 @@ attestation or to resolve selectors.
...
 | `agent_path_template`       | Optional | A URL path portion format of Agent's SPIFFE ID. Describe in text/template format.                                                                                                                                                                                                      | `"/{{ .PluginName }}/{{ .TenantID }}/{{ .SubscriptionID }}/{{ .VMID }}"` |
 | `allowed_metadata_domains`  | Optional | A list of allowed Azure metadata domains for certificate validation. Each domain accepts the base domain (e.g., `metadata.azure.com`) and automatically validates certificates with that domain or any subdomain (e.g., `eastus.metadata.azure.com`, `sub.eastus.metadata.azure.com`). | `["metadata.azure.com"]`                                                 |
+| `trust_bundle_path`         | Optional | Path to a PEM file with one or more additional root CA certificates to trust when validating the signing certificate chain. These are added to (not a replacement for) the roots embedded in SPIRE, so operators can trust a new Azure root CA without a new SPIRE release.            |                                                                          |
```
MarcosDY, line 23: The `## Security Considerations section` covers the IMDS/TOFU risk but isn't updated for this option. `trust_bundle_path` adds trust anchors for node attestation, which is worth calling out explicitly there.
maybe something like:
```
### Additional Root CAs

Certificates supplied via trust_bundle_path become trust anchors for node attestation. They are added to the root CAs embedded in SPIRE, which remain trusted and cannot be removed or overridden through configuration.

Any party able to issue certificates under one of these roots can produce an attested document that this attestor accepts. The Subject Alternative Name check compares against allowed_metadata_domains without binding to a particular issuer, so a certificate issued under an operator-supplied root carrying a SAN such as metadata.azure.com will pass. The subscription and tenant allowlists limit which tenants and subscriptions may be claimed, but within an allowed tenant the holder of such a key can attest as an arbitrary virtual machine.

Configure this option only with root CAs you have verified are operated by Microsoft. Protect the bundle file accordingly, since write access to it is sufficient to introduce a new trust anchor.

The bundle is read once, when the plugin is configured. Changes to the file take effect only after the SPIRE Server is restarted.
```

pkg/server/plugin/nodeattestor/azureimds/imds.go
```diff
@@ -202,11 +206,21 @@ func (p *IMDSAttestorPlugin) buildConfig(coreConfig catalog.CoreConfig, hclText
 		allowedMetadataDomains = []string{DefaultMetadataDomain}
 	}
 
+	var additionalRoots []*x509.Certificate
+	if newConfig.TrustBundlePath != "" {
```
MarcosDY, line 210: this is only loaded when configuring attestor, can you plz update documentation about that? (on md file)

MarcosDY (commented):
Great work!! 
Just some comments

pkg/server/plugin/nodeattestor/azureimds/imds.go
ravishen: Sorry, I'm not sure if i follow. Currently, `util.LoadCertificates` already rejects a file that isn't a usable PEM certificate bundle it errors with no certificates found or unable to parse certificate in block

pkg/server/plugin/nodeattestor/azureimds/doc_validation.go
ravishen: That was me acting on an [earlier Copilot comment
](https://github.com/spiffe/spire/pull/7143#discussion_r3570132721)
I fixed the wrong end though only the `Verify()` path was duplicated, and the other two paths (`signing certificate was not issued by intermediate certificate`, embedded-root parse failure) lost their context. So I've restored the wrapper here and reworded the inner message to `unable to verify signing certificate against root CAs` instead.

Both wrappers pre-date this PR, so happy to just restore the outer one if you'd prefer a narrower diff.

doc/plugin_server_nodeattestor_azure_imds.md
ravishen: Updated the doc

pkg/server/plugin/nodeattestor/azureimds/imds.go
ravishen: Documented in the new `### Additional Root CAs` section under Security Considerations

>  the bundle is read once when the plugin is configured, and changes don't take effect until it's configured again.

pkg/server/plugin/nodeattestor/azureimds/imds.go
MarcosDY: yeap, it is a Certificate, but I was worried someone add a Leaf by error

ravishen: Added the check in 537c437bd0693d37ecb153dc268c37125b7db298

MarcosDY (approved):

PR #7149, opened 2026-07-17
Move the testing functionality out of the main DataStore interface

sorindumitru:
This simplifies the interface a bit and makes it impossible to use these functions in core code.


amartinezfayo (approved):
Thank you @sorindumitru!

PR #7150, opened 2026-07-17
Fix unrecognized parameters warnings

mweigel:
**Affected functionality**
Currently, when using the hashicorp_vault KeyManager plugin with Vault or OpenBao, there are warnings in the audit log - "warnings":["Endpoint ignored these unrecognized parameters: [marshalling_algorithm]".

The correct parameter is "marshaling_algorithm". Note, this doesn't cause any issues as asn1 is the default anyway and the plugin works well. This fix just removes the warnings.

- https://developer.hashicorp.com/vault/api-docs/secret/transit#marshaling_algorithm
- https://openbao.org/api-docs/secret/transit/#parameters-20

**Description of change**
Get rid of unnecessary warning messages.


sorindumitru (approved):
LGTM, thanks @mweigel!

Issue #7153, opened 2026-07-20
slurm workload attestor support


kfox1111:
Slurm is the most common HPC job scheduler. Its to HPC what Kubernetes is to container orchestration.

We would like to use SPIRE on our HPC systems.

According to https://slurm.schedmd.com/cgroup_v2.html
when using its cgroupv2 support, will encode the (SLURM jobid or SLUID) along with the step.
This info alone is useful along with a spire-controller-manager like process running on the slurm control node to target svids to sllurm jobs.

While there is support for v1, its on the way out. I'd propose we simply support v2 in the initial implementation.

This functionality matches closely the other in tree plugins that support workload attestation by cgroup naming.

amartinezfayo:
Hey @kfox1111, we have discussed this within the maintainers group, and we think that it would be good for the project to add Slurm workload attestation support in SPIRE.
- Do you have an idea of the set of selectors that the plugin would discover? Our main interest is to look into the security properties of the process of the discovery and make sure they come from a trusted source.
- Do you think we could have an integration test that exercises this plugin?

Thank you!

kfox1111:
I was thinking just
* the (SLURM jobid or SLUID)
* the step.

slurm encodes both of those in the cgroup path directly. anything else I think is less strong. We could gather other options with a slurm hook, but not sure they are needed. Maybe we start with just this much.

As for an integration test, there are slurm ubuntu packages, so an integration test should be fairly easy to do.

sorindumitru:
> * the (SLURM jobid or SLUID)
> 
> * the step.

How do these look and what do they represent? I don't think any of us are familiar with slurm and it's terminology.

kfox1111:
The structure is in the link:
https://slurm.schedmd.com/cg_hierarchy.jpg

slurm jobids are just integers.

steps are generally integers, but can also be the string "batch" for the main batch script itself on the first node, or "extern" used by slurm support processes.  unlikely for slurm itself ot be calling spire. but maybe some day.

for a sluid, https://slurm.schedmd.com/MISC26/Slurm_GTC_2026.pdf has some details. its generally intended to be a unique opaque string.

So, I think we'd just have 3 selectors, one for each of those 3 things. only 2 would ever be used at once though.

---

Pointing cluade at the upstream code and asking it about the sluid generation code:
```
Found it. The core lives in src/common/sluid.c and src/common/sluid.h, with the type in slurm/slurm.h:62.

What a SLUID is
[...]

PR #7160, opened 2026-07-23
Slurm workload attestor

kfox1111:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
NA

**Description of change**
Adds a new slurm workload attestor

**Which issue this PR fixes**
https://github.com/spiffe/spire/issues/7153


pkg/agent/plugin/workloadattestor/slurm/slurm_posix.go
```diff
@@ -0,0 +1,121 @@
...
+// slurmCgroupRe matches the Slurm cgroup/v2 hierarchy created by slurmstepd:
+//
+//	/system.slice/[<nodename>_]slurmstepd.scope/<JOBDIR>/step_<STEP>/{slurm|user}/task_<n>
+//
+// <JOBDIR> is either "job_<numeric>" (when CgroupJobIdPaths=yes) or a bare SLUID,
+// which is Crockford Base32 with a leading 's'. <STEP> is a numeric step id or one
+// of the special names (batch, extern, interactive, TBD). Anchoring on
+// "slurmstepd.scope/" also matches the "<nodename>_slurmstepd.scope" form used with
+// --enable-multiple-slurmd, since that segment ends in "slurmstepd.scope".
+var slurmCgroupRe = regexp.MustCompile(`slurmstepd\.scope/(job_\d+|s[0-9A-Za-z]+)/step_([^/]+)`)
```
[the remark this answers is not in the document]
kfox1111: This is a great catch. Will fix.

test/integration/suites/slurm-x509/conf/job/fetch-svid.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+#SBATCH --job-name=spire-fetch
```
sorindumitru, line 2: Is it not possible to get this job name as a `selector` in some way? I imagine this would be a bit more useful than the IDs.

sorindumitru: https://slurm.schedmd.com/rest_api.html#slurmV0045GetJob seems to have some more details about the job, the name, some group name, and maybe others.

kfox1111: the job name is a slurm user specifiable field for the user to help keep track of their own jobs. It will be insecure when used alone as a selector. So not super useful to spire setups I think.

I was trying to keep things simple. slurm-rest is not a commonly deployed component, and there is no standard way for a node to auth to it.

The easiest extension point would be to add a slurm prolog script on the compute nodes that pushes any of the metadata that slurm pushes to the hook:

> Prolog gives you a decent set: SLURM_JOB_ID, SLURM_JOB_UID/GID/USER, SLURM_JOB_NODELIST, SLURM_JOB_PARTITION, SLURM_JOB_QOS, SLURM_JOB_ACCOUNT, SLURM_JOB_START_TIME, SLURM_JOB_WORK_DIR, SLURM_JOB_STDOUT/STDERR, SLURM_JOB_CONSTRAINTS, SLURM_JOB_GPUS, SLURM_JOB_SLUID, SLURMD_NODENAME, SLURM_CLUSTER_NAME

then get that data to the spire-agent plugin somehow. like dump it to json files. job name was left out I think as its a user field, so not for hooks.

Of those though, I think the only really useful ones are SLURM_CLUSTER_NAME and SLURM_JOB_ACCOUNT. USER info is attestable via the unix attestor already.
I'm not convinced any of those are worth making workload selectors for though. Just my 2c.

So, none of this helps with eliminating the need of of plumbing that makes the registration entries dynamically per job in spire. The most useful thing would be if you could do it per slurm account, but HPC systems often run into lots of scaling issues due to their size. Blanket adding x509 entries on nodes where the jobs are not registered to will make the scaling issue much worse. So an account selector, being the most useful other selector, doesnt really help much. Its super useful as a template item for the spiffe_id though.

So my suggestion would be, lets merge the functionality as is. Its simple and doesn't require extra scripts or advanced configuration of slurm on the compute nodes to get it to work. As we get more users of it, maybe some will think the extra functionality is worth the effort, and we can enhance it then. Or, after I try and use it for a while, maybe I'll change my mind and need it too. But, this current level of functionality feels like a good starting point to me.

sorindumitru: Thanks for the information. So how would the SPIFFE IDs for these entries look like? Something like `spiffe://<trust domain>/sluid/<sluid>job/<id>`? Also, how would the nodes where the job runs be identified?

kfox1111: I'm planning on using `spiffe://<trust domain>/<clustername>/<slurm account>`

As previously mentioned, a spire-controller-manager like process that runs on the control plane node and collects job information from the slurm controller and pushes in the registration entries to spire. Hence the SSC request for another repo for that code.

As a taste though, this generates the list of running jobs and the node sets the job is running on:
> squeue --json | jq -r '.jobs[] | select(IN(.job_state[]; "RUNNING")) | .job_id as $id | .account as $account | .job_resources | select(IN(.select_type[]; "CORE")) | "\($account) \($id) \(.nodes.list)"'

Example output:
> project1 819721 t28
> project1 819752 t[24,27,35-36]
> project2 819846 t[53,101],ta18
> project3 819912 t77
> project1 820052 t86

doc/plugin_agent_workloadattestor_slurm.md
```diff
@@ -0,0 +1,40 @@
+# Agent plugin: WorkloadAttestor "slurm"
```
amartinezfayo, line 1: I would be good to also have this new plugin in the plugin table in the SPIRE Agent documentation and the full agent reference config.

doc/plugin_agent_workloadattestor_slurm.md
```diff
@@ -0,0 +1,40 @@
...
+
+General selectors:
+
+| Selector       | Value                                                                                   |
+| -------------- | --------------------------------------------------------------------------------------- |
+| `slurm:job_id` | Numeric Slurm job id when `CgroupJobIdPaths=yes` is set (e.g. `slurm:job_id:3385`).     |
+| `slurm:sluid`  | Slurm SLUID of the job; the default job identifier (e.g. `slurm:sluid:s5K1KKYAYG5D00`). |
+| `slurm:step`   | Job step: a number or `batch`/`extern`/`interactive` (e.g. `slurm:step:batch`).         |
+
+Exactly one of `slurm:job_id` or `slurm:sluid` is produced for a given workload, depending
```
amartinezfayo, line 26: I think this framing may only hold for Slurm 26.05 and later. Before 26.05 the cgroup/v2 layout always used job_<id> directories (that is what the suite's Slurm 23.11 produces), and SLUID naming + the CgroupJobIdPaths option were introduced in 26.05 (https://slurm.schedmd.com/cgroup.conf.html). I'm not sure what's the best way to handle this, would it make sense to qualify this by version so operators on older Slurm don't go looking for a knob that doesn't exist?

doc/plugin_agent_workloadattestor_slurm.md
```diff
@@ -0,0 +1,40 @@
...
+A sample configuration:
+
+```hcl
+    WorkloadAttestor "slurm" {
+    }
+```
+
+## Platform support
+
+This plugin is only supported on Unix systems and requires that Slurm is configured to use
```
amartinezfayo, line 39: It could be worth documenting the deployment requirement here as well. When the agent runs in a container it needs to share the host PID and cgroup namespaces (as the suite's compose file does), since (I think?) otherwise /proc/<pid>/cgroup won't show the host-rooted /system.slice path and Slurm workloads would attest with no selectors.

test/integration/suites/slurm-x509/00-setup
```diff
@@ -0,0 +1,82 @@
...
+cid=$(docker create spire-agent:latest-local)
+docker cp "${cid}:/opt/spire/bin/spire-agent" ./spire-agent
+docker rm "${cid}" >/dev/null
+sudo install -m 0755 ./spire-agent /usr/local/bin/spire-agent
+
+# --- Host-shared directories for the workload socket and job output --------
+sudo rm -rf /tmp/slurm-spire-sockets /tmp/slurm-spire-jobs
+sudo install -d -m 0777 /tmp/slurm-spire-sockets /tmp/slurm-spire-jobs
+
+# --- Install and configure a single-node Slurm cluster on the host ---------
```
amartinezfayo, line 23: Maybe we should guard this step against a machine that already runs Slurm?
It overwrites /etc/slurm/slurm.conf, installs a binary into /usr/local/bin, and restarts munge/slurmctld/slurmd, and the teardown stops those daemons and cancels all jobs, which could be destructive on a developer host with a real Slurm install. Failing early when slurmd or munge is already active (or unless running in CI) would keep that from happening by accident, what do you think?

amartinezfayo (commented):
Thank you @kfox1111 for this contribution!

test/integration/suites/slurm-x509/00-setup
kfox1111: Hmm, I was assuming no one would run the integration tests on their own boxes... Maybe this was a wrong assumption?

Should we do something like look for the GITHUB ACTIONS env vars and just deny the test if they are not in the environment?

doc/plugin_agent_workloadattestor_slurm.md
kfox1111: I took a stab at it.

doc/plugin_agent_workloadattestor_slurm.md
kfox1111: Updated.

doc/plugin_agent_workloadattestor_slurm.md
kfox1111: Done

test/integration/suites/slurm-x509/00-setup
amartinezfayo: Checking for GITHUB_ACTIONS and refusing to run otherwise sounds reasonable to me. I'll just note that this would diverge from how the other suites handle host dependencies, which is to run everything in containers. Since the regex requires the slurmstepd scope to live under the real host /system.slice, I think containerizing Slurm isn't really an option here, so gating on the CI environment seems like a fair trade-off. Maybe we could also accept an explicit opt-in (for example SLURM_SUITE_ALLOW_HOST_CHANGES=1) so the suite can still be run deliberately outside CI for debugging, say in a disposable VM.
It would be good to update the test suite README with these details.

conf/agent/agent_full.conf
```diff
@@ -669,6 +669,12 @@ plugins {
             # workload_size_limit = 0
         }
     }
+
+    # WorkloadAttestor "slurm": A workload attestor which allows selectors based
```
amartinezfayo: nit: the WorkloadAttestor entries here are in alphabetical order, so this one would go between k8s and systemd. The other Unix-only attestors also include a "Supported on Unix only." line in their description, which would be worth adding here too.

doc/plugin_agent_workloadattestor_slurm.md
amartinezfayo: Thanks for updating this. I think the job_id row now conflicts with itself though. CgroupJobIdPaths only exists on 26.05 and later, where setting it to yes emits numeric job ids, while before 26.05 numeric ids are emitted unconditionally, so the row shouldn't say the selector only applies to versions older than 26.05. Maybe something like "Numeric Slurm job id. Emitted on Slurm older than 26.05, or on 26.05 and later when CgroupJobIdPaths=yes". There's also a small typo in the sluid row (verison).

test/integration/suites/slurm-x509/00-setup
amartinezfayo: > Hmm, I was assuming no one would run the integration tests on their own boxes... Maybe this was a wrong assumption?

Our posture has been that the integration tests can be run on dev machines without needing CI infrastructure. `make integration` runs all the suites by default and supports running individual ones with SUITES.

doc/plugin_agent_workloadattestor_slurm.md
kfox1111: oh. yeah.

I just took another stab at it. How does this look?

test/integration/suites/slurm-x509/00-setup
kfox1111: ok. Maybe I tweak the job to skip the job but pass then?

kfox1111: Ok. I took a stab at fixing this too.

amartinezfayo (approved):
Thank you @kfox1111! 🚀

PR #7165, opened 2026-07-26
oidc-discovery-provider: soften allow_insecure_scheme warning wording

### LINKED ISSUE #7155: oidc-discovery-provider: allow_insecure_scheme WARN is misleading and unsuppressible behind a TLS-terminating load balancer
### Summary

Since #6968 / #6970, `oidc-discovery-provider` logs a WARN on every startup when `allow_insecure_scheme = true`, stating the setting "must only be used in development environments." 

Terminating TLS at a load balancer / reverse proxy and forwarding plaintext to the provider is a legitimate **production** topology, but there is no way to acknowledge that intent or downgrade the warning, and the pattern is undocumented for a TCP `insecure_addr` backend.

### Version

- SPIRE `oidc-discovery-provider` **1.15.2**

### Deployment

The provider runs behind an AWS ALB that terminates TLS and forwards plaintext to the pod:

- ALB listener: **HTTPS :443** with an ACM certificate
- `backend-protocol: HTTP`, `target-type: ip` → forwards to the pod's `:8080`
- Only the ALB→pod hop is plaintext; it is inside the VPC and carries only public JWKS / OIDC-discovery data

```hcl
allow_insecure_scheme = true
insecure_addr         = ":8080"
```

### Current behavior

On each process start the provider logs at WARN:

```
{"level":"warning","msg":"allow_insecure_scheme is enabled. JWKS keys will be served over HTTP. ..
...This setting must only be used in development environments."...}
```

- The message asserts the setting is dev-only.
- There is no config option to acknowledge intent, downgrade the level, or suppress it.

### Expected behavior

An operator running behind a trusted TLS-terminating proxy / load balancer can declare that intent so the provider does not emit a production WARN on every boot, while still leaving an audit trail that the insecure scheme is active. Or, a softer WARN message.

### Proposed solutions

Any one of these would resolve it:

1. **Acknowledgement flag.** Add a config field, e.g. `insecure_scheme_acknowledged = true` (or `behind_tls_terminating_proxy = true`). When set alongside
`allow_insecure_scheme`, log the same content once at INFO instead of WARN — preserves the audit signal, stops tripping WARN alerts, and requires an explicit opt-in.
2. **Soften the wording.** Change "This setting must only be used in development environments." to e.g. "Only enable this when the network path to the provider is trusted end-to-end (for example, TLS terminated at a trusted reverse proxy or load balancer on a private network)."

Happy to send a PR for based on the agreed approach.

### References

- #6968 — feature request that introduced the warning
- #6970 — implementation (`fixes #6968`)

pujitha24:
Motivation:
allow_insecure_scheme logs a WARN on every startup stating the setting
"must only be used in development environments." This is inaccurate:
terminating TLS at a load balancer or reverse proxy and forwarding
plaintext to the provider over a private network is a legitimate
production topology, not just a dev-only pattern. The wording gave
operators no way to reconcile a correct, trusted-network production
deployment with a message that implies they're doing something wrong.

Approach:
Reword the WARN log message (and the matching README footnotes for
allow_insecure_scheme) to describe the actual safety condition instead
of asserting an environment: only enable this when the network path to
the provider is trusted end-to-end (e.g. TLS terminated at a trusted
reverse proxy or load balancer on a private network). This matches the
wording a SPIRE maintainer (sorindumitru) explicitly agreed to on the
issue. The log stays at WARN level and still fires on every startup
when the setting is enabled; only the message text changes. This is a
wording/documentation clarity fix, not a behavior change.

Validation:
- go build ./support/oidc-discovery-provider/...
- go test ./support/oidc-discovery-provider/... -> ok

Fixes #7155

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/support/oidc-discovery-provider/README.md
+++ b/support/oidc-discovery-provider/README.md
@@ -66,13 +66,13 @@ The configuration file is **required** by the provider. It contains
 
 [1]: One of `acme`, `serving_cert_file` or `listen_socket_path` must be defined.
 
-[3]: The `allow_insecure_scheme` should only be enabled when the network path to the provider is trusted end-to-end (for example, TLS terminated at a trusted reverse proxy or load balancer on a private network). It only works in conjunction with `insecure_addr` or `listen_socket_path`.
+[3]: The `allow_insecure_scheme` should only be enabled when the network path to the provider is trusted end-to-end (for example, when TLS is terminated at a trusted reverse proxy or load balancer on a private network). It only works in conjunction with `insecure_addr` or `listen_socket_path`.
 
 #### Considerations for Windows platforms
 
 [1]: One of `acme`, `serving_cert_file` or `listen_named_pipe_name` must be defined.
 
-[3]: The `allow_insecure_scheme` should only be enabled when the network path to the provider is trusted end-to-end (for example, TLS terminated at a trusted reverse proxy or load balancer on a private network). It only works in conjunction with `insecure_addr` or `listen_named_pipe_name`.
+[3]: The `allow_insecure_scheme` should only be enabled when the network path to the provider is trusted end-to-end (for example, when TLS is terminated at a trusted reverse proxy or load balancer on a private network). It only works in conjunction with `insecure_addr` or `listen_named_pipe_name`.
 
 #### Considerations for all platforms
 
--- a/support/oidc-discovery-provider/README.md
+++ b/support/oidc-discovery-provider/README.md
@@ -66,13 +66,13 @@ The configuration file is **required** by the provider. It contains
 
 [1]: One of `acme`, `serving_cert_file` or `listen_socket_path` must be defined.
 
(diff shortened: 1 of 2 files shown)


sorindumitru:
Thanks @pujitha24, could you have a look over the copilot suggestion? After that this should be good to merge.

pujitha24:
Thanks for the nudge, @sorindumitru! I've addressed both Copilot nits:

- Dropped the trailing period from the `main.go` warning message, matching the convention of the other log messages in this file.
- Fixed the "for example, TLS terminated..." sentence fragment to "for example, when TLS is terminated..." in the warning and in both README footnotes (Unix and Windows sections).

Build and tests pass locally. Let me know if there's anything else you'd like adjusted.

sorindumitru (approved):
LGTM, thanks @pujitha24!

PR #7167, opened 2026-07-27
add tls config support for TLS Terminating endpoints

### LINKED ISSUE #7168: Enhancement: configurable TLS config (min TLS version, cipher suites, curves) for terminating endpoints
## Summary

SPIRE currently exposes `experimental.require_pq_kem`, which provides a mechanism to enforce post-quantum key exchange on selected TLS connections.

While this addresses one aspect of TLS hardening, there is currently no first-class configuration for broader TLS security settings such as:

- Minimum TLS version
- TLS 1.2 cipher suites
- TLS curve / group preferences

For compliance-oriented and security-hardened deployments, users often require explicit control over these parameters on TLS terminating endpoints.

## Motivation

Many a times users need a configurable TLS security config that allows administrators to define the TLS characteristics of inbound endpoints.

Providing similar functionality in SPIRE would enable operators to:

- enforce a minimum TLS version
- restrict TLS 1.2 cipher suites
- control preferred classical or hybrid ML-KEM key exchange groups
- expose a predictable and auditable TLS configuration for serving endpoints

This complements the existing `experimental.require_pq_kem` feature rather than replacing it.

## Proposed Direction

Introduce a `tls_config` HCL block for TLS terminating endpoints.

Example:

```hcl
tls_config {
    min_tls_version = "VersionTLS13"

    cipher_suites = [
        "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
    ]

    curve_preferences = [
        "X25519MLKEM768",
        "X25519",
        "secp256r1",
    ]
}
```

The existing `experimental.require_pq_kem` option should remain supported for backward compatibility. When both are configured, `experimental.require_pq_kem` would retain its current behavior by overriding the relevant TLS config settings where applicable.

## Scope

This proposal intentionally applies `tls_config` only to **TLS terminating endpoints**.

rationale being that the SPIRE components also initiate outbound TLS connections to systems whose TLS configuration is outside SPIRE's administrative control. Examples include:

- UpstreamAuthority "spire" plugin — nested SPIRE talks to a parent server; client-side config enforcement can break nested topologies unless explicitly coordinated on both sides.
- Federation bundle fetch client — remote federated servers can run their own listener configs; local client config enforcement does not replace the need for bilateral compatibility.

Applying the same TLS config to outbound clients could unnecessarily reduce interoperability with independently managed systems. Restricting the initial scope to TLS listeners provides explicit control over the security posture of SPIRE's exposed endpoints while preserving compatibility for outbound communication.

If future use cases require outbound TLS policy enforcement, client-side support can be introduced independently on a per-surface, opt-in basis where interoperability requirements are well understood.

nhegde07:
## Summary

Adds a configurable `tls_config` HCL block for TLS-terminating endpoints in SPIRE components. Operators can set minimum TLS version, TLS 1.2 cipher suites, and curve/group preferences beyond the existing `experimental.require_pq_kem` option.

`tls_config` is currently applied only to inbound TLS listeners. Outbound TLS clients are unchanged due to risk of  breaking interoperability with external systems.

When both `tls_config` and `experimental.require_pq_kem` are set, `require_pq_kem` takes precedence (curves and minimum TLS version).

## Components

| Component | Configuration | TLS surfaces |
|-----------|---------------|--------------|
| SPIRE Server | `server { tls_config { ... } }` | gRPC TCP API, Federation Bundle HTTPS endpoint, Prometheus HTTPS |
| SPIRE Agent | `agent { tls_config { ... } }` | Prometheus HTTPS |
| OIDC Discovery Provider | `tls_config { ... }` | HTTPS listeners (disk certificate and ACME modes) |

## Implementation

- Add `TLSConfig / ParsedTLSConfig` and extend Policy in `pkg/common/tlspolicy`.
- Parse and validate `tls_config` once at startup via `NewPolicy() / ParseTLSConfig()`.
- Extend `ApplyPolicy()` with `WithServerTLSConfig()` to opt in listener-only settings; `RequirePQKEM` is always applied when enabled and thus retaining backward compatibility.
- Wire `tls_config` into SPIRE Server, SPIRE Agent, and OIDC Discovery Provider configuration.
- Update docs, sample configs, and unit tests.

## Example

```hcl
server { # or agent
    tls_config {
        min_tls_version = "VersionTLS13"

        cipher_suites = [
            "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
            "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
        ]

        curve_preferences = [
            "X25519MLKEM768",
            "X25519",
            "secp256r1",
        ]
    }
}
```

Fixes #7168


nhegde07:
@sorindumitru 
@MarcosDY 
Requesting for review

sorindumitru:
Hi @nhegde07, Is this related to some issue? If not, it might be good to first open one so we can discuss about the need for it and the solution.

nhegde07:
@sorindumitru My bad. Here is the issue: https://github.com/spiffe/spire/issues/7168

nhegde07:
@MarcosDY / @sorindumitru 
Gentle followup on the PR review :pray: 
Looking forward to your feedback!

nhegde07:
@amartinezfayo  / @sorindumitru
Gentle followup on the PR review 🙏
Looking forward to your feedback!

nhegde07 (commented):
self reviewed

nhegde07:
@MarcosDY / @sorindumitru / @amartinezfayo / @evan2645 
Requesting you guys to take a look at the PR and issue being addressed! :pray: 

I have improvised the flow to read and parse the tls config once at start and read it!
Looking forward to your feedback/blessings!

pkg/common/tlspolicy/tlspolicy.go
```diff
@@ -24,25 +64,167 @@ func LogPolicy(policy Policy, logger hclog.Logger) {
...
+			return nil, fmt.Errorf("invalid minTLSVersion %q: %w", cfg.MinTLSVersion, err)
 		}
+		parsedTLSCfg.MinTLSVersion = minVersion
+	}
 
-		// Require TLS 1.3, as all PQ-safe KEMs require it anyway.
-		if config.MinVersion < tls.VersionTLS13 {
-			config.MinVersion = tls.VersionTLS13
+	if len(cfg.CipherSuites) > 0 {
+		cipherSuites, err := cliflag.TLSCipherSuites(cfg.CipherSuites)
```
MarcosDY: cliflag.TLSCipherSuites resolves against allCiphers(), which deliberately merges Go's tls.InsecureCipherSuites() ([component-base/cli/flag/ciphersuites_flag.go:74-84](https://github.com/kubernetes/component-base/blob/master/cli/flag/ciphersuites_flag.go#L75-L84)). 

So we can set:
```
TLS_RSA_WITH_RC4_128_SHA      → ids=[5]  err=<nil>
TLS_RSA_WITH_3DES_EDE_CBC_SHA → ids=[10] err=<nil>
```

k8s accepts these because `kube-apiserver` must support legacy clients, and surfaces them as "insecure" in flag help. 

The package exposes InsecureTLSCipherNames() precisely so callers can reject them.
Can you verify them? if not we must add a warning to prevent unexpected configuration.

pkg/common/tlspolicy/tlspolicy.go
```diff
@@ -24,25 +64,167 @@ func LogPolicy(policy Policy, logger hclog.Logger) {
...
-			tls.SecP384r1MLKEM1024,
+		applyRequirePQKEM(config)
+	}
+
+	return nil
+}
+
+func (p Policy) applyServerTLSConfig(cfg *tls.Config) {
+	if p.TLSCfg != nil && p.TLSCfg.MinTLSVersion != 0 {
+		cfg.MinVersion = p.TLSCfg.MinTLSVersion
```
MarcosDY, line 117: this can allow `TLS 1.0/1.1` is this expected? may we avoid those versions? SPIRE has multiple places where we have documented 1.2 as minimal

support/oidc-discovery-provider/main.go
```diff
@@ -255,7 +260,10 @@ func newACMEListener(log logrus.FieldLogger, config *Config) (net.Listener, erro
 	}
 
 	tlsConfig := m.TLSConfig()
-	tlsConfig.MinVersion = tls.VersionTLS12
+	if err := applyTLSPolicy(tlsConfig, config.TLSConfig); err != nil {
```
MarcosDY: same comment as before about minimal version.
As long i know server is going to just reject any value less than 1.2 (except we use ` GODEBUG=tls10server=1`) if htat is the case users may be waiting to have a TLS version that is not being set.

pkg/common/tlspolicy/tlspolicy.go
```diff
@@ -24,25 +64,167 @@ func LogPolicy(policy Policy, logger hclog.Logger) {
...
+		cipherSuites, err := cliflag.TLSCipherSuites(cfg.CipherSuites)
+		if err != nil {
+			return nil, fmt.Errorf("invalid cipherSuites: %w", err)
 		}
+		parsedTLSCfg.CipherSuites = cipherSuites
 	}
 
-	return nil
+	if len(cfg.CurvePreferences) > 0 {
+		curves, err := parseCurvePreferences(cfg.CurvePreferences)
```
MarcosDY: parseCurvePreferences accepts arbitrary raw decimal curve IDs with no validation against curves Go actually supports, so a typo'd or unsupported ID passes startup and breaks the handshake at connection time.

pkg/common/tlspolicy/tlspolicy.go
```diff
@@ -24,25 +64,167 @@ func LogPolicy(policy Policy, logger hclog.Logger) {
 	if policy.RequirePQKEM {
 		logger.Debug("Experimental option 'require_pq_kem' is enabled; all TLS connections will require use of a post-quantum safe KEM")
 	}
+	if policy.TLSCfg == nil {
+		logger.Debug("No TLS config configured")
+		return
+	}
+	if policy.TLSCfg.MinTLSVersion != 0 {
+		logger.Debug("TLS config min TLS version configured", "min_tls_version", policy.TLSCfg.MinTLSVersion)
```
MarcosDY: MinTLSVersion is the raw version `MinTLSVersion    uint16`, can you update it to the configured name? (easier to read from logs)

support/oidc-discovery-provider/main.go
```diff
@@ -255,7 +260,10 @@ func newACMEListener(log logrus.FieldLogger, config *Config) (net.Listener, erro
...
+		return nil, err
+	}
+	log.Infoln("TLS Policy applied", config.TLSConfig)
```
MarcosDY: this is going to print a pointer, can you plz display values instead?

support/oidc-discovery-provider/config.go
```diff
@@ -352,6 +357,14 @@ func ParseConfig(hclConfig string) (_ *Config, err error) {
 	if c.JWKSURI == "" && c.JWTIssuer != "" {
 		fmt.Printf("Warning: The jwt_issuer configuration will also affect the jwks_uri behavior when jwks_url is not set. This behaviour will be changed in 1.13.0.")
 	}
+
+	if c.TLSConfig != nil {
+		_, err = tlspolicy.ParseTLSConfig(c.TLSConfig)
```
MarcosDY: this is parsing twice right? on startup and here

pkg/agent/broker/endpoints_test.go
```diff
@@ -2,19 +2,81 @@ package broker
...
+	svid *x509svid.SVID
+}
+
+func (s staticSVIDSource) GetX509SVID() (*x509svid.SVID, error) {
+	return s.svid, nil
+}
+
+// brokerListenerTLSConfig mirrors the TLS setup in Endpoints.ListenAndServe so
+// profile application can be tested without duplicating production helpers.
+func brokerListenerTLSConfig(svidSource x509svid.Source, bundleSource x509bundle.Source, brokerIDs []spiffeid.ID, policy tlspolicy.Policy) (*tls.Config, error) {
```
MarcosDY: brokerListenerTLSConfig copies the production TLS setup into the test, so the test cannot catch a regression in the production call site.

MarcosDY (commented):
Looks great! 
I added some questions and recommendations about validations to be done

nhegde07 (commented):
self reviewed.
@MarcosDY I have addressed all of your feedback & also I have increased unit tests to improve the validation coverage.
Please review the PR! 🙏

pkg/common/tlspolicy/tlspolicy.go
```diff
@@ -4,17 +4,58 @@ package tlspolicy
 
 import (
 	"crypto/tls"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
 
 	"github.com/hashicorp/go-hclog"
+	cliflag "k8s.io/component-base/cli/flag"
```
sorindumitru: Looks like this module does not provide and [compatibility guarantees](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/component-base#compatibility). Is there something else that we can use for this?

nhegde07: Thanks for raising this, @sorindumitru .

In this PR we only use k8s.io/component-base/cli/flag as a helper for parsing TLS-related strings:

- TLSVersion() for min_tls_version names (e.g. VersionTLS12, VersionTLS13)
- TLSCipherSuites() and InsecureTLSCiphers() for cipher suite name resolution and insecure-suite filtering
- TLSCurvePreferences() for validating curve IDs against what Go’s crypto/tls supports
We are not using component-base for anything beyond that. No flags, metrics, or other APIs.

That said, I agree we shouldn’t take a dependency on an unstable package for a small amount of parsing logic. We have two reasonable paths:

- Keep it - accept the dependency, pin it to the same k8s minor version SPIRE already uses, and treat any upstream changes as part of our regular k8s dependency bumps; or
- Replace it - implement equivalent helpers in SPIRE (using crypto/tls where possible) and drop the component-base import.

I’m happy to go with whichever the maintainers prefer. If you’d rather avoid the unstable dependency, I can follow up with a commit that inlines these helpers so the change should be fairly small given our existing tests.

Let me know your preference and I’ll update the PR accordingly.

nhegde07:
@sorindumitru / @MarcosDY gentle request for PR review!

nhegde07:
Requesting for PR review @MarcosDY / @sorindumitru 
I have addressed all of your feedback.
Unit tests and docs are upto date as well.

MarcosDY (approved):
LGTM!
Sorry about delay

nhegde07:
thanks a lot @MarcosDY 
Sorry if I bothered you :pray:

PR #7169, opened 2026-07-28
Retry NewJWTSVID with backoff on transient errors

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Fetching JWT-SVIDs

**Description of change**
NewJWTSVID previously failed on the first gRPC error. Transient server or network hiccups therefore failed a workload's JWT-SVID request even though a quick retry would have succeeded.

Retry the NewJWTSVID RPC with an exponential backoff, bounded by the existing per-call timeout (which now covers the whole retry loop).>


pkg/agent/client/client.go
```diff
@@ -351,20 +361,74 @@ func (c *client) NewX509SVIDs(ctx context.Context, csrs map[string][]byte) (map[
...
+			IssuedAt:  time.Unix(svid.IssuedAt, 0).UTC(),
+			ExpiresAt: time.Unix(svid.ExpiresAt, 0).UTC(),
+		}, spiffeId, nil
+	}
+}
+
+// fetchJWTSVID performs a single NewJWTSVID RPC. Any returned error is the raw
+// error from the connection or gRPC call so the caller can classify it for
+// retries.
+func (c *client) fetchJWTSVID(ctx context.Context, entryID string, audience []string) (*types.JWTSVID, error) {
```
MarcosDY, line 429: fetchJWTSVID keeps the pre-existing `c.release(connection)` on the RPC error path, but that path is now executed once per retry attempt instead of once per NewJWTSVID call:

```
resp, err := svidClient.NewJWTSVID(ctx, ...)
if err != nil {
    c.release(connection)
    return nil, err
}
```

release isn't per-call cleanup, it tears down the client's single shared connection:

```
func (c *client) release(conn *nodeConn) {
      c.m.Lock()
      defer c.m.Unlock()
      if c.connections != nil && (conn == nil || conn == c.connections) {
              c.connections.Release()
              c.connections = nil
      }
}
```

That c.connections is the same connection used by FetchUpdates, SyncUpdates, RenewSVID, NewX509SVIDs and PostStatus. So during an outage, each retry attempt nils it out and the next attempt's 

Possible solutoin: drop the c.release(connection) from the per-attempt path and let the retry loop decide. 

Either release once when the loop finally gives up, or only release for the codes where a stale connection is plausibly the cause (Unavailable) rather than on every error. The defer connection.Release() already handles the per-attempt refcount correctly.

pkg/agent/client/client.go
```diff
@@ -351,20 +361,74 @@ func (c *client) NewX509SVIDs(ctx context.Context, csrs map[string][]byte) (map[
...
+	fetchBackoff := backoff.NewBackoff(c.clk, jwtSVIDRetryInterval)
+	for {
+		svid, err := c.fetchJWTSVID(ctx, entryID, audience)
+		if err != nil {
+			if !isRetriable(err) {
+				c.withErrorFields(err).Error("Failed to fetch JWT SVID")
+				return nil, spiffeid.ID{}, fmt.Errorf("failed to fetch JWT SVID: %w", err)
+			}
+
+			c.withErrorFields(err).Debug("Failed to fetch JWT SVID; retrying")
```
MarcosDY, line 390: I'm not sure about set this as Debug, users are going to lose context about why they are not getitng errors, 
what do you think?

MarcosDY (dismissed):

MarcosDY (approved):

PR #7177, opened 2026-07-30
Add exclude_completed_pods option to k8s workload attestor

### LINKED ISSUE #7161: SPIRE Agent Memory Balloons With ContainerStatusUnknown Pods
We've recently encountered an issue where the SPIRE agent, when using the k8s workloadattestor, dramatically increases in memory utilization when there are pods that have failed but are not yet garbage collected on the node.  This problem can be papered over by increasing memory limits, and fixed by manually deleting the lost pods.  It appears that this is being caused by the pod cache in the plugin.  I'd like to propose that instead of caching all pods, a check be performed on each pod, and any which are in `status.phase: Failed` be discarded to prevent the waste of memory.  For the sake of not breaking existing cases, this could be put behind a flag on the plugin.

Happy to open a PR with a proposed change for this if that'd be acceptable.

isugimpy:
Signed-off-by: Jayme Howard <jayme@reddit.com>

**Pull Request check list**

- [X] Documentation updated?

**Affected functionality**
Agent k8s WorkloadAttestor optionally excludes completed pods from its cache, defaulted off.

**Description of change**
Pods in a terminal status.phase ("Failed" or "Succeeded") are always stored in the kubelet pod-list cache, even though they have no running containers and can never be the target of workload attestation. On nodes that accumulate large numbers of terminated pods (e.g. hundreds stuck in ContainerStatusUnknown, or completed jobs), this balloons the agent's memory footprint.

Add an opt-in exclude_completed_pods flag (default false) that drops Failed and Succeeded pods before they enter the cache, bounding memory usage while leaving attestation behavior unchanged for running pods.

**Which issue this PR fixes**
fixes #7161


sp3nx0r:
bumping this as we (Reddit) have had a few incidents due to unschedulable pods causing OOM bloat in spire server. we're talking about forking this internally to resolve, would love to not have to do that

sorindumitru:
Sorry for not following up here, @isugimpy. Any reason for making this opt-in? I think it could be good to have the flag for now, just in case it causes issues, but it seems like we could default it to opt-out. The two states are terminal and nothing should be running as the pod in those states.

isugimpy:
@sorindumitru Just as an abundance of caution, as technically this is a breaking change if anyone is somehow relying on this behavior.  I'm certainly open to changing it to opt-out if that's what is preferred by maintainers.  Unfortunately, I won't have access to my computer to be able to push up the change for a few days.

isugimpy:
@sorindumitru Finally back to a computer, so I've switched this over to opt-out as requested.  I did need one additional commit to fix merge conflicts, because there was a significant rewrite merged between when I opened the PR and today.  This should be ready for review.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -361,6 +371,7 @@ func (p *Plugin) buildConfig(coreConfig catalog.CoreConfig, hclText string, stat
 			kubeletCAPath:              kubeletCAPath,
 			nodeName:                   nodeName,
 			reloadInterval:             reloadInterval,
+			excludeCompletedPods:       newConfig.ExcludeCompletedPods == nil || *newConfig.ExcludeCompletedPods,
```
MarcosDY: this is changing defaults.
I think it should it be:

```  newConfig.ExcludeCompletedPods != nil && *newConfig.ExcludeCompletedPods?```

so only exclude when true

File: pkg/agent/plugin/workloadattestor/k8s/podlistfetcher.go [hunk 2]
```diff
@@ -373,6 +379,20 @@ func (f *podListFetcher) parsePodList(podListBytes []byte) (map[string]*fastjson
...
 		}
+
+		// Pods in a terminal phase (Failed or Succeeded) have no running
+		// containers and can never host a workload being attested. Failed pods
+		// include evicted pods and pods left in ContainerStatusUnknown after a
+		// node problem; Succeeded pods include completed jobs. Optionally drop
+		// them so they do not balloon the cache on nodes with many terminated
+		// pods.
+		if excludeCompletedPods {
+			phase := string(podValue.Get("status", "phase").GetStringBytes())
```
MarcosDY, line 390: anyone who opts in: the filter drops terminal pods during parse, so `broker pod references` stop resolving them too. 
Under pod_reference_scope = "agent_node" there's no API-server fallback, so a reference to a Job pod that just hit Succeeded fails with pod ... not found on agent node.

Could we either apply the phase filter at the PID-attestation consumer instead of in parsePodList (so reference lookups still see terminal pods), or document that enabling this makes terminal pods unresolvable under `agent_node` and point operators at `pod_reference_scope = "cluster"`?

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -165,6 +165,16 @@ type HCLConfig struct {
...
 
+	// ExcludeCompletedPods, when true, causes pods that have reached a
+	// terminal status.phase ("Failed" or "Succeeded") to be dropped from the
+	// kubelet pod-list cache instead of being tracked. Such pods (e.g. evicted
+	// pods, pods left in ContainerStatusUnknown after a node problem, or
+	// completed jobs) have no running containers and are never the target of
+	// workload attestation, so excluding them reduces the plugin's memory
+	// footprint on nodes that accumulate large numbers of terminated pods.
+	// Defaults to true if unset.
+	ExcludeCompletedPods *bool `hcl:"exclude_completed_pods"`
```
MarcosDY: `disable_container_selectors` exists to fall back to pod selectors when the container isn't visible. 
Since the filter drops terminal pods during parse, that fallback can't fire for them (they're gone before the loop sees them). 
Is there a window where an evicted or ContainerStatusUnknown pod is marked Failed while its workload still calls the Workload API? If so, these two flags conflict and it's worth a note in the docs.

doc/plugin_agent_workloadattestor_k8s.md
```diff
@@ -53,6 +53,7 @@ since [hostprocess](https://kubernetes.io/docs/tasks/configure-pod-container/cre
...
 | `disable_kubelet_client`                | If true, disables kubelet client setup and kubelet pod-list calls. PID-based workload attestation and `pod_reference_scope = "agent_node"` pod references are unavailable. Mutually exclusive with kubelet client settings.                 |
 | `enable_namespace_labels`               | If true, namespace labels are fetched from the Kubernetes API server and produced as selectors. Requires RBAC `get`, `list` and `watch` on `namespaces`. Disabled by default.                                                               |
+| `exclude_completed_pods`                | If true, pods in a terminal `status.phase` of `Failed` (e.g. evicted pods or pods left in `ContainerStatusUnknown` after a node problem) or `Succeeded` (e.g. completed jobs) are dropped from the kubelet pod-list cache instead of being tracked. Reduces memory usage on nodes with many terminated pods. Enabled by default. Set to `false` to track terminal-phase pods. |
```
MarcosDY: This contradicts what the PR says, are we supposed to exclude by default or only when true is set?
If it is on by default, that's a behavior change for existing deployments, we may call it "include_completed_pods"

pkg/agent/plugin/workloadattestor/k8s/k8s.go
isugimpy: @sorindumitru asked me to make the default true above, which is what this is intended to reflect.  I'm fine with either, happy to make whatever change here.

pkg/agent/plugin/workloadattestor/k8s/podlistfetcher.go
isugimpy: I'd prefer if we could do the documentation.  Moving the filter, I think, would defeat the goal here of preventing the memory balloon.

doc/plugin_agent_workloadattestor_k8s.md
isugimpy: Whoops, missed updating this when I switched it from opt-in to opt-out.  I'm reverting to opt-in anyway based on your earlier feedback though.  That's what I originally wanted.

isugimpy:
Added the documentation caveat around enabling `exclude_completed_pods`, as well as returning this to being opt-in.

Truly apologize for the force push, but I don't think I had another good option.  I missed signing off on my previous commit, and after pulling in the head of master as a merge commit I didn't have a good way to amend the commit to fix it.  If necessary, I could create a new branch and open a new PR to start fresh if that's what maintainers would prefer.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
sorindumitru: Yes, sorry about the back-and-forth, I'm ok with this. Probably best to not break anything for users, even if next release is a minor release.

pkg/agent/plugin/workloadattestor/k8s/podlistfetcher.go
```diff
@@ -373,6 +379,20 @@ func (f *podListFetcher) parsePodList(podListBytes []byte) (map[string]*fastjson
...
+			if phase == string(corev1.PodFailed) || phase == string(corev1.PodSucceeded) {
+				continue
+			}
```
MarcosDY, line 393: Thanks for this!!! 
I tested it, and measure it and I want to display what i found: your diagnosis is awesome and the shape of the PR is right. 
Terminated pods do accumulate, they shouldn't be cached, and a config flag is the right way to expose that. Two things need adjusting, and one of them is a non-obvious property of fastjson that nothing in the codebase documents... I'd have missed it too :( 

cached pods                  - retained    -     heap 
main today (no filter)          2000        47908 KiB
your filter alone                        20        47771 KiB     
arena fix alone                     2000       29363 KiB  (I'll explain it later)
both together                           20           300 KiB 

(node with 20 live pods among 1980 reaped terminal pods, 2628 KiB kubelet response)

Your filter is doing most of the work in the final row, it just can't cash in until the `arena issue` it is fixed. The two changes are complementary, not alternatives. Plus one predicate correction so live workloads aren't dropped.

**Issue 1**: the cached pods all alias one parser, so exclusion frees nothing
podlistfetcher.go:389,  the continue removes a map entry but frees none of the pod's JSON. Every value in the map points into the single fastjson.Parser at podlistfetcher.go:352, which backs all its values with two allocations:
- Parser.b — a full working copy of the response (parser.go:32, p.b = append(p.b[:0], s...))
- cache.vs — one contiguous []Value holding every parsed node (parser.go:67-74)

Possible fix:
```
Fix

      result := make(map[string]*fastjson.Value, len(items))
      var scratch []byte

      for _, podValue := range items {
              uid := string(podValue.Get("metadata", "uid").GetStringBytes())
              if uid == "" {
                      f.log.Warn("Pod has no UID", "pod", podValue)
                      continue
              }

              // Terminated pods whose containers have already been reaped can never
              // be attested again.
              if excludeCompletedPods && podIsUnattestable(podValue) {
                      continue
              }

              // Values from a single parser all alias that parser's working copy of
              // the response and its shared value slice, so retaining any pod keeps
              // the whole response alive. Re-parse each pod into its own parser so a
              // cached pod costs only its own JSON.
              scratch = podValue.MarshalTo(scratch[:0])
              podParser := new(fastjson.Parser)
              compactPod, err := podParser.ParseBytes(scratch)
              if err != nil {
                      f.log.Warn("Unable to re-parse pod from kubelet response", "pod_uid", uid, "error", err)
                      continue
              }

              result[uid] = compactPod
      }
```

**Issue 2**: status.phase alone drops pods with running containers
Your own fixture demonstrates this. In testdata/pod_list_failed.json the attestation target blog-24ck7 is:

phase: Failed
  conditions: Ready=True
  containerStatuses:
    blog:       state=running, ready=true
    ghostunnel: state=running, ready=true

A Failed pod with two running, ready containers. Attestation never consults phase — k8s.go:723 matches solely on container ID via lookUpContainerInPod (k8s.go:1506). 
So TestAttestExcludeCompletedPodsFailedWhenEnabled currently asserts that a workload with live containers fails with DeadlineExceeded, which is the regression written down as an expectation.

Fix: two conditions, not one
```
func podIsUnattestable(podValue *fastjson.Value) bool {
      phase := string(podValue.Get("status", "phase").GetStringBytes())
      if phase != string(corev1.PodFailed) && phase != string(corev1.PodSucceeded) {
              return false
      }

      // Mirror the status lists lookUpContainerInPod searches. Ephemeral
      // container statuses are deliberately excluded, since attestation never
      // matches against them.
      for _, statusField := range [...]string{"containerStatuses", "initContainerStatuses"} {
              for _, containerStatus := range podValue.GetArray("status", statusField) {
                      if len(containerStatus.GetStringBytes("containerID")) > 0 {
                              return false
                      }
              }
      }

      return true
}
```

Here is a test i used to validate all of this:
```
func buildLargePodListResponse(t *testing.T, podCount int) []byte {
      t.Helper()

      items := make([]map[string]any, 0, podCount)
      for i := range podCount {
              containerStatuses := make([]map[string]any, 0, 4)
              for c := range 4 {
                      containerStatuses = append(containerStatuses, map[string]any{
                              "name":        fmt.Sprintf("container-%d", c),
                              "containerID": fmt.Sprintf("containerd://%064d", i*10+c),
                              "image":       "registry.example.com/some/fairly/long/image/path:v1.2.3",
                              "imageID":     fmt.Sprintf("sha256:%064d", i),
                              "ready":       true,
                              "state":       map[string]any{"running": map[string]any{"startedAt": "2026-01-01T00:00:00Z"}},
                      })
              }
              items = append(items, map[string]any{
                      "metadata": map[string]any{
                              "uid":       fmt.Sprintf("pod-uid-%d", i),
                              "name":      fmt.Sprintf("some-workload-deployment-abcdef-%05d", i),
                              "namespace": "production",
                              "labels": map[string]any{
                                      "app": "some-workload", "pod-template-hash": "abcdef", "team": "platform",
                              },
                              "annotations": map[string]any{
                                      "kubectl.kubernetes.io/last-applied-configuration": `{"apiVersion":"v1","kind":"Pod"}`,
                              },
                      },
                      "spec": map[string]any{"nodeName": "k8s-node-1", "serviceAccountName": "default"},
                      "status": map[string]any{
                              "phase": "Running", "hostIP": "10.0.0.1", "podIP": "10.1.2.3",
                              "containerStatuses": containerStatuses,
                      },
              })
      }

      response, err := json.Marshal(map[string]any{"items": items})
      require.NoError(t, err)
      return response
}

func liveHeapBytes() uint64 {
      runtime.GC()
      runtime.GC()
      var stats runtime.MemStats
      runtime.ReadMemStats(&stats)
      return stats.HeapAlloc
}

func TestParsePodListDoesNotRetainWholeResponse(t *testing.T) {
      const podCount = 2000

      fetcher := podListFetcher{log: hclog.NewNullLogger()}
      response := buildLargePodListResponse(t, podCount)

      before := liveHeapBytes()
      pods, err := fetcher.parsePodList(response, false)
      require.NoError(t, err)
      require.Len(t, pods, podCount)

      // Retain a single pod and drop the rest, as a cache holding only live pods
      // on a node full of terminated ones would.
      retained := pods["pod-uid-0"]
      require.NotNil(t, retained)
      pods = nil

      after := liveHeapBytes()
      runtime.KeepAlive(retained)
      runtime.KeepAlive(response)

      // One pod is a few KiB. Allow generous headroom for allocator noise while
      // still failing loudly if the whole multi-MiB response is pinned.
      const maxRetainedBytes = 256 * 1024
      retainedBytes := int64(after) - int64(before)
      t.Logf("response %d KiB across %d pods; retaining 1 pod holds %d KiB live",
              len(response)/1024, podCount, retainedBytes/1024)
      require.Less(t, retainedBytes, int64(maxRetainedBytes),
              "retaining one pod pinned %d KiB; the parser arena is being held alive by the cached pod",
              retainedBytes/1024)

      // The retained pod must still be fully usable after being decoupled.
      require.Equal(t, "some-workload-deployment-abcdef-00000",
              string(retained.Get("metadata", "name").GetStringBytes()))
      require.Len(t, retained.GetArray("status", "containerStatuses"), 4)
      require.NoError(t, json.Unmarshal(retained.MarshalTo(nil), &map[string]any{}))
}
```

MarcosDY (commented):
Looks great, but i found an issue on fastjson :(, can you plz verify that? and resolve the issues on MD? 

Thank you very much this change is going to be veeeeeeery useful!!!!

pkg/agent/plugin/workloadattestor/k8s/podlistfetcher.go
isugimpy: Think I've got this all addressed now.

MarcosDY (approved):

MarcosDY:
@isugimpy can you resolve the lint issue?

PR #7178, opened 2026-07-31
doc/telemetry: drop empty targets from the two status piecharts

### LINKED ISSUE #6918: doc/telemetry: empty-expr refId-B targets cause Grafana parse errors on dashboard render
## Summary

The maintained dashboard at `doc/telemetry/spire_grafana_dashboard.json` contains two panels with malformed Prometheus targets that trigger Grafana query parse errors at render time:

- Line ~570: "SPIRE Agent: GET JWT status" panel
- Line ~660: "SPIRE Agent: GET x509 status" panel

Both panels include a second Prometheus target (`refId: B`) with:

```json
"expr": "",
"hide": false
```

Grafana attempts to evaluate the empty PromQL expression and surfaces a parse error on the panel, breaking visualization.

## Reproduction

Import `doc/telemetry/spire_grafana_dashboard.json` (any commit through master HEAD) into Grafana with a Prometheus datasource and observe the parse error on the two listed panels.

## Suggested fix

The simplest non-invasive fix is to flip `"hide": false` → `"hide": true` on those two refId-B targets. The structure is preserved for future PromQL backfill (presumably the original intent), and the panels render their refId-A target cleanly without an evaluation error.

## Context

Discovered while vendoring the dashboard JSON downstream at `bodesystems/platform` (PR [#227](https://github.com/bodesystems/platform/pull/227)) — caught by Qodo's panel-correctness lint. We've patched locally to `hide: true` and will revert to byte-identity once a fix lands here.

Happy to open a small PR if maintainers prefer.

arpitjain099:
Fixes #6918.

`doc/telemetry/spire_grafana_dashboard.json` has two piechart panels, `SPIRE Agent: GET JWT status` and `SPIRE Agent: GET x509 status`, that each carry a second target with `"expr": ""` under `refId: B`. Grafana still issues that target, Prometheus rejects an empty query, and the panel renders a query error next to the series it did manage to fetch.

Nothing references `refId: B` in either panel: no transformation, no override, no legend entry. It looks like a leftover from editing the panels in the Grafana UI.

The other empty `expr` values in the file are on `row` panels, which do not query anything, so I left those alone. After the change the only remaining empty exprs in the dashboard are on rows.

Checked with a small script over the JSON before and after: two non-row panels with an empty target before, none after, and the file still parses.

### DIFF
--- a/doc/telemetry/spire_grafana_dashboard.json
+++ b/doc/telemetry/spire_grafana_dashboard.json
@@ -555,18 +555,6 @@
               "interval": "",
               "legendFormat": "",
               "refId": "A"
-            },
-            {
-              "datasource": {
-                "type": "prometheus",
-                "uid": "${DS_PROMETHEUS}"
-              },
-              "exemplar": true,
-              "expr": "",
-              "hide": false,
-              "interval": "",
-              "legendFormat": "",
-              "refId": "B"
             }
           ],
           "title": "SPIRE Agent: GET JWT status",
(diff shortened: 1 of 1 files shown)


sorindumitru (approved):
LGTM, thanks @arpitjain099!

PR #7185, opened 2026-08-02
ci: fix nightly AWS CA certs push and limit to main repo

### LINKED ISSUE #7184: Nightly `update-aws-ca-certs` job has been failing since #6917 landed
* **Version**: `main` @ c9c0e1307
* **Platform**: CI runner `ubuntu-22.04` (reproduced locally on Darwin 25.5.0 arm64, git 2.54.0)
* **Subsystem**: CI (`.github/workflows/nightly_build.yaml`), affects `nodeattestor/awsiid`

## What happens

The `update-aws-ca-certs` job in the nightly workflow has failed 23 nights in a row. It succeeded for the first several runs after #6917 landed, then started failing and has not recovered.

The failure is at the push step:

```
Switched to a new branch 'bot/update-aws-ca-certs'
[bot/update-aws-ca-certs 10c054c] chore: update AWS EC2 CA certificates
 2 files changed, 1469 insertions(+), 1024 deletions(-)
To https://github.com/spiffe/spire
 ! [rejected]  bot/update-aws-ca-certs -> bot/update-aws-ca-certs (stale info)
error: failed to push some refs to 'https://github.com/spiffe/spire'
```

Because nightly failures do not gate PRs, this has been quiet.

## Why

`git push --force-with-lease` with no arguments needs a remote tracking ref to compare against. The checkout step uses the default settings, so only `main` is fetched and `refs/remotes/origin/bot/update-aws-ca-certs` never exists locally. The remote branch does exist, created by the one run that succeeded, so the push is not a fast forward and git refuses it with `stale info`.

That also explains the pattern. The very first run had no remote branch to protect, so the push created it and passed. Every run since has had a branch to overwrite and no lease to do it with.

## Reproducing it

Outside of CI, in a throwaway repo:

```bash
git init --bare remote.git
git clone remote.git seed && cd seed
echo v1 > f.txt && git add . && git commit -m init && git push origin HEAD:main
git checkout -b bot/x && echo v2 > f.txt && git commit -am r1
git push --force-with-lease origin bot/x     # succeeds, branch did not exist

cd .. && git clone --single-branch --branch main remote.git run2 && cd run2
git checkout -b bot/x && echo v3 > f.txt && git commit -am r2
git push --force-with-lease origin bot/x     # ! [rejected] ... (stale info)
```

## Fix

Fetching all branches at checkout gives the lease something to compare against:

```yaml
      - name: Checkout
        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
        with:
          fetch-depth: 0
```

The push line then works unchanged. I tested the alternatives too, in case a lighter option was preferable:

| checkout | push | result |
|---|---|---|
| full | `--force-with-lease` | ok |
| full | `--force-with-lease=bot/update-aws-ca-certs` | ok |
| single branch | `--force-with-lease=bot/update-aws-ca-certs` | rejected, `stale info` |

Naming the ref on its own is not enough. Git only accepts a tracking ref that the configured fetch refspec covers, and a single branch clone does not cover it. Adding the refspec with `git config --add remote.origin.fetch` and then fetching also works, if fetching full history on the nightly is not wanted.

## Why it is worth fixing rather than leaving

The certificates really are drifting. I ran the parser on a clean checkout of `main` and got the same diff CI reports, 1469 insertions and 1024 deletions across the two `cacerts.go` files. Running it a second time produced byte identical output, so the generator is deterministic and that diff is genuine drift from what AWS publishes now, not churn.

#7138 is still open from the single run that worked, so its contents are from 2026-07-09 and are now stale themselves.

Happy to send the one line workflow change if you want it, or leave it to you since you were working on this. Either is fine, just say which.

samuraixwandering:
## Summary

Fixes the nightly `update-aws-ca-certs` job that has been failing with `git push --force-with-lease` rejected as `stale info`.

1. **`fetch-depth: 0`** on the job checkout so the remote tracking ref for `bot/update-aws-ca-certs` exists and `--force-with-lease` can compare against it (default single-branch checkout only has `main`).
2. **`if: github.repository == 'spiffe/spire'`** so the job does not run on forks.

As discussed on #7184.

## Test plan

- [x] Diff limited to `.github/workflows/nightly_build.yaml`
- [ ] After merge: nightly (or `workflow_dispatch`) succeeds at the push step when certs drift
- [ ] On a fork, `update-aws-ca-certs` is skipped

Fixes #7184

### DIFF
--- a/.github/workflows/nightly_build.yaml
+++ b/.github/workflows/nightly_build.yaml
@@ -40,6 +40,7 @@ jobs:
         run: ./.github/workflows/scripts/push-images.sh nightly
 
   update-aws-ca-certs:
+    if: github.repository == 'spiffe/spire'
     runs-on: ubuntu-22.04
 
     permissions:
@@ -49,6 +50,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
       - name: Setup go
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:


sorindumitru (approved):
LGTM, thanks @samuraixwandering

PR #7187, opened 2026-08-03
Simplify k8s workload attestor broker configuration

sorindumitru:
We can apply some sensible defaults to the values here so that users do not have to specify brokers twice in a lot of cases.

### DIFF
--- a/pkg/agent/plugin/workloadattestor/k8s/k8s.go
+++ b/pkg/agent/plugin/workloadattestor/k8s/k8s.go
@@ -812,7 +812,7 @@ func (p *Plugin) checkBrokerImpersonationForReference(ctx context.Context, confi
 	if brokerEntry == nil {
 		return nil
 	}
-	if config.Broker == nil || config.Broker.AccessPolicy != brokerAccessPolicyEnforced {
+	if config.Broker.AccessPolicy != brokerAccessPolicyEnforced {
 		return nil
 	}
 	objRef := result.ObjectReference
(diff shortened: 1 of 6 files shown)


pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -854,16 +855,32 @@ func (p *Plugin) getBrokerEntryIfPresent(ctx context.Context, config *k8sConfig)
...
 	}
-	brokerEntry, ok := config.Broker.Brokers[callerID.String()]
-	if !ok {
-		return nil, status.Errorf(codes.PermissionDenied, "broker %q is not configured", callerID.String())
+
+	// Synthesize the default entry: node scope, which limits the broker to
+	// PID and pod references. Node scope needs the kubelet client, so fall
+	// back to cluster scope when it is disabled.
+	scope := podReferenceScopeAgentNode
+	if config.DisableKubeletClient {
```
MarcosDY, line 874: If config.Broker == nil -> Internal: broker configuration missing guard, so an agent with no experimental.broker block now accepts any broker the endpoint forwards instead of rejecting it. 
Two consequences (that i got at least):

1. No SubjectAccessReview is performed. checkBrokerImpersonationForReference early-returns when config.Broker == nil (k8s.go:820), and the config-time access_policy requirement only fires when the block is present (buildBrokerAccessPolicy). So the unconfigured path skips authorization at both ends.

2. With disable_kubelet_client = true, the synthesized default is cluster scope (k8s.go:877) — the widest setting, combined with the missing SAR from '1'.

so,  on a broker-only agent with disable_kubelet_client = true and no broker block, any broker reaching the endpoint can resolve any pod on any node, plus non-pod cluster-scoped objects (Deployments, Kustomizations, CRDs), and receive SPIFFE selectors for them.
may we scope it?

pkg/agent/plugin/workloadattestor/k8s/k8s.go
```diff
@@ -808,6 +804,11 @@ func (p *Plugin) attestByKubernetesObjectReference(ctx context.Context, brokerEn
 	case objType.Plural == "pods" && objType.Group == "core":
 		return p.attestByPodReference(ctx, brokerEntry, objRef)
 	default:
+		// Non-pod objects are cluster-wide, so they require cluster scope;
+		// node-scoped brokers are limited to pods.
+		if brokerPodReferenceScope(brokerEntry) != podReferenceScopeCluster {
```
MarcosDY: If i'm reading this fine, we are going to reject every time we get on default and `brokerEntry` is nil, 
since `brokerPodReferenceScope(nil)` returns `agent_node`, so a non-broker `AttestReference` caller that previously resolved a non-pod object now gets `PermissionDenied`.

pkg/agent/plugin/workloadattestor/k8s/k8s.go
sorindumitru: Yes, that's true. You can still enable it, but you have to explicitly add it in there. I don't think we can default to enabled for that check because it's not part of the SPIFFE spec.

Alternatively, we could default to requiring a broker block and require specifying if the SAR check should be done or not.

MarcosDY (approved):

PR #7189, opened 2026-08-04
fixes ordering of cache events

### LINKED ISSUE #7181: Skipped event ID spikes after full cache rebuild delay SVID issuance (part 2)
Unfortunately, updating to the latest Spire version (v1.15.2) did not bring the expected improvements from #7022  and #6994 , so I am opening this issue (but feel free to re-open also #6876 if you like).

We had an even bigger "skipped event ID spike" now that went over a longer period of time. The DB was very busy with the batch queries, too, so it did not bring improvement. Maybe the page size is too high, or the query needs to be optimized further to find a middle-ground between thousands of small queries and few big ones.

Any one else has some experience with this new version and the impact of the batch query on the DB?

Query:

```
WITH
  filtered_nodes AS (
  SELECT
    *
  FROM
    attested_node_entries
  WHERE
    $10001
    AND spiffe_id IN ($1,
      $2,
      $3,
      $4,
...
      $207,
      $208
```

nweisenauer-sap:
## Summary
Addresses a root cause of the false skipped-event growth seen in #7181.
The events-based cache assumes `List*Events` returns rows ordered by ID and tracks every gap as a potentially late event.
With GORM v1, `Find()` executes immediately, so the existing trailing `Order("id asc")` had no effect.
An unordered startup scan can therefore classify a large portion of the event table as missing.
This causes unnecessary event lookups and can delay normal cache-update processing.
Apply `Order("id asc")` before `Find()` for attested-node and registration-entry event queries.

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**

**Description of change**

**Which issue this PR fixes**
fixes #7181

### DIFF
--- a/pkg/server/datastore/sqlstore/sqlstore.go
+++ b/pkg/server/datastore/sqlstore/sqlstore.go
@@ -1807,11 +1807,11 @@ func listAttestedNodeEvents(db *sqlDB, req *datastore.ListAttestedNodeEventsRequ
 			return nil, sqlcommon.NewWrappedSQLError(err)
 		}
 
-		if err := db.Find(&events, query.String(), id).Order("id asc").Error; err != nil {
+		if err := db.Order("id asc").Find(&events, query.String(), id).Error; err != nil {
 			return nil, sqlcommon.NewWrappedSQLError(err)
 		}
 	} else {
-		if err := db.Find(&events).Order("id asc").Error; err != nil {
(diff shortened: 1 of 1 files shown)


sorindumitru (approved):
Thanks @nweisenauer-sap! This matches what the [docs](https://v1.gorm.io/docs/query.html#Order) specify, so LGTM

PR #7190, opened 2026-08-04
Broker API Selector support

### LINKED ISSUE #7007: Selector based version of the spiffe broker api
For the spire-identity-exchange, we'd like to use the new, spiffe broker api. Our workloads are not running locally though, so we would need to use selector based equivalent of the spire-agent delegated api. Can that please be added to the broker api implementation? It sounds like it can be done while being compatible with the existing broker api specification. Once done that would make the delegation api completely redundant and can be deprecated.

kfox1111:
- [x] Documentation updated?

**Affected functionality**
Broker api changed

**Description of change**
Broker api supports selectors now. Similar to the delegated api's selector support

**Which issue this PR fixes**
fixes #7007


cmd/spire-agent/cli/run/run.go
```diff
@@ -205,6 +248,13 @@ func (c *agentConfig) brokerBindAddrs() ([]net.Addr, error) {
 	return addrs, nil
 }
 
+// brokerReferenceTypeRequiresExplicitGrant reports whether a reference type URL
+// is excluded from the `"*"` wildcard and so must be named verbatim. Kept in
+// sync with the api package's requiresExplicitGrant.
+func brokerReferenceTypeRequiresExplicitGrant(typeURL string) bool {
```
MarcosDY, line 254: Config validation and the runtime disagree on how they match `SelectorReference`. 

`brokerReferenceTypeRequiresExplicitGrant` compares the type URL by exact string, but the runtime's `requiresExplicitGrant` -> `isSelectorReference` compares only the segment after the last `/`, deliberately, since that's what anypb resolves on. 

So with a non-canonical prefix like `example.com/spire.api.types.SelectorReference`, two things go wrong: combining it with `"*"` is rejected as not requiring an explicit grant even though the runtime says it does, and omitting the `selector_assertion` block passes validation entirely,
the agent starts cleanly, then fails every request with `PermissionDenied: broker %q is not configured to assert selectors`. 

Possible fix: export the runtime helper (`requiresExplicitGrant` → `RequiresExplicitGrant`), re-export it in `endpoints.go` beside `SelectorReferenceTypeURL`, and route both config call sites through it, deleting the duplicate,

pkg/agent/broker/api/service.go
```diff
@@ -358,12 +391,26 @@ func (s *Service) SubscribeToJWTBundles(req *broker.SubscribeToJWTBundlesRequest
...
 		log.Error("No workload reference provided")
 		return nil, status.Error(codes.InvalidArgument, "workload reference must be provided")
 	}
 
+	// Reference types that carry selectors are resolved here rather than by the
+	// attestor catalog. A workload attestor plugin cannot express them: the
+	// plugin facade stamps its own name as the type of every selector it
+	// returns (see workloadattestor.V1.selectorsFrom), so a plugin can never
+	// round-trip an asserted set spanning arbitrary selector types.
+	if isSelectorReference(ref.GetReference().GetTypeUrl()) {
```
MarcosDY, line 405: Can you cover this wiht tests?

pkg/agent/broker/api/selectorreference.go
```diff
@@ -0,0 +1,163 @@
...
+		if _, ok := policy.AllowedSelectorTypes[selector.Type]; !ok {
+			log.WithField(telemetry.Type, selector.Type).Error("Permission denied; broker may not assert this selector type")
+			return nil, status.Errorf(codes.PermissionDenied, "broker %q is not allowed to assert selector type %q", caller, selector.Type)
+		}
+	}
+
+	// Audit at Info, not Debug. This is the only record that identities were
+	// issued on the broker's word rather than on the agent's attestation, and
+	// the workload attestor metrics do not cover this path.
+	log.WithFields(logrus.Fields{
```
MarcosDY, line 146: `SelectorReferenceTypeURL`  is hardcoded intead of the type URL the caller actually sent, Same on service.go:410 with metric label. 
Can you use `ref.GetTypeUrl()` here?; keeping the constant as the metric label is good.

pkg/agent/broker/api/service.go
```diff
@@ -86,11 +101,16 @@ type Service struct {
...
 }
 
 // authorizeReferenceType applies the per-broker reference type policy. UDS
 // requests only need the type to be allowed. TCP requests additionally need
-// that same allowed type to opt in to TCP use.
+// that same allowed type to opt in to TCP use via allow_over_tcp.
+//
+// Reference types requiring an explicit grant (see requiresExplicitGrant) are
+// never granted by the "*" wildcard, and are denied outright for callers with
+// no policy at all.
```
MarcosDY, line 113: All four denial paths in `authorizeReferenceType` return a status error without logging. 

The gRPC status reaches the caller, not the agent operator, so the agent-side record of "a broker was denied SelectorReference over TCP", including the fail-closed denial of an unconfigured caller attempting selector assertion exists only if the deployment happens to run a status-logging interceptor.

cmd/spire-agent/cli/run/run.go
```diff
@@ -205,6 +248,13 @@ func (c *agentConfig) brokerBindAddrs() ([]net.Addr, error) {
...
+	return typeURL == broker.SelectorReferenceTypeURL
+}
+
 func brokerAllowedReferenceTypesFromHCL(brokerID string, in []brokerAllowedReferenceTypeHCLEntry) ([]broker.AllowedReferenceType, error) {
```
MarcosDY, line 258: Duplcated type_url entries aren't rejected, and buildAllowedReferenceTypeMap writes into a map, so the last one silently wins.
I think  rejecting duplicates per broker here seems worth it, mirroring the duplicate-broker-ID pre-pass we already have.

PR #7191, opened 2026-08-04
sshpop: optionally verify client IP against source-address

samuraixwandering:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Server-side `sshpop` node attestor. Optional config only; default behavior is unchanged.

**Description of change**

Adds `verify_client_ip` to the sshpop server plugin (default `false`), following the x509pop option introduced in #6911.

When enabled, if the host certificate includes an OpenSSH `source-address` critical option, the connecting peer IP (from `X-Forwarded-Client-IP` on the node attestor facade) is checked against that option. The option is a comma-separated list of IPs and/or CIDRs, matching `golang.org/x/crypto/ssh`.

`CheckCert` accepts `source-address` but only enforces it during user authentication. sshpop validates host certificates via `CheckHostKey`, so the option was previously ignored after being accepted. Certificates without `source-address` are unaffected even when the flag is on.

This is the first of two remaining plugins from #6769 (httpchallenge to follow in a separate PR, per maintainer guidance).

**Test plan**

- [x] `go test ./pkg/common/plugin/sshpop/ ./pkg/server/plugin/nodeattestor/sshpop/ ./pkg/agent/plugin/nodeattestor/sshpop/`
- [x] Unit coverage for flag off, no option, exact/CIDR match and mismatch, missing/invalid client IP

Part of #6769


pkg/server/plugin/nodeattestor/sshpop/sshpop.go
```diff
@@ -55,7 +57,7 @@ func (p *Plugin) Attest(stream nodeattestorv1.NodeAttestor_AttestServer) error {
 	}
 
 	handshaker := p.sshserver.NewHandshake()
-	if err := handshaker.VerifyAttestationData(payload); err != nil {
+	if err := handshaker.VerifyAttestationData(payload, clientIPFromContext(stream.Context())); err != nil {
```
amartinezfayo: It could be worth adding a test that exercises this path through the facade, along the lines of TestVerifyClientIP in the x509pop plugin tests, which builds a context with peer.NewContext and asserts the allow and deny outcomes. The new tests cover the handshake layer well, but as I read it nothing currently verifies that the IP observed by the server actually flows through the metadata into clientIPFromContext.

pkg/server/plugin/nodeattestor/sshpop/sshpop.go
```diff
@@ -116,3 +118,11 @@ func (p *Plugin) Validate(_ context.Context, req *configv1.ValidateRequest) (*co
 		Notes: notes,
 	}, nil
 }
+
+func clientIPFromContext(ctx context.Context) string {
```
amartinezfayo: nit: since the x509pop plugin already has the same extraction inline and httpchallenge will follow, would it make sense to add a shared helper next to XForwardedClientIPKey in the nodeattestor package so the three plugins read the metadata the same way? x509pop also warns when multiple values are present, so a shared helper would keep that behavior consistent too.

doc/plugin_server_nodeattestor_sshpop.md
```diff
@@ -21,6 +21,9 @@ spiffe://<trust_domain>/spire/agent/sshpop/<fingerprint>
 | `cert_authorities_path` | A file that contains a list of trusted CAs in ssh `authorized_keys` format.                                                    |                                         |
 | `canonical_domain`      | A domain suffix for validating the hostname against the certificate's valid principals. See CanonicalDomains in ssh_config(5). |                                         |
 | `agent_path_template`   | A URL path portion format of Agent's SPIFFE ID. Describe in text/template format.                                              | `"{{ .PluginName}}/{{ .Fingerprint }}"` |
+| `verify_client_ip`      | If `true`, match peer IP to host-cert `source-address` when set (IPs/CIDRs). No-op if absent. Peer may be a load balancer.     | false                                   |
```
amartinezfayo: nit: the row reads a bit terse, especially "Peer may be a load balancer." Maybe we could phrase it closer to the x509pop row, something like "Reflects the immediate connecting peer, which may be a load balancer rather than the agent behind it."

pkg/common/plugin/sshpop/handshake_test.go
```diff
@@ -234,6 +235,130 @@ func marshalAttestationData(t *testing.T, cert []byte) []byte {
...
+			}
+			handshaker := s.NewHandshake()
+			attestation := marshalAttestationData(t, tp.Certificate.Marshal())
+			err := handshaker.VerifyAttestationData(attestation, tt.clientIP)
+			spiretest.RequireGRPCStatusContains(t, err, tt.expectCode, tt.expectMsg)
+		})
+	}
+}
+
+func TestCheckSourceAddress(t *testing.T) {
```
amartinezfayo: nit: I think these cases all duplicate scenarios already covered by the table-driven test above, so this could probably be dropped, or kept table-driven with named cases if testing the helper in isolation is valuable.

pkg/common/plugin/sshpop/handshake.go
```diff
@@ -146,6 +156,51 @@ func (s *ServerHandshake) VerifyAttestationData(data []byte) error {
...
 
+// verifyClientIPAgainstSourceAddress enforces the source-address critical option
+// when verify_client_ip is enabled. If the option is absent, attestation is unchanged.
+func (s *ServerHandshake) verifyClientIPAgainstSourceAddress(cert *ssh.Certificate, clientIPStr string) error {
+	if !s.s.verifyClientIP {
+		return nil
+	}
+	sourceAddrs, ok := cert.CriticalOptions[sourceAddressCriticalOption]
+	if !ok || sourceAddrs == "" {
+		return nil
```
amartinezfayo: I think enabling verify_client_ip should reject certificates that lack the source-address option, matching x509pop where a cert with no matching IP SAN fails closed. As written, the flag silently provides no IP verification for certs issued without the option, so an operator can believe the check is active when it isn't. That would also keep the same option name meaning the same guarantee across the two plugins.

amartinezfayo (commented):
Thank you for this contribution, @samuraixwandering!

pkg/server/plugin/nodeattestor/sshpop/sshpop.go
samuraixwandering: Added `TestVerifyClientIP` in the plugin tests, modelled on the x509pop one. It mints a certificate carrying `source-address` (the fixtures have none) and attests through the facade with a peer context, covering the matching, mismatching and missing client IP cases. I checked it actually exercises the plumbing by stubbing the metadata forwarding in the facade, which makes the test fail.

pkg/server/plugin/nodeattestor/sshpop/sshpop.go
samuraixwandering: Added `ClientIPFromContext` next to `XForwardedClientIPKey` and moved sshpop onto it, including the warning when the metadata carries more than one value. I left the x509pop call site alone to keep this change scoped to sshpop. Happy to migrate x509pop too, here or separately, if you would prefer all three plugins on the helper from the start.

doc/plugin_server_nodeattestor_sshpop.md
samuraixwandering: Reworded closer to the x509pop row, and it now also states that a certificate without the option is rejected. Realigned the table since the longer row broke the aligned pipe style.

pkg/common/plugin/sshpop/handshake_test.go
samuraixwandering: Dropped `TestCheckSourceAddress`. Its cases were all covered by the table-driven test above.

pkg/common/plugin/sshpop/handshake.go
samuraixwandering: Agreed, and changed. A certificate without the `source-address` option is now rejected when `verify_client_ip` is enabled, so enabling the option always means the client IP was verified. You are right that it was silently verifying nothing otherwise, which is the worse failure mode because the operator has no signal. Updated the docs and inverted the test case that asserted the old no-op behaviour.

pkg/common/plugin/sshpop/sshpop.go
```diff
@@ -95,6 +96,10 @@ type ServerConfig struct {
 	// the certificate's valid principals. See CanonicalDomains in ssh_config(5).
 	CanonicalDomain   string `hcl:"canonical_domain"`
 	AgentPathTemplate string `hcl:"agent_path_template"`
+	// VerifyClientIP, when true, checks the connecting peer's IP against a
+	// source-address critical option on the host certificate when that option
+	// is present. Certificates without source-address are unaffected.
```
amartinezfayo: I think this comment still describes the earlier behavior, where a certificate without source-address was unaffected. We should  update it to match the current behavior.

doc/plugin_server_nodeattestor_sshpop.md
```diff
@@ -50,6 +53,9 @@ If both `cert_authorities` and `cert_authorities_path` are configured, the resul
 
             # Change the agent's SPIFFE ID format
             # agent_path_template = "static/{{ index .ValidPrincipals 0 }}"
+
+            # Optional: when true, enforce source-address critical option if present
```
amartinezfayo: nit: the commented example still says the option is enforced "if present". Maybe we could reword it along the lines of "# Optional: when true, require and enforce the source-address critical option".

amartinezfayo (commented):
Thank you @samuraixwandering for addressing the comments. A couple of very small things and we should be good to merge 🚀

pkg/common/plugin/sshpop/sshpop.go
samuraixwandering: Updated. It now says the option is required and that certificates without `source-address` are rejected. I also swept the other mentions of the option across `pkg/` and `doc/` and the remaining four already describe the current behaviour, so this and the example below were the last two.

doc/plugin_server_nodeattestor_sshpop.md
samuraixwandering: Reworded to your suggestion.

amartinezfayo (approved):
🚀

PR #7197, opened 2026-08-06
Make the vault and broker-api-k8s integration suites runnable on macOS

### LINKED ISSUE #7174: Integration suites require GNU coreutils and bash 4+ to run on a macOS host
* **Version**: `941bd35f1` (main)
* **Platform**: `Darwin 25.5.0 arm64`, `/bin/bash` 3.2.57, LibreSSL 3.3.6
* **Subsystem**: integration test suites

### Summary

Three integration suites cannot run on a macOS host without first installing GNU replacements for tools that macOS either ships in a BSD flavour or does not ship at all. CI only runs these on `ubuntu-22.04`, so the gap is invisible there.

I hit all three while working on #6127. Each one fails at the point of use rather than up front, so the failure surfaces several minutes into a suite run and does not obviously point at the host environment.

### What fails

**1. `timeout` is not present on macOS**

`key-manager-vault/01-setup-vault:70` and `upstream-authority-vault/01-setup-vault:70`:

```bash
timeout 30s bash -c 'while [[ -z $(./bin/kubectl get csr -n vault vault.svc -o jsonpath="{.status.certificate}" 2>/dev/null) ]]; do sleep 1; done'
```

```
01-setup-vault: line 70: timeout: command not found
```

**2. `date -d` is GNU only**

`key-manager-vault/04-verify-token-renewal.sh:7` and `upstream-authority-vault/04-verify-token-renewal.sh:7`:

```bash
timeout=$(date -ud "1 minute 30 second" +%s)
```

BSD `date` has no `-d`, so this fails even once `timeout` is available.

**3. `readarray` is a bash 4+ builtin**

`broker-api-k8s/02-create-entries:7`:

```bash
readarray -t NODEUIDS < <(./bin/kubectl get nodes -o jsonpath='...')
```

```
02-create-entries: line 7: readarray: command not found
```

macOS ships bash 3.2.57 as `/bin/bash` for licensing reasons. `run-step()` in `test-one.sh` invokes `bash -s` resolved from `PATH`, so this one is fixed by putting a newer bash earlier on `PATH`, without any change to the repo.

### Workaround

For anyone else hitting this, the suites all pass on macOS after:

```
brew install coreutils bash
PATH="/opt/homebrew/opt/coreutils/libexec/gnubin:/opt/homebrew/bin:$PATH"
```

The `gnubin` path supplies `timeout` and a GNU `date`. Worth noting that putting `gnubin` on `PATH` permanently also shadows BSD `ls`, `sed` and friends, which is a bigger change than most people want in their shell profile, so scoping it to the test run is preferable.

(Unrelated to portability, but in the same area: `make load-images` needs `regctl`, which is not mentioned in `CONTRIBUTING.md`. `brew install regclient` covers it.)

### Possible directions

I do not want to assume which of these you would prefer, and `01-setup-vault` is shared CI-critical infrastructure, so I would rather ask than send a patch:

1. **Replace the GNU-specific calls with portable equivalents.** The repo already has a portable waiting idiom in `k8s/01-apply-config`, which polls with `for ((i=0; i<${MAXROLLOUTCHECKS}; i++))` instead of wrapping in `timeout`. The `date -ud` case can be expressed as `$(( $(date +%s) + 90 ))`, and `readarray` as a `while IFS= read -r` loop.
2. **Document a Linux-only requirement** for the integration suites, and optionally fail fast with a clear message when the host lacks the tools, instead of dying mid-suite.
3. **Leave as is** if running these on macOS is out of scope.

Happy to send a PR for whichever direction you prefer.

samuraixwandering:
**Pull Request check list**

- [x] Proper tests/regressions included? (the changed files are the integration suites themselves; all three were run end to end, below)
- [ ] Documentation updated? (not needed)

**Affected functionality**

The `key-manager-vault`, `upstream-authority-vault` and `broker-api-k8s` integration suites. No production code is touched and there is no behaviour change on Linux.

**Description of change**

These suites cannot run on a macOS host without installing GNU replacements first. CI only runs them on `ubuntu-22.04`, so the gap is invisible there.

- `timeout 30s bash -c '...'` in both vault suites becomes the bounded polling idiom `wait-for-rollout` already uses in `k8s/01-apply-config`, since macOS does not ship `timeout`. One deliberate difference: the budget is a check count now rather than wall-clock time, matching that suite, and it calls `fail-now` with a reason instead of dying at bare exit 124 under `set -e`.
- `date -ud "1 minute 30 second" +%s` becomes `$(( $(date -u +%s) + 90 ))`, since BSD `date` has no `-d`. Line 10 of the same script already compares against `date -u +%s`, so both sides agree.
- `readarray` becomes a `while IFS= read -r` loop, since it is a bash 4+ builtin and macOS ships bash 3.2 as `/bin/bash`. Kept as a separate commit: unlike the other two it has a workaround needing no repo change, so it is easy to drop if you would rather leave `broker-api-k8s` alone.

**Verification**

Run on Darwin 25.6.0 arm64 with `PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin`, which carries no GNU tools. Gated before running: `bash` is 3.2.57, `timeout` is absent, `date -d` is rejected, `readarray` is unavailable.

On unmodified main (`6215e5995`) the reported failure reproduces, after the kind cluster comes up healthy:

```
01-setup-vault: line 70: timeout: command not found
```

On this branch, same host and same PATH:

| Suite | Result |
| --- | --- |
| `upstream-authority-vault` | passed |
| `key-manager-vault` | passed |
| `broker-api-k8s` | passed |

The changed paths are exercised rather than skipped. The poll loop obtains a real certificate from k8s, `04-verify-token-renewal.sh` runs a real deadline loop and reports `token renewal is verified`, and `broker-api-k8s` comes up as a three node cluster so the read loop parses three distinct node UIDs.

**Which issue this PR fixes**

fixes #7174

### DIFF
--- a/test/integration/suites/broker-api-k8s/02-create-entries
+++ b/test/integration/suites/broker-api-k8s/02-create-entries
@@ -4,7 +4,10 @@ set -e
 
 source init-kubectl
 
-readarray -t NODEUIDS < <(./bin/kubectl get nodes -o jsonpath='{range .items[*]}{.metadata.uid}{"\n"}{end}')
+NODEUIDS=()
+while IFS= read -r nodeUID; do
+    NODEUIDS+=("${nodeUID}")
+done < <(./bin/kubectl get nodes -o jsonpath='{range .items[*]}{.metadata.uid}{"\n"}{end}')
 PARENTS=()
 for nodeUID in "${NODEUIDS[@]}"; do
     if [ -n "${nodeUID}" ]; then
--- a/test/integration/suites/key-manager-vault/01-setup-vault
+++ b/test/integration/suites/key-manager-vault/01-setup-vault
@@ -67,7 +67,16 @@ log-info "installing hashicorp vault..."
 ./bin/kubectl create -f csr.yaml
 ./bin/kubectl certificate -n vault approve vault.svc
 log-debug "waiting for certificate to be issued by k8s..."
-timeout 30s bash -c 'while [[ -z $(./bin/kubectl get csr -n vault vault.svc -o jsonpath="{.status.certificate}" 2>/dev/null) ]]; do sleep 1; done'
+MAXCERTCHECKS=30
+for ((i=1; i<=MAXCERTCHECKS; i++)); do
+    if [[ -n $(./bin/kubectl get csr -n vault vault.svc -o jsonpath="{.status.certificate}" 2>/dev/null) ]]; then
+        break
+    fi
+    if [[ $i -eq $MAXCERTCHECKS ]]; then
+        fail-now "certificate was not issued by k8s after ${MAXCERTCHECKS} checks"
+    fi
+    sleep 1
+done
 ./bin/kubectl get csr -n vault vault.svc -o jsonpath='{.status.certificate}' | openssl base64 -d -A -out vault.pem
 ./bin/kubectl config view --raw --minify --flatten -o jsonpath='{.clusters[].cluster.certificate-authority-data}' \
               | base64 -d > vault_ca.pem
--- a/test/integration/suites/key-manager-vault/04-verify-token-renewal.sh
+++ b/test/integration/suites/key-manager-vault/04-verify-token-renewal.sh
@@ -4,7 +4,7 @@ set -eo pipefail
 
 log-debug "verifying token renewal..."
 
-timeout=$(date -ud "1 minute 30 second" +%s)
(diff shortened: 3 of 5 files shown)


sorindumitru (approved):
LGTM, thanks @samuraixwandering!

PR #7202, opened 2026-08-10
Make LRU cache generic

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
This should behave the same before and after the change

**Description of change**
Parameterise the LRU cache on SVID type to make it possible to reuse for WIT-SVIDs


pkg/agent/manager/cache/lru_cache.go
```diff
@@ -977,43 +832,16 @@ func (c *LRUCache) matchingEntries(set selectorSet) []*common.RegistrationEntry
...
-					telemetry.FederatedBundle: federatesWith,
-				}).Warn("Federated bundle contents missing")
-			}
-		}
-	}
-
-	return w
+// GetRecordsForSelectors returns the set of records matching the selector set.
+// Exported for use by buildUpdate callbacks.
+func (c *LRUCache[SVID, Update]) GetRecordsForSelectors(set selectorSet) (lruCacheRecordSet, func()) {
```
amartinezfayo: Since BuildUpdate takes the unexported selectorSet type, a buildUpdate callback can only be implemented inside this package, so I think GetRecordsForSelectors (buildX509WorkloadUpdate calls the unexported version directly), Bundles, TrustDomain, and Log could stay unexported or be dropped until an external consumer exists. Bundles in particular returns the internal map that UpdateEntries mutates in place, so a caller reading it after the read lock is released would race with a sync. Would it make sense to trim these until the WIT-SVID cache actually needs them?

pkg/agent/manager/cache/lru_cache.go
```diff
@@ -469,17 +461,18 @@ func (c *LRUCache) UpdateEntries(update *UpdateEntries, checkSVID func(*common.R
 	}
 }
 
-func (c *LRUCache) UpdateSVIDs(update *UpdateSVIDs) {
+// UpdateSVIDs updates SVIDs in the cache for the given entries.
+func (c *LRUCache[SVID, Update]) UpdateSVIDs(svids map[string]*SVID) {
 	c.mu.Lock()
-	defer func() { agentmetrics.SetSVIDMapSize(c.metrics, c.CountX509SVIDs()) }()
+	defer func() { agentmetrics.SetSVIDMapSize(c.metrics, c.CountSVIDs()) }()
```
amartinezfayo: I wonder if the gauges emitted here and in UpdateEntries (SVIDMapSize and RecordMapSize) should gain a cache-type dimension as part of making the cache reusable. They carry no label today, so once a WIT-SVID instance of the generic cache exists, both instances would write to the same gauge and overwrite each other's values.

pkg/agent/manager/cache/lru_cache.go
```diff
@@ -792,8 +672,8 @@ func (c *LRUCache) updateOrCreateRecord(newEntry *common.RegistrationEntry) (*lr
 	return record, existingEntry
 }
 
-func (c *LRUCache) diffSelectors(existingEntry, newEntry *common.RegistrationEntry, added, removed selectorSet) {
-	// Make a set of all the selectors being added
+// Make a set of all the selectors being added
```
amartinezfayo: nit: it looks like the refactor hoisted this step comment into doc-comment position; it describes only the first block of diffSelectors, and the same happened with the dedup comment above getRecordsForSelectors. Maybe we could move them back inside the function bodies so they don't read as doc comments for the whole function.

pkg/agent/manager/cache/x509_lru_cache.go
```diff
@@ -0,0 +1,236 @@
...
+
+// X509SVIDLRUCache wraps LRUCache with X509-specific operations.
+type X509SVIDLRUCache struct {
+	*LRUCache[X509SVID, X509WorkloadUpdate]
+
+	processingBatchSize     int
+	taintedBatchProcessedCh chan struct{}
+}
+
+func NewX509LRUCache(config LRUCacheConfig[X509SVID, X509WorkloadUpdate]) *X509SVIDLRUCache {
```
amartinezfayo: nit: since BuildUpdate is unconditionally overwritten here, a caller-supplied value is silently discarded. It could be worth taking a narrower config, or an alias like type X509LRUCacheConfig = LRUCacheConfig[X509SVID, X509WorkloadUpdate], so call sites don't have to spell out the type parameters and can't pass a BuildUpdate that gets ignored.

pkg/agent/manager/cache/lru_cache.go
sorindumitru: Yes, that's true, these should get a label with the type

pkg/agent/manager/cache/x509_lru_cache.go
sorindumitru: Yes, I think this makes sense. We may end up in the situation where the different caches might need cache specific behaviour at some point, so having a dedicated config makes sense.

pkg/common/telemetry/agent/lru.go
```diff
@@ -18,8 +18,10 @@ func SetEntriesMapSize(m telemetry.Metrics, recordMapSize int) {
 	m.SetGauge([]string{telemetry.RecordMapSize}, float32(recordMapSize))
```
amartinezfayo: nit: it looks like the entries map gauge kept its unlabeled form, so two cache instances would still overwrite each other's record map size even though the SVID map gauge is now dimensioned. Should we consider giving SetEntriesMapSize the same svid_type label treatment (plus the matching telemetry.md row)? I guess doing it in a follow-up PR that introduces the second cache instance would also be fine.

pkg/agent/manager/cache/lru_cache.go
```diff
@@ -977,43 +839,10 @@ func (c *LRUCache) matchingEntries(set selectorSet) []*common.RegistrationEntry
...
-	}
-
-	return w
-}
-
-func (c *LRUCache) getRecordsForSelectors(set selectorSet) (lruCacheRecordSet, func()) {
-	// Build and dedup a list of candidate entries. Don't check for selector set inclusion yet, since
-	// that is a more expensive operation, and we could easily have duplicate
-	// entries to check.
+// Build and dedup a list of candidate entries. Don't check for selector set inclusion yet, since
```
amartinezfayo: nit: this one still reads as a doc comment for the whole function, and since the second loop does perform the inclusion check, the phrase about not checking inclusion yet only describes the first loop. Maybe we could move it back above the first loop like the diffSelectors one.

amartinezfayo (dismissed):

amartinezfayo (approved):

PR #7203, opened 2026-08-11
test: add tpm_devid integration suite backed by a software TPM

### LINKED ISSUE #6466: Integration coverage for TPM attestation
SPIRE makes use of a TPM simulator (https://github.com/spiffe/spire/blob/main/test/tpmsimulator/simulator.go) for whitebox unit testing of the `tpm_devid` plugin. This is useful as it provides an ephemeral TPM fake that runs quickly with the test suite and covers both attestation and session management primitives.

However, this is limited to the `tpm_devid` plugin only and does not exercise the SPIFFE external plugin [`spire-tpm-plugin`](https://github.com/spiffe/spire-tpm-plugin) (or any plugins that might be added in the future).

There's also something of a maintenance burden on having our own TPM2.0 compliant simulator (although perhaps this isn't a huge problem given the likely operations the Agent uses).

An improvement here could be to begin to add test coverage at the integration level by using a software TPM - something like [swtpm](https://github.com/stefanberger/swtpm). This is an approach used by TPM-centric projects like [Keylime](https://github.com/keylime/keylime/blob/0b4335d05446588a0cd1e60ba9b6eec45ce809c9/.ci/test_wrapper.sh#L6-L10), and e.g. [library code for TPM2.0 bindings](https://github.com/parallaxsecond/rust-tss-esapi/blob/2df045ff9cada406caf59d0ed9284b663507668b/tss-esapi/tests/all-fedora.sh#L20-L41) in Rust.

Happy to contribute with this, but wanted to raise for opinions before submitting any diff

samuraixwandering:
## Summary

Adds an integration suite that attests an agent with the `tpm_devid` node
attestor against a software TPM, so the plugin is covered end to end and not
only by the whitebox unit tests against the in-process simulator.

Two commits: the provisioner under `test/integration/setup/tpmdevid`, then the
suite itself under `test/integration/suites/tpm-devid`.

## How it works

swtpm runs in its own container and exposes the TPM as a unix socket on a
shared volume. go-tpm dials that socket and speaks the emulator protocol, so
the suite needs no TPM device, no kernel module and no extra container
privileges.

The provisioner takes the key templates and the NV index from SPIRE rather than
restating them. The agent regenerates the endorsement key from
`client.DefaultEKTemplateRSA()` and loads the DevID key under
`tpmutil.SRKTemplateHighRSA()` at attestation time, so credentials provisioned
under different parameters would not attest. Taking them from the plugin means
the suite follows if those ever change.

The test derives the expected agent ID from the DevID certificate it
provisioned, so it fails if the agent attests with some other credential rather
than merely reporting that an agent appeared.

The in-process simulator in `test/tpmsimulator` is untouched and still backs the
unit tests. This is additive coverage, per the discussion on the issue.

## Verification

- The suite passes end to end against images built from `main`.
- Positive control: pointing `devid_ca_path` at a CA that did not sign the
  agent's DevID certificate fails the suite with
  `nodeattestor(tpm_devid): unable to verify DevID signature: verification
  failed: x509: certificate signed by unknown authority`, so the test is
  actually exercising attestation.
- The provisioner's output was checked against the plugin's own session code:
  `tpmutil.NewSession` loads the DevID key and regenerates the EK,
  `GetEKCert` reads the certificate back from NV, and both
  `SolveDevIDChallenge` and `CertifyDevIDKey` succeed.
- The agent ID the test derives in shell was confirmed to match the server's
  `Fingerprint()` for the same certificate.
- The TPM is provisioned fresh per run; the agent ID differs between runs.

## Notes

Suites are discovered automatically by `split.sh`, so no workflow change is
needed. The suite runs on both the x64 and arm64 integration runners, so it
needs no `IGNORE_SUITES` entry.

Coverage for the external `spire-tpm-plugin` is deliberately left out. It adds a
cross-repo dependency to the integration CI and that plugin currently builds
against `spire-plugin-sdk` v1.14.4 while SPIRE is on 1.15, so it is worth a
maintainer decision. Happy to follow up with it separately.

Closes #6466


test/integration/suites/tpm-devid/swtpm-entrypoint.sh
```diff
@@ -0,0 +1,33 @@
...
+    i=$((i + 1))
+    if [ "${i}" -gt 100 ]; then
+        echo "swtpm did not create ${SOCKET_DIR}/swtpm.sock" >&2
+        exit 1
+    fi
+    sleep 0.1
+done
+chmod 0666 "${SOCKET_DIR}/swtpm.sock"
+
+wait "${swtpm_pid}"
```
amartinezfayo: I think the shell stays PID 1 here with no TERM handler, so the SIGTERM from docker compose down would be ignored and each teardown would sit out the 10 second kill timeout before the SIGKILL. Would it be worth trapping TERM to kill swtpm, or moving the socket-wait and chmod into a backgrounded subshell and exec'ing swtpm so the signal reaches it directly?

test/integration/setup/tpmdevid/provision.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	ekHandle, ekPublicBlob, _, _, _, _, err := tpm2.CreatePrimaryEx(rwc, tpm2.HandleEndorsement,
+		tpm2.PCRSelection{}, "", "", client.DefaultEKTemplateRSA())
+	if err != nil {
+		return nil, fmt.Errorf("cannot create endorsement key: %w", err)
+	}
+	defer func() {
+		if err := tpm2.FlushContext(rwc, ekHandle); err != nil {
+			log.Printf("warning: cannot flush endorsement key handle: %v", err)
```
amartinezfayo, line 129: nit: the two flush defers treat failure differently: this one logs a warning while the one in provisionDevID promotes the error into the named return. A flush failure doesn't invalidate the material either way, so would it make sense to give both the same treatment?

test/integration/setup/tpmdevid/provision.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot provision endorsement certificate: %w", err)
+	}
+
+	devIDCA, devIDCert, privateBlob, publicBlob, err := provisionDevID(rwc)
+	if err != nil {
+		return fmt.Errorf("cannot provision DevID: %w", err)
+	}
+
+	for dir, files := range map[string]map[string][]byte{
```
amartinezfayo: nit: I wonder if two direct writeFiles calls, one for agentDir and one for serverDir, would read a bit more directly than iterating a map of maps here.

test/integration/setup/tpmdevid/provision.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return nil, nil, nil, nil, fmt.Errorf("cannot create storage root key: %w", err)
+	}
+	defer func() {
+		if flushErr := tpm2.FlushContext(rwc, srkHandle); flushErr != nil && err == nil {
+			err = fmt.Errorf("cannot flush storage root key handle: %w", flushErr)
+		}
+	}()
+
+	devIDTemplate := client.AKTemplateRSA()
```
amartinezfayo, line 212: The plugin also supports ECC DevID keys, and the in-process simulator generates both flavors. RSA is a great start for the suite, but ECC coverage would be worth a follow-up maybe as a flag on this provisioner. What do you think?

amartinezfayo (commented):
This is a great contribution, thank you @samuraixwandering!

test/integration/suites/tpm-devid/swtpm-entrypoint.sh
samuraixwandering: Good catch, and confirmed. The shell was PID 1 with `SigCgt` showing only SIGINT and SIGCHLD, so SIGTERM was discarded and `docker stop` took the full 10 seconds and ended in a SIGKILL (exit 137). I took your second suggestion: the socket wait and chmod moved into a backgrounded subshell and swtpm is now exec'd, so it is PID 1 and catches SIGTERM itself. Same check after the change: `SigCgt` has SIGTERM, `docker stop` returns in 0s with exit 0, and the socket still comes out `srw-rw-rw-`. The suite's teardown went from about 11 seconds to 1.

test/integration/setup/tpmdevid/provision.go
samuraixwandering: Agreed. Both now log, since a flush failure does not invalidate the material either way. That also let `provisionDevID` drop its named results, which only existed so the defer could write the error out. The change landed on the storage root key side, which is why this comment still points at unchanged lines.

test/integration/setup/tpmdevid/provision.go
samuraixwandering: Changed to two direct `writeFiles` calls. It reads better and it also makes the order deterministic, which the map iteration was not.

test/integration/setup/tpmdevid/provision.go
samuraixwandering: Agreed that it is worth covering. I would rather do it as a follow-up than grow this one: a `-key-type` flag alone would be untested code, so it wants a suite step that re-provisions and re-attests with ECC. One thing I want to check before committing to a shape is that ECC works over the swtpm transport, since the RSA path needed a spike to prove the emulator protocol worked at all.

While addressing these I also looked at what else the suite does not cover, in case you want any of it folded into the same follow-up. Three things: it never asserts the selectors the plugin emits (`subject:cn`, `issuer:cn`, `ca:fingerprint`), there is no negative case proving attestation is rejected when the DevID certificate is not signed by the configured CA, and the three password options (`devid_password`, `owner_hierarchy_password`, `endorsement_hierarchy_password`) are unexercised. Happy to add any of them here instead if you would prefer.

samuraixwandering:
The three failures on the last run are not from this PR. The failing suite is `upgrade`, at `02-verify-codebase-version-is-updated`, and the `tpm-devid` suite did not run in either of those shards.

It looks like the post-release window: `v1.15.3` was tagged today at 13:34 UTC, and `pkg/common/version/version.go` on `main` still has `Base = "1.15.3"`, so the commit version equals the latest tag and the step fails with "has the version been bumped?". Other open PRs are green, but their last runs were before the tag; this one happened to run at 14:28 UTC.

So I do not think there is anything to change here. Is a version bump on `main` already planned? Happy to re-run once it lands.

**Edit**: Seems fine now, will be reworking.

test/integration/setup/tpmdevid/provision.go
amartinezfayo: Thanks for laying this out. I agree ECC is better as a follow-up for the reason you gave, so let's keep this PR as is. Of the three gaps, the selector assertions and the negative case are the ones worth doing, since the suite proves only the agent ID today, and the password options seem lower value. Feel free to open an issue to track them.

amartinezfayo (approved):
🚀

PR #7211, opened 2026-08-13
Don't warn on expected profiling server shutdown

c4rlo:
`http.Server.ListenAndServe` always returns a non-nil error, returning the `ErrServerClosed` sentinel once the server has been shut down. Both the agent and server logged this unconditionally, so every shutdown with a profiling port configured emitted a spurious warning:

    Unable to serve profiling server  error="http: Server closed"

Filter out `ErrServerClosed`.

While here, give `Shutdown` its own context: it was being passed the parent context, which is already cancelled by the time the shutdown goroutine runs, so shutdown dropped in-flight connections and returned `context.Canceled` instead of draining.


pkg/agent/agent.go
```diff
@@ -395,13 +395,15 @@ func (a *Agent) setupProfiling(ctx context.Context) (stop func()) {
...
 		wg.Go(func() {
-			if err := server.ListenAndServe(); err != nil {
+			if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
 				a.c.Log.WithError(err).Warn("Unable to serve profiling server")
 			}
 		})
 		wg.Go(func() {
 			<-ctx.Done()
-			if err := server.Shutdown(ctx); err != nil {
+			shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
```
amartinezfayo: I'm wondering if the drain should hold up process exit: stopProfiling is deferred in Run, so an in-flight pprof profile or trace would now keep the agent alive for the full timeout instead of exiting right away. Would a shorter budget be enough for a debug-only endpoint?

pkg/server/server.go
```diff
@@ -317,13 +317,15 @@ func (s *Server) setupProfiling(ctx context.Context) (stop func()) {
...
 		wg.Go(func() {
-			if err := server.ListenAndServe(); err != nil {
+			if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
 				s.config.Log.WithError(err).Warn("Unable to serve profiling server")
 			}
 		})
 		wg.Go(func() {
 			<-ctx.Done()
-			if err := server.Shutdown(ctx); err != nil {
+			shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
```
amartinezfayo: nit: would it be worth naming this duration a constant in both the agent and the server, like gracefulStopTimeout in the endpoints package, instead of repeating the literal? The warning just below could also name the profiling server, since it reads as if the SPIRE server failed to shut down.

amartinezfayo (commented):
Thank you @c4rlo!

pkg/agent/agent.go
c4rlo: Yes, fair. Reduced to 500 ms.

pkg/server/server.go
c4rlo: Pulled timeouts into constants and improved log messages.

amartinezfayo (approved):
🚀

PR #7213, opened 2026-08-14
feat(node-attestor/gcp_iit): allow alternative service account email whitelist

jeengbe:
With https://github.com/spiffe/spire/pull/6869, this attestor always has access to the email address, which allows a cheap whitelist check for it too.

So now, either project ID or service account email have to be whitelisted. Service accounts are already a granular subject whitelist, so no TOFU check is needed for that path.


jeengbe:
I forgot which original use case I had in mind for this and I really can't think of any scenario in which this would be useful now..

PR #7218, opened 2026-08-14
Skip redundant federation queries on entry writes

### LINKED ISSUE #7215: Entry writes issue redundant federation queries on every create, update, and delete
## What we see

In our deployment at Uber, entry write RPCs spend a large share of database time on federation queries that either return nothing or repeat work the transaction has already done. Measured over a six hour window across roughly 18 MySQL instances, federation work accounts for 22.8% of total database time, and two of the top fingerprints by database time are federation queries against tables that hold no rows for us.

The two fingerprints are:

- `SELECT bundles.* JOIN federated_registration_entries ... FOR UPDATE` at 13.8% of database time, ranked first overall. This is the `FederatesWith` association read in `modelToEntry`.
- `SELECT * FROM bundles WHERE trust_domain IN (NULL) FOR UPDATE` at 4.1%. This is `makeFederatesWith` called with an empty trust domain list.

Our primaries sit near 90% CPU on 48 cores against a few hundred thousand registration entries, so this share is material.

## Why these are redundant

Three federation touch points on the write path do work that is unnecessary regardless of whether a deployment uses federation.

`makeFederatesWith` queries `bundles` for the requested trust domains, then verifies each requested id was found. When the caller passes an empty list the query cannot match anything and the verification loop does not execute, so the function can return early without issuing SQL. Both `createRegistrationEntry` and `updateRegistrationEntry` call it unconditionally.

On create, the resolved bundles are appended to the `FederatesWith` association. The entry row was just inserted and has no existing associations, so appending an empty set writes nothing.

`modelToEntry` re-reads the `FederatesWith` association to populate its response. On create and update the caller resolved the exact bundle set a few lines earlier, so the read asks the database to return a value the transaction already holds. This is the 13.8% fingerprint. Delete is different, because it needs the associations as they existed before the row was removed.

## What we would like to change

1. Return early from `makeFederatesWith` when the id list is empty.
2. Skip the association append on create when there is nothing to append.
3. Pass the already resolved federation set into `modelToEntry` on create and update rather than re-reading it.

Each of these is behavior preserving in any deployment, federated or not, and none require cached state or new configuration. There is no schema change, so the CONTRIBUTING.md constraint about datastore changes preceding dependent code does not apply here.

We would preserve the `FederatesWith` ordering established in #2645, and keep the existing federation coverage in the datastore conformance suite green.

The association `Replace` on update and `Clear` on delete cannot be skipped without knowing whether an entry currently has associations, so we are leaving those out of scope.

We are happy to open a PR for this.

azimt:
**Pull Request check list**

- [ ] Documentation updated? (no user facing behavior, schema, or configuration change)

**Affected functionality**

The SQL datastore write path for registration entries: `CreateRegistrationEntry`, `UpdateRegistrationEntry`, and `DeleteRegistrationEntry`. There is no schema change and no configuration change.

**Description of change**

Entry writes issue two federation queries that are not needed.

`makeFederatesWith` queries `bundles` for the requested trust domains and then verifies each requested id was found. When the caller passes an empty list the query cannot match a row and the verification loop has nothing to check, so it now returns early without issuing SQL. Both `createRegistrationEntry` and `updateRegistrationEntry` call it unconditionally today.

`modelToEntry` reads the `FederatesWith` association to populate its response. Create and update resolve that exact bundle set a few lines earlier, so the read asks the database to return a value the transaction already holds. The converter is split into `modelToEntry`, which reads the association as before, and `modelToEntryWithFederatesWith`, which takes the trust domains from the caller. Create and update use the latter. Delete keeps reading, because it reports the associations as they existed before the row was removed.

Both changes are behavior preserving whether or not a deployment uses federation, and neither introduces cached state or new configuration.

Two details worth flagging for review:

The trust domains are derived from the resolved bundle rows rather than the requested ids. Bundle rows are unique per trust domain, so this keeps the current behavior of collapsing a trust domain named more than once into a single association, which returning the request verbatim would not.

No ordering guarantee changes. `FederatesWith` ordering is unspecified today on both paths, since the association read has no `ORDER BY` and the entry read queries order by `e_id, selector_id, dns_name_id` with no federation column. This change neither adds nor removes a guarantee, and the new multi-bundle test compares trust domains as a set rather than asserting an order the queries do not promise.

The issue also proposed guarding the association append on create. That turned out to remove nothing, because GORM's `saveAssociations` iterates the slice it is passed and the loop body never runs for an empty set, so it is not included here.

**Measurements**

From our deployment at Uber, over a six hour window across roughly 18 MySQL instances, ranked by database time:

| Query | Share of DB time | Removed by |
|---|---|---|
| `SELECT bundles.* JOIN federated_registration_entries ... FOR UPDATE` | 13.8%, ranked first overall | the `modelToEntry` split |
| `SELECT * FROM bundles WHERE trust_domain IN (NULL) FOR UPDATE` | 4.1% | the `makeFederatesWith` early return |

**Tests**

The existing federation coverage in the datastore conformance suite passes unchanged, including `TestRegistrationEntriesFederatesWithSuccess`, `TestRegistrationEntriesFederatesWithAgainstMissingBundle`, the `Update FederatesWith` mask cases in both directions, and the `DeleteBundle*RegistrationEntries` cases.

Three suite tests are added for paths that were not covered before: `TestRegistrationEntriesFederatesWithMultipleBundles`, `TestRegistrationEntriesFederatesWithDuplicateTrustDomain`, and `TestRegistrationEntriesWithoutFederatesWith`.

This targets the gormv1 datastore, since that is the only implementation in the tree today. Happy to port it if the gormv2 migration lands first.

**Which issue this PR fixes**

fixes #7215

### DIFF
--- a/pkg/server/datastore/sqlstore/sqlstore.go
+++ b/pkg/server/datastore/sqlstore/sqlstore.go
@@ -2692,7 +2692,9 @@ func createRegistrationEntry(tx *gorm.DB, entry *common.RegistrationEntry) (*com
 		}
 	}
 
-	registrationEntry, err := modelToEntry(tx, newRegisteredEntry)
+	// The federation associations were just written from federatesWith, so the
+	// trust domains are already known and do not need to be read back.
+	registrationEntry, err := modelToEntryWithFederatesWith(tx, newRegisteredEntry, bundlesToTrustDomains(federatesWith))
 	if err != nil {
 		return nil, err
 	}
@@ -4193,15 +4195,15 @@ func updateRegistrationEntry(tx *gorm.DB, e *common.RegistrationEntry, mask *com
 		if err := tx.Model(&entry).Association("FederatesWith").Replace(federatesWith).Error; err != nil {
 			return nil, err
 		}
-		// The FederatesWith field in entry is filled in by the call to modelToEntry below
+
+		// The associations were just replaced with federatesWith, so the trust
+		// domains are already known and do not need to be read back.
+		return modelToEntryWithFederatesWith(tx, entry, bundlesToTrustDomains(federatesWith))
 	}
 
-	returnEntry, err := modelToEntry(tx, entry)
-	if err != nil {
-		return nil, err
-	}
-
-	return returnEntry, nil
+	// FederatesWith was not part of this update, so the existing associations are
+	// unchanged and have to be read to build the response.
+	return modelToEntry(tx, entry)
 }
 
 func deleteRegistrationEntry(tx *gorm.DB, entryID string) (*common.RegistrationEntry, error) {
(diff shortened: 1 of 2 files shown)


sorindumitru (approved):
LGTM, thanks @azimt!

PR #7221, opened 2026-08-18
Remove use_sync_authorized_entries option

sorindumitru:
Has been deprecated for a while and was supposed to be removed in 1.13, we should do it now in 1.16


pkg/agent/client/client.go
```diff
@@ -489,25 +429,6 @@ func (c *client) newServerGRPCClient() (*grpc.ClientConn, error) {
...
 }
 
-func (c *client) fetchEntries(ctx context.Context) ([]*types.Entry, error) {
-	entryClient, connection, err := c.newEntryClient()
-	if err != nil {
-		return nil, err
-	}
-	defer connection.Release()
-
-	resp, err := entryClient.GetAuthorizedEntries(ctx, &entryv1.GetAuthorizedEntriesRequest{
```
sorindumitru, line 552: We can probably also remove this API at some point, but it will have to remain until 1.17 at least

MarcosDY (dismissed):
Looks great can you rebased? and I can approve this again

MarcosDY (approved):

PR #7222, opened 2026-08-18
Deprecate Notifier plugin

sorindumitru:
**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
Just deprecation messages.

**Description of change**
The Notifier plugin is being replaced by the BundlePublisher plugin. We should let our users know to switch.

This change marks the plugin as deprecated and also changes the internal plugin interface to expose some additional message, so we can let users know that they need to switch to the bundle publisher.


amartinezfayo (approved):
LGTM!

Issue #7176, opened 2026-07-30
bootstrapping trust bundle from configmap


kfox1111:
Since we are now supporting using the k8s psat node attestor with the tcp broker api, there are now multiple different ways to run the agent. Running it per node requires extra privilege but the tcp broker variants do not. Because the Kubernetes PSS needs to be applied per namespace, you cant really mix the two and have maximum security in your configurations, so must be in different namespaces. When using the file delivery method of k8s configmaps, you can only read the configmap from the same namespace.

So, lets add the ability to specify the namespace/configmap names where to fetch the configmap, and have the spire-agent able to fetch it directly. Then we dont have to manage syncing the bundle to configmaps in multiple namespaces.

sorindumitru:
Another option here, with fewer new configuration knobs, would be add a ClusterTrustBundle publisher plugin. Those can be read from multiple namespaces/service accounts.

kfox1111:
That could work, but GA'ed only very recently. if we supported beta too, I think that would solve it for me.

sorindumitru:
I think it should be ok to support both beta and GA, I think they're basically the same apart from the version.

kfox1111:
do we fork the configmap plugin or try and share as much code between plugins as possible?

sorindumitru:
> do we fork the configmap plugin or try and share as much code between plugins as possible?

If there are a lot of common things, yes, it would be nice to share the common code between them.

PR #7223, opened 2026-08-18
Trust bundle from configmap

kfox1111:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**
Bootstrapping/rebootstrapping

**Description of change**
spire-agent can use a Kubernetes configmap to bootstrap/rebootstrap from directly

**Which issue this PR fixes**
https://github.com/spiffe/spire/issues/7176

### DIFF
--- a/cmd/spire-agent/cli/run/run.go
+++ b/cmd/spire-agent/cli/run/run.go
@@ -68,36 +68,39 @@ type Config struct {
 }
 
 type agentConfig struct {
-	DataDir                       string    `hcl:"data_dir"`
-	AdminSocketPath               string    `hcl:"admin_socket_path"`
-	InsecureBootstrap             bool      `hcl:"insecure_bootstrap"`
-	RebootstrapMode               string    `hcl:"rebootstrap_mode"`
-	RebootstrapDelay              string    `hcl:"rebootstrap_delay"`
-	JoinToken                     string    `hcl:"join_token"`
-	JoinTokenFile                 string    `hcl:"join_token_file"`
-	LogFile                       string    `hcl:"log_file"`
-	LogFormat                     string    `hcl:"log_format"`
-	LogLevel                      string    `hcl:"log_level"`
-	LogSelectors                  []string  `hcl:"log_selectors"`
-	LogSourceLocation             bool      `hcl:"log_source_location"`
-	SDS                           sdsConfig `hcl:"sds"`
-	ServerAddress                 string    `hcl:"server_address"`
-	ServerPort                    int       `hcl:"server_port"`
-	SocketPath                    string    `hcl:"socket_path"`
-	DisableWorkloadAPI            bool      `hcl:"disable_workload_api"`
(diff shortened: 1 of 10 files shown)


cmd/spire-agent/cli/run/run.go
```diff
@@ -114,6 +117,56 @@ type agentConfig struct {
...
+//	    # running in a pod with usable in-cluster credentials.
+//	    kubeconfig_path = "/etc/spire/agent/kubeconfig"
+//	}
+type trustBundleConfigMapHCLConfig struct {
+	Namespace      string `hcl:"namespace"`
+	Name           string `hcl:"name"`
+	Key            string `hcl:"key"`
+	KubeConfigPath string `hcl:"kubeconfig_path"`
+
+	UnusedKeyPositions map[string][]token.Pos `hcl:",unusedKeyPositions"`
```
MarcosDY, line 144: It is depclared but never validated, can you add validations and return error in case some is found (similar to another cases like this one)

doc/spire_agent.md
```diff
@@ -145,16 +146,76 @@ The second case is when trust is lost and must be reestablished, known as Reboos
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: spire-bundle-reader
+subjects:
+  - kind: ServiceAccount
+    name: spire-agent
+    namespace: spire-agents
+```
+
+The bundle is read from the ConfigMap on every bootstrap and rebootstrap attempt, so an updated ConfigMap is picked up on the next attempt without restarting the agent.
```
MarcosDY, line 218: can you plz verify this? my reading is than both rebootstrap paths terminate the process (agent.go:514, manager.go:385), and GetBundle()'s only caller is agent.go:229 at startup. So a restart is required.

pkg/agent/trustbundlesources/configmap_test.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/spiffe/spire/test/spiretest"
+	"github.com/spiffe/spire/test/util"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestGetBundleFromConfigMap(t *testing.T) {
```
MarcosDY, line 21: I thing there are 2 test gaps that we may add:
- Nothing pins which ConfigMap is actually read: Every case in the table uses the same `spire-server`/`spire-bundle` pair, so hardcoding them is indistinguishable from reading the config. The "config map in another namespace" case doesn't catch it, because it only asserts the negative direction (a bundle in the wrong namespace isn't found). I think it worth one positive case with values distinct from the defaults.
- The empty-bundle guard is untested: The existing "empty bundle" case uses `""`, which fails earlier in PEM parsing and never reaches the guard.

PR #7225, opened 2026-08-18
draft: implement CMP based upstreamAuthority

### LINKED ISSUE #6711: Pluggable ServerCA / RA mode for external per-SVID signing
FIPS 140-3 compliance (NIST 800-57, ISO 19790) requires CA signing keys to be generated inside validated cryptographic modules via witnessed ceremony and to never leave the module boundary. Currently SPIRE's architecture doesn't accommodate this, instead the UpstreamAuthority plugin handles getting the intermediate CA certificate signed, but the intermediate CA key still has to be available to SPIRE at runtime for SVID signing.

Organizations with existing PKI infrastructure (Venafi, Keyfactor, EJBCA, etc.) already have FIPS-validated signing services with ceremony-established keys. SPIRE should be able to delegate per-SVID signing to those services rather than holding a signing key itself.

### Proposal

An extension point that allows SPIRE to delegate individual SVID signing to an external service:

- Per-SVID mode for UpstreamAuthority — extend UpstreamAuthority so it can be invoked per-SVID rather than only per-CA-rotation.
- ServerCA as a plugin type — a gRPC plugin interface alongside KeyManager and UpstreamAuthority. Implementations would receive signing requests and return signed certificates, calling external CAs (Venafi, EJBCA, cert-manager, etc.).

There are real tradeoffs (signing latency, availability dependency on the external CA, throughput limits), but the current architecture forces a choice between FIPS compliance and using SPIRE.

### Related
#4482 — Custom ServerCA
#525 — ServerCA HSM plugin
#2238 — Define SPIRE Server's CA explicitly

### LINKED ISSUE #7018: FR: Sign X.509-SVIDs with a static, externally-anchored CA (no SPIRE-minted downstream intermediate)

RufusJWB:
@maintainer: please feel free to comment / discuss! I'm happy to incorporate your feedback.

**Pull Request check list**

- [ ] Documentation updated?

**Affected functionality**
This PR implements a new UpstreamAuthority which issues x509-svids utilizing the CMP (RFC 4210) protocol.

**Description of change**
Implementation of new UpstreamAuthority

**Which issue this PR fixes**
fixes #6711
fixes #7018


sorindumitru:
@RufusJWB, checking in to see if you're still planning on implementing this.

RufusJWB:
> @RufusJWB, checking in to see if you're still planning on implementing this.

yes, I do. Currently I'm fighting with my CMP implementation in go

PR #7229, opened 2026-08-19
Add log file rotation

### LINKED ISSUE #7097: SPIRE: add support for [automatically] rotating log files
It would be good if SPIRE got some support for automatically rotating log files. This can be currently done through a signal on systems that support it, but it won't work on Windows. It would be good if SPIRE would either:

* Get support for automatically rolling files, maybe through something like [lumberjack](https://github.com/natefinch/lumberjack)
* Add some other way to trigger re-opening the file, maybe in the Logger API.

kerlenton:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Logging in spire-server, spire-agent and the OIDC discovery provider, specifically how the file named by `log_file` gets rotated.

**Description of change**

Rotation today means wiring up logrotate or newsyslog per host type and having it signal SPIRE with `SIGUSR2`. There is no equivalent on Windows. Since #7248 an external tool can move the log aside there, but nothing can tell a running SPIRE to start a new file, so it keeps writing into the file that was moved.

This adds an opt-in `log_file_rotation` block that has SPIRE rotate the file itself.

```hcl
log_file = "/var/log/spire/server.log"
log_file_rotation {
    max_size_mb  = 100
    max_files    = 7
    max_age_days = 30
}
```

Rotation moves the accumulated content aside to a timestamped sibling and keeps writing to `log_file`, so anything tailing that path keeps working. The same code path runs everywhere, since SPIRE closes its own descriptor before renaming. `SIGUSR2` still works and now forces an immediate rotation, which pairs with `max_size_mb = 0` when rotation should be driven entirely on demand. Rotating an already empty file is a no-op, so a nightly signal on an idle service does not spend the `max_files` budget on zero byte files.

Without the block nothing changes, so existing deployments are unaffected.

As you put it in the review, this is a convenience chosen for all platforms rather than the only thing that can work on Windows. It is still the only way to bound log growth there, since the reopen trigger does not exist yet, but on POSIX it is squarely a convenience that removes the need to configure an external rotator per host.

The writer lives in `pkg/common/log/rotate.go`. No new dependency, `go.mod` and `go.sum` are untouched. The issue mentions lumberjack and I went the other way after seeing #7064 replace an abandoned dependency with a small in repo package. lumberjack has had no commits since August 2024. Writing it here also avoids three behaviors I did not want, namely no time based rotation, dropping a write larger than MaxSize, and an unexported clock a wrapper cannot fake in tests.

**Open questions**

1. You mentioned wanting a minimal surface. Three keys is my guess at minimal, but I am happy to cut. `max_age_days` is the one I would drop first, since `max_files` already bounds disk use.
2. The OIDC discovery provider is included. It has never had rotation on any platform and the writer was already there. It is also where naming is awkward, since its option is `log_path` but the block is still `log_file_rotation`. Happy to rename or drop it from this PR.
3. `max_size_mb` is an int with the unit in the name. There is precedent for plain int counts like `x509_svid_cache_max_size` but no byte size key anywhere, so I avoided inventing a string parser.
4. Not gated behind `experimental`.
5. Rotated files inherit the directory ACL on Windows, matching what `log_file` already does.

**Which issue this PR fixes**

Fixes #7097


amartinezfayo (commented):
Thank you @kerlenton for taking this on. I would like to revisit one premise first, because I think it could change the shape of this.

As far as I can tell, what stops an external rotator from moving log_file aside on Windows is not really Windows, it is that Go's os.OpenFile never passes FILE_SHARE_DELETE, so our own handle is what denies the rename. Opening through windows.CreateFile with that flag added would let a tool rename the file while SPIRE keeps writing to the handle, the way a rename already behaves on POSIX. Windows also has a way to trigger the reopen, since a service can be sent a control code in the user defined range of 128 to 255 and x/sys/windows/svc forwards control codes to the handler unfiltered, so `sc.exe control spire-agent 128` would reach the Execute loop in pkg/common/entrypoint. Neither piece needs a configurable, and I would want the share mode fixed regardless, since today nobody can archive or move the log of a running SPIRE on Windows.

That does not remove the case for built-in rotation, since Windows ships nothing like logrotate and one file based story everywhere is what would let you drop your wrapper. It does make it a convenience we choose for all platforms rather than the only thing that can work on Windows, which matters for how much configuration it carries and for the Windows specific wording in the docs.

Would it make sense to land the platform gap first, the share mode and the control code with no new config, and then take rotation on its own merits with a minimal surface? I have notes on the writer itself, but I would rather hold them until we settle that. The validate change looks worth splitting out either way, as you offered.

kerlenton:
@amartinezfayo You are right, that is our own handle and not Windows. Fixing the wording in the description now.

Yes, that order makes sense to me. I will do the share mode and the control code first with no new config, then come back to rotation separately. Do you want those as separate PRs with this one parked as a draft in the meantime, or should I reuse this one?

One thing worth knowing for that first piece, the control code only exists when SPIRE runs as a service, so a console run still has no trigger. The share mode fix stands on its own either way.

For the trigger I was thinking of mirroring how POSIX already works, a process global broadcast out of the Execute loop that `pkg/common/log` subscribes to the way it subscribes to signal.Notify today. configure_windows.go has a TODO about driving its reconfigure from an event, so it might be able to reuse the same thing later. Say if you had something else in mind.

kerlenton:
@amartinezfayo rebased on main now that #7247 and #7248 have landed, and reworked this on top of them.

Fixed the premise you corrected. The docs and the annotated confs no longer claim Windows denies the rename, they now say an external tool can move the log aside there but cannot make SPIRE start a new one. The rotating writer also goes through the same `openLogFile` from #7248, so it gets `FILE_SHARE_DELETE` too.

The shape question is the one still open. I left three config keys and wrote up in the description where I would cut first, but I would rather hear your notes on the writer before trimming blind.

pkg/common/log/rotate.go
```diff
@@ -0,0 +1,409 @@
...
+	// disables size based rotation, leaving it to Reopen.
+	MaxSizeMB int `hcl:"max_size_mb"`
+
+	// MaxFiles is how many rotated files to retain, not counting the active
+	// one. Zero retains all of them.
+	MaxFiles int `hcl:"max_files"`
+
+	// MaxAgeDays is the age beyond which a rotated file is removed. Zero
+	// retains rotated files regardless of age.
+	MaxAgeDays int `hcl:"max_age_days"`
```
amartinezfayo: Retention runs only on rotation, so an idle service keeps aged files indefinitely and max_age_days does not hold its bound. Dropping it seems right, since max_files already bounds disk use.

pkg/common/log/rotate.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+	if c.MaxAgeDays > maxAgeDaysLimit {
+		return fmt.Errorf("max_age_days (%d) must not be greater than %d", c.MaxAgeDays, maxAgeDaysLimit)
+	}
+	return nil
+}
+
+// NewOutputFile opens the log file, returning a RotatableFile when cfg asks for
+// rotation and a ReopenableFile otherwise. A nil cfg means no rotation.
+func NewOutputFile(name string, cfg *RotationConfig) (ReopenableWriteCloser, error) {
```
amartinezfayo, line 88: An empty block, or one leaving max_size_mb unset, passes validation and then never rotates on size. Would a non-zero default help, with max_size_mb as a pointer so an explicit 0 still means on demand only?

pkg/common/log/rotate.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+	f, err := NewRotatableFile(name, *cfg)
+	if err != nil {
+		return nil, err
+	}
+	return f, nil
+}
+
+// RotatableFile rotates the log file in process rather than relying on an
+// external tool to rename it. That is the only option on Windows, where the
```
amartinezfayo: Since openLogFile now passes FILE_SHARE_DELETE, I think an external tool can rename or delete the file while SPIRE holds it open. Could this match the wording already corrected in the docs and confs?

pkg/common/log/rotate.go
```diff
@@ -0,0 +1,409 @@
...
+	if err := r.open(); err != nil {
+		return nil, err
+	}
+	return r, nil
+}
+
+// Write rotates first if this write would take the file past MaxSizeMB. A
+// failed rotation is not returned: logrus hands this writer to plugins as a
+// bare io.Writer, and reporting through the logger would re-enter logrus while
+// it holds its own lock. The write is attempted regardless, so a failure costs
```
amartinezfayo, line 158: ReopenOnSignal is a no op on Windows, so nothing calls Reopen there and a repeatedly failing rotation seems to go unreported while the file grows. Could we surface it outside the writer lock?

pkg/common/log/rotate.go
```diff
@@ -0,0 +1,409 @@
...
+			r.size = 0
+			r.prune()
+			return nil
+		}
+	}
+
+	// Ignore errors closing old file descriptor since the alternative is to
+	// keep writing to a descriptor we already decided to abandon.
+	if r.f != nil {
+		_ = r.closeFunc(r.f)
```
amartinezfayo: With FILE_SHARE_DELETE in place, do we still need to close before renaming? Renaming first, then opening the fresh path, then closing the old handle, would make a failed rename a no op.

doc/spire_server.md
```diff
@@ -75,7 +75,8 @@ This may be useful for templating configuration files, for example across differ
...
-| `log_file`                         | File to write logs to                                                                                                                                                                                                                                                                                                                                                                  |                                                                |
+| `log_file`                         | File to write logs to. On POSIX platforms, SPIRE Server reopens this file on receipt of `SIGUSR2`, so an external tool like logrotate can move it aside. That is not possible on Windows; use `log_file_rotation` there (see below)                                                                                                                                                    |                                                                |
```
amartinezfayo: This row still reads as though moving the file aside is impossible on Windows, which openLogFile changed. Could it say the reopen is what Windows lacks? The agent table reads the same way.

amartinezfayo (commented):
Thanks for creating #7247 and #7248! I would keep this PR for rotation and do the trigger as its own. The broadcast shape you describe sounds right, a process global request out of the Execute loop that pkg/common/log subscribes to the way it subscribes to signal.Notify today.

On the shape, trimming keys is not what would help most. Every key defaults to zero and zero means that bound is off, so a bare block never rotates on size and one with only max_size_mb rotates forever without pruning. If unset meant a sane default, with the keys as pointers so an explicit 0 still means on demand only, a bare block would be bounded and the keys would become tuning. I would not reject a block that cannot rotate though, since max_size_mb = 0 with SIGUSR2 is a real configuration on POSIX, and erroring would stop one config file serving all three platforms. A startup warning where nothing can trigger a rotation would fit better. I would still drop max_age_days and leave max_size_mb spelled as it is. That also fixes the provider, where nothing wires ReopenOnSignal so its block never rotates without max_size_mb.

The rest of the notes are inline, though the close before rename one is reasoned from the share mode rules rather than run on Windows. One for the trigger PR, ReopenableFile.Reopen opens before closing, and CreateFile on a file pending deletion fails with ERROR_ACCESS_DENIED, so an external delete would not be recoverable once a trigger exists.

pkg/common/log/rotate.go
kerlenton: Agreed, and the docs already admit it by saying retention applies on rotation rather than on a timer. Dropping the key.

pkg/common/log/rotate.go
kerlenton: Yeah, and that is the better framing.

pkg/common/log/rotate.go
kerlenton: Missed that one, it still says the file cannot be renamed.

pkg/common/log/rotate.go
kerlenton: Right, on Windows nothing calls Reopen so the error goes nowhere.

pkg/common/log/rotate.go
kerlenton: That ordering only existed because of the old share mode. Renaming first, then opening the fresh path, then closing the old handle, makes a failed rename a no op and matches what ReopenableFile.Reopen does.

doc/spire_server.md
kerlenton: Fixed in both tables, they now say the reopen is what Windows lacks rather than the move.

kerlenton:
All taken.

The keys are pointers now, with 100 MiB and 7 files when unset, so a bare block is bounded and an explicit 0 still means on demand only. Nothing errors when a block cannot rotate, there is a startup warning instead, which also covers the provider since it never wires ReopenOnSignal. max_age_days is gone.

Rotation failures are no longer silent. The writer records the failure and the LogReopener task drains and logs it, so the Windows slot that did nothing before now earns its keep.

On ReopenableFile.Reopen, that one is unreachable today since Reopen is only called from the SIGUSR2 handler and Windows has no trigger. It is also not a straight flip, the current order deliberately keeps a working descriptor when the open fails, so it wants a close and retry rather than a swap. Happy to open an issue so it is not lost. Are you picking up the trigger, or should I?

pkg/common/log/rotate.go
```diff
@@ -0,0 +1,423 @@
...
+		return nil, err
+	}
+	return r, nil
+}
+
+// Write rotates first if this write would take the file past MaxSizeMB. A
+// failed rotation is not returned: logrus hands this writer to plugins as a
+// bare io.Writer, and reporting through the logger would re-enter logrus while
+// it holds its own lock. The write is attempted regardless, so a failure costs
+// an oversized file rather than a lost line. Reopen surfaces the error.
```
amartinezfayo: nit: Reopen is no longer the only path, since TakeRotateError is what reports a size triggered failure now. Could this sentence mention the drain?

pkg/common/log/rotate.go
```diff
@@ -0,0 +1,423 @@
...
+	defer r.mu.Unlock()
+
+	if r.shouldRotate(int64(len(b))) {
+		_ = r.rotateAndRecord()
+	}
+
+	// An earlier rotation failed after closing the descriptor. Reopen rather
+	// than drop this line and every line after it, ignoring the backoff since
+	// the point is to restore logging, not to rotate.
+	if r.f == nil {
```
amartinezfayo: Now that rotate restores the old descriptor on failure, nothing sets f to nil, so this branch looks unreachable. Could it go, along with the test that fabricates the nil?

pkg/common/log/rotateerr.go
```diff
@@ -0,0 +1,45 @@
...
+// report a failure from inside Write.
+type rotateErrorSource interface {
+	TakeRotateError() error
+}
+
+// reportRotateErrors logs rotation failures until ctx is done. Without it a
+// self rotating file that keeps failing would grow silently, since Reopen is
+// the only other path that surfaces the error and nothing triggers it on
+// Windows.
+func reportRotateErrors(ctx context.Context, logger *Logger, reopener Reopener, interval time.Duration) {
```
amartinezfayo: Nothing exercises this or TakeRotateError, so the path that makes a Windows failure visible is untested. Could we add one that fails a rotation, then asserts the error is logged and cleared?

pkg/common/log/reopen_posix.go
```diff
@@ -20,7 +20,14 @@ func ReopenOnSignal(logger *Logger, reopener Reopener) func(context.Context) err
 	return func(ctx context.Context) error {
 		signalCh := make(chan os.Signal, 1)
 		signal.Notify(signalCh, reopenSignal)
-		return reopenOnSignal(ctx, logger, reopener, signalCh)
+		defer signal.Stop(signalCh)
+
+		// Rotation failures are reported alongside the signal handling, since a
+		// size triggered rotation can fail without anyone sending a signal.
+		go reportRotateErrors(ctx, logger, reopener, rotateErrorInterval)
```
amartinezfayo: nit: the Windows path calls this inline, while here it runs in a goroutine. Could the ticker fold into the select below, so both platforms read the same way?

doc/spire_server.md
amartinezfayo: Both log_file rows still read the same as before, byte for byte. The prose below them was already right, so I think this one is still open.

amartinezfayo: I think the log_file rows are unchanged in both docs, so "That is not possible on Windows" is still there. Could that sentence say the reopen is what Windows lacks rather than the move?

amartinezfayo (commented):
I will pick up the trigger, so no need to take that on, thank you!
Notes are inline, and the docs one from last round still looks open, so I left a note there. None of it blocks this PR, since the trigger builds on the share mode change that already landed, so this can go in first and the trigger PR can pick up what it makes stale in the constant, the warning, and the docs.
I would fold ReopenableFile.Reopen in there rather than track it separately, since the ordering is only observable once something can call Reopen on Windows, so no separate issue needed. I would keep open first as the primary path so POSIX does not change, adding the close and retry only for the delete pending case.

pkg/common/log/rotate.go
kerlenton: Updated, it now says Reopen returns it and TakeRotateError drains it for the size triggered case.

pkg/common/log/rotate.go
kerlenton: Branch gone, and both tests that fabricated the nil went with it.

pkg/common/log/rotateerr.go
kerlenton: Added. One fails a rotation through a size triggered write, then asserts the error is logged and that a second drain comes back clean. Another checks a ReopenableFile produces nothing, since it has no rotation to report.

pkg/common/log/reopen_posix.go
kerlenton: Folded. There is one watchLog now with the ticker in the same select, and the signal channel is nil on Windows, so a receive on it blocks forever and both platforms run the same loop. The goroutine is gone.

doc/spire_server.md
kerlenton: Sorry, I fixed the sub table last round and said both were done. The log_file rows are actually changed now, in both docs.

kerlenton:
Thanks for taking the trigger, and for the pointer that this can go in first.

All four inline notes are in. The nil branch and its two tests are gone, watchLog is one loop with the signal channel nil on Windows, and the drain has tests now. The log_file rows really are fixed this time.

pkg/common/log/rotate.go
```diff
@@ -0,0 +1,415 @@
...
+
+	if err := r.renameCurrent(); err != nil {
+		return err
+	}
+
+	old := r.f
+	if err := r.open(); err != nil {
+		// The old descriptor still works, it just writes into the file that was
+		// moved aside. Keeping it loses less than dropping it.
+		r.f = old
```
amartinezfayo: Removing the fabricated nil test also dropped the only cover for this restore, which coverage now shows unexercised. Could we exercise a failed open after the rename, since this is what makes the nil branch unnecessary?

pkg/common/log/rotateerr_test.go
```diff
@@ -0,0 +1,92 @@
...
+	logger, hook := newNullLogger()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		watchLog(ctx, logger, rf, nil, time.Millisecond)
+	}()
+
+	time.Sleep(20 * time.Millisecond)
```
amartinezfayo: This can pass without a tick ever firing, so it may not be testing much. Could it use require.Never, the counterpart to the Eventually above?

pkg/common/log/rotateerr.go
```diff
@@ -0,0 +1,54 @@
...
+
+// watchLog reopens the log on every value from signalCh and reports rotation
+// failures that the writer could not report itself. signalCh is nil on
+// platforms without a reopen signal, and a receive on a nil channel blocks
+// forever, so both platforms run the same loop.
+func watchLog(ctx context.Context, logger *Logger, reopener Reopener, signalCh <-chan os.Signal, interval time.Duration) {
+	// Only a self rotating file has failures to drain.
+	source, _ := reopener.(rotateErrorSource)
+
+	ticker := time.NewTicker(interval)
```
amartinezfayo: nit: the ticker runs even when there is nothing to drain. Could it be nil for that case, the way signalCh already is, which would also drop the continue?

pkg/common/log/rotate.go
amartinezfayo, line 283: TestRotatableFileWriteAfterUnusableRotation was the only thing reaching this restore, and coverage now shows the block unexercised. Could a case that fails the open after the rename come back, since this restore is what makes the nil branch unnecessary?

pkg/common/log/rotateerr_test.go
amartinezfayo: This can pass without a tick ever firing, so it may not be exercising the guard. Would taking the tick channel as a parameter help, so tests drive ticks rather than sleep for them, and the ticker can be nil when there is nothing to drain?

amartinezfayo (commented):
Thank you @kerlenton, I think we are pretty close!
Something I never answered is the experimental question from the description. I don't think we need an experimental section for these settings, the surface is two keys in a shape operators already know and rotation is settled enough that I am comfortable committing to those names.
I left a couple of comments. Thank you for your patience going through this!

pkg/common/log/rotate.go
kerlenton: Good catch, I took the coverage out with the test. Added one that fails the open after a successful rename through an openFunc seam, next to the existing closeFunc and renameFunc ones. It asserts the line still lands in the file that was moved aside, and that the writer returns to the configured path once opening works again.

pkg/common/log/rotateerr_test.go
kerlenton: Both, thank you. watchLog takes the tick channel now, so tests send ticks instead of sleeping for them, and rotateErrorTicker returns nil when the writer has nothing to drain, so a ReopenableFile never gets a ticker at all. The nil channel blocks forever in the select, same trick as the signal channel, so no extra branch. Split into three cases, the ticker choice, a tick with a failure recorded, and a tick with none.

kerlenton:
Both notes are in. The restore has a test again, and the drain is driven by an injected tick channel rather than a sleep.

pkg/common/log/rotate.go
```diff
@@ -0,0 +1,417 @@
...
+	f  *os.File
+	// size is seeded from disk so content written before this process started,
+	// or removed by an external copytruncate, is accounted for.
+	size int64
+	// rotateFailedAt is zero unless the last rotation attempt failed, and
+	// rotateErr carries that failure until something reports it.
+	rotateFailedAt time.Time
+	rotateErr      error
+
+	// now, closeFunc and renameFunc are intended for injecting errors and a
```
amartinezfayo: nit: openFunc joined this group but not the sentence above it. Could it be listed with the others, since it is also called while holding the lock?

pkg/common/log/rotateerr.go
```diff
@@ -0,0 +1,60 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-signalCh:
+			if err := reopener.Reopen(); err != nil {
+				// never fail; best effort to log to old file descriptor
+				logger.WithError(err).Error(failedToReopenMsg)
+			}
+		case <-tickCh:
+			source, ok := reopener.(rotateErrorSource)
```
amartinezfayo: nit: rotateErrorTicker already made this decision, so a non-nil tickCh means the assertion holds, and coverage shows this branch unreached. Could the source come back from rotateErrorTicker instead?

amartinezfayo (approved):
Thank you @kerlenton! 🚀

amartinezfayo:
@kerlenton, heads-up that I've opened https://github.com/spiffe/spire/pull/7269 to add support for triggering a log file reopen on Windows. If you have a chance to take a look, that would be great!

kerlenton:
Yeah, I'll take a look.

PR #7239, opened 2026-08-22
Add BundlePublisher migration guidance to the Notifier plugin docs

amartinezfayo:
The Notifier plugin type was deprecated in #7222, and both notifier docs now point at the BundlePublisher plugin that replaces them. The pointer alone leaves out the parts of the migration that are not mechanical, so an operator following it has to diff two reference docs to work out what actually changes.

This PR adds a migration section to each notifier doc, with the configuration mapping and the behavior differences that matter when switching.

**`gcs_bundle`**

- `bucket` and `object_path` become `bucket_name` and `object_name`, while `service_account_file` keeps its name and the authentication behavior is unchanged.
- `format` is required and has no equivalent in the notifier. Setting it to `pem` keeps the object contents unchanged, since `gcs_bundle` uploads the X.509 authorities PEM encoded, while `spiffe` and `jwks` include the JWT authorities as well.

**`k8sbundle`**

- Every cluster is configured under the `clusters` map, including the local one, and `namespace`, `configmap_name`, `configmap_key` and `format` are all required where `k8sbundle` defaulted them to `spire`, `spire-bundle` and `bundle.crt`.
- `k8s_configmap` writes the ConfigMap with a server-side apply that creates it when it does not exist, so the Service Account needs `create` on ConfigMaps in addition to `get` and `patch`.
- An empty `clusters` map is a valid configuration under which the bundle is never published, which is an easy state to land in when moving the settings over.

Both sections end with a configuration equivalent to the notifier sample it replaces, and the deprecation warning at the top of each doc now links to the section.

While writing the `gcs_bundle` mapping I noticed that the two sample configurations in the `gcp_cloudstorage` doc set `bucket`, but the plugin reads `bucket_name`. HCL drops the unknown key, so configuring from either sample fails with `configuration is missing the bucket name`. This PR fixes both samples.

### DIFF
--- a/doc/plugin_server_notifier_gcs_bundle.md
+++ b/doc/plugin_server_notifier_gcs_bundle.md
@@ -3,7 +3,7 @@
 > [!WARNING]
 > The Notifier plugin type is deprecated and will be removed in a future
 > release. Use the [`gcp_cloudstorage` BundlePublisher](/doc/plugin_server_bundlepublisher_gcp_cloudstorage.md)
-> plugin instead.
+> plugin instead. See [Migrating to the `gcp_cloudstorage` BundlePublisher](#migrating-to-the-gcp_cloudstorage-bundlepublisher).
 
 The `gcs_bundle` plugin responds to bundle loaded/updated events by fetching and
 pushing the latest root CA certificates from the trust bundle to an object in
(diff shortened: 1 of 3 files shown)


doc/plugin_server_notifier_gcs_bundle.md
```diff
@@ -59,3 +59,34 @@ credentials found in the `/path/to/service/account/file` file.
         }
     }
 ```
+
+## Migrating to the `gcp_cloudstorage` BundlePublisher
+
+The [`gcp_cloudstorage` BundlePublisher](/doc/plugin_server_bundlepublisher_gcp_cloudstorage.md)
+uploads the trust bundle to a Google Cloud Storage object as well, and
+authenticates the same way, but the configuration options are named differently
+and the bundle format has to be set explicitly.
```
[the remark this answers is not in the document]
amartinezfayo: Fixed the sample configs to use `bucket_name`. The samples were not just inconsistent with the migration section: the plugin reads `bucket_name`, so HCL dropped the unknown `bucket` key and configuring from either sample failed with "configuration is missing the bucket name".

doc/plugin_server_notifier_k8sbundle.md
```diff
@@ -163,3 +164,51 @@ server to
...
+writes the trust bundle to a Kubernetes ConfigMap as well, but every cluster is
+configured under the `clusters` map, including the local one, and the settings
+that `k8sbundle` fills in by default have to be set explicitly.
+
+| `k8sbundle`             | `k8s_configmap`   | Notes                                                |
+|-------------------------|-------------------|------------------------------------------------------|
+| `namespace`             | `namespace`       | Required. `k8sbundle` defaults it to `spire`.        |
+| `config_map`            | `configmap_name`  | Required. `k8sbundle` defaults it to `spire-bundle`. |
+| `config_map_key`        | `configmap_key`   | Required. `k8sbundle` defaults it to `bundle.crt`.   |
+| `kube_config_file_path` | `kubeconfig_path` | In-cluster credentials are used when unset.          |
```
sorindumitru: Maybe we can specify something like `clusters.namespace` for these to make it a bit clearer. But I'm also happy with what you have here, it's clear from the example.

sorindumitru (dismissed):

doc/plugin_server_notifier_k8sbundle.md
amartinezfayo: Thanks for the suggestion, I went with `clusters.namespace` and the same prefix on the other per-cluster keys. I also reworded the sentence below the table to say each cluster's `format`, so the table and the prose agree on where those keys live.

sorindumitru (approved):
LGTM, thanks @amartinezfayo!

PR #7240, opened 2026-08-24
httpchallenge: optionally verify client IP against hostname DNS

samuraixwandering:
**Pull Request check list**

- [x] Documentation updated?

**Affected functionality**

Server-side `http_challenge` node attestor. Optional config only; default behavior is unchanged.

**Description of change**

Adds `verify_client_ip` to the http_challenge server plugin (default `false`), following the x509pop option introduced in #6911 and the sshpop work in #7191.

When enabled, the connecting peer IP (from `X-Forwarded-Client-IP` on the node attestor facade) is checked against the DNS addresses for the attested hostname. Lookup failure or no matching address fails attestation. The check runs after hostname validation and before the HTTP challenge.

This is the second of the two remaining plugins from #6769 (sshpop merged as #7191). Split per maintainer guidance.

The IP comes from `ClientIPFromContext`, the helper added in #7191, so sshpop and this plugin read the metadata the same way. The DNS lookup gets a 10s timeout on the stream context. That's the same budget as the challenge fetch, but derived from the stream so an agent hanging up cancels the lookup.

**Test plan**

- [x] `go test ./pkg/server/plugin/nodeattestor/httpchallenge/`
- [x] Unit coverage for flag off (lookup not called), missing/mismatched/unresolvable client IP, IPv4/IPv6 match, and match among several DNS answers. Fail-closed cases use a transport that fatals if the HTTP challenge runs.

Part of #6769

### DIFF
--- a/pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
+++ b/pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"net"
 	"net/http"
 	"regexp"
 	"sync"
@@ -17,8 +18,10 @@ import (
 	"github.com/spiffe/spire/pkg/common/catalog"
 	"github.com/spiffe/spire/pkg/common/plugin/httpchallenge"
 	"github.com/spiffe/spire/pkg/common/pluginconf"
+	"github.com/spiffe/spire/pkg/server/plugin/nodeattestor"
 	nodeattestorbase "github.com/spiffe/spire/pkg/server/plugin/nodeattestor/base"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 )
 
@@ -39,10 +42,19 @@ func disableHTTPRedirects(req *http.Request, via []*http.Request) error {
 }
 
 func BuiltInTesting(client *http.Client, forceNonce string) catalog.BuiltIn {
+	return BuiltInTestingWithLookup(client, forceNonce, nil)
+}
+
+// BuiltInTestingWithLookup is BuiltInTesting with a DNS lookup hook.
+// A nil lookupIPs uses the system resolver.
+func BuiltInTestingWithLookup(client *http.Client, forceNonce string, lookupIPs func(context.Context, string) ([]net.IP, error)) catalog.BuiltIn {
 	plugin := New()
 	plugin.client = client
 	plugin.client.CheckRedirect = disableHTTPRedirects
 	plugin.forceNonce = forceNonce
+	if lookupIPs != nil {
+		plugin.lookupIPs = lookupIPs
+	}
 	return builtin(plugin)
 }
 
@@ -59,6 +71,7 @@ type configuration struct {
 	allowNonRootPorts bool
 	dnsPatterns       []*regexp.Regexp
 	tofu              bool
+	verifyClientIP    bool
 }
 
 func buildConfig(coreConfig catalog.CoreConfig, hclText string, status *pluginconf.Status) *configuration {
@@ -111,6 +124,7 @@ func buildConfig(coreConfig catalog.CoreConfig, hclText string, status *pluginco
 		requiredPort:      hclConfig.RequiredPort,
 		allowNonRootPorts: allowNonRootPorts,
 		tofu:              tofu,
+		verifyClientIP:    hclConfig.VerifyClientIP,
 	}
 }
 
@@ -119,6 +133,7 @@ type Config struct {
 	RequiredPort       *int     `hcl:"required_port"`
 	AllowNonRootPorts  *bool    `hcl:"allow_non_root_ports"`
 	TOFU               *bool    `hcl:"tofu"`
+	VerifyClientIP     bool     `hcl:"verify_client_ip"`
 }
 
 type Plugin struct {
@@ -133,6 +148,7 @@ type Plugin struct {
 
 	client     *http.Client
 	forceNonce string
+	lookupIPs  func(context.Context, string) ([]net.IP, error)
 }
 
 func New() *Plugin {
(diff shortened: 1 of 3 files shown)


pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -270,6 +292,49 @@ func (p *Plugin) getConfig() (*configuration, error) {
...
+	clientIP := net.ParseIP(ips[0])
+	if clientIP == nil {
+		return status.Errorf(codes.Internal, "invalid client IP %q", ips[0])
+	}
+
+	lookup := p.lookupIPs
+	if lookup == nil {
+		lookup = defaultLookupIPs
+	}
+	resolved, err := lookup(ctx, hostName)
```
amartinezfayo: Attest bounds the challenge fetch with an explicit 10 second timeout, but this resolver call seems to inherit the stream context instead. Would a short context.WithTimeout here keep the two outbound calls consistent?

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -270,6 +292,49 @@ func (p *Plugin) getConfig() (*configuration, error) {
...
+	}
+	if len(ips) > 1 && p.log != nil {
+		p.log.Warn("Multiple client IP values found in metadata, using first value")
+	}
+	clientIP := net.ParseIP(ips[0])
+	if clientIP == nil {
+		return status.Errorf(codes.Internal, "invalid client IP %q", ips[0])
+	}
+
+	lookup := p.lookupIPs
```
amartinezfayo: nit: I think New could set lookupIPs to the default resolver, dropping this per-call fallback. net.Resolver.LookupIP already returns []net.IP, so defaultLookupIPs and its nil filtering could go too.

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -39,10 +42,19 @@ func disableHTTPRedirects(req *http.Request, via []*http.Request) error {
 }
 
 func BuiltInTesting(client *http.Client, forceNonce string) catalog.BuiltIn {
+	return BuiltInTestingWithLookup(client, forceNonce, nil)
+}
+
+// BuiltInTestingWithLookup is BuiltInTesting with a DNS lookup hook.
+// A nil lookupIPs uses the system resolver.
+func BuiltInTestingWithLookup(client *http.Client, forceNonce string, lookupIPs func(context.Context, string) ([]net.IP, error)) catalog.BuiltIn {
```
amartinezfayo: nit: BuiltInTesting has one caller, in this package's test. I wonder if adding the lookup parameter to it would be simpler than a second exported constructor, with callers that do not care passing nil.

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge_test.go
```diff
@@ -504,7 +508,131 @@ func TestAttestSucceeds(t *testing.T) {
...
+		HostName:  "foo",
+		AgentName: "default",
+		Port:      80,
+	})
+	challengeFn := func(context.Context, []byte) ([]byte, error) {
+		return nil, nil
+	}
+	failingTransport := &http.Client{
+		Transport: roundTripFunc(func(*http.Request) (*http.Response, error) {
+			t.Fatal("http challenge must not run when client IP verification fails")
```
amartinezfayo: This round trip runs on the plugin's gRPC handler goroutine, where t.Fatal would abort the handler rather than the test and surface as a confusing stream error. Could we use t.Error and return an error, here and in the lookup hooks?

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
```diff
@@ -270,6 +292,49 @@ func (p *Plugin) getConfig() (*configuration, error) {
 	return p.config, nil
 }
 
+func (p *Plugin) verifyConnectingClientIP(ctx context.Context, hostName string) error {
+	ips := metadata.ValueFromIncomingContext(ctx, nodeattestor.XForwardedClientIPKey)
```
amartinezfayo: nit: nodeattestor.ClientIPFromContext landed with the sshpop change and already does the metadata read and the multiple-value warning. Could this collapse into a call to it, keeping the existing split between the empty and unparseable errors?

amartinezfayo (commented):
Thank you @samuraixwandering!

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
samuraixwandering: Yes. Lookup now uses `context.WithTimeout` on the stream context, 10s, same as the challenge fetch.

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
samuraixwandering: `New()` sets `lookupIPs` to `DefaultResolver.LookupIP`. Dropped `defaultLookupIPs` and the per-call fallback.

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
samuraixwandering: Folded the hook into `BuiltInTesting`. The extra constructor is gone. Callers that do not care pass nil.

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge_test.go
samuraixwandering: Hooks now `t.Error` and return an error. The hostname check in the mismatch case runs after `Attest`, not on the handler goroutine.

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge.go
samuraixwandering: Now `ClientIPFromContext`. Empty is still "client IP not available", unparseable is still "invalid client IP".

pkg/server/plugin/nodeattestor/httpchallenge/httpchallenge_test.go
samuraixwandering: Confirmed, and it's worse than a stream error: it hangs.

I put the old `t.Fatal` back and made `Attest` skip the verification so the challenge would actually run. `Goexit` unwinds the handler goroutine inside `RoundTrip`, so the handler never sends a status and the client stream never returns. The dump has the subtest parked in `nodeattestor.(*V1).Attest` at `v1.go:91`, inside `Recv`. No output, no failure message. I killed the first run at 131s.

With `t.Error` the same break fails in under a second and points at the transport.

amartinezfayo (approved):
Thank you @samuraixwandering!

PR #7242, opened 2026-08-25
Set Status in WIT-SVID reply and fix dead test

sorindumitru:
The server WIT fetching APIs didn't return the status that is required by spire-agent. Also fix a test that was unintentionally disabled.


amartinezfayo (approved):
🚀

PR #7243, opened 2026-08-25
Make sure bundle contains wit keys

sorindumitru:
Fix some leftover pieces now that go-spiffe has the support for wit keys.

### DIFF
--- a/pkg/common/bundleutil/bundle.go
+++ b/pkg/common/bundleutil/bundle.go
@@ -99,7 +99,16 @@ func SPIFFEBundleToProto(b *spiffebundle.Bundle) (*common.Bundle, error) {
 		})
 	}
 
-	/* TODO: Parse WIT authorities once go-spiffe adds support */
+	for kid, key := range b.WITAuthorities() {
+		pkixBytes, err := x509.MarshalPKIXPublicKey(key)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal public key: %w", err)
+		}
+		bundle.WitSigningKeys = append(bundle.WitSigningKeys, &common.PublicKey{
+			PkixBytes: pkixBytes,
+			Kid:       kid,
+		})
+	}
 
 	return bundle, nil
 }
@@ -113,12 +122,10 @@ func SPIFFEBundleFromProto(b *common.Bundle) (*spiffebundle.Bundle, error) {
 	if err != nil {
 		return nil, err
 	}
-	/* TODO: Set WIT authorities once go-spiffe adds support
 	witSigningKeys, err := WITSigningKeysFromBundleProto(b)
 	if err != nil {
 		return nil, err
 	}
-	*/
 	td, err := spiffeid.TrustDomainFromString(b.TrustDomainId)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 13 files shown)


pkg/common/bundleutil/bundle.go
```diff
@@ -127,6 +134,7 @@ func SPIFFEBundleFromProto(b *common.Bundle) (*spiffebundle.Bundle, error) {
 	bundle := spiffebundle.New(td)
 	bundle.SetX509Authorities(rootCAs)
 	bundle.SetJWTAuthorities(jwtSigningKeys)
+	bundle.SetWITAuthorities(witSigningKeys)
```
amartinezfayo, line 137: I think this makes the agent's JWT bundle responses carry WIT keys: parseBundles feeds the cache through here, and composeJWTBundlesResponse marshals with only NoX509SVIDKeys and StandardJWKS, so go-spiffe's jwtbundle.Parse would accept them as JWT authorities. Should those call sites also pass NoWITSVIDKeys?

pkg/server/api/bundle.go
```diff
@@ -143,15 +150,38 @@ func ParseJWTAuthorities(keys []*types.JWTKey) ([]*common.PublicKey, error) {
 		}
 
 		jwtKeys = append(jwtKeys, &common.PublicKey{
-			PkixBytes: key.PublicKey,
-			Kid:       key.KeyId,
-			NotAfter:  key.ExpiresAt,
+			PkixBytes:  key.PublicKey,
+			Kid:        key.KeyId,
+			NotAfter:   key.ExpiresAt,
+			TaintedKey: key.Tainted,
```
amartinezfayo: I think setting TaintedKey here changes AppendBundle: a caller can now persist a tainted flag on the local trust domain bundle without the checks TaintJWTAuthority applies, and ParseX509Authorities still drops the flag. Was that intended, or should the parse helpers keep ignoring tainted on input?

pkg/server/api/bundle.go
```diff
@@ -92,12 +92,18 @@ func ProtoToBundle(b *types.Bundle) (*common.Bundle, error) {
...
+		return nil, fmt.Errorf("unable to parse WIT authority: %w", err)
+	}
+
 	commonBundle := &common.Bundle{
 		TrustDomainId:  td.IDString(),
 		RefreshHint:    b.RefreshHint,
 		SequenceNumber: b.SequenceNumber,
 		RootCas:        rootCas,
 		JwtSigningKeys: jwtSigningKeys,
+		WitSigningKeys: witSigningKeys,
```
amartinezfayo, line 106: nit: Now that WIT authorities are accepted on this path, should FieldsFromBundleProto also emit them in the audit fields? The federated bundle RPCs would otherwise log JWT and X.509 authorities but nothing for WIT ones. Fine as a follow-up if new telemetry names are needed.

pkg/common/bundleutil/bundle.go
```diff
@@ -99,7 +99,16 @@ func SPIFFEBundleToProto(b *spiffebundle.Bundle) (*common.Bundle, error) {
 		})
 	}
 
-	/* TODO: Parse WIT authorities once go-spiffe adds support */
+	for kid, key := range b.WITAuthorities() {
+		pkixBytes, err := x509.MarshalPKIXPublicKey(key)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal public key: %w", err)
```
amartinezfayo: nit: This error reads the same as the one in the JWT loop above. Would naming the key type, as the FromProto side does with "unable to parse WIT signing key", make failures easier to tell apart?

pkg/common/bundleutil/bundle.go
sorindumitru: yikes, thanks for catching that. I'll add a test for it to make sure we catch this in the future.

pkg/server/api/bundle.go
sorindumitru: Ah, I see. I'll remove the flag from these

pkg/server/api/bundle.go
sorindumitru: Yes, added it to the audit logs

amartinezfayo (approved):
🚀

